PR #9, opened 2015-06-05
Olog

danielballan:
See most recent commit only. (This is built on the huge stack of commits in `more-scans`.)

Is Olog integration really this simple?


danielballan:
This was superceded long ago.

PR #60, opened 2015-06-09
WIP: Filestore validation callback.

danielballan:
@tacaswell can probably do this better, but I'll leave this PR here so we remember that it exists. Feel free to supercede.


ericdill:
This needs to be revisited in light of new api's in dataportal, metadatastore and filestore

danielballan:
It's like we also broke some other bits of broker_callbacks.

danielballan:
Superceded by a succinct implementation in #311

Issue #73, opened 2015-06-11
Message validation


danielballan:
Various docstrings and docs claim that certain messages and message sequences are not allowed. We don't actually validate all of them yet, so the RunEngine is likely to fail in nasty ways if users break the rules.

A running list of things to validate:

- [x] 'checkpoint' must not come between 'create' and 'save
- [ ] 'subscribe' cannot subscribe to 'start'
- [ ] 'save' must follow 'create'
- [ ] 'deconfigure' must follow a 'configure' with the same target

danielballan:
(For the record, as @ericdill knows from offline conversations, we now do allow 'subscribe' to start.)

danielballan:
@ericdill If you are interested in revisiting this, here's one way it could be done with surer accuracy and less repeated logic and code:

Use `RE.unregister_command` to remove some of the built-in methods and replace that with mocked-up alternatives methods that don't actually do anything. The internal logic of RunEngine itself -- both the state machine and the numerous flags -- will be exercised. They won't need to be mocked up separately. The mocked-up methods can watch for things like `IllegalMessageSequence`.

tacaswell:
Closing this as we have many validation checks in place.

PR #97, opened 2015-06-15
Scan Validation

ericdill:
Scan Validation. The plan here is to pass a scan generator and a run engine to a validation class.  This class will use one or more state machines internally to validate all the message that it receives. It will also count the number of messages of each type that it has received.  If everything behaves, ScanValidator.validate() will execute without raising. ScanValidator.report() will then print out the number of times each message was called.

IPython notebook in progress: http://nbviewer.ipython.org/7bda5523b33f7a611b86

When done, this will take care of #73. As of now (https://github.com/danielballan/bluesky/commit/319a502f907d8e00868b39812f3fa24f643b616c) all it does is count messages.

### DIFF
--- a/bluesky/broker_examples.py
+++ b/bluesky/broker_examples.py
@@ -3,6 +3,7 @@ from .examples import Reader
 from .run_engine import Msg
 from filestore.file_writers import save_ndarray
 import tempfile
+from .examples import *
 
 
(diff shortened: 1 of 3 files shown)


danielballan:
Per conversation @ericdill, punting for now as low-prio.

PR #133, opened 2015-06-30
Sundry

tacaswell:


danielballan:
One minor line note. Merge at will.

PR #139, opened 2015-07-15
Implement a SPEC-like API layer

danielballan:
After lengthy discussion with @stuwilkins, @tacaswell, @ericdill, we are implementing Python functions with the same names as popular SPEC macros. They do not of course use SPEC's *syntax*, but they have matching function signatures.

We think of this as a analogous to matplotlib's "pyplot," an API with quick but less flexible usage. By designing the object-oriented API first, we hope we have avoided some of the negative aspects of the design of pyplot.


danielballan:
This now incorporates commits from #137, which I needed, and multidim scans.

bluesky/brittle_scans.py
```diff
@@ -0,0 +1,247 @@
+"""
```
tacaswell: We have to rename thing module, we are being brats.

I suggest we go back to 'simple'

bluesky/brittle_scans.py
```diff
@@ -0,0 +1,247 @@
...
+These "scans" bundle a Message generator with an instance of the RunEngine,
+combining two separate concepts -- instructions and execution -- into one
+object. This makes the interface less flexible and somewhat less "Pythonic"
+but more convenient to users who are not acclimated to using tab-completion and
```
tacaswell: remove commentary

bluesky/brittle_scans.py
danielballan: Way to talk me into this and then pull the high ground. :- P

tacaswell: Yeah, sorry, I blame @stuwilkins for pulling the highground on me.

danielballan:
Status update: A `GlobalState` singleton is implemented with descriptors that validate the settings at assignment time. ("Hey, that's not a detector!")

All tests pass. Simple scans are documented but need unit tests.

Also, older documentation sections (live feedback, pausing, etc.) will need to be updated once we setting on how `GlobalState`, `RE`, and others are exposed in the IPython session.

bluesky/global_state.py
```diff
@@ -0,0 +1,69 @@
...
+                            "a '{1}' method'".format(det, method))
+
+
+class GlobalState:
+    "A bucket of validated global state used by the simple scan API"
+    RE = RunEngineDescriptor(RunEngine(dict()))
+    DETS = DetectorListDescriptor([])
+    MASTER_DET = DetectorDescriptor(None)
+    MASTER_DET_FIELD = DetectorDescriptor(None)
+    H_MOTOR = DetectorDescriptor(None)
```
tacaswell: Shouldn't these be `MovabelDescriptor`?

danielballan:
Merging own with @tacaswell's blessing.

PR #144, opened 2015-07-21
TST: Test logbook

danielballan:
- Not to be confused with Python logging
- This does not actually use the olog, but it checks that a
  correct message is passed to the handler that is normally
  plugged into the olog.


danielballan:
@tacaswell ready to merge

PR #146, opened 2015-07-21
Snake

danielballan:
For my last trick before vacation... "snaked" mesh scans:

```
# The final argument here means, "Move motor2 on a snaked trajectory."
# The first motor, the "slow axis", is never repeated and therefore never snaked.

In [11]: dscan = OuterProductDeltaScan([det], motor1, 1, 3, 3, motor2, 1, 2, 2, True)

In [12]: gs.RE(dscan)
+------------+-------------------+----------------+----------------+
|   seq_num  |             time  |        motor1  |        motor2  |
+------------+-------------------+----------------+----------------+
|         1  |  18:32:18.920597  |          1.00  |          1.00  |
|         2  |  18:32:19.191593  |          1.00  |          2.00  |
|         3  |  18:32:19.360474  |          2.00  |          2.00  |
|         4  |  18:32:19.629678  |          2.00  |          1.00  |
|         5  |  18:32:19.800756  |          3.00  |          1.00  |
|         6  |  18:32:20.070489  |          3.00  |          2.00  |
+------------+-------------------+----------------+----------------+
```

Notes:
* This works generally for any number of dimensions.
* Specify whether to "snake" or simply go left-to-right on a motor by motor basis.
* As illustrated above, an odd number of steps is not a problem.
* It is possible to mix "outer product" (mesh) scans and "inner product" (a2, d2) scans and without rewriting the snaking logic.

### DIFF
--- a/bluesky/scans.py
+++ b/bluesky/scans.py
@@ -5,7 +5,7 @@ from cycler import cycler
 from lmfit.models import GaussianModel, LinearModel
 import numpy as np
 from .run_engine import Msg
-from .utils import Struct
+from .utils import Struct, snake_cyclers
 
 
 class ScanBase(Struct):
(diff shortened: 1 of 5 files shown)


danielballan:
@tacaswell This solidly tested, so unless you have revisions, merge away!

@gjwillms Look OK to you?

Issue #153, opened 2015-08-05
Validate uniqueness of gs.DETS


danielballan:
Inspired by @ericdill in NSLS-II/ophyd#138

danielballan:
To state the obvious, gs.DETS is a traitlet so validation is very easy.

PR #155, opened 2015-08-06
Prohibit duplicates in gs.DETS and during event bundling

danielballan:
We can do this elsewhere as well (in the RunEngine, perhaps also in the scan `detectors` setters) but this is a start.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -739,7 +739,7 @@ class RunEngine:
             keys = data_keys.keys()  # that is, field names
             for known_obj, known_data_keys in self._describe_cache.items():
                 known_keys = known_data_keys.keys()  # that is, field names
-                if len(set(known_keys) & set(keys)) > 0:
+                if set(known_keys) & set(keys):
                     raise ValueError("Data keys (field names) from {0!r} "
                                      "collide with those from {1!r}"
                                      "".format(obj, known_obj))
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -742,7 +742,7 @@ class RunEngine:
                 if len(set(known_keys) & set(keys)) > 0:
                     raise ValueError("Data keys (field names) from {0!r} "
                                      "collide with those from {1!r}"
-                                     "".format(obj, old_obj))
+                                     "".format(obj, known_obj))
             self._describe_cache[obj] = data_keys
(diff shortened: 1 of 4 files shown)


tacaswell:
This does not check during bundling, only at the global state level.

danielballan:
@tacaswell I changed the title before pushing the commits. :- ) Check again.

danielballan:
Closes #153

danielballan:
What is "QuantifiedCode" and why is it yelling at me?

bluesky/run_engine.py
```diff
@@ -734,7 +734,16 @@ def _read(self, msg):
         obj = msg.obj
         self._objs_read.append(obj)
         if obj not in self._describe_cache:
-            self._describe_cache[obj] = obj.describe()
+            # Validate that there is no data key name collision.
+            data_keys = obj.describe()
+            keys = data_keys.keys()  # that is, field names
+            for known_obj, known_data_keys in self._describe_cache.items():
+                known_keys = known_data_keys.keys()  # that is, field names
+                if len(set(known_keys) & set(keys)) > 0:
```
tacaswell: Don't really need the `>0`

bluesky/run_engine.py
```diff
@@ -734,7 +734,16 @@ def _read(self, msg):
         obj = msg.obj
         self._objs_read.append(obj)
         if obj not in self._describe_cache:
-            self._describe_cache[obj] = obj.describe()
+            # Validate that there is no data key name collision.
+            data_keys = obj.describe()
+            keys = data_keys.keys()  # that is, field names
+            for known_obj, known_data_keys in self._describe_cache.items():
```
tacaswell: This is a _bit_ more restrictive than we need it to be, we should be able to use objects with conflicting key names in the same scan as long as we don't bundle them into the same event.  On the other hand, we will deal with that complication when we find a beamline that needs it.

I like that this is a one time check.

bluesky/run_engine.py
danielballan: done

bluesky/run_engine.py
danielballan: I agree.

danielballan:
@tacaswell I took

> On the other hand, we will deal with that complication when we find a beamline that needs it.

to mean, "Good enough." Satisfied with this?

PR #160, opened 2015-08-13
WIP: serving Documents over sockets

danielballan:
This allows any of our callbacks to run on a separate machine, receiving a stream of Documents via a socket.

This is very much a work in progress, shared prematurely so that @ericdill can take a look. It relies on [my python-3-compatible branch on jsonsocket](https://github.com/danielballan/jsonsocket/tree/py3-compat).


danielballan:
This is a reasonable MVP, but it doesn't look much more difficult to implemented using pyzmq. Note to self, start here: http://zeromq.github.io/pyzmq/api/zmq.html#zmq.Socket.send_json

danielballan:
Superceded by #211

PR #164, opened 2015-08-17
ENH: Support a Shutter interface with open and close.

danielballan:
Sending booleans to shutter PVs is confusing. ("Is 0 open or close?") Methods called `open` and `close` are not. The Shutter interface is expected to support `read()`, `describe()`, `open()`, `close()`. All of these methods take no arguments.

It might be useful to implement the properties `is_open` and `is_closed`. Those need not be known to the RunEngine, so they are not addressed in this PR.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -233,6 +233,8 @@ class RunEngine:
             'pause': self._pause,
             'collect': self._collect,
             'kickoff': self._kickoff,
+            'open': self._open,
+            'close': self._close,
             'logbook': self._logbook,
             'configure': self._configure,
             'deconfigure': self._deconfigure,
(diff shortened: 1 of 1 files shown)


tacaswell:
:+1: Holding off merging in hope that one of us will write a toy shutter on the train ride in tomorrow.

danielballan:
Alternate proposal by @dchabot that helps us avoid an ever-creeping number of interfaces:

```
shutter.set('open')
shutter.set('close)
```

:+1: from me -- any other reason to prefer `Shutter` that I am forgetting, @tacaswell? Specific methods are cleaner in a way, but I think that is outweighed by the advantage of limiting the number of bluesky interfaces.

tacaswell:
One advantage of the specialized interfaces is that we can statically (ish)
check the validity of open/close messages by checking if their objects have
the right methods.

klauer:
The specific point brought up in the original post is that "booleans are confusing". I'd argue that if this is the real concern, it should be solved on the EPICS layer with configuring a `bo` record (a binary output) to have the right string names for zero and one ('open'/'close'). You could then do a `caput(PV, "open")` instead. I'd be very surprised if this weren't the case in most instances already.

@dchabot makes a great point about the "ever-creeping number of interfaces". If you go down this path, you'll be adding a hundred different messages for all different types of devices. Devices also tend to have their own little differences, so applying a single interface to a group of devices doesn't usually work out so well.

If you want to say that "bluesky knows nothing about EPICS so your point is invalid", then fine. Make this whole thing configurable in a simple way -- allow the user to define all of the interfaces he or she desires so that you can "statically-ish" determine the validity of their messages. Call `RE.configure_interface('shutter', ['open', 'close'])` and it automatically adds the messages for it.

danielballan:
The other downside of special interfaces is that scan logic is not so easily transferrable between different experiments, especially the cycler-based stuff. I think there are other ways we could make interfaces verifiable, like adding validation methods.

```
shutter.validate({'shutter': 'open'})
```

tacaswell:
The details of the validation will need to be worked out, but I am pretty
sure that using Signature we can do a reasonable job introspecting the
object methods and checking that the payload of the `set` methods is valid
with out actually running call.

I have come around and am now also :-1: on adding this interface.

danielballan:
Signature is awesome.

PR #168, opened 2015-08-26
API: Change callbacks signature.

danielballan:


danielballan:
@klauer Update the signature of `print_scan_ids` in `50-scans.py` to `print_scan_ids(name, doc)` but do nothing with the name. :- )

tacaswell:
I suggest something like

```python
class HxnMDPrinter:
    def __init__(self):
        self._last_start = None
    def call(self, name, doc):
        if name == 'start':
           self._last_start = doc
        if name in ('start, 'stop'):
            print('blah blah ', self._last_start['scan_id'])) 
```

PR #170, opened 2015-08-26
BUG: callback signature changed with PR 168

tacaswell:


tacaswell:
Super seceded by #171

PR #172, opened 2015-08-27
API: Expose doc names as strings, not enum values.

danielballan:
Internally, we pass around document names (start, stop, event, descriptor) as values of an Enum, `DocumentTypes`. This is for strictness and probably a minor performance boost. But we should not expose this to the users via subscriptions; we should expose the string names.

### DIFF
--- a/bluesky/callbacks.py
+++ b/bluesky/callbacks.py
@@ -71,7 +71,7 @@ class CallbackCounter:
     # Wrap itertools.count in something we can use as a callback.
     def __init__(self):
         self.counter = count()
-        self(None)  # Start counting at 1.
+        self(None, 'start')  # Start counting at 1.
 
     def __call__(self, name, doc):
         self.value = next(self.counter)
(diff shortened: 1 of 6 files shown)


danielballan:
(This is a follow-up on #171, making the same fix somewhat differently.)

tacaswell:
This also needs to be fixed in line 998 of runengine for the _scan_cb

danielballan:
Heads up, @klauer, I am reverting back to ignoring exceptions raised by the callbacks. In production, we don't want callback errors to interrupt long scans or block users from running scans when no developers are available. But I added `RE.ignore_callback_exceptions`. Set it to `True` to debug callbacks.

danielballan:
P.S. to self, tests were failing at some point. Now all tests pass. We should get Travis running soon.

klauer:
@danielballan Thanks for the heads-up. Will also have to switch my subscriptions to compare against the strings.

tacaswell:
Sorry about this, we normally are more careful about this sort of thing :disappointed:

PR #173, opened 2015-08-27
ENH: Add stop method; stop things.

danielballan:
This keeps a list (strictly speaking a set) of all objects that we ever called `set` or `kickoff` on during a given call to the RunEngine. It calls `stop` on them all at the end, whether the call concludes successfully or with errors.

See forthcoming issue for thoughts on how this might be refined in the future. It looks like all the relevant ophyd objects have a `stop` method, so this is ready to try @klauer.


klauer:
This may be part of your forthcoming issue, but:
Will `collect` still get called for interrupted fly-scans that were `kickoff`'d? I'm not familiar enough with the code base as of yet to tell from the PR diff.

tacaswell:
Dan is currently working on implementing that in this PR.

danielballan:
Done.

tacaswell:
Looks good to me, should we hold of on merging this until it is tested at HXN?

klauer:
Great. I'll give it a go tomorrow and let you know the results.

klauer:
Seems to be working with a couple minor modifications from @tacaswell and me, see [here](https://github.com/klauer/bluesky/commits/stop).

klauer:
Working well enough in my tests.

... also, clicking merge so I can feel important.

PR #176, opened 2015-09-01
ENH: Post-run table + ScanBase open-run metadata

klauer:
A couple things from the hxn_work branch:

* Allow metadata in `open_run` (as we previously discussed)
* Splits LiveTable into TableBase + PostRunTable/LiveTable. 
    * LiveTables were unusable for fly scans due to all of the events coming in at once - rows were printed out of order at an excruciatingly slow rate (~1s/row).
    * Added option to not query filestore for data at all.
    * Added a mutex to handle multiple events coming in at once for all classes.

If the table stuff doesn't belong here, I'll stick it in hxntools.

### DIFF
--- a/bluesky/callbacks.py
+++ b/bluesky/callbacks.py
@@ -5,6 +5,8 @@ import sys
 from itertools import count
 import asyncio
 import warnings
+import operator
+import threading
 from prettytable import PrettyTable
 
 import matplotlib.backends.backend_qt5
@@ -205,7 +207,7 @@ class LivePlot(CallbackBase):
 
 
 def format_num(x, max_len=11, pre=5, post=5):
-    if (abs(x) > 10**pre or abs(x) < 10**-post) and x != 0:
+    if (abs(x) > 10 ** pre or abs(x) < 10 ** -post) and x != 0:
         x = '%.{}e'.format(post) % x
     else:
         x = '%{}.{}f'.format(pre, post) % x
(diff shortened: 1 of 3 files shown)


danielballan:
Without having read the code, I like those changes to LiveTable and I'm fine with them being in bluesky for now. We can discuss splitting off callbacks when they grow numerous or pick up weird dependencies.

Yes, let's allow metadata to be passed in the 'open_run' command, as you have written here. That allows some metadata to be determined after the call has begun, which will be especially useful for multi-run calls. But let's not process the `md` attribute by adding logic to `ScanBase`. Instead, let's make the `RunEngine` check for an `md` property, the same way it checks for a `subs` property. If that doesn't make sense we can talk about it in person when I get back. I'm OK if this gets merged in the meantime: the refactor I have in mind will not break the API.

bluesky/callbacks.py
```diff
@@ -392,6 +351,147 @@ def stop(self, stop_document):
...
+                    self.num_events_since_last_header = 0
+                self.num_events_since_last_header += 1
+            else:
+                # print the whole table
+                print(self.table)
+
+            sys.stdout.flush()
+
+
+class PostRunTable(TableBase):
```
ericdill: This is still going to be a lossy table. Does it make any sense for this to go to mds/fs and grab **all** the events and then format them into this table?

klauer: It definitely make sense to do it that way. A couple reasons why I chose this way as a first pass are:
* Re-using most of the `LiveTable` implementation made this easy
* "Collect data, close run, write to db, read from db, print to screen" is a rather inefficient way of doing things when you have most of the data (as you say, it can be lossy) already in memory. Perhaps once the caching layer is implemented on the db side doing this won't be such an issue.

klauer:
This doesn't make sense anymore, as mds has been sped up significantly. I also was previously unaware of the examples in `broker_callbacks.py` which will make this easier. Closing.

PR #177, opened 2015-09-04
ENH: mds bulk events integration

klauer:
I spent a bit of time to see how bulk events from `collect` could be passed around, and here's what I came up with:

* Added a new document name `"bulk_events"` 
    * It's a dictionary that's keyed on `descriptor_uid`, with lists of events as values
    * It's verified with jsonschema (thanks @arkilic for the input)
* Individual events are by default not emitted, unless `emit_events` is set in the `collect` message (as in `Msg('collect', flyer, emit_events=True)`)

Feel free to reject it outright - I honestly have no clue how you intended to implement this. 

In any case, it works on my end and allows me to run tests. The speed-up is significant: 5000 events are stored in 3.5 seconds, giving an estimate of 45 seconds for our larger 65k event scans. I don't think that's an unreasonable number.

### DIFF
--- a/bluesky/callbacks.py
+++ b/bluesky/callbacks.py
@@ -54,16 +54,19 @@ class CallbackBase(object):
         return getattr(self, name)(doc)
 
     def event(self, doc):
-        logger.debug("CallbackBase: I'm an event with doc = {}".format(doc))
+        logger.debug("CallbackBase: I'm an event with doc = %r", doc)
+
+    def bulk_events(self, doc):
+        logger.debug("CallbackBase: I'm an event with a big doc")
 
     def descriptor(self, doc):
-        logger.debug("CallbackBase: I'm a descriptor with doc = {}".format(doc))
+        logger.debug("CallbackBase: I'm a descriptor with doc = %r", doc)
 
     def start(self, doc):
-        logger.debug("CallbackBase: I'm a start with doc = {}".format(doc))
+        logger.debug("CallbackBase: I'm a start with doc = %r", doc)
 
     def stop(self, doc):
-        logger.debug("CallbackBase: I'm a stop with doc = {}".format(doc))
+        logger.debug("CallbackBase: I'm a stop with doc = %r", doc)
 
 
 class CallbackCounter:
--- a/bluesky/register_mds.py
+++ b/bluesky/register_mds.py
@@ -20,7 +20,7 @@ known_run_start_keys = ['time', 'scan_id', 'beamline_id', 'uid', 'owner',
                         'group', 'project']
 
 
-def _insert_runstart(name, doc):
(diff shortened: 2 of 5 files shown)


tacaswell:
This seems more-or-less reasonable, but that still seems slow.  Is that time just the document emission or the full scan?

I was trying to sort out a way to do the bulk insertions without having to keep a list of them (ie, pass a generator straight out of the flyer into the bulk insert call), but that has issues with registering the callbacks as generators are not rewindable.  This is a much better first pass.

Allowing both bulk and single event emission could be problematic as it might hit mds twice.

tacaswell:
Sorry for you having to deal with the API breakage in MDS.

klauer:
The time I mentioned was just for the `register_mds._insert_bulk_events` call to complete. You have been testing with mongodb on localhost, right? In this case, bluesky is running on a workstation and saving on the channel archiver server. It's also with an already-populated database. I can't say for certain either of those are the cause, of course, but they are worth noting.

You're definitely right about the data hitting mds twice. It should be removed. My line of reasoning for it was that some callback handlers might not want to differentiate between bulk and single events. What do you think of `CallbackBase` turning a `bulk_events` document into regular `event` documents (if not handled by the subclass)? ... Actually, that's also not a great approach as it turns a blocking `bulk_events` into a bunch of blocking `event` documents which really should be non-blocking. I guess the idea should be scrapped altogether.

tacaswell:
We should check which version of bson is getting imported.  Testing this between two computers I found that on my laptop the version of bson shipped with mongoengine (?!) ends up being deeply recursive where as the version installed at the 'top' level is not (which saves 2-3s on 10k events for me).

Doing 10k inserts over wi-fi takes ~4-7s and the bulk of that time is in waiting on network traffic.  I really hope that we have better network between your workstations and the CA than my consumer-grade wi-fi router at home.

klauer:
With a bit of profiling, I noticed that pymongo was using its legacy wire protocol when connecting to the server. As it turns out, the old mongodb version from Debian wheezy (2.0.6-1.1) is at least partly to blame for the slower bulk insertion speed. 

A temporary installation of mongodb 3.0.6, with a copy of the previous database as a starting point, shows that a bulk_event comprised of 65k events should be stored in about 15.6 seconds (compared to 45 seconds above). The estimate did not appear to change when using different storage engines ("wiredtiger").

@tacaswell also verified the bson library version is correct. Profiling showed that bson calls aren't contributing significant overhead.

danielballan:
I'm on board with this. @klauer, merge at will, but first merge @tacaswell's new PR into this branch, basically undoing the MDS API changes. Apologies for the flailing.

klauer:
@danielballan Done - merged and merged.

PR #179, opened 2015-09-10
Look for metadata in md attribute of plan passed to RE

danielballan:
Like subscriptions, metadata can come from:

* kwargs to `RE.__call__` for metadata
* a special attribute of the plan passed to `RE.__call__`
* a Msg, wherein metadata can be determined after the call has begun

Unlike subscriptions, we don't need a new type of command to pass in metadata in a Msg. It is always good enough to put it in the 'open_run' Msg: we can't do it any later than that (it has to get into the RunStart) and it doesn't help to set it any earlier than that.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -495,7 +495,10 @@ class RunEngine:
                 scan_args[field] = repr(getattr(plan, field))
             metadata['scan_args'] = scan_args
 
-        self._metadata_per_call = metadata
+        if hasattr(plan, 'md'):
+            self._metadata_per_call.update(plan.md)
+        # If kwargs to __call__ collide with plan.md, kwargs win.
+        self._metadata_per_call.update(metadata)
 
         self.state = 'running'
         gen = iter(plan)  # no-op on generators; needed for classes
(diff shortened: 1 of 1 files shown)


danielballan:
for @klauer

ericdill:
looks fine

bluesky/run_engine.py
```diff
@@ -464,9 +464,9 @@ def __call__(self, plan, subs=None, **metadata):
 
         # Register temporary subscriptions. Save tokens to unsubscribe later.
         subs = normalize_subs_input(subs)
-        try:
+        if hasattr(plan, 'subs'):
```
tacaswell: Why `hasattr` vs `try...except`?

danielballan: This part of the code does not have to be high-perf, and this pattern avoids accidentally snarfing AttributeErrors that might bubble up from `normalize_subs_input` or something. The right way to do this is to put that in else, but at that point the code just seems obfuscated. I like this better.

danielballan: ...But if this is annoying to you I'll reverse it in both cases. Readability should be the priority here.

tacaswell: fair enough, just checking if you had a reason

klauer:
Confirmed this works, finally (sorry).

tacaswell:
@klauer See https://github.com/NSLS-II/filestore/pull/36 for (at least my) typical level of promptness.

danielballan:
@tacaswell @klauer Thanks for testing. Somebody hit the button?

PR #180, opened 2015-09-10
DOC: Update docstrings.

danielballan:
A follow-up to #177 with some docstring fixes I didn't catch in the review.


danielballan:
cc @klauer (are you getting emails about all issues or just the ones I tag you in?)

klauer:
Oops. Looks good to me.

(I'm now watching so you shouldn't need to tag me each time.)

PR #181, opened 2015-09-11
Make MDS tests skippable if MDS is not installed

danielballan:


bluesky/tests/test_vertical_integration.py
```diff
@@ -1,17 +1,33 @@
...
 
 def test_scan_and_get_data():
+    try:
+        import metadatastore
+    except ImportError:
+        raise SkipTest
+    try:
+        from dataportal import DataBroker as db
+    except ImportError:
+        raise SkipTest
```
ericdill: Why not combine these two try/except blocks?

bluesky/tests/test_vertical_integration.py
```diff
@@ -1,17 +1,33 @@
...
+from nose import SkipTest
 from bluesky.examples import stepscan, det, motor
-from bluesky.standard_config import gs
 
 def setup():
-    mds_setup()
+    try:
+        from metadatastore.utils.testing import mds_setup
+    except ImportError:
+        pass  # test will be skipped
```
ericdill: in-line comment is incorrect

bluesky/hardware_checklist.py
```diff
@@ -17,10 +16,12 @@
 
 
 def connect_mds_mongodb():
+    import metadatastore.conf
     _connect_mongodb('metadatastore', metadatastore.conf.connection_config)
 
 
 def connect_fs_mongodb():
+    import metadatastore.conf
```
ericdill: this should be `import filestore.conf`

bluesky/hardware_checklist.py
```diff
@@ -5,7 +5,6 @@
 import logging
 from collections import namedtuple
 import requests
-import metadatastore.conf
 import filestore.conf
```
ericdill: I imagine you want to remove this line too

bluesky/tests/test_vertical_integration.py
danielballan: How?

bluesky/tests/test_vertical_integration.py
danielballan: could make more informative errors later, individualized

ericdill:
Do you want to follow the same pattern and make pyolog optional in hardware_checklist too?

bluesky/tests/test_vertical_integration.py
ericdill: ok

bluesky/tests/test_vertical_integration.py
ericdill: I guess it is not incorrect if enough context is placed around the in-line comment. I initially assumed that this was just a copy/paste error from the `test_scan_and_get_data` function

danielballan:
Good point about pyolog. Added a commit.

ericdill:
looks good to me

ericdill:
tests are passing locally?

PR #183, opened 2015-09-11
REF: Do not import lmfit at the top of scans.

danielballan:
lmfit raises on import in IPython 4.0. Since it's only used in one scan (`Center`) let's import it in the `__init__`. I don't usually like this pattern, but this is one time I think it's justified. I'll fix this on the lmfit side shortly.


ericdill:
Seems sensible enough

PR #184, opened 2015-09-11
Add some deps to setup.py, meta.yaml

danielballan:


setup.py
```diff
@@ -20,7 +20,7 @@
     packages=['bluesky', 'bluesky.tests', 'bluesky.testing'],
     package_data={'bluesky': ['schema/*.json']},
     requires=['jsonschema', 'traitlets', 'prettytable', 'cycler', 'history',
```
ericdill: isn't this supposed to be `install_requires`?

conda-recipe/meta.yaml
```diff
@@ -24,6 +24,8 @@ requirements:
     - filestore
     - matplotlib
     - prettytable
+    - cycler
+    - traitlets
```
ericdill: do we have both of these on our internal anaconda server?

setup.py
ericdill: This might cause a merge conflict with #181

conda-recipe/meta.yaml
ericdill: I only ask because I think the plan is to change `.condarc` on each beamline so that they point to `anaconda`, which hits the internal server only, not `defaults`, which hits the public server too.  If we don't have these two deps on our internal server then the install will fail (yeah i guess that's obvious, sorry for the stupid comment)

danielballan: Yeah, the internal server is way behind.

PR #185, opened 2015-09-14
Doc edits

danielballan:


danielballan:
self-merging; only contents of doc/ are touched

PR #186, opened 2015-09-14
ENH: Add PeakStats callback

danielballan:
This adds a class for computing peak stats. It works like this:

```
ps = PeakStats(motor, det)
RE(ascan, subs=ps)  # or, in the simple SPEC-like API, ascan(motor, start, stop, interval, subs=ps)
```

When the scan finishes, the statistics are automatically computed, and the results are available via attributes of `ps`. So far, center-of-mass (`ps.com`) and min and max (`ps.min`, `ps.max`) are implemented. I need guidance from @cmazzoli or @ambarb on exactly how `cen` is supposed to work.

Since the pattern, "compute some things when a scan finishes" is likely to come up again, I implemented `PeakStats` on a base class `CollectThenCompute` which is more general. I'm interested in thoughts from non-CSX-ers on whether this is the right generalization. Notice that this does not use databroker; we have the `post_run` callback for that.

### DIFF
--- a/bluesky/callbacks.py
+++ b/bluesky/callbacks.py
@@ -3,6 +3,7 @@ Useful callbacks for the Run Engine
 """
 import sys
 from itertools import count
+from collections import deque
 import warnings
 from prettytable import PrettyTable
 
@@ -383,3 +384,34 @@ def _get_obj_fields(fields):
                                  "'describe' method that return a dict.")
             string_fields.extend(field_list)
     return string_fields
+
+
+class CollectThenCompute(CallbackBase):
+
+    def __init__(self):
+        self._start_doc = None
+        self._stop_doc = None
+        self._events = deque()
+        self._descriptors = deque()
+
+    def start(self, doc):
+        self._start_doc = doc
+
+    def descriptor(self, doc):
+        self._descriptors.append(doc)
+
+    def event(self, doc):
+        self._events.append(event)
+
+    def stop(self, doc):
+        self._stop_doc = doc
+        self.compute()
+
+    def reset(self):
+        self._start_doc = None
+        self._stop_doc = None
+        self._events.clear()
+        self._descriptors.clear()
+
+    def compute(self):
+        raise NotImplementedError("This method must be defined by a subclass.")
--- a/bluesky/callbacks.py
+++ b/bluesky/callbacks.py
@@ -401,7 +401,7 @@ class CollectThenCompute(CallbackBase):
         self._descriptors.append(doc)
 
     def event(self, doc):
-        self._events.append(event)
+        self._events.append(doc)
 
     def stop(self, doc):
         self._stop_doc = doc
--- a/bluesky/scientific_callbacks.py
+++ b/bluesky/scientific_callbacks.py
@@ -61,12 +61,12 @@ class PeakStats(CollectThenCompute):
                 y.append(_y)
         x = np.array(x)
         y = np.array(y)
+        self.x_data = x
+        self.y_data = y
         # Compute x value at min and max of y
         self.max = x[np.argmax(y)]
         self.min = x[np.argmin(y)]
-        self.com = np.interp(center_of_mass(y), x, y)
-        self.x_data = x
-        self.y_data = y
+        self.com = np.interp(center_of_mass(y), np.arange(len(x)), x)
 
 
 def plot_peak_stats(peak_stats, ax=None):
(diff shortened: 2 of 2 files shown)


danielballan:
I forgot to mention there is also plotting code in here. We can work at individual beamlines to get the right kind of automation, maybe involving automatic plotting of the various peak stats in some circumstances.

bluesky/callbacks.py
```diff
@@ -413,3 +414,62 @@ def _get_obj_fields(fields):
...
+        self._start_doc = None
+        self._stop_doc = None
+        self._events.clear()
+        self._descriptors.clear()
+
+    def compute(self):
+        raise NotImplementedError("This method must be defined by a subclass.")
+
+
+class PeakStats(CollectThenCompute):
```
tacaswell: This code is in here twice

bluesky/scientific_callbacks.py
```diff
@@ -0,0 +1,83 @@
...
+        dictionary of matplotlib Artist objects, for further styling
+    """
+    ps = peak_stats  # for brevity
+    if ax is None:
+        fig, ax = plt.subplots()
+    # Plot points, vertical lines, and a legend. Collect Artist objs to return.
+    points, = ax.plot(ps.x_data, ps.y_data, 'o')
+    vlines = []
+    for attr in ['cen', 'com', 'max', 'min']:
+        vlines.append(ax.axvline(getattr(ps, attr), label=attr))
```
tacaswell: in things that make me happy, if we make the `peak_stats` object support `__getitem__` and we upgrade to mpl 1.5 this can read as `ax.axvline(attr, data=peak_stats)`

bluesky/callbacks.py
danielballan: Read the copy in scientific_callbacks.py. I'll delete this in a force-push in a moment.

bluesky/scientific_callbacks.py
```diff
@@ -0,0 +1,83 @@
...
+                pass
+            try:
+                y.append(event['data'][self.y])
+            except KeyError:
+                pass
+        x = np.array(x)
+        y = np.array(y)
+        # Compute x value at min and max of y
+        self.min = x[np.argmax(y)]
+        self.max = x[np.argmin(y)]
```
tacaswell: I think these two lines are swapped

bluesky/scientific_callbacks.py
```diff
@@ -0,0 +1,83 @@
...
+        self.max = None
+        self.min = None
+        super().__init__()
+
+    def compute(self):
+        "This method is called at run-stop time by the superclass."
+        x = []
+        y = []
+        for event in self._events:
+            try:
```
tacaswell: I would do this as

```python
try:
    _x = event['data'][self.x]
   _y = event['data'][self.y]
except KeyError:
    pass
else:
    x.append(_x)
    y.append(_y)
```

so that they never get out of sync

klauer:
Does it make sense to do scientific calculations on potentially lossy data? If I'm not mistaken, using this with `RE(ascan, subs=ps)` can miss `Event` documents -- however unlikely that may be.

Also pinging @skalbfleisch to get his input on `cen()`.

bluesky/scientific_callbacks.py
danielballan: There is no reason that x and y have to be in the same event stream. As long as those event streams have the same length, we want this to work.

tacaswell: if they are in different event streams then they are asynchronous and matching them up by sequence number is scary.

danielballan:
@klauer This implementation avoids mixing the scientific logic with any assumptions about where the Documents come from. It's up to users to choose whether to subscribe to a lossy live Event stream, subscribe to a lossless live stream (as MDS does), or pull a lossless stream out of the broker at the end of the scan like so:

```
# Convert any live callback into a broker-fed callback, kicked off by the live RunStop.
from bluesky.broker_callbacks import post_run
RE(ascan, subs=post_run(ps))
```

Granted, you could make a case that we flatly shouldn't allow users to apply scientific logic to a lossy stream. But given how easy it is to attach a generic Document-consumer to any kind of source, I'd vote for flexibility (and the potential for users to shoot themselves in the foot).

bluesky/scientific_callbacks.py
danielballan: Fair point. Revised.

danielballan:
Can anyone chime in on how `cen` is supposed to be computed? @skalbfleisch @arkilic @cmazzoli?

Even without `cen`, I'd like to see this merged soon so we can test the workflow ASAP. It's easy to add more stats later. Here's an example of it working:

peak-stats

And a notebook: https://gist.github.com/anonymous/8b4d3826f05ad95514fd

PR #192 will make it possible to do this with less typing. :- )

PR #190, opened 2015-09-22
REF: Put hardware-checklist deps in function-level imports.

danielballan:
The hardware_checklist.py will soon move outside the package. We definitely don't want to pick up dependency on epics, requests, and pymongo just for these checks, so I'm making them function-level imports for now.


tacaswell:
There is something fishy with this PR

tacaswell:
have a look at gitk, somehow something got rebased really badly.

tacaswell:
gitk_bad

PR #192, opened 2015-09-24
Simple (SPEC-like) scans are more customizable

danielballan:
In current master, the subscriptions attached to the simple scans are hard-coded. Customizing them is tricky because they inspect the scan to automatically do things like use `motor` attribute as the x axis in `LivePlot`.

This introduces two new user-modifiable attributes: `.subs`, which works the same way as usual, and `.sub_factories`, which expects a *function that accepts a scan instance and returns a callback*. It's the job of the function to inspect the scan and pass the right arguments to, for example, `LivePlot`. A collection of these "sub factories" is included in the PR. See for example `plot_first_motor`.

I don't imagine that users will be writing very many "sub factories," but they are crucial to providing the magical scan inspection in a way that lets users add and remove other subs (like peak stats) at will.

### DIFF
--- a/bluesky/simple_scans.py
+++ b/bluesky/simple_scans.py
@@ -26,14 +26,51 @@ from bluesky import scans
 from bluesky.callbacks import LiveTable, LivePlot
 from boltons.iterutils import chunked
 from bluesky.standard_config import gs
+from bluesky.utils import normalize_subs_input
 
 
-class _PrimitiveScan:
+### Factory functions acting a shim between scans and callbacks ###
+
+
+def table_from_motors(scan):
+    "Setup a LiveTable by inspecting a scan and gs."
+    # > 1 motor
+    return LiveTable(scan.motors + gs.TABLE_COLS)
+
+
+def table_from_motor(scan):
+    "Setup a LiveTable by inspecting a scan and gs."
+    # 1 motor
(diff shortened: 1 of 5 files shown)


danielballan:
P.S. Tests pass. This is ready once it survives review.

tacaswell:
bike shedding the name, `_BoundScan` might be better?

klauer:
:+1: for sub-factories (I've been doing the same in a less-than-ideal way...)

bluesky/simple_scans.py
```diff
@@ -47,12 +84,18 @@ def __call__(self, *args, subs=None, **kwargs):
...
 
-        self.scan = self._scan_class(gs.DETS, *args, **scan_kwargs)
-        if subs is None:
-            subs = self.subs
+        self.scan = self.scan_class(gs.DETS, *args, **scan_kwargs)
+        # Combine subs passed as args and subs set up in subs attribute.
+        _subs = normalize_subs_input(subs)
+        _subs.update(normalize_subs_input(self.subs))
+        # Create a sub from each sub_factory.
+        _subs.update({k: [sf(scan) for sf in v] for k, v in
```
klauer: `scan` -> `self.scan`, this line and next

danielballan: Good catch. Fixed and force-pushed.

PR #197, opened 2015-09-25
FIX: SUB FACTORIES

klauer:
Sorry for the caps -- well, not really. This one line bug caused way more confusion than I'd like to admit.


danielballan:
All caps was warranted. Sorry the sub_factories PR got merged in a rush. The demo on Monday has us off our game over here.

PR #198, opened 2015-09-25
ENH: add cen methodology

tacaswell:


tacaswell:
Self-merging because I need this for the demo.

PR #200, opened 2015-09-26
Enh ps stuff

tacaswell:


bluesky/simple_scans.py
```diff
@@ -66,8 +67,8 @@ class _BundledScan:
 
     def __init__(self):
         # subs and sub_factories can be set individually per instance
-        self.subs = self.default_subs
-        self.sub_factories = self.default_sub_factories
+        self.subs = list(self.default_subs)
```
danielballan: I'd prefer `.copy()` so we can allow document-specific subscriptions like `self.default_subs = {'start': cb}`

tacaswell: I really don't like that these things can be dicts or lists.  We should to settle on all subscriptions need to be able to take everything (and gracefully ignore those they can't handle) and this is a list or you must always be specific about these being dicts.

danielballan: Sigh. We had this conversation really carefully because I know it runs again your general taste. At the time you said it made sense to accept users' "Do what I mean!" code at this layer. There is no ambiguity.

danielballan: As a reminder: the callbacks that don't need Events really shouldn't be called on event Event, which is why we accept dicts. On the other hand, forcing users to understand the Document Model just to add a table seems unnecessary. Since our built-in callbacks can safely accept every document, we can afford to be flexible and guess.

tacaswell: There are internal data structures to convenience classes, if we are going to do DWIM here, then it needs to happen a layer up and be stored in the attributes normalized.

If this is supposed to be user facing in this case it is even worse as now we are exposing either, a dict _or_ a list to the user, how are they supposed to know what method to use to add to it?

danielballan: DWIM is always safe in the args to the Run Engine or the simple scans.

I appreciate the downside of doing DWIM in the attributes. We could do the normalization in the setter. That means pushing it into ScanBase, which I like to keep as light as possible, but it might be the least evil thing to do.

danielballan:
The `Subs` descriptor for validation/normalization makes me happy. :- )

PR #205, opened 2015-09-28
Make overplotting configurable through global switch.

danielballan:
`gs.OVERPLOT`, True by default

### DIFF
--- a/bluesky/simple_scans.py
+++ b/bluesky/simple_scans.py
@@ -241,7 +241,7 @@ class AbsScan(_StepScan):
 
 class OuterProductAbsScan(_OuterProductScan):
     "mesh"
-    default_sub_factories = {'all': [raster]}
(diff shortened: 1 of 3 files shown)


danielballan:
P.S. also contains two unrelated minor fixes -- see commit msgs

bluesky/simple_scans.py
```diff
@@ -54,12 +54,18 @@ def table_gs_only(scan):
 
 def plot_first_motor(scan):
     "Setup a LivePlot by inspecting a scan and gs."
-    return LivePlot(gs.PLOT_Y, scan.motors[0]._name)
+    fig = None
+    if gs.OVERPLOT:
```
tacaswell: This logic is inverted

danielballan: fixed and force-pushed

PR #207, opened 2015-09-28
FIX: only wait after starting all moves

tacaswell:
Also only open the event bundle once.


danielballan:
For posterity / collaborators following along, the symptom of this is that we were failing to move motors concurrently in InnerProduct scans.

danielballan:
Note: tested successfully on real motors at CSX

PR #210, opened 2015-09-28
Update recipe to python >=3.4.3

ericdill:


ericdill:
This was a bad idea and causes conda-build to fail.  Fixing before I tag v0.2.4.

PR #217, opened 2015-09-30
API/MNT: make qt_kicker not install on import

### LINKED ISSUE #214: Fix qt kicker to work smarter
@tacaswell

tacaswell:
There is now a function install_qt_kicker in qt_kicker which must be
called to install the kicker.

THIS WILL REQUIRE A CHANGE TO THE BEAMLINE CONFIG

The suggested pattern is:

from bluesky.qt_kicker import install_qt_kicker
install_qt_kicker()

closes #214

### DIFF
--- a/bluesky/qt_kicker.py
+++ b/bluesky/qt_kicker.py
@@ -1,35 +1,45 @@
-"""
-A utility that, once imported, kicks the qt event loop at regular
-intervals. This lets qt and asyncio peacefully coexist, and it is 
-necessary for matplotlib < 1.5.0.
-"""
-import asyncio
-import matplotlib.backends.backend_qt5
-from matplotlib.backends.backend_qt5 import _create_qApp
+import sys
+
+_INSTALLED = None
 
 
-_create_qApp()
-qApp = matplotlib.backends.backend_qt5.qApp
+def install_qt_kicker():
+    """Install a periodic callback to integrate qt and asyncio event loops
 
-from matplotlib._pylab_helpers import Gcf
+    If a version of the qt bindings are not already imported, this function
+    will do nothing.
 
-try:
-    _draw_all = Gcf.draw_all  # mpl version >= 1.5
-except AttributeError:
-    # slower, but backward-compatible
-    def _draw_all():
-        for f_mgr in Gcf.get_all_fig_managers():
-            f_mgr.canvas.draw_idle()
+    It is safe to call this function multiple times.
+    """
 
+    global _INSTALLED
+    if _INSTALLED is not None:
+        return
+    if not any(p in sys.modules for p in ['PyQt4', 'pyside', 'PyQt5']):
+        return
+    import asyncio
+    import matplotlib.backends.backend_qt5
+    from matplotlib.backends.backend_qt5 import _create_qApp
+    from matplotlib._pylab_helpers import Gcf
 
-def _qt_kicker():
-    # The RunEngine Event Loop interferes with the qt event loop. Here we
-    # kick it to keep it going.
-    _draw_all()
+    _create_qApp()
+    qApp = matplotlib.backends.backend_qt5.qApp
 
-    qApp.processEvents()
-    loop.call_later(0.1, _qt_kicker)
+    try:
+        _draw_all = Gcf.draw_all  # mpl version >= 1.5
+    except AttributeError:
+        # slower, but backward-compatible
+        def _draw_all():
+            for f_mgr in Gcf.get_all_fig_managers():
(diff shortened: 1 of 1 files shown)


klauer:
:+1: much prefer the explicit call

bluesky/qt_kicker.py
```diff
@@ -1,35 +1,47 @@
...
+    """
 
-def _qt_kicker():
-    # The RunEngine Event Loop interferes with the qt event loop. Here we
-    # kick it to keep it going.
-    _draw_all()
+    global _INSTALLED
+    if _INSTALLED is not None:
+        return
+    print("HERE")
```
klauer: `HERE`

bluesky/qt_kicker.py
```diff
@@ -1,35 +1,47 @@
...
+    if not any(p in sys.modules for p in ['PyQt4', 'pyside', 'PyQt5']):
+        return
+    print("ALSO HERE")
```
klauer: `ALSO HERE`

bluesky/qt_kicker.py
```diff
@@ -1,35 +1,47 @@
...
 
-loop = asyncio.get_event_loop()
-loop.call_soon(_qt_kicker)
+    def _qt_kicker():
+        # The RunEngine Event Loop interferes with the qt event loop. Here we
+        # kick it to keep it going.
+        _draw_all()
+
+        qApp.processEvents()
+        loop.call_later(0.03, _qt_kicker)
```
ericdill: why change the time from 0.1 to 0.03?

bluesky/qt_kicker.py
```diff
@@ -1,35 +1,47 @@
...
 
-    qApp.processEvents()
-    loop.call_later(0.1, _qt_kicker)
+    _create_qApp()
+    qApp = matplotlib.backends.backend_qt5.qApp
 
+    try:
+        _draw_all = Gcf.draw_all  # mpl version >= 1.5
+    except AttributeError:
+        # slower, but backward-compatible
```
ericdill: Do you just want to pin bluesky to matplotlib >= 1.5?

ericdill:
Previously the qt-kicker loop started on import. Now you it does not start until you explicitly tell it to. Why the change in behavior?  I am not advocating for one over the other but I would like to understand what the rationale for the change is

bluesky/qt_kicker.py
klauer: Does it improve the responsiveness at all? Not so great currently.

tacaswell: Yes, this it should

bluesky/qt_kicker.py
tacaswell: debugging print statements!

tacaswell:
Because registering something like this on import as a bit surprising.

ericdill:
Cool. Thanks for the explanation.  Makes sense to change the behavior so that you have to explicitly activate the kicker

bluesky/qt_kicker.py
danielballan: Not until 1.5 is tagged at least. What's the advantage?

ericdill: We've pinned to matplotlib >=1.5rc1 in our collections environment already. Seems like we're committing to mpl >= 1.5. As such this try/except feels like unnecessary extra code

tacaswell:
Removed debugging statements and try/except logic

bluesky/qt_kicker.py
ericdill: Sorry for not making this clear (and doubly sorry if this is obvious) but I am not considering this to be a blocking issue. 

I can now more clearly articulate where my questioning of this came from:

**A.** Not having even realized that we might need to write backwards compatible matplotlib code 
**B.** As a result of **A**, me starting to think about in which projects it might be important to write backwards compatible matplotlib code.

tacaswell: none of them :imp:

ericdill: I could see a reasonable argument for keeping plotting code in `xray_vision` backwards compatible as that is the project that seems most likely to pick up external collaborators the soonest and is a plotting library.  I can't think of any other library where we would need to particularly care (at least right now) since we ~~control the world!~~ are the only users of the rest of the stack

danielballan:
To quote my comment above:

> Not until 1.5 is tagged at least.

@tacaswell Why remove the code that let this work on < 1.5? If you want to assume >= 1.5 in some callbacks, that's fine with me, but I don't see the advantage in making this utility less useful.

PR #219, opened 2015-10-02
Configuration

danielballan:
This PR adds an `objects` property to `ScanBase`. It is a dict mapping the ophyd objects to their data keys. It addresses two issues at once:

1. Downstream consumers can get a list of the objects involved without needing to know a scan's specific layout (motor or motors plural).
2. They can know which objects are associated with which data keys. This information was previously not knowable.

This PR also pushes all configure/deconfigure logic into ScanBase. The scan's `num` and `shape` attributes, if present, are passed. Users can pass custom configuration logic through the `.configuration` attribute.

This PR does not support a configuration kwarg. I'd like to postpone that until we need it.


danielballan:
P.S. Do not merge. I want to add tests first. But please let me know if you think this is off track.

danielballan:
Lots of stuff in this one PR now.

All OO scans and SPEC-like scans now have:
* a `flyers` attribute, a list of objects to kickoff at the beginning of the scan and collect at the end. This can be used to monitor things concurrent with any scan.
* a `configuration` attribute, as discussed. It is a dict mapping objects to configuration that will be passed to their `configure` methods. Everything is the `md` property (e.g., `shape`) is also included.

All OO scans additionally have:
* an `objects` property, mapping each object used by the scan to its data keys
* an `md` (metadata) property, a dict with all the scan's fields. It includes:
    * raw fields like `'detectors'`
    * derived fields like `'shape'`
    * `'objects'`, the above property but passed through `repr`
    * `'scan_args'`, a dict of valid kwargs that can fully recreate the scan object and therefore the scan command; this is an intentionally redundant subset of the above

The SPEC-like scans still create `md` and pass it to the RunEngnine; it's just not possible to expose it to the user as a property. Not a big deal.

This PR removes a lot code from the RunEngine, notably making metadata both simpler and more user-configurable. It maintains better separation between the RunEngine and ScanBase.

bluesky/tests/test_callbacks.py
```diff
@@ -57,69 +57,25 @@ def cb():
...
     def cb1(name, doc):
-        called['cb1'] = True
+        pass
+
     def cb2(name, doc):
-        called['cb2'] = True
+        pass
+
     def cb3(name, doc):
-        called['cb3'] = True
```
danielballan: These tests play badly with OSX for some reason. @tacaswell suggested a better way to test this aspect anyway.

bluesky/run_engine.py
```diff
@@ -154,6 +152,10 @@ def __init__(self, md=None, logbook=None):
             history.History, a simple interface to a sqlite file. Any object
             supporting `__getitem__`, `__setitem__`, and `clear` will work.
 
+        md_validator : callable, optional
+            a function that raises and prevents starting a run if it deems
```
tacaswell: can you provide the signature and expectation that it raises on failure?

bluesky/run_engine.py
```diff
@@ -1009,8 +995,8 @@ def _configure(self, msg):
         _, obj, args, kwargs = msg
         if not hasattr(obj, 'configure'):
             return None
-        result = obj.configure(*args, **kwargs)
-        self._configured.append(obj)
+        self._configured.add(obj)  # add first in case of failure below
+        result = obj.configure(kwargs.get('state'))
```
tacaswell: don't need the `result` local anynare

bluesky/run_engine.py
```diff
@@ -746,28 +734,26 @@ def _open_run(self, msg):
         self._clear_run_cache()
         self._run_start_uid = new_uid()
         self._run_start_uids.append(self._run_start_uid)
+        self.increment_scan_id()
```
tacaswell: does this have sideeffects now?

bluesky/tests/test_simple_api.py
```diff
@@ -18,6 +18,7 @@ def test_basic_usage():
     gs.TEMP_CONTROLLER = motor
     gs.TH_MOTOR = motor1
     gs.TTH_MOTOR = motor2
+    gs.RE.md['owner'] = 'whatever'
```
tacaswell: our default owner should be a bit less snarky

bluesky/run_engine.py
danielballan: It always has. That is needlessly obfuscated, isn't it? It's only a four-line method. I will delete the method and put its code inline here.

PR #223, opened 2015-10-07
Tst add travis

tacaswell:


danielballan:
Attn @dchabot this .travis.yml file with epics will be a useful starting point for doing the same to ophyd.

PR #224, opened 2015-10-07
Fix cb exceptions and tests; increase metadata test coverage

danielballan:
This is a bit of a hodgepodge -- see commit msgs for details.

Note that `RE.ignore_callback_exceptions`, a boolean flag, has never actually worked until now. Sorry, @klauer.

### DIFF
--- a/bluesky/broker_callbacks.py
+++ b/bluesky/broker_callbacks.py
@@ -69,6 +69,8 @@ def post_run(callback):
     +------------+-------------------+----------------+----------------+
     """
     def f(name, stop_doc):
+        if name != 'stop':
+            return
         uid = stop_doc['run_start']
         start = run_start_given_uid(uid)
         descriptors = descriptors_by_start(uid)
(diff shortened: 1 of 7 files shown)


klauer:
Thought I had been using it incorrectly or something - I had `catch_exceptions` decorators in place of that strewn about to help debug...

danielballan:
We had a test covering it, but the test was wrong. Tricky!

danielballan:
The `test_post_run` test has always been broken, but we're only now seeing it because the RunEngine/Dispatcher are now propagating callback errors.

It's still broken, but the problem is *definitely* confined to `post_run` specifically and *possibly* just a problem with the test code itself.

I'd like to get this merged for all the API changes and address the pre-existing issue with `test_post_run` separately. I marked the test a Known Failure in the latest commit.

klauer:
A+, would click merge again. Thanks, @danielballan.

Issue #225, opened 2015-10-08
Update scan documentation


danielballan:
All minor, just noting so these don't get forgotten:

- [x] DeltaListScan fails if the input is a list; it requires a numpy array
- [x] The API documentation for `OuterProductAbsScan` and probably others renders badly
- [x] The API documentation for `InnerProduct*Scan` seems to be missing.
- [ ] It might be worth implementing `*Product*ListScan`. One could just use `ScanND` and cyclers, but it's easy enough to provide a convenience class for this.

cc @vivekthampy

danielballan:
Punting on the last point. I think it is a good "gateway" to the more powerful ScanND way of doing things. @vivekthampy has been using ScanND successfully. See #258 for the docs fixes.

PR #229, opened 2015-10-14
FIX/TST: DeltaListScan should accept a list, not only an ndarray.

danielballan:
Straightforward fix to one issue in #225

### DIFF
--- a/bluesky/scans.py
+++ b/bluesky/scans.py
@@ -226,7 +226,7 @@ class DeltaListScan(Scan1D):
             raise NotImplementedError("Can't DScan this motor")
         key, = ret.keys()
         self.init_pos = ret[key]['value']
-        self._abs_steps = self.steps + self.init_pos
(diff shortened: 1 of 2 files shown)


danielballan:
cc @vivekthampy -- this is one of the problems I worked around when we built a higher-dimensional scan

bluesky/scans.py
```diff
@@ -226,7 +226,7 @@ def _pre_scan(self):
             raise NotImplementedError("Can't DScan this motor")
         key, = ret.keys()
         self.init_pos = ret[key]['value']
-        self._abs_steps = self.steps + self.init_pos
+        self._abs_steps = np.asarray(self.steps) + self.init_pos
```
ericdill: is `self.init_pos` guaranteed to be an int/float?

ericdill: Ok, dan says "yes".

ericdill:
LGTM

PR #233, opened 2015-10-15
REF: post-run cleanup re-ordering

klauer:
This PR re-orders the post-run clean-up to be: stop(), collect(), deconfigure()

Fix for `deconfigure` in areadetector filestore clearing the `filestore_resource`, causing any further `collect()` calls to fail.


danielballan:
It makes sense to me that `stop()` should be the first thing we do, and I can't think of anything that this will break.

Issue #208, opened 2015-09-28
raise exception if opening an event while an event is open


tacaswell:

danielballan:
Closed by #235

PR #235, opened 2015-10-16
WIP: Validate order or create/save.

danielballan:
Add straightforward validation in `_create` and `_save` methods on `RunEngine` and tests to match. One of our *old* tests tripped the validation, exposing a pre-existing issue in the generator stack where the first Message is run twice. In every one of our built in scans, this Message happens to be 'create', so no harm was done.

Will revisit. This PR is full of debug prints and stuff to be removed before merging.


danielballan:
Closes #208

PR #236, opened 2015-10-16
FIX: plan metadata should be accessed at open_run

klauer:
Most current scan plan metadata code is failing silently when their `md` properties are accessed prior to open_run. This should fix it. After a brief chat with @danielballan. 

I'd like to require an `md` attribute on all plans so we don't ever have missing metadata due to buggy code. However, according to @danielballan this won't work with Hugo's code...

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -230,6 +230,7 @@ class RunEngine:
         self._exit_status = 'success'  # optimistic default
         self._reason = ''  # reason for abort
         self._task = None  # asyncio.Task associated with call to self._run
+        self._plan = None  # the scan plan instance from __call__
         self._command_registry = {
             'create': self._create,
             'save': self._save,
@@ -292,6 +293,7 @@ class RunEngine:
         self._exit_status = 'success'
         self._reason = ''
         self._task = None
+        self._plan = None
 
         # Unsubscribe for per-run callbacks.
         for cid in self._temp_callback_ids:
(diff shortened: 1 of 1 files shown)


bluesky/run_engine.py
```diff
@@ -486,11 +488,9 @@ def __call__(self, plan, subs=None, **metadata):
...
                 self._temp_callback_ids.add(self.subscribe(name, func))
 
-        metadata['scan_type'] = getattr(type(plan), '__name__')
-        if hasattr(plan, 'md'):
-            self._metadata_per_call.update(plan.md)
-        # If kwargs to __call__ collide with plan.md, kwargs win.
-        self._metadata_per_call.update(metadata)
+        metadata_kw['scan_type'] = getattr(type(plan), '__name__')
+        self._plan = plan
+        self._metadata_per_call = metadata_kw
```
danielballan: We used `.update()` instead of `=` so that any references to this object are not blown away. It is guaranteed to be an empty dict because we call `self._clear_call_cache()` above.

bluesky/run_engine.py
```diff
@@ -748,6 +748,8 @@ def _open_run(self, msg):
 
         # Metadata can come from history, __call__, or the open_run Msg.
         self._metadata_per_run = dict(self.md)
```
danielballan: This is my mistake, not yours, but can we change this to `.update()` as well for philosophical consistency?

danielballan:
> I'd like to require an `md` attribute on all plans so we don't ever have missing metadata due to buggy code. However, according to @danielballan this won't work with Hugo's code...

In case others are reading along, I specifically meant that the input "plan" should only be required to be a generic iterable. Our built-in OO plans happen to be classes with certain useful attributes, but Python generators and lists should also work. This has makes it easy to test, write very simple examples, and to generalize how Messages are sent. cc @hhslepicka

klauer:
Updated to using `dict.update` as per @danielballan's request.

PR #237, opened 2015-10-22
REF: Make more things properties; .md is inspectable before run

danielballan:
Tests pass, but I want to discuss the approach before merging.

### DIFF
--- a/bluesky/scans.py
+++ b/bluesky/scans.py
@@ -223,13 +223,19 @@ class DeltaListScan(Scan1D):
     """
     _derived_fields = ['num', 'init_pos']
 
+    @property
+    def init_pos(self):
+        "None unless a scan is running"
+        return getattr(self, '_init_pos', None)
+
     def _pre_scan(self):
+        "Get current position for the motor."
         self.num = len(self.steps)
         ret = yield Msg('read', self.motor)
         if len(ret.keys()) > 1:
-            raise NotImplementedError("Can't DScan this motor")
+            raise NotImplementedError("Can't DScan a multiaxis motor")
         key, = ret.keys()
-        self.init_pos = ret[key]['value']
+        self._init_pos = ret[key]['value']
         self._abs_steps = np.asarray(self.steps) + self.init_pos
         yield from super()._pre_scan()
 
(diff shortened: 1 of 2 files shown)


danielballan:
tl;dr We were using `_pre_scan` to set a lot of things that could have just been properties. Now that they are, one can inspect `plan.md` at any time.

This would also have solved #236. I still think the changes in #236 were good, but this should be done as well. Does this seem like a step in the right direction, @klauer?

danielballan:
Test pass.

bluesky/scans.py
```diff
@@ -782,21 +806,22 @@ def args(self):
...
             # Special case first pass because their is no
             # mutliplicative identity for cyclers.
-            if self.cycler is None:
-                self.cycler = c
+            if result is None:
+                result = c
             else:
-                self.cycler += c
-        yield from super()._pre_scan()
+                result += c
```
tacaswell: Probably better to make a list of cyclers and then run the list through a reduce.

bluesky/scans.py
```diff
@@ -812,9 +837,12 @@ class InnerProductAbsScan(_InnerProductScanBase):
     motor1, start1, stop1, ..., motorN, startN, stopN : list
         motors can be any 'setable' object (motor, temp controller, etc.)
     """
-    def _pre_scan(self):
-        self._offsets = defaultdict(lambda: 0)
-        yield from super()._pre_scan()
+    _derived_fields = ['motors', 'extents']
+
+    @property
+    def init_pos(self):
```
tacaswell: make private(ish)

bluesky/scans.py
```diff
@@ -866,9 +909,11 @@ class OuterProductAbsScan(_OuterProductScanBase):
         is a boolean indicating whether to following snake-like, winding
         trajectory or a simple left-to-right trajectory.
     """
-    def _pre_scan(self):
-        self._offsets = defaultdict(lambda: 0)
-        yield from super()._pre_scan()
+    _derived_fields = ['motors', 'shape', 'num', 'extents', 'snaking']
+    @property
+    def init_pos(self):
```
tacaswell: private(ish)

danielballan:
I did more cleanup. This code is less confusing to read now, I hope. Ready for one last review.

danielballan:
@tacaswell has reviewed this a couple times and I have @klauer's humorously curmudgeonly blessing.

Issue #13, opened 2015-06-05
hdf5 and tiff export via RunStop callback


danielballan:
```
def hdf5_export(template):
    def callback(run_stop):
       filename = template.format(**header)
```

tacaswell:
Also see suitcase

danielballan:
See http://nbviewer.ipython.org/gist/anonymous/a3ee62067986e73b2d2b, which demonstrates automatic TIFF export. This should make its way into the sphinx documentation before we close the issue.

danielballan:
This is documented.

PR #239, opened 2015-10-23
ENH: Add tiff export function factory.

danielballan:
@sbillinge @timoHMM This is how I would implement TIFF export.

```python
In [2]: from bluesky.broker_callbacks import exporter_factory

In [3]: f = exporter_factory('img', 'test/{h.start.scan_id}_{N}.tiff')

In [4]: header = db['0b96536a']  # example data with 20 images in a field named `'img'`

In [6]: !mkdir test

In [7]: f(header)

In [8]: !ls test
1_0.tiff   1_11.tiff  1_14.tiff  1_17.tiff  1_2.tiff   1_5.tiff   1_8.tiff
1_1.tiff   1_12.tiff  1_15.tiff  1_18.tiff  1_3.tiff   1_6.tiff   1_9.tiff
1_10.tiff  1_13.tiff  1_16.tiff  1_19.tiff  1_4.tiff   1_7.tiff
```

As you can see, it is simple to customize the filename based on any piece of metadata. It's all contained in the Header. See docstring for more.

### DIFF
--- a/bluesky/broker_callbacks.py
+++ b/bluesky/broker_callbacks.py
@@ -4,6 +4,9 @@ from metadatastore.commands import run_start_given_uid, descriptors_by_start
 import matplotlib.pyplot as plt
 from xray_vision.backend.mpl.cross_section_2d import CrossSection
 from .callbacks import CallbackBase
+import tifffile
+import numpy as np
+from databroker import get_images
 
 
 class LiveImage(CallbackBase):
@@ -86,4 +89,48 @@ def post_run(callback):
         for e in events:
             callback('event', e)
         callback('stop', stop_doc)
+    return f
+
+
+def exporter_factory(field, template):
+    """
+    Build a function that, given a header, exports tiff files.
+
+    The file names will incorporate the contents of the Header.
+
+    Parameters
+    ----------
+    field : str
+        a data key, e.g., 'pe1_image_lightfield'
+    template : str
+        A templated file path, where curly brackets will be filled in with
+        the attributes of 'h', a Header, and 'N', a sequential number.
+        e.g., "dir/scan{h.start.scan_id}_by_{h.start.experimenter}_{N}.tiff"
+
+    Returns
+    -------
+    f : function
+        a function that accepts a header and saves TIFF files
+    """
+    # validate user input
+    if not '{N}' in template:
+        raise ValueError("format_stirng must include '{N}'")
+
+    def f(h):
+        imgs = get_images(h, field)
+        # Fill in h, defer filling in N.
+        _template = template.format(h=h, N='{N}')
+        for i, img in enumerate(imgs):
+            filename = _template.format(N=i)
+            tifffile.imsave(filename, np.asarray(img))
+
+    # Write a customized docstring for f based on what is specifcally does.
+    f.__doc__ = """
+Export sequentially-numbered TIFF files from the {field} field.
+
+Parameters
+----------
+h : Header
+    a header from the databroker
+""".format(field=field)
     return f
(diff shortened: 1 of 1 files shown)


danielballan:
cc @pavoljuhas

pavoljuhas:
I think the example should say

```python
In [3]: f = exporter_factory('pe1_image_lightfield', 'test/{h.start.scan_id}_{N}.tiff')
```

Is this intended for end users, ie, people with a little Python/command-line knowledge, possibly stressing out at 3AM at the beamline?  If so I would recommend some bells and whistles from the [TiffSaver](https://github.com/NSLS-II-XPD/ipython_ophyd/blob/master/uiophyd/tiffsaver.py), i.e.,

* support for dry-run that would let users check output names before writing.
* exporter should have an attribute for filename-template that could be checked or changed without
  having to instantiate a new one.
* timestamping of the written tiff files by the scan time.
* ability to overwrite wrongly-named files identified by the same timestamp.
* dark image subtraction.  Possibly a boolean flag for carrying it out or not.

Is there a potential for the factory to support more output formats?  If it will be tiff-only, perhaps the factory might be renamed to `tiff_exporter_factory` or `make_tiff_exporter`.

danielballan:
My example was applied to example data on my computer, generated from
databroker.examples.sample_data. Your example makes sense for XPD's pe1.

+1 on the rename and any bells, whistles people want. I'm just suggesting
this pattern.

pavoljuhas:
Factory pattern is nifty and would be very useful if there were several exporter types, for example for several output formats.  I am however not sure what is its benefit for a single exporter type.  The factory call seems to be the only way of configuring the exporter and I don't see any way of accessing its configuration later.  It feels a bit clunky having to pass all of exporter parameters to the factory if user needs to change just one.  Am I perhaps missing something?

danielballan:
Did you notice that the configuration is a template on `h`? If I call `f` repeatedly, using several different headers, each header's metadata fills in the template.

There are only two things to configure here -- which field (which in your case will probably never change) and which template, as in `'test/{h.start.scan_id}_{N}.tiff'`. I emphasize that `h` is a stand-in for any Header, not a specific Header. If you need to modify the template, you'd have to retype it anyway, right?

I don't expect non-sophisticated or overtired users will need to customize the template very often either, so all-purpose `f` could be defined in the IPython profile, ready to use. The simplicity of a function appeals to me.

bluesky/broker_callbacks.py
```diff
@@ -87,3 +90,47 @@ def f(name, stop_doc):
...
+    """
+    # validate user input
+    if not '{N}' in template:
+        raise ValueError("template must include '{N}'")
+
+    def f(h):
+        imgs = get_images(h, field)
+        # Fill in h, defer filling in N.
+        _template = template.format(h=h, N='{N}')
+        for i, img in enumerate(imgs):
```
tacaswell: This should generate and check if any of the files exist.  If _any_ exist it should raise.

danielballan:
New commits add a `dryrun` argument to the exporter function. Also, the function now always returns a list of filepaths to the files it wrote (or, if dryrun *would* write).

I like the idea of the timestamps -- can you open up a PR to incorporate that since you've done it before?

The rest, particularly exporting *processed* data, is a more complex task, out of scope for this light-weight utility. The functionality is needed, but I'd like to build it separately.

danielballan:
P.S. One useful thing I want to highlight: it is possible to hook this up to bluesky subscriptions such that TIFFs are automatically exported at the end of a scan. Users don't have to intervene or know anything about how it works.

danielballan:
Useful notebook. I already sent this over email, but I'm posting the link on github so we can find it later. http://nbviewer.ipython.org/gist/anonymous/a3ee62067986e73b2d2b

tacaswell:
That notebook should go into our examples or something.

danielballan:
See #13

pavoljuhas:
Took a while to get to this.  See #263.

Issue #226, opened 2015-10-09
Double control C to abort a scan


ambarb:
Aborted the scan using by hitting control C twice.  Opened new figure. Could not launch scan.  gs.RE.abort did not fix. Close plotting windows (generates error report).  Scan would still not launch.  Restart bluesky works.

See below.

```python
In [194]: grabit
Select area of screen to grab .........

In [195]: ascan(epu2_gap,21,22,20)
Transient Scan ID: 52140
Persistent Unique Scan ID: 'a502454f-8f5f-48a6-89bb-1a084522d081'
+------------+--------------+----------------+----------------+
|   seq_num  |        time  |      epu2_gap  |    sclr_chan2  |
+------------+--------------+----------------+----------------+
^C
^C+------------+--------------+----------------+----------------+
/opt/conda_envs/collection/lib/python3.4/site-packages/numpy/core/_methods.py:59: RuntimeWarning: Mean of empty slice.
  warnings.warn("Mean of empty slice.", RuntimeWarning)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/opt/conda_envs/collection/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py in _make_request(self, conn, method, url, timeout, **httplib_request_kw)
    371             try:  # Python 2.7, use buffering of HTTP responses
--> 372                 httplib_response = conn.getresponse(buffering=True)
    373             except TypeError:  # Python 2.6 and older

TypeError: getresponse() got an unexpected keyword argument 'buffering'

During handling of the above exception, another exception occurred:

KeyboardInterrupt                         Traceback (most recent call last)
<ipython-input-195-e9cbc2869e74> in <module>()
----> 1 ascan(epu2_gap,21,22,20)

/opt/conda_envs/collection/lib/python3.4/site-packages/bluesky/simple_scans.py in __call__(self, motor, start, finish, intervals, time, subs, **kwargs)
    209         original_times = _set_acquire_time(time)
    210         result = super().__call__(motor, start, finish, intervals + 1,
--> 211                                   subs=subs, **kwargs)
    212         _unset_acquire_time(original_times)
    213         return result

/opt/conda_envs/collection/lib/python3.4/site-packages/bluesky/simple_scans.py in __call__(self, subs, sub_factories, *args, **kwargs)
    136         # Any remainging kwargs go the RE. To be safe, no args are passed
    137         # to RE; RE args effectively become keyword-only arguments.
[...]

PR #242, opened 2015-10-26
FIX: Catch KeyboardInterrupt from badly-behaving external code.

danielballan:
This addresses another subtle case that may have caused #226.

If some external code (in the case of #226, something in `pyOlog` or `requests`) takes control of SIGINT / Ctrl+C away from the RunEngine and then raises `KeyboardInterrupt`, the run will fail on a generic error. This PR captures `KeyboardInterrupt` and then calls `RunEngine.request_pause` as we normally do when we capture SIGINT.

If external code is behaving properly, a `KeyboardInterrupt` should never be raised. The external code should defer to the previously-registered SIGINT handler (that is, the RunEngine's handler) instead of raising KeyboardInterrupt. Therefore, this should be fixed upstream if the problem can be identified. At the same time, we can mitigate the issue in general --- it might come up again from other external code.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -607,31 +607,40 @@ class RunEngine:
         self._reason = ''
         try:
             while True:
-                if self._exception is not None:
-                    raise self._exception
-                # Send last response; get new message but don't process it yet.
                 try:
-                    msg = self._genstack[-1].send(
-                        response if not self._new_gen else None)
+                    if self._exception is not None:
+                        raise self._exception
+                    # Send last response;
+                    # get new message but don't process it yet.
+                    try:
+                        msg = self._genstack[-1].send(
+                            response if not self._new_gen else None)
 
-                except StopIteration:
-                    self._genstack.pop()
-                    if len(self._genstack):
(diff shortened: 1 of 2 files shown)


ericdill:
Ahh this makes a lot of sense.  Nice catch!

ericdill:
I don't suppose there is any way to add a test for this? It would be nice if we could protect ourselves from having to deal with **this** particular bug again

danielballan:
It would be complicated.

danielballan:
With enough peer pressure I would do it, but I'd rather fry other fish and assume no one will go in and delete this `except` block. I would be hard to break this, I think.

tacaswell:
I think the test would be easy, just make a plan generator that after yielding a few innocuous messages, raises `KeyboardInterupt`

danielballan:
Fair enough. I was thinking of something more complicated -- registering a "rogue" SIGINT handler. What you suggest is better than nothing. Test added.

Let's discuss more though. Do not merge.

danielballan:
OK, I'm happy with this again. Ready to merge.

PR #245, opened 2015-10-26
Log when objects are done setting, triggering.

danielballan:


ericdill:
:+1: on logging more info!

Totally separate question... What are your feelings on adding a logging framework that would log stuff to a file as well as optionally print to stdout?

danielballan:
Definitely. Not sure whether to use traitlets.LoggingConfigurable or roll
our own so punting for a little while longer.

ericdill:
Ok cool. Thanks for your thoughts

PR #251, opened 2015-11-03
ENH: Write new Event Descriptor when config is changed.

danielballan:
1. Capture `obj.configuration` in the Event Descriptor document.
2. Create a new Event Descriptor for the same set of data keys whenever that configuration is changed through a 'configure' Msg.

Notes:
* In the first pass I called `obj.configuration` `obj.state`. No objects have actually implemented this yet, so it is safe to change the name.
* Notice that we currently do not record in the Documents which data keys came from the same object. They are all lumped together in one dictionary. Now, recording this association is now important, because we need to record which data key is associated with which configuration. Thus, I generate a unique ID by hashing the set of the object's field names. This is safe: we have never allowed two objects to use the same field name.


bluesky/run_engine.py
```diff
@@ -1027,13 +1046,28 @@ def _logbook(self, msg):
 
     @asyncio.coroutine
     def _configure(self, msg):
+        if self._bundling:
+            raise IllegalMessageSequence(
+                "Cannot configure after 'create' but before 'save'"
+                "Aborting!")
```
dchabot: Does this not have implications for the "gain switching" scheme proposed for CSX?
At least semantically?

danielballan: Yes. Maybe later we can relax this, but I think it makes sense for now. No currently-implemented scans would violate, and if one did, incorrect metadata would be recorded. I vote for generalizing this in a future, more comprehensive effort.

bluesky/run_engine.py
```diff
@@ -841,11 +877,25 @@ def _save(self, msg):
...
+                    dk['group'] = group
+                data_keys.update(dks)
+                configuration[group] = self._configuration_cache[obj]
+                configuration_times[group] = \
+                    self._configuration_times_cache[obj]
             _fill_missing_fields(data_keys)  # TODO Move this to ophyd/controls
             descriptor_uid = new_uid()
             doc = dict(run_start=self._run_start_uid, time=ttime.time(),
-                       data_keys=data_keys, uid=descriptor_uid)
+                       data_keys=data_keys, uid=descriptor_uid,
```
tacaswell: @arkilic @klauer @dchabot 

This is the interesting bit of this PR in that it adds a path to add configuration information into mds.  The basic scheme is that each data key picks up another entry `'group'` and there is a `'configuration'` and `'configuration_times'` are dicts keyed on those groups which store what ever that object thought was important to store.

@danielballan is working on code to 'pivot' any thing in the configuration dict up into the events (ex, every event would have an entry with the exposure time).

bluesky/run_engine.py
```diff
@@ -415,13 +421,41 @@ def request_pause(self, defer=False, name=None, callback=None):
                 print("Cannot pause from {0} state. "
                       "Ignoring request.".format(self.state))
 
-    def _register_scan_callback(self, name, func):
-        """Register a callback to be processed by the scan thread.
+    def _subscribe_lossless(self, name, func):
```
klauer: :+1: on clarifying the 'lossless' callbacks (but it's not really related to the PR and also not mentioned in your summary...)

danielballan: Yeah, sorry, it was added in a later commit. You can see my line of thought in the commit messages... "Well as long as I'm touching `register_mds.py` to fix `insert_descriptors` I might as well fix this thing...."

tacaswell: @danielballan sometimes needs blinders :smiling_imp:  Or it may be a campaign to get us to merge his stuff quickly "this looks good ... MERGE IT BEFORE HE STARTS RENAMING THINGS".

danielballan: I just laughed out loud in a coffee shop. NERD!!

bluesky/run_engine.py
danielballan: @arkilic @klauer @dchabot This is where the configuration discussion was kicked off (but never picked up, obviously). I was misremembering that this PR was merged. Fortunately, it was in fact abandoned, waiting for consensus -- which is as it should be.

I will make some revisions in light of our discussion today.

danielballan:
I iterated on the organization of the configuration field a bit. Here's where a landed, trying to mirror the organization of Event as much as possible:

```python
event.descriptor.{object_name}.data
event.descriptor.{object_name}.timestamps
event.descriptor.{object_name}.data_keys
```

We used the word "configuration descriptor" in our conversation, but more precisely it's configuration "data keys."

danielballan:
Tests pass -- ready for review.

danielballan:
post-vacation bump: this needs to be reviewed and merged ahead of the ophyd deployment.

bluesky/register_mds.py
```diff
@@ -35,6 +35,24 @@ def _insert_run_start(name, doc):
     return mds.insert_run_start(**doc)
 
 
+known_descriptor_keys = ['run_start', 'data_keys', 'time', 'uid']
```
tacaswell: bikeshed: make this a `set`

bluesky/run_engine.py
tacaswell: These functions should be made public.  Bikeshedding the name a bit `subscribe_reliable` `subscribe_blocking`, `subscribe_guaranteed`.

bluesky/run_engine.py
```diff
@@ -1028,25 +1087,38 @@ def _logbook(self, msg):
...
+        old, new = obj.configure(kwargs.get('configuration'))
+
+        self._cache_config(obj)
+        return old, new
 
     @asyncio.coroutine
-    def _deconfigure(self, msg):
-        # If an object has no 'deconfigure' method, assume it does not need
-        # deconfiguring.
+    def _stage(self, msg):
```
tacaswell: This is missing the call to `obj.stage`

PR #254, opened 2015-11-04
Add gs.FLYERS

danielballan:
Closes https://github.com/NSLS-II/wishlist/issues/79

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -115,6 +115,9 @@ class Mover(Base):
     def stop(self):
         pass
 
+    def trigger(self):
+        return None
+
 
 class SynGauss(Reader):
     """
@@ -364,6 +367,8 @@ class FlyMagic(Base):
             ttime.sleep(0.01)
         self._time = None
 
+    def stop(self):
+        pass
 
 motor = Mover('motor', ['motor'])
 motor1 = Mover('motor1', ['motor1'], sleep_time=.1)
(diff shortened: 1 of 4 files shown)


danielballan:
attn @stuwilkins

bluesky/examples.py
```diff
@@ -115,6 +115,9 @@ def settle(self):
     def stop(self):
         pass
 
+    def trigger(self):
```
tacaswell: I don't think moveables actually have a trigger

bluesky/tests/test_simple_api.py
```diff
@@ -59,3 +59,6 @@ def test_basic_usage():
     tscan(1, 2, 2, time=0.1)
     dtscan(1, 2, 2, time=0.1)
     th2th(1, 2, 2, time=0.1)
+    flyer = FlyMagic('wheee', motor, det1, det2)
```
tacaswell: I like the name.

bluesky/examples.py
danielballan: As I remember it, we allow for the possibility that users might want to set a motor and then trigger that setting as a separate step. Imagine the case where set performs some moderately expensive computation.

danielballan: Also, Movable's interface is a subset of the Readable interface.

tacaswell: oh, fair point.

PR #259, opened 2015-11-05
Fix collect orderofops

tacaswell:


tacaswell:
```python


class fake_collector:
    def __init__(self, n):
        self.n = n
        
    def kickoff(self):
        sts = StatusBase()
        sts._finished()
        return sts

    def stop(self):
        pass

    def describe(self):
        return [{'dat': {'source': 'syn', 'dtype': 'number', 'shape': ()}}]
    
    def collect(self):
        # for j in range(self.n):
        for j in self.n:   
             yield {'data': {'dat': j,},
                   'timestamps': {'dat': ttime.time()},
                   'time': ttime.time()}
                   
```

Example flyer that exposed these issues.

tacaswell:
```python
flyer = fake_collector(10000)

def fly_gen(flyer):
    yield Msg('open_run')
    yield Msg('kickoff', flyer)
    yield Msg('wait', None, 'fly')
    yield Msg('collect', flyer)
    yield Msg('close_run')

```

PR #260, opened 2015-11-05
FIX: do not try bulk insertion with no events

tacaswell:
If a flyer yeilds no events, an empyt list bubbles it's way through
to pymongo which raises on a no-op bulk operation.

This is deployed as a hot fix at CSX.


klauer:
:+1:

tacaswell:
hurray for commit messages with typos!

PR #261, opened 2015-11-07
CLN/API: remove unused class

tacaswell:


danielballan:
Travis failed on an Epics hiccup, but all unit tests passed.

PR #263, opened 2015-11-14
ENH: switch to a configurable tiff exporter

pavoljuhas:
Add configurable TiffExporter class with a similar defaults as the exporter function.  Move output name generation to a DataNaming class.  Support custom data-extraction function when image data need to be transformed. @danielballan, @chiahaoliu, @sbillinge

## demo session at XPD

```python
In [1]: from bluesky.broker_callbacks import make_tiff_exporter
        from dataportal.broker import DataBroker as db
        h = db['97064560']
        te = make_tiff_exporter(prefix='/tmp')
Out[1]: <TiffExporter 
  fetch:functools.partial(<function get_images at 0x7fab5b840b70>, name='pe1_image_lightfield')
  naming:DataNaming(template='apples{scan_id:05d}_{N:03d}<-T{e.data[cs700]:03.1f}>.tiff', prefix='/tmp')
  use_mtime:True>

In [2]: te.naming(h)
Out[2]:
['/tmp/scan00015_000.tiff',
 '/tmp/scan00015_001.tiff',
 '/tmp/scan00015_002.tiff',
 '/tmp/scan00015_003.tiff',
 '/tmp/scan00015_004.tiff',
 '/tmp/scan00015_005.tiff']

In [3]: te.findExistingOutputs(h)
Out[3]: []

In [4]: te(h, dryrun=1);
[dryrun] write image data to /tmp/scan00015_000.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:12.031946
[dryrun] write image data to /tmp/scan00015_001.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:23.855054
[dryrun] write image data to /tmp/scan00015_002.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:36.101679
[dryrun] write image data to /tmp/scan00015_003.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:48.093346
[dryrun] write image data to /tmp/scan00015_004.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:59.882183
[dryrun] write image data to /tmp/scan00015_005.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:14:11.803296

In [5]: te(h, dryrun=1, select=slice(3));
[dryrun] write image data to /tmp/scan00015_000.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:12.031946
[dryrun] write image data to /tmp/scan00015_001.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:23.855054
[dryrun] write image data to /tmp/scan00015_002.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:36.101679

In [6]: te(h, select=range(3));
Out[6]:
['/tmp/scan00015_000.tiff',
 '/tmp/scan00015_001.tiff',
 '/tmp/scan00015_002.tiff']

In [7]: te(h, dryrun=1)
[dryrun] skip /tmp/scan00015_000.tiff already saved as /tmp/scan00015_000.tiff
[dryrun] skip /tmp/scan00015_001.tiff already saved as /tmp/scan00015_001.tiff
[dryrun] skip /tmp/scan00015_002.tiff already saved as /tmp/scan00015_002.tiff
[dryrun] write image data to /tmp/scan00015_003.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:48.093346
[dryrun] write image data to /tmp/scan00015_004.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:59.882183
[dryrun] write image data to /tmp/scan00015_005.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:14:11.803296

In [8]: import re
        te.naming.template = re.sub('^scan', 'apples', te.naming.template)
        te(h, dryrun=1)
[dryrun] skip /tmp/apples00015_000.tiff already saved as /tmp/scan00015_000.tiff
[dryrun] skip /tmp/apples00015_001.tiff already saved as /tmp/scan00015_001.tiff
[dryrun] skip /tmp/apples00015_002.tiff already saved as /tmp/scan00015_002.tiff
[dryrun] write image data to /tmp/apples00015_003.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:48.093346
[dryrun] write image data to /tmp/apples00015_004.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:13:59.882183
[dryrun] write image data to /tmp/apples00015_005.tiff
[dryrun] adjust image file mtime to 2015-09-24 14:14:11.803296
Out[8]:
['/tmp/apples00015_003.tiff',
 '/tmp/apples00015_004.tiff',
 '/tmp/apples00015_005.tiff']

In [9]: te(h)
skip /tmp/apples00015_000.tiff already saved as /tmp/scan00015_000.tiff
skip /tmp/apples00015_001.tiff already saved as /tmp/scan00015_001.tiff
skip /tmp/apples00015_002.tiff already saved as /tmp/scan00015_002.tiff
Out[9]:
['/tmp/apples00015_003.tiff',
 '/tmp/apples00015_004.tiff',
 '/tmp/apples00015_005.tiff']

In [10]: te.findExistingOutputs(h)
Out[10]:
['/tmp/scan00015_000.tiff',
 '/tmp/scan00015_001.tiff',
 '/tmp/scan00015_002.tiff',
 '/tmp/apples00015_003.tiff',
 '/tmp/apples00015_004.tiff',
 '/tmp/apples00015_005.tiff']

In [11]: te(h, select=range(3), overwrite=1)
Out[11]:
['/tmp/apples00015_000.tiff',
 '/tmp/apples00015_001.tiff',
 '/tmp/apples00015_002.tiff']

In [12]: te.findExistingOutputs(h)
Out[12]:
['/tmp/apples00015_000.tiff',
 '/tmp/apples00015_001.tiff',
 '/tmp/apples00015_002.tiff',
 '/tmp/apples00015_003.tiff',
 '/tmp/apples00015_004.tiff',
 '/tmp/apples00015_005.tiff']

```

### DIFF
--- a/bluesky/broker_callbacks.py
+++ b/bluesky/broker_callbacks.py
@@ -1,13 +1,8 @@
-import os
 from databroker import DataBroker as db, get_events
 import filestore.api as fsapi
 from metadatastore.commands import run_start_given_uid, descriptors_by_start
-import matplotlib.pyplot as plt
 from xray_vision.backend.mpl.cross_section_2d import CrossSection
 from .callbacks import CallbackBase
-import tifffile
-import numpy as np
-from databroker import get_images
 
 
 class LiveImage(CallbackBase):
@@ -25,9 +20,10 @@ class LiveImage(CallbackBase):
     relevant commit is a951b7.
     """
     def __init__(self, field):
+        from matplotlib.pyplot import figure
         super().__init__()
         self.field = field
-        fig = plt.figure()
+        fig = figure()
         self.cs = CrossSection(fig)
         self.cs._fig.show()
 
--- a/bluesky/broker_callbacks.py
+++ b/bluesky/broker_callbacks.py
@@ -1,6 +1,7 @@
 from databroker import DataBroker as db, get_events
 import filestore.api as fsapi
 from metadatastore.commands import run_start_given_uid, descriptors_by_start
+import matplotlib.pyplot as plt
 from xray_vision.backend.mpl.cross_section_2d import CrossSection
 from .callbacks import CallbackBase
 
@@ -20,10 +21,9 @@ class LiveImage(CallbackBase):
     relevant commit is a951b7.
     """
     def __init__(self, field):
-        from matplotlib.pyplot import figure
         super().__init__()
         self.field = field
-        fig = figure()
+        fig = plt.figure()
         self.cs = CrossSection(fig)
         self.cs._fig.show()
 
--- a/bluesky/broker_callbacks.py
+++ b/bluesky/broker_callbacks.py
@@ -93,7 +93,7 @@ def post_run(callback):
     return f
 
 
-def make_tiff_exporter(field, template):
+def make_tiff_exporter(fetch=None, template=None, prefix=None, **kwargs):
     """
     Build a function that, given a header, exports tiff files.
 
@@ -101,54 +101,38 @@ def make_tiff_exporter(field, template):
 
     Parameters
     ----------
-    field : str
-        a data key, e.g., 'pe1_image_lightfield'
-    template : str
-        A templated file path, where curly brackets will be filled in with
-        the attributes of 'h', a Header, and 'N', a sequential number.
-        e.g., "dir/scan{h.start.scan_id}_by_{h.start.experimenter}_{N}.tiff"
+    fetch : str or callable object, optional
+        Set function that will generate 2D arrays from databroker Header.
+        When string, obtain arrays using ``get_images(h, NAME)``.
+        See the TiffExporter class for more details.
+    template : str, optional
+        Set filename template where curly brackets are filled with entries
+        from databroker Header.  See the DataNaming class for more details.
+    prefix : str, optional
+        Set filename prefix, a constant output directory that is prepended
+        to generated filenames.
+        Default is ''.
+    kwargs : TiffExporter attributes, optional
+        Use to set any additional attributes of the constructed TiffExporter
+        object.  Raise AttributeError when TiffExporter does not have such
+        attribute.
 
     Returns
     -------
-    f : function
-        a function that accepts a header and saves TIFF files
+    TiffExporter
+        a callable object that accepts a header and saves TIFF files.
+
+    See Also
+    --------
+    bluesky.datanaming.TiffExporter : configurable TIFF export from a header.
+    bluesky.datanaming.DataNaming : file name generation from a header.
     """
-    # validate user input
-    if not '{N}' in template:
-        raise ValueError("template must include '{N}'")
-
-    def f(h, dryrun=False):
-        imgs = get_images(h, field)
-        # Fill in h, defer filling in N.
-        _template = template.format(h=h, N='{N}')
-        filenames = [_template.format(N=i) for i in range(len(imgs))]
-        # First check that none of the filenames exist.
-        for filename in filenames:
-            if os.path.isfile(filename):
-                raise FileExistsError("There is already a file at {}. Delete "
-                                      "it and try again.".format(filename))
-        if not dryrun:
-            # Write files.
-            for filename, img in zip(filenames, imgs):
-                tifffile.imsave(filename, np.asarray(img))
-        return filenames
-
-    # Write a customized docstring for f based on what is specifcally does.
-    f.__doc__ = """
-Export sequentially-numbered TIFF files from the {field} field.
-
-Parameters
-----------
-h : Header
-    a header from the databroker
-dryrun : bool
-    Set to True to return list of filenames without actually writing files.
-    False by default.
-
-Returns
--------
-filenames : list
-    list of filenames where files were written (or, if dryrun, *would* be
-    written)
-""".format(field=field)
-    return f
+    from bluesky.tiffexporter import TiffExporter
+    te = TiffExporter(fetch=fetch, template=template, prefix=prefix)
+    # validate and apply keyword arguments
+    emsg = "Invalid keyword argument {}, TiffExporter has no such attribute."
+    for n, v in kwargs.items():
+        if not hasattr(te, n):
+            raise AttributeError(emsg.format(n))
+        setattr(te, n, v)
+    return te
--- a/bluesky/tiffexporter.py
+++ b/bluesky/tiffexporter.py
@@ -1,0 +1,263 @@
+#!/usr/bin/env python
+
+
+"""Export of 2D image data from databroker header in TIFF format.
+"""
+
+import os.path
+import datetime
+
+from bluesky.datanaming import DataNaming
+
+
+class TiffExporter(object):
+    """
+    Export detector image data from databroker header in a TIFF format.
+
+    Build a callable object which saves image data from a databroker
+    header.  Output files are built from a template that may include values
+    from the header and its associated events.  TiffExporter can either
+    skip or overwrite already existing files.  The TIFF file modification
+    time is by default set to the timestamp of the exported event.  The
+    timestamp is also used to identify already exported images, possibly
+    under a different naming scheme.
+
+    Attributes
+    ----------
+    default_fetch : str or callable object, class attribute
+        Default value for the `fetch` argument in initialization.
+        Default is 'pe1_image_lightfield'.
+    fetch : callable object
+        Function which generates 2D output arrays from databroker Header.
+        When set to a string NAME, extract arrays using
+        ``databroker.get_images(h, NAME)``.
+    naming : DataNaming instance
+        Callable object which generates output file names from databroker
+        Header.  Use ``naming(h)`` to obtain a list of output names.
+        To adjust naming scheme, modifying the `naming.template` and
+        `naming.prefix` attributes.
+    use_mtime : bool
+        When True (default), write output files with modification time set
+        according to the exported Event.  Also use modification time to
+        identify existing outputs.  When False, write files with current
+        modification time and only check if target paths exist.
+        Default is `True`.
+    mtime_window : float
+        Precision in file modification times when searching for existing
+        output files.   Ignored when `use_mtime` is `False`.
+
+    Parameters
+    ----------
+    fetch : str or callable object, optional
+        Set function that will generate 2D arrays from databroker Header.
+        When string, obtain arrays using ``get_images(h, NAME)``.
+        Use the `default_fetch` value when not specified.
+    template : str, optional
+        Set `naming.template`, a filename template where curly brackets
+        are filled with entries from databroker Header.  See the DataNaming
+        class for more details.
+    prefix : str, optional
+        Set `naming.prefix` a constant output directory that is prefixed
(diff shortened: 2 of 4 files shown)


bluesky/broker_callbacks.py
```diff
@@ -101,54 +101,38 @@ def make_tiff_exporter(field, template):
...
-    False by default.
-
-Returns
--------
-filenames : list
-    list of filenames where files were written (or, if dryrun, *would* be
-    written)
-""".format(field=field)
-    return f
+    from bluesky.tiffexporter import TiffExporter
```
tacaswell: top level imports please.

bluesky/tiffexporter.py
```diff
@@ -0,0 +1,263 @@
...
+
+        Returns
+        -------
+        list
+            Exported output files.
+        """
+        import tifffile
+        from databroker import get_events
+        if dryrun:
+            dryordo = lambda msg, f, *args, **kwargs:  print('[dryrun]', msg)
```
tacaswell: Assigning lambdas to variables is discouraged.  Can you do this with two private module level functions?

bluesky/broker_callbacks.py
pavoljuhas: This import happens in a factory function and I feel is better delayed until truly needed.  
What if user only wants LiveImage and is not exporting any TIFFs?  
Likewise, matplotlib.pyplot needed in LiveImage is expensive, 
needs DISPLAY and is better delayed to `LiveImage.__init__()`.

There were also few unused modules and unnecessarily local imports which I cleaned in a42b4e606bab8aa9377b5c8e6bdb29b714ae22da.

bluesky/tiffexporter.py
pavoljuhas: done in 8647b4e92d7244c013629d046b6dc29c720c60e4.

danielballan:
This is a lot to digest. It does more, but it is also ten times as much code. There are more parameters for the user to understand and much more code to maintain. The original approach plus basic command-line competency seems easier to me.

pavoljuhas:
I don't see it that way, here are some points:
- It is more code, because it does more for the user.  Users need their mental bandwidth for running the experiment and evaluating the data, not for constructing template strings or cleaning up misnamed files.
- a standard usage is easier for the end user as they can run
`te = make_tiff_exporter()` (ie, no need to know Python format language) and get file naming according to the `DataNaming.default_template` class attribute.  The `DataNaming.default_template` can be adjusted in IPython startup file to a preferred naming scheme for the beamline.
- In its standard configuration `TiffSaver.__call__` skips any existing outputs and writes those that do not exist.  This is important for 8-hour temperature scans where users need to check the first few measurements while the scan is still in progress.  They can do the same export call later to save any new events and skip those that already exist.  The previous exporter would throw FileExist exception if they tried to use it in the same way.
- the TiffExporter instance has introspection (`print(te); print(te.naming)`) and can be configured and tested after creation.  This is not quite possible with a factory-produced function.
- the DataNaming class can be reused wherever else might be a need to generate names from databroker fields.
- the TiffExporter class could be easily generalized for other file formats by converting it to a GenericExporter base that would call `self.write` instead of `tifffile.imsave`.  The derived classes, say `TiffExporter`, `RawExporter` or `BzippedPickleExporter` would just override the `write` method.

danielballan:
I should have said above that I agree that *parts* of this are definitely a good idea.

We are less interested/excited about full-featured export facilities because the operation is lossy: one loses the power of the data broker and full metadata. We are more interested in hooking up the analysis code (e.g., XPDF Suite) to a live stream of Documents. But this is not to say that good export tools are not desirable.

I think we reach a satisfactory compromise. We will need some time to digest your comments and look more closely at the code.

pavoljuhas:
I agree the best situation is to use the databroker directly, but until there is some way of using it remotely or exporting a full experiment, some export facility will be needed so users can take something home.

I'd be happy to chat / explain any parts of the code you might have questions about.  Thank you for checking.

tacaswell:
At a minimum the extension of the format mini-language should be pulled out into it's own class (maybe subclass `str`, maybe just provide `+`, `[]`, and `format`) which will reduce the apparent complexity of this code.  It might be better to just use jinja here (maybe with a translation layer from `<>` -> jinja).

I am not convinced that the complexity of the mtime based file finding or the optional templating is worth the complexity it brings.

pavoljuhas:
The purpose of conditional template segments is to allow beamlines to set standard naming scheme which would work most of the time.  For example, the `T{e.data[cs700]:03.1f}` would normally throw exception for measurements without the CryoStream 700 attached and force users to switch between temperature-included and temperature-free templates.  With conditional templates the cryostream part would just get omitted.  Several conditional segments may be included, for example, 
`LBCO<-pressure={e.data[pressure]}><-cs700temperature={e.data[cs700]}>.tiff`.<br>I feel this is a minimum extension that takes care of the non-permanent data items.  I know the parsing code is not easy on eye, but to my knowledge it works and I'd be glad to fix cases where it does not behave as advertised.

I agree the formatting extension belongs to a subclass or perhaps some `conditional_format(fmt, *a, **kw)` function.  I can do so if you anticipate a need for it outside the DataNaming class.  Otherwise we may as well hold on with crossing that bridge.

The mtime timestamps allow to find already exported files from databroker header regardless of their naming scheme.  The users can easily clean things up with `te(header, overwrite=1)` if they happened to export under incorrect naming.  They can also produce a list of header-tiffs with `te.findExistingOutput(header)` and use it however they see fit.

Maybe I over-engineered and there is no compelling need for conditional-naming or file-timestamps  functionality.  Perhaps @sghose or @sbillinge could comment?

sbillinge:
I sympathise with both sides here.

I would much rather that our effort was going into building rich metadata
libraries with the data collection (well our effort is going in there, but
such rich templates for writing the metadata may be helpful) and then being
able to do rich searches on that and pass it around with the data all the
way to reduced data (or at the very least facile ways to trace back to the
metadata in the database from the reduced data).  But there are two
impediments to this being received with open arms by the users.  First, is
infrastructural.....it is not yet easy to get at the database from a remote
location.  Second is sociological.....scientists are conservative beasts
and they will _all_ want their tiffs in their pockets with some human
readable name until they have used the other data-retrieval infrastructure
a few times and built confidence that it serves their needs.    Until the
day that both these criteria are met, I am afraid that we will be living
with tiff-writing.

Given that that is the case, offering users very easy, powerful and robust
ways to get their fancy-named tiffs will make them very happy, and build
their confidence in our ability to deliver them the code they want.  Later
on I wouldn't be completely surprised if you found them using the file
naming scheme but it is writing a text file containing data-set UID....just
because they are used to using microsoft explorer as their main data
curation tool......Actually I was teasing my students about this recently,
but they kind of demonstrated to me how rapidly they could find any
data-set I asked them about from a recent experiment using fuzzy search on
richly composed filenames, so I kind of ate my words.  They are pretty much
not going to fully embrace our new database system, until the search
capabilities, and access to the database, are as easy, quick and reliable
to use as microsoft explorer.  I am afraid that this is kind of reality :(

One caveat to all this is that when the data volumes become immense, they
will quickly want to switch over to a system where they are not passing the
actual Tbytes of data around.

Can the tiff-name composing templates be reused later in any way as
datacase-search composition templates?  Also, it is not clear to me from
the discussion about whether the templates are already coded up.  My sense
from what Pavol is saying is that it is already coded and it is just
whether it is pulled into bluesky or not?  Would one suggestion be to make
it a separate package and not pull it into bluesky, in which case the
software group is committed to maintain and document it?

On Tue, Nov 17, 2015 at 5:50 PM, Pavol Juhas <notifications@github.com>
wrote:

> The purpose of conditional template segments is to allow beamlines to set
> standard naming scheme which would work most of the time. For example, the
> T{e.data[cs700]:03.1f} would normally throw exception for measurements
> without the CryoStream 700 attached and force users to switch between
> temperature-included and temperature-free templates. With conditional
> templates the cryostream part would just get omitted. Several conditional
> segments may be included, for example,
> LBCO<-pressure={e.data[pressure]}><-cs700temperature={e.data[cs700]}>.tiff
> .
> I feel this is a minimum extension that takes care of the non-permanent
> data items. I know the parsing code is not easy on eye, but to my knowledge
> it works and I'd be glad to fix cases where it does not behave as
> advertised.
>
> I agree the formatting extension belongs to a subclass or perhaps some conditional_format(fmt,
> *a, **kw) function. I can do so if you anticipate a need for it outside
> the DataNaming class. Otherwise we may as well hold on with crossing that
> bridge.
>
> The mtime timestamps allow to find already exported files from databroker
> header regardless of their naming scheme. The users can easily clean things
> up with te(header, overwrite=1) if they happened to export under
> incorrect naming. They can also produce a list of header-tiffs with
> te.findExistingOutput(header) and use it however they see fit.
>
> Maybe I over-engineered and there is no compelling need for
> conditional-naming or file-timestamps functionality. Perhaps @sghose
> <https://github.com/sghose> or @sbillinge <https://github.com/sbillinge>
> could comment?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/NSLS-II/bluesky/pull/263#issuecomment-157535311>.
>



-- 
Prof. Simon Billinge
Applied Physics & Applied Mathematics
Columbia University
500 West 120th Street
Room 200 Mudd, MC 4701
New York, NY 10027
Tel: (212)-854-2918 (o) 851-7428 (lab)

Condensed Matter Physics and Materials Science Dept.
P.O. Box 5000
Upton, NY 11973-5000
(631)-344-5661

email: sb2896 at columbia dot edu
home: http:// <http://nirt.pa.msu.edu/>bgsite.apam.columbia.edu/

bluesky/broker_callbacks.py
```diff
@@ -25,9 +20,10 @@ class LiveImage(CallbackBase):
     relevant commit is a951b7.
     """
     def __init__(self, field):
+        from matplotlib.pyplot import figure
```
tacaswell: Please revert this change.  It is better to fail on import than first use.

pavoljuhas: Here is a use case that I try to address:  The user logs in over ssh or su and forgets to forward the DISPLAY.  This is not trivial - especially with su, where one needs to adjust the .Xauthority permissions.  The user wants to export some tiff files and does not need any plotting.  The user would get stuck, because the import of broker_callbacks would fail together with a top-level import of matplotlib.pyplot.

I feel it is better if user can accomplish jobs that can be accomplished without DISPLAY.

pavoljuhas: I was mistaken, matplotlib.pyplot can be imported without DISPLAY.  Fixed in 807a52b783d9bb7998780be5b33ed7cdcb9bc389.

tacaswell:
This code should probably go to something like xpdAcq, not in the main bluesky repo.

PR #265, opened 2015-11-17
Enh gs fig management

tacaswell:
closes https://github.com/NSLS-II/Bug-Reports/issues/61

attn @hanfeiyan

### DIFF
--- a/bluesky/callbacks.py
+++ b/bluesky/callbacks.py
@@ -118,9 +118,10 @@ class LivePlot(CallbackBase):
         super().__init__()
         if fig is None:
             # overplot (or, if no fig exists, one is made)
-            fig, ax = plt.gcf(), plt.gca()
-        else:
-            ax = fig.gca()
+            fig = plt.gcf()
+
+        ax = fig.gca()
+
         if legend_keys is None:
             legend_keys = []
         self.legend_keys = ['scan_id'] + legend_keys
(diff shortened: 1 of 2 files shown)


ericdill:
I was working with @lwiegart yesterday regarding a similar issue.  Lutz pointed out that there are three common ways that a live plot could behave.

1. Plot to the same figure by adding a new line to the existing lines and updating the legend. I believe that this is the current behavior of `gs.OVERPLOT = True`
2. Plot to the same figure by clearing the existing axes and plotting a new line.
3. Plot to a new figure and leave the existing figure. I believe that this is the current behavior of `gs.OVERPLOT = False`.

I implemented the following options in the IPython startup files at CHX in [this commit] (https://github.com/NSLS-II-CHX/ipython_ophyd/commit/e3d1c8bbf238516d7a10e03aea0fbd378c4a4f6f) in the file 99-bluesky.py as opposed to patching the bluesky source.

danielballan:
Nice to see sub_factories in the wild. Would you consider implementing the modes as strings or enums rather than ints?

danielballan:
In the terms @ericdill outlined above, @tacaswell's changes make `gs.OVERPLOT = True1` behave like (2) instead of (1). (Roughly speaking -- he makes a new figure if the axes labels need to change.)

I think this is a better default.

PR #268, opened 2015-12-03
DOC: Add docs for __call__ on _StepScan

ericdill:


ericdill:
dscan? now shows helpful info
```
In [6]: dscan?
Type:           DeltaScan
String form:    <bluesky.simple_scans.DeltaScan object at 0x104c4b320>
File:           ~/dev/python/bluesky/bluesky/simple_scans.py
Signature:      dscan(self, motor, start, finish, intervals, time=None, subs=None, **kwargs)
Docstring:      dscan (also known as lup)
Call signature: dscan(motor, start, finish, intervals, time=None, subs=None, **kwargs)
Call docstring:
Invoke the scan

Parameters
----------
motor
start : number
    The start point of the motion
finish : number
    The finish point of the motion
intervals : int
    The number of steps between `start` and `finish`
time : number
    The acquire time of the detector(s)?
subs : dict
    The temporary subscriptions to add to **this scan only**. These
    subscriptions are **not** persistent
```

ericdill:
Whereas previously:
```
In [3]: dscan?
Type:           DeltaScan
String form:    <bluesky.simple_scans.DeltaScan object at 0x104a1ad30>
File:           ~/dev/python/bluesky/bluesky/simple_scans.py
Signature:      dscan(self, motor, start, finish, intervals, time=None, subs=None, **kwargs)
Docstring:      dscan (also known as lup)
Call signature: dscan(motor, start, finish, intervals, time=None, subs=None, **kwargs)
```

bluesky/simple_scans.py
```diff
@@ -228,6 +228,23 @@ class _StepScan(_BundledScan):
...
+        ----------
+        motor
+        start : number
+            The start point of the motion
+        finish : number
+            The finish point of the motion
+        intervals : int
+            The number of steps between `start` and `finish`
+        time : number
+            The acquire time of the detector(s)?
```
ericdill: I'm not sure exactly how this works. Does this set **all** the acquire times of the detectors?

tacaswell: It is set through the magical DWIM interface which seems to do what _noone_ wants.  For AD it sets the total acquisition time to this value by changing the exposure time (iirc) but for scalers just sets the count time.

danielballan:
Good. Thanks for adding this.

ericdill:
Any blocking issues on this being merged?

tacaswell:
kicked to restart tests

ericdill:
It is literally only a documentation change

bluesky/simple_scans.py
danielballan: @ericdill I assumed you wanted to remove the '?' or elaborate before merging.

danielballan: (This is due for a refactor once ophyd is ready to re-deploy though, so we can just fix it at that time.)

PR #269, opened 2015-12-14
WIP: Add CHX spec callback

ericdill:
This is a callback that when instantiated with the following code in the ipython profile will dump spec files to disk for all scalar data that is coming in during the scan.  In addition, the position of **all** positioners is recorded at the start of the scan and included in the spec file.  Our users and beamline staff can then make use of all the tooling that is built around spec to help perform basic operations like fitting a curve to determine where to position the slits.

```python
from bluesky.callbacks import LiveSpecFile
live_specfile_callback = LiveSpecFile()
gs.specpath = os.path.expanduser('~/specfiles/test.spec')
gs.RE.subscribe('all', live_specfile_callback)
```

This PR as it stands now requires the metadata that is in the RunStart document that apparently only CHX is getting.  See https://github.com/NSLS-II/Bug-Reports/issues/70.  I have refactored it for CSX, as seen in NSLS-II-CSX/ipython_ophyd#1, but now I am not sure how much of that code is unique to CSX.  I could use some insight from @tacaswell @danielballan or @dchabot


danielballan:
I like this in the spirit of "build bridges, not walls."

Some quick thoughts on the implementation:
* Take a look at PeakStats for an example of a design that *can* use globalstate but can also be used in a more object-oriented way. I would like this construction to also be possible:

    ```python
    live_specfile_callback = LiveSpecFile(os.path.expanduser('~/specfiles/test.spec'))
    gs.RE.subscribe('all', live_specfile_callback)
    ```
* The massive ophyd refactor removes the session manager, but I think it still provides a way to ask for a list of the positioners. If not, we should certainly add one.

klauer:
Regarding the positioner list:
[ophyd.commands.get_all_positioners](https://github.com/klauer/ophyd/blob/ophyd0_rebase/ophyd/commands.py#L64)

ericdill:
I dont see anywhere in peak stats where global state is used. Csn you elaborate?

danielballan:
Right, PeakStats doesn't use GlobalState, but GlobalState can always carry a reference to a particular PeakStats instance. We don't ship one by default (maybe we should) but we monkey-patch one on in the CSX beamline config.

I'm suggesting:

```python
class LiveSpecFile(CallbackBase):
    def __init__(filepath):
       self.filepath = filepath
    ...
```

In the config:

```python
gs.specfile = LiveSpecFile(os.path.expanduser('~/specfiles/test.spec'))
```

Then, editing the filepath can be done like this:

```python
gs.specfile.filepath = 'newpath'
```

Thus, the notion of global state doesn't leak into the specfile-writing logic. And, potentially, other attributes (customizing the contents of the file in some way?) would then also be accessible as attributes `gs.specfile`.

danielballan:
@ericdill If you have to revisit the global state aspect and rebase, I'd be happy to see this merged for this cycle.

ericdill:
@danielballan Alright, it has been rebased and I removed the global state stuff.  This should be tested with Lutz at CHX. Whenever it passes muster with them, this is good to go.  

I am not having much luck coming up with useful tests for this code.  The only thing that I can think of would be to just run some scan through this callback and then save the output as part of the test data.  Then just run the same simulated scan through this callback and make sure it produces the same output?  Is that even a useful test?

bluesky/callbacks/core.py
```diff
@@ -540,3 +541,139 @@ def event(self, doc):
...
+            f.write('#P0 {0}\n'.format(' '.join(positions)))
+        # writing the list of motor names and their current values in
+        # a more human readable way. Apparently "#M" is not a used key in
+        # spec. Fingers crossed....
+        with open(self.specpath, 'a') as f:
+            for idx, (name, positioner) in enumerate(sorted(pos.items())):
+                f.write('#M%s %s %s\n' % (idx, name, str(positioner.position)))
+        print("RunStart document received in LiveSpecFile")
+        #raise
+        self.motorname = eval(doc['motor']).name
```
tacaswell: Is this going to create a motor instance to get it's name?

ericdill: I think so. Is there a better way to get the name?

klauer: I'm not saying this is pretty, sensibly named, or how you should do it, but I mocked up "`NamedObjects`" to extract object names from bluesky `scan_args` in hxntools like so:
```python
def _eval(scan_args):
    '''Evaluate scan arguments, replacing OphydObjects with NamedObjects'''

    class NamedObject:
        def __init__(self, name):
            self.name = name

    def no_op():
        def no_op_inner(*args, name=None, **kwargs):
            if name is not None:
                return NamedObject(name)

        return no_op_inner

    return eval(scan_args, collections.defaultdict(no_op))
```

danielballan:
About tests, I'm comfortable leaving this untested. A blunt test of equality with a block of test is likely to fail on a false alarm. If it becomes a problem we can revisit.

bluesky/callbacks/core.py
danielballan: I have to say this is pretty damn clever. It's a little scary, but definitely *less* scary that instantiating live Devices.

danielballan:
@ericdill We are deploying at CHX this afternoon. If you want to incorporate @klauer's trick, we can merge at that point.

ericdill:
Ok, so I got rid of the `eval(hdr.start.motor)`, in favor of using `ast` to parse the eval-able code in start.motor.  I see that currently `scan_args` is being captured as a key in RunStart, but also being unpacked into the top level of the RunStart so that RunStart.scan_args.motor exists, but so does RunStart.motor.  Is one of these methods of accessing the scan args deprecated?

I am still eval-ing the detectors (line 647) so that I can get the acquire time.  I need an active object to get this acquire time because it is not part of the kwarg list that bluesky dumps into the RunStart document.  I could get around this by doing something like:

(forgive any code errors, this is being typed direectly into the github comments section)
```python
tree = ast.parse(doc['detectors'])
self._name_finder.visit(tree)
detname = self._name_finder.name
self.acquisition_time = get_ipython().user_global_ns[detname].acquire_time
```

I suppose it is important to note that doing it this way would not work if the user wanted to use this callback outside of an active data collection ipython instance to output their scan data into a spec file for whatever reason (hey, I'm not judging if the user wants to use something outside of our stack!).  In the case where get_ipython() fails, we could still fall back to just eval-ing the detector I suppose.  Though now I'm a little confused, does the detector "remember" what it's acquire time was?  Oh wait, @klauer didn't you just close an issue from Hanfei regarding something about the acquire time?  Are we stashing that somewhere in the RunStart or Descriptor documents now?

klauer:
@ericdill referring to https://github.com/NSLS-II/wishlist/issues/93, this is fixed only locally for the HXN (on the previous ophyd/bluesky versions). I'm not aware of a fix for the new version on all beamlines as of yet. 

Here's my hxn-only branch: https://github.com/klauer/bluesky/commit/ca5c493f6824e65c06574a0cada32d0f16319c7e

ericdill:
Ah ok, thanks @klauer . IIUC this put the acquire time into the repr?  It would then naturally go into the RunStart, right?

klauer:
It sends off the `time` argument to the RunEngine as `exposure_time`, which interprets it as a custom field to be put into the RunStart document.

ericdill:
Ah ok, thanks.  I would imagine this will go in to bluesky at some point?

ericdill:
@danielballan how did the testing of this go at CHX the other day?

danielballan:
No chance to test it. We still need to get their Eigers set up.

danielballan:
I have sketch of how to do this without touching IPyhon or ophyd or making live hardware objects on my spec-callback branch. It's not quite there, but it has all the necessary pieces from bluesky and Documents. Care to try to finish it off?

ericdill:
@danielballan In theory, this now works... :rocket: 

https://gist.github.com/ericdill/2056d44123627f7853f1

Oh wait, I still need to format the scan command. Stand by. :clock10:

danielballan:
Thanks for pushing this across the finish line. Is this good to merge?

ericdill:
As far as I can tell it is. Pymca correctly loads these spec files

danielballan:
Great. Thanks for testing it. And thanks for jumping back in to craft my approach into something that actually works.

Just for the record, despite co-authoring this, I am a little hesitant to merge this callback into bluesky on principle because I don't necessarily want DAMA to own the responsibility for maintaining it long-term.

Our group does and should work assiduously to make this kind of thing *possible* through carefully designing flexible interfaces. But presently we do not have the manpower to build and maintain a large variety of bespoke document processors. In my opinion, we need to strive to make beamline scientists, post-docs, and Python-savvy users the primary authors of downstream utilities like this -- while of course providing support.

That said, this is needed and we have it, so I'll step down from my soap box and click the green button.

ericdill:
I am :+1: on your soapbox

PR #270, opened 2015-12-14
ENH: Add example for customized Olog callback

ericdill:
This is how I customized the logbook entries at CHX.  I am confident that this is not what everyone else will want, but it does serve as a template with which others could modify this to suit their needs.  Thoughts?

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -6,7 +6,6 @@ from itertools import count
 from collections import deque
 import warnings
 from prettytable import PrettyTable
-from pprint import pformat
 
 import matplotlib.pyplot as plt
 import matplotlib.colors as mcolors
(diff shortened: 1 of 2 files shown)


tacaswell:
Can you move this into https://github.com/NSLS-II/bluesky/blob/master/bluesky/callbacks/olog.py and rebase?

ericdill:
@tacaswell rebased and moved to olog.py.  Not sure if this is still terribly useful if you can just make a jinja template to do this.

tacaswell:
This way is conceptually simpler (no templating, no function factories).
Has obvious path to add olog integration on other documents.

If this does not go here, it should go in examples some place.

ericdill:
Ah ok, that makes sense.  If it makes sense for it to be in the library then let's just merge this in if it looks ok

PR #272, opened 2015-12-17
REF: Get schemas from external repo.

danielballan:
Wait to merge this until the new event_model repo is ready and built.


danielballan:
https://github.com/NSLS-II/event-model

cowanml:
This is on the road to what I've been thinking for "Types", where the schema
is stored in, I guess it would be "typestore"...

Would enable us to deal with different types of: analysis results,
samples/containers/requests, etc.

danielballan:
OK, good. Maybe you could start by copying this repo and making a `sample_model` with a `sample.json`, `container.json`, etc.

danielballan:
OK, this is finally ready. The event-model package is available internally on nsls2-tag and externally on anaconda.org/lightsource2. (Thanks again, @ericdill.) Travis passes. It's one more package to juggle, but it means changes occur in one place instead of three (mdserver, mdstore, bluesky).

Since this is critical, I'd appreciate eyes and :+1:s from @klauer, @tacaswell, and @arkilic before we merge.

danielballan:
@ericdill Please merge [this](https://github.com/NSLS-II/staged-recipes-dev/pull/3) first so that the next nsls2-dev build for bluesky includes the event-model dependency.

tacaswell:
:+1:

setup.py
```diff
@@ -15,7 +15,7 @@
     package_data={'bluesky': ['schema/*.json']},
     install_requires=['jsonschema', 'traitlets', 'cycler',
                       'numpy', 'matplotlib', 'super_state_machine',
-                      'historydict'],
+                      'historydict', 'event_model'],
```
ericdill: this needs to be `event-model`

danielballan: Hmm. Then why do the tests pass? I think `event_model` is correct -- we want the Python package name, not the conda package name, and `event-model` is not a valid Python package name. This is confusing, and I'm not 100% sure I'm right about this, but I were wrong I would expect the Travis build to fail when bluesky is installed.

ericdill: Is event_model on pypi?

tacaswell: I am with Dan, if this were wrong I would expect the install to fail, but if this is right, there is evil magic / heuristics going on

ericdill: You guys are right. *If* we include `event-model` then it needs a hyphen (`-`).  Given that we do not have it as a pypi package, we should not include it in the install_requires section.  For example:

```
conda create -p /tmp/foo bluesky python=3.5
source activate bluesky
conda remove bluesky
cd /path/to/this/bluesky/branch
python setup.py develop
```
The above will barf because it cannot find event-model on pypi:

```
$ psd
running develop
running egg_info
writing bluesky.egg-info/PKG-INFO
writing requirements to bluesky.egg-info/requires.txt
writing top-level names to bluesky.egg-info/top_level.txt
writing dependency_links to bluesky.egg-info/dependency_links.txt
reading manifest file 'bluesky.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'bluesky.egg-info/SOURCES.txt'
running build_ext
Creating /tmp/foo/lib/python3.5/site-packages/bluesky.egg-link (link to .)
bluesky 0.4.3+87.gd35b09f.dirty is already the active version in easy-install.pth

Installed /home/edill/dev/python/bluesky
Processing dependencies for bluesky==0.4.3+87.gd35b09f.dirty
Searching for event-model
Reading https://pypi.python.org/simple/event-model/
Couldn't find index page for 'event-model' (maybe misspelled?)
Scanning index of all packages (this may take a while)
Reading https://pypi.python.org/simple/
No local packages or download links found for event-model
error: Could not find suitable distribution for Requirement.parse('event-model')
```

Therefore I would strongly suggest that we do not include this package in the setup.py.  We *could* also publish it to pypi, but then we have issues connecting to pypi from behind the NSLS-II firewall.  Overall, I think this causes more trouble than it solves.

ericdill: Oh, sorry, the above was with my edits to setup.py for `event-model` instead of `event_model`.  Though the result is the same with `event_model` in the install_requires section:

```
(/tmp/foo)edill@edill-810g:~/dev/python/bluesky (remove-schema $ u=)$ psd
running develop
running egg_info
writing top-level names to bluesky.egg-info/top_level.txt
writing dependency_links to bluesky.egg-info/dependency_links.txt
writing requirements to bluesky.egg-info/requires.txt
writing bluesky.egg-info/PKG-INFO
reading manifest file 'bluesky.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'bluesky.egg-info/SOURCES.txt'
running build_ext
Creating /tmp/foo/lib/python3.5/site-packages/bluesky.egg-link (link to .)
bluesky 0.4.3+87.gd35b09f is already the active version in easy-install.pth

Installed /home/edill/dev/python/bluesky
Processing dependencies for bluesky==0.4.3+87.gd35b09f
Searching for event_model
Reading https://pypi.python.org/simple/event_model/
Reading https://pypi.python.org/simple/event-model/
Couldn't find index page for 'event_model' (maybe misspelled?)
Scanning index of all packages (this may take a while)
Reading https://pypi.python.org/simple/
^[	No local packages or download links found for event_model
error: Could not find suitable distribution for Requirement.parse('event_model')
```

danielballan: The advantage of including `event-/_model` in `install_requires`, even though it's not on PyPI, is that it blows up at install time instead of at run time. This not a very big advantage, so I think we can remove it for now. We are inconsistent with `install_requires` in all our projects (e.g., databroker, to name one). I will remove `event_model`. Perhaps we should revisit the PyPI question offline.

klauer:
Looks good to me.

PR #274, opened 2016-01-15
Add pre_run, post_run hooks

danielballan:
This provides a nice way to add messages to the beginning or end of a plan, perhaps nicer than making a subclass.


bluesky/scans.py
```diff
@@ -64,18 +68,34 @@ def __iter__(self):
...
-            yield Msg('configure', obj, conf)
+        """
+        Subclasses use this to inject some messages to be processed
+        before the run is opened -- for example, configuration or
+        taking some preliminary readings needed for metadata.
+        """
+        # prerun is expected to be a callable that takes the Scan object
+        # itself as its argument and returns a generator of messages
+        if self.pre_run is not None:
+            yield from self.pre_run(self)
```
tacaswell: We had discussed doing this as `self.pre_run(self, *self.stashed_args, **self.stashed_kwargs)` to allow injecting arbitrary logic into these functions

danielballan: As suggested by @klauer, I would rather make the user refactor their function so that

```python
f(scan, *args, **kwrags)
   ...
```

becomes

```python
f(scan)
   args = scan.stashed_args
   kwrags = scan.stashed_kwargs
   ...
```

Less API surface area is good.

bluesky/utils.py
```diff
@@ -274,21 +275,25 @@ class Struct(metaclass=StructMeta):
...
 
     def __init__(self, *args, **kwargs):
+        # Now bind default values of optional arguments.
+        # If it seems like there should be a cleaner way to do this, see
+        # http://bugs.python.org/msg221104
         bound = self.__signature__.bind(*args, **kwargs)
+        for name, param in self.__signature__.parameters.items():
+            if (name not in bound.arguments and
+                    param.default is not inspect._empty):
+                bound.arguments[name] = param.default
```
tacaswell: should this raise if there is a missing value + no default?

bluesky/scans.py
tacaswell: I do not see where the args/kwargs get stashed?  It looks like `Struct` will only save args/kwargs it knows about.  I thought the point of this was to let the user inject semi-arbitrary stuff.  Suspect that this is a rabbit hole.

klauer: I think the point is 'stash them if you want to and how you want to'. The original suggestion of a tuple `(fcn, args, kwargs)` for both pre and post-run is annoying to deal with (in my opinion).

tacaswell: Given that this exists to make @klauer life easier I am going to stop arguing

bluesky/utils.py
danielballan: That is the job of `bind`.

PR #275, opened 2016-01-15
Rename scan -> plan

danielballan:
Renamed, at the request of @klauer, to differentiate: a *plan* is an iterable object that yields messages. A *scan* is a call to process those messages.


danielballan:
I did some careful `git commit -p` work here and I have decent faith in the tests. The docs will need work -- didn't want to apply sed heavily to that.

bluesky/plans.py
```diff
@@ -98,12 +98,12 @@ def _post_scan(self):
             yield from self.post_run(self)
 
     def _gen(self):
-        raise NotImplementedError("ScanBase is a base class, you must "
+        raise NotImplementedError("PlanBase is a base class, you must "
```
tacaswell: this means we can not do MI with these classes!

bluesky/plans.py
```diff
@@ -161,8 +161,8 @@ def _gen(self):
             yield Msg('sleep', None, delay)
 
 
-class Scan1D(ScanBase):
-    "Use AbsListScan or DeltaListScan. Subclasses must define _abs_steps."
+class Scan1D(PlanBase):
```
tacaswell: Scan1D -> Plan1D?

bluesky/plans.py
```diff
@@ -698,7 +698,7 @@ def _gen(self):
             self.output_mutable['model'] = res
 
 
-class ScanND(ScanBase):
+class ScanND(PlanBase):
```
tacaswell: ScanND -> PlanND

bluesky/plans.py
danielballan: PRs welcome

bluesky/plans.py
```diff
@@ -736,8 +736,8 @@ def _gen(self):
             yield Msg('save')
 
 
-class _OuterProductScanBase(ScanND):
-    # We define _fields not for Struct, but for ScanBase.log* methods.
+class _OuterProductPlanBase(ScanND):
+    # We define _fields not for Struct, but for PlanBase.log* methods.
```
tacaswell: didn't we do in those methods?

klauer:
:+1: for slightly less confusion

klauer:
How about `simple_scans.py` -> `scans.py`?

danielballan:
@klauer That rename is pretty simple, but it's worrying that `scans.AbsScan` in the current master would be replaced by `scans.AbsScan` that is actually a different class. Could we do this later -- say in a cycle or so -- after beamline configs and user code have been updated to scans -> plans?

klauer:
You're absolutely right, but you'll get much more flak for bikeshedding/renaming things as the userbase grows. Either way, this PR is already an improvement.

tacaswell:
If we do want to do the `simple_scan` -> `scan` rename before the beam comes up we should do it in a second pass around the ring once we have updated everything else in their beam line configs to work.

PR #276, opened 2016-01-15
FIX: scans->plans over-aggressive sed job

klauer:
Might be more, but this is what I've found so far


danielballan:
Gah. Sorry. I am pretty sure those objects are not referenced outside of those two files. This should be it.

PR #277, opened 2016-01-17
history.History -> historydict.HistoryDict

danielballan:
Also, now that historydict is on PyPI, we can include it as a dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,8 @@ setup(
               'bluesky.callbacks'],
     package_data={'bluesky': ['schema/*.json']},
     install_requires=['jsonschema', 'traitlets', 'prettytable', 'cycler',
-                      'numpy', 'matplotlib', 'super_state_machine'],
+                      'numpy', 'matplotlib', 'super_state_machine',
+                      'historydict'],
     classifiers=[
         "Development Status :: 3 - Alpha",
         "Programming Language :: Python :: 3.4",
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -21,7 +21,8 @@
               'bluesky.callbacks'],
     package_data={'bluesky': ['schema/*.json']},
     install_requires=['jsonschema', 'traitlets', 'prettytable', 'cycler',
-                      'numpy', 'matplotlib', 'super_state_machine'],
+                      'numpy', 'matplotlib', 'super_state_machine',
+                      'historydict'],
```
ericdill: I am 95% sure that all of these should be in the `build` section of the `meta.yaml`.

danielballan: Good catch. Updated history -> historydict in new commit. I think everything else in the recipe is already up to date. (Does it have to be the `build` section? I think the `run` section works -- at least that's how it is set up currently.)

tacaswell: I do not think _any_ of these should be in the `install_requires` section, but should all be the `requires` section.  The `install_requires` is for stuff that is imported in `setup.py` (ex if you are build c-extensions with numpy you need to have numpy installed at install/build time to be able to use numpy to find it's headers (and to have the header)).

danielballan: @tacaswell I don't think that's right. I think you are thinking of `setup_requires`. The terminology is awful, but I believe `install_requires` the most closely means "runtime requirements." [Ref](https://pythonhosted.org/setuptools/setuptools.html#new-and-changed-setup-keywords)

tacaswell: Fair enough.  :sheep:

ericdill: Does conda build for this recipe work? I remember getting errors in conda build about `downloading with setuptools is disabled during conda build` but I never tracked down exactly what caused it. I know that removing a bunch of stuff from `install_requires` made the error go away. That was the source of my initial comment

PR #280, opened 2016-01-19
Plans are iterables

danielballan:
In two ways, the RunEngine breaks design principle, "plans are just iterables." The RunEngine looks for an optional `md` attribute (with metadata it feeds to `open_run`) and it looks for a optional `subs` attribute (with callbacks that it subscribes and then unsubscribes).

But if a user composes two plans:

```python
from itertools import chain
master_plan = chain(ct(), ascan(motor, 1, 5, 5), ct())
```

or tacks an extra message onto a plan:

```python
modified_plan = chain([Msg('configure', det, {'num_points': 5})], ascan(motor, 1, 5, 5))
```

the result is a generator, and the original `subs` and `md` attributes are not accessible to the RunEngine.

Fortunately, there is a better way: `PlanBase` can inspect its own `subs` and `md` and send them to the RunEngine by yielding `Msg('subscribe', ...)` and `Msg('open_run', ...)`. This way, the RunEngine does not need to know about these special attributes.

This PR mostly involves cutting code from `run_engine.py` and moving it into `plans.py`. It simplifies the RunEngine, and it brings us closer to that "Zen of Python" principle, "There should be one-- and preferably only one --obvious way to do it." The only way to set metadata or subscriptions from a plan is via a Message.

I really want users to be using `chain`. It provides a relatively simple way to start making custom plans without starting from scratch. And it provides a simple way to specify multi-run jobs, handing supervision to the RunEngine rather than writing custom error-handling in every user script.


bluesky/run_engine.py
```diff
@@ -1109,6 +1095,21 @@ def _subscribe(self, msg):
         return token
 
     @asyncio.coroutine
+    def _unsubscribe(self, msg):
```
tacaswell: Does this need an entry in the default mapping?

bluesky/plans.py
```diff
@@ -51,16 +51,35 @@ def md(self):
...
+        for name, funcs in subs.items():
+            if not isinstance(funcs, Iterable):
+                # Take funcs to be a single function.
+                funcs = [funcs]
+            for func in funcs:
+                if not callable(func):
+                    raise ValueError("subs values must be functions or lists "
+                                     "of functions. The offending entry is\n "
+                                     "{0}".format(func))
+                token = yield Msg('subscribe', name, func))
```
tacaswell: and one too many `)`

bluesky/run_engine.py
```diff
@@ -1109,6 +1095,21 @@ def _subscribe(self, msg):
...
+        """
+        Remove a subscription during a call -- useful for a multi-run call
+        where subscriptions are wanted for some runs but not others.
+        """
+        logger.debug("Removing subsription %r", msg)
```
klauer: `subsription`

klauer:
Any changes requiring less searching on my part between the plan/scan/runengine files get a +1 from me. Much easier to figure out flow and how things work this way.

bluesky/run_engine.py
danielballan: Yes. Done.

bluesky/plans.py
```diff
@@ -51,16 +51,27 @@ def md(self):
...
         """
-        yield from self._pre()
         # Some metadata is compute in _pre, so it must be done before
         # the RunStart is generated by open_run, below.
-        yield Msg('open_run')
+        yield from self._pre()
+
+        # If the plan has a 'subs' attribute, yield a 'subscribe' Msg for
+        # each one. The RunEngine will remove these at the end if the
+        # event of termination.
```
dchabot: "..._in_ the event of termination." ?

tacaswell:
https://github.com/NSLS-II/bluesky/pull/280/files#diff-3e4b24e3ff848e88df36ab0f5cd648bcL42 these docs now lie

danielballan:
fixed

PR #282, opened 2016-01-20
HOTFIX: Apply key-guessing heuristic to more places

danielballan:
Following up on #281, a little less hack-ish. Bluesky looks at the *order* of `read_attrs` and assumes the first one is the one it should use to (1) compute relative trajectories and (2) plot against in the subscription factories in `simple_scans.py`.

### DIFF
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -393,7 +393,8 @@ def first_key_heuristic(device):
     """
     Get the fully-qualified data key for the first entry in read_attrs.
 
-    This will raise is that entry is not a Signal.
+    This will raise is that entry's `describe()` method does not return a
+    dictionary with exactly one key.
     """
     first_attr = device.read_attrs[0]
     key, = getattr(device, first_attr).describe().keys()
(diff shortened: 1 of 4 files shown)


bluesky/utils.py
```diff
@@ -387,3 +387,29 @@ def snake_cyclers(cyclers, snake_booleans):
             expanded = v2[:total_length]
             new_cyclers.append(cycler(k, expanded))
     return reduce(operator.add, new_cyclers)
+
+
+def first_key_heuristic(device):
+    """
+    Get the fully-qualified data key for the first entry in read_attrs.
+
+    This will raise is that entry is not a Signal.
```
tacaswell: will raise if entry has not exactly 1 key in describe.

danielballan: True, I shouldn't refer to `Signal` here.

klauer:
I think this is the path of least resistance, so for now it's good enough.

A minor side-note: `Device.read` in ophyd returns `OrderedDict`s whereas `Signal.read` returns a dict. Maybe we should be more strict about that API - then `read_attrs` would not need to be checked by bluesky.

danielballan:
@klauer That sounds good. Please merge?

PR #288, opened 2016-01-24
FIX: Do not hold onto a stale run_uid after run is closed.

danielballan:
If the RunEngine goes back to a checkpoint inside a run after processing a `close_run`, it attempts to write a second RunStart Document (and then fails on a MDS exception if MDS is hooked up). 
We did anticipate this problem, as you can see in the code, but we didn't quite handle it right.

Notice that two RunStop documents with the same 'start' uid are generated:

```python
In [50]: RE([Msg('open_run'), Msg('checkpoint'), Msg('close_run'), Msg('pause')])
(...)
*** Emitted RunStop:
{'time': 1453652515.306092, 'run_start': '00587fb4-816e-42eb-95aa-c5c46c3c2d0f', 'uid': 'df48f306-4e87-4c6d-87d4-7141e1ecb9fa', 'reason': '', 'exit_status': 'success'}
RE.state: running
msg: close_run: (None), (), {}
  response: None
About to process: <bound method RunEngine._pause of <bluesky.run_engine.RunEngine object at 0x104e5b080>>, pause: (None), (), {}
Pausing...
RE.state: paused
msg: pause: (None), (), {}
  response: None
Out[50]: ['00587fb4-816e-42eb-95aa-c5c46c3c2d0f']

In [51]: RE.resume()
About to process: <bound method RunEngine._close_run of <bluesky.run_engine.RunEngine object at 0x104e5b080>>, close_run: (None), (), {}
*** Emitted RunStop:
{'time': 1453652522.148735, 'run_start': '00587fb4-816e-42eb-95aa-c5c46c3c2d0f', 'uid': '409eb109-ff6f-4061-b87c-853f7d6fba65', 'reason': '', 'exit_status': 'success'}
RE.state: running
msg: close_run: (None), (), {}
  response: None
Out[51]: ['00587fb4-816e-42eb-95aa-c5c46c3c2d0f']
```

With this PR, the same process now raises:

```python
IllegalMessageSequence: A 'close_run' message was received but there is no run open. If this occurred after a pause/resume, add a 'checkpoint' message after the 'close_run' message.
```

As an aside, we could correct plans on the fly and prevent users from hitting this error, but I would rather do that in the future with a "preprocessor" generator than with yet more state in the RunEngine.


bluesky/run_engine.py
```diff
@@ -270,7 +270,7 @@ def _clear_run_cache(self):
         self._metadata_per_run.clear()
         self._bundling = False
         self._run_is_open = False
-        self._msg_cache = None  # checkpoints can't rewind into a closed run
```
danielballan: We *do* want to support checkpoints between runs during a multi-run plan. ("If interrupted redo the last 5 samples....") It's better to handle the rewinding issue in `_close_run` itself.

tacaswell: why generate a new uid now, rather that when the runstart is created?

danielballan:
The "suspenders" tests are being flaky again (ran twice, failed both times) but otherwise Travis passes.

bluesky/run_engine.py
danielballan: This could be `self._run_start_uid = None` if you prefer. The point is to erase the old uid -- might as well make the next one while we are here. But now that I say this I think `None` is safer.

danielballan: In effect we should have `self._run_is_open == self._run_start_uid is not None` at all times.

tacaswell: We should probably reset this to `None` here and add checks other places we use it.

danielballan:
OK, now `self._run_is_open` is a property that reflects whether `self._run_start_uid is None`, and every method / coroutine that uses that uid first checks that a run is open.

bluesky/run_engine.py
```diff
@@ -199,7 +199,6 @@ def __init__(self, md=None, *, md_validator=None):
         self._metadata_per_run = {}  # for one run, incorporating Msg metadata
         self._panic = False
         self._bundling = False  # if we are in the middle of bundling readings
-        self._run_is_open = False  # if we have emitted a RunStart, no RunStop
```
tacaswell: sweet, nuked duplicate state!

bluesky/run_engine.py
```diff
@@ -807,12 +810,18 @@ def _open_run(self, msg):
 
     @asyncio.coroutine
     def _close_run(self, msg):
+        if not self._run_is_open:
+            raise IllegalMessageSequence("A 'close_run' message was received "
+                                         "but there is no run open. If this "
+                                         "occurred after a pause/resume, add "
+                                         "a 'checkpoint' message after the "
+                                         "'close_run' message.")
```
tacaswell: maybe close run should always _create_ a checkpoint?

danielballan: If close_run always creates a checkpoint, then we rule out the possibility of placing checkpoints, say, between every five runs of a multi-sample plan. A more useful thing to do would be to insert a checkpoint at the end of a run *if and only if that run contained a check point*. But I guess I would rather do this with a "preprocessor" than in the RunEngine. So far the RunEngine is not in the business of fixing plans on the fly, and doing so would add even more state.

tacaswell: fair enough.

tacaswell:
Need to remove all setting of `self._run_is_open`

bluesky/run_engine.py
danielballan: ...but I did just add a checkpoint message at the very end of `PlanBase`, so our built-in plans will address the common case.

danielballan:
Thanks. All tests except suspenders pass again.

bluesky/plans.py
```diff
@@ -81,6 +81,7 @@ def __iter__(self):
         for token in tokens:
             yield Msg('unsubscribe', token)
         yield from self._post()
+        yield Msg('checkpoint')
```
tacaswell: is planbase otherwise emitting checkpoints?

tacaswell: Do we also have an edge case with `chekpoint->open_run->pause` where we will double emit the run start?

Do we need to add a `clear_checkpoint` message?

danielballan: > is planbase otherwise emitting checkpoints?

No, but all the mid-level base classes (Plan1D, PlanND, _AdaptivePlanBase) do, so all the plans do.

> Do we need to add a clear_checkpoint message?

Yes.

PR #244, opened 2015-10-26
For now, don't implicitly update RE.md

danielballan:
In #240, several of us agreed that the *default* behavior should be that kwarg-based metadata should not implicitly update persistent metadata. Some advocated for a switch to turn implicit updating on and off. I came out somewhat against this. I would reconsider if users seem to want it. So far, the only users I've heard from want the explicit way, so that is what this PR does.

PR #289, opened 2016-01-24
WIP: Switching to pytest.

ericdill:
Opening a PR so I can see how travis is behaving.  This is still a WIP.

@danielballan, can you leave a note in this PR about sending random messages
to the run engine with pytest?


danielballan:
On Slack, I suggested that it would be interesting to "noise test" the RunEngine: that is, generate random sequences of Messages and make sure the RunEngine always lands in a recoverable state. Maybe py.test offers nice ways to do this kind of thing.

To be clear, I am not suggesting that this belongs in our automated test suite. We do not want randomized tests running on Travis. Rather it should be a script that we run manually -- perhaps 100, 1000, 10000 times -- to discover edge cases, leading to bug fixes and then *targeted* tests for those cases in the automated test suite.

Something like:

```python
def smoke_test(plan)
    try:
        RE = setup_test_run_engine()
        RE(plan)
        while True:
            if RE.state == 'paused':
                RE.resume()
            elif RE.state == 'idle':
                break
    except IllegalMessageSequnece:
        # This is the only kind of error we should except. Swallow it.
        # Any other errors probably indicate a flaw in the RunEngine's logic.

for i in range(100):
    random_plan = ...
    try:
        noise_test(plan)
    except Exception:
        # re-raise for inspection, or log the traceback and `random_plan` to a file
```

ericdill:
Thanks for adding this and thanks for clarifying because I was definitely planning on doing this with travis. Though as you point out, random bugs on travis sounds terrrrrible

cowanml:
fuzzing! cool!

bluesky/tests/test_examples.py
```diff
@@ -166,58 +154,68 @@ def print_event_time(name, doc):
...
-    assert_raises(KeyError, RE, scan, owner='dan')
+    with pytest.raises(KeyError):
+        RE(scan, owner='dan')
     scan = simple_scan(motor)
     RE(scan, owner='dan', beamline_id='his desk',
        group='some group', config={})  # this should work
     scan = simple_scan(motor)
-    assert_raises(KeyError, RE, scan)  # this should fail; none was persisted
+    with pytest.raises(KeyError):
+        # TODO: Is this persistent thing still true?  I thought that something
```
danielballan: Good question, but yes it is still true. You are probably thinking of #244. This aspect has not changed.

bluesky/tests/test_examples.py
```diff
@@ -360,24 +359,24 @@ def sim_kill():
...
+    assert RE.state == 'idle'
     stop = ttime.time()
 
     assert mid - start > 1
     assert stop - start < 2
 
-
 def test_abort():
-    raise KnownFailureTest
+    pytest.xfail("This needs a better reason why we know it fails."
```
danielballan: This is for some reason not successful on OSX. But the issue is not reproducible interactively. I concluded it was an issue with nose. Perhaps py.test fares better?

bluesky/tests/test_examples.py
ericdill: Yes that is exactly what I was thinking of.  Great, thanks.  Removing this comment in a future commit

danielballan:
Yeoman's work. Thanks @ericdill, this is indeed simpler.

PR #290, opened 2016-01-26
ENH: a few simple tools for inspecting plans

tacaswell:
These should be a step up from just `list(plan)`
```python
from bluesky.examples import motor, motor1, det, det1
from bluesky.plans import OuterProductAbsScanPlan
plan = OuterProductAbsScanPlan([det, det1], motor, -1, 1, 10, motor1, -5, -3, 15, 1)

path_plotter(plan, 'motor', 'motor1', probe_size=.1)
```
so

A bunch of nicities missing (like axes labels, where the start is, legend, etc) but this gets the general idea across.

### DIFF
--- a/bluesky/plan_tools.py
+++ b/bluesky/plan_tools.py
@@ -1,0 +1,82 @@
+import matplotlib.pyplot as plt
+from matplotlib import collections as mcollections
+from matplotlib import patches as mpatches
+
+
+def path_plotter(plan, x_motor, y_motor, ax=None, probe_size=None):
+    """Plot the raster path for this plan
+
+    Parameters
+    ----------
+    plan : iterable
+       Must yield `Msg` objects and not be a co-routine
+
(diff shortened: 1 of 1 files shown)


danielballan:
Very nice. It would be nice to have 1D version as well.

klauer:
Looks cool!

bluesky/plan_tools.py
```diff
@@ -0,0 +1,82 @@
+import matplotlib.pyplot as plt
+from matplotlib import collections as mcollections
+from matplotlib import patches as mpatches
+
+
+def path_plotter(plan, x_motor, y_motor, ax=None, probe_size=None):
```
danielballan: Can I bikeshed this name to `plot_path`? This sounds like a returns a factory. I'm perfectly happy with the module name `plan_tools` though.

PR #291, opened 2016-01-26
REF: Companion PR to ophyd StatusBase API change

danielballan:
Changes to `run_engine.py` are minimal. Most of this is cleaning up old "done" logic in `examples.py`, some of which predates status objects.


tacaswell:
Both of these PRs need a rebase.

danielballan:
😢

PR #292, opened 2016-01-26
ENH: add clear_checkpoint message

tacaswell:
remove a checkpoint with out creating a new one


danielballan:
This looks good. I guess you should remove the aborted attempt to fix the suspenders tests -- we'll leave marking them as known failures to @ericdill and his py.test overhaul.

PR #296, opened 2016-01-27
Explicitly import scans into top level

ericdill:


ericdill:
This is per an in-person conversation with @tacaswell . Options seemed to be to 

1. add a `__all__` to the `plans.py` module 
2. just explicitly import everything into the top level `__init__.py`. 

@tacaswell was in favor of just explicitly importing into the top level, so this PR is doing that.

The test failures are from the suspenders tests and will be fixed as soon as #289 is in

klauer:
By 'scans' you mean 'plans', ... right?

ericdill:
Yeah I suppose I do :cry:

PR #300, opened 2016-01-28
ENH/TST: Warn if subscriptions are producing errors.

danielballan:


danielballan:
The passed except for the LiveTable test, which we have seen is unstable under py.test. I wonder if we are encountering the lossiness of event subscriptions and missing a row.

Anyway, I think this is good to merge once someone reads through it.

PR #304, opened 2016-01-29
Include object names in Start document

danielballan:
This a change to PlanBase. ~~I'd like to get it in before users start taking data tomorrow.~~ (There are no users tomorrow after all.)

It would be nice to be able to search on object name:

```python
db(start_time='2016-01-19', detectors='fccd')
```

where mongo should interpret this as, "all of yesterday's scans that included that fccd as a detector."

This PR adds top-level fields to the Start Document for every category of object ('motor' or 'motors', 'detector' or 'detectors'). This is addition to storing the full reprs in 'plan_args', as we have been doing for some time now.

This also strips out some gratuitous complication from PlanBase. The 'derived_fields' concept is no more. It was supposed to make custom metadata "easy," but it really only made things more complicated.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -24,9 +24,6 @@ class PlanBase(Struct):
       set, which is used to stage/unstage everything
     - a class level ``_fields`` attribute which is used to
       construct the init signature via meta-class magic
-    - optionally, a class level ``_derived_fields`` attribute which is used to
-      indicate other attributes, beyond _fields, which should be included in
-      metadata
 
     If you do not use the class-level ``_fields`` and write a custom
     ``__init__`` (which you need to do if you want to have optional kwargs)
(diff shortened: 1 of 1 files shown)


danielballan:
P.S. This is what the RunStart document looks like with this PR. Unrelated: I really like that `RE(plan, print)` just works.

```python
In [8]: RE(plan, print)
start {'time': 1454099701.357652, 'scan_id': 2, 'owner': 'test_owner', 'group': 'Grant No. 12345', 'plan_type': 'AbsScanPlan', 'uid': '4e0e91f6-3468-4b69-9ee3-06932630491f', 'plan_args': {'num': '5', 'start': '1', 'detectors': '[reader: det]', 'motor': 'mover: motor', 'stop': '5'}, 'detectors': ['det'], 'beamline_id': 'test_beamline', 'motor': 'motor', 'config': {'detector_model': 'XYZ', 'pixel_size': 10}}
```

PR #306, opened 2016-01-30
ENH: Support infinite ct: Count(num=None)

danielballan:
This is a little piece of SPEC functionality that we are missing. The intention here is that the scan will end with a user-initiated pause + stop or abort.

@cmazzoli and @ambarb asked for `ct(num=0)`, but in Python we prefer not to use numbers as sentinel values. Instead we can use `ct(num=None)`, which is, if you think about it, a more accurate description of what is happening. I hope this is suitable for you. :- )

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -73,11 +73,11 @@ class PlanBase(Struct):
         yield Msg('open_run', **self.md)
         for flyer in self.flyers:
             yield Msg('kickoff', flyer, block_group='_flyers')
-            yield Msg('wait', None, '_flyers')
+        yield Msg('wait', None, '_flyers')
         yield from self._gen()
         for flyer in self.flyers:
             yield Msg('collect', flyer, block_group='_flyers')
-            yield Msg('wait', None, '_flyers')
+        yield Msg('wait', None, '_flyers')
         yield Msg('close_run')
         for token in tokens:
             yield Msg('unsubscribe', None, token)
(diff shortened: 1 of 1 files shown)


bluesky/plans.py
```diff
@@ -158,19 +158,28 @@ def _objects(self):
...
-                yield Msg('wait', None, 'A')
-            for det in dets:
-                yield Msg('read', det)
-            yield Msg('save')
-            yield Msg('sleep', None, delay)
+        yield Msg('checkpoint')
+        yield Msg('create')
+        for det in dets:
+            yield Msg('trigger', det, block_group='A')
+        for det in dets:
```
tacaswell: only need one wait message.

danielballan: fixed here and elsewhere in the file

danielballan:
@tacaswell As soon as this is merged, I believe that @cmazzoli and @ambarb would like to start using it (this weekend if possible). Would you rebuild the conda package and ping them to `conda update bluesky`?

bluesky/plans.py
danielballan: Travis failed on the table again -- otherwise all OK

danielballan:
P.S. @cmazzoli and @ambarb: when you do upgrade bluesky, once Tom makes the new build available, you should remove the new plotting code I wrote at the bottom of 99-settings.py or each `ct` might spawn *two* plots.

PR #307, opened 2016-01-30
TST: Fix LiveTable test by using lossless subs.

danielballan:
The LiveTable test was failing because, I think, the lossy subscription was doing its job, dropping an event or two, and thus missing a row.

Also, this makes a non-breaking API change that was previously discussed: `RunEngine.subscribe_lossless` and `RunEngine.unsubscribe_lossless` are now public methods. Aliases for the private names are maintained in case someone (HXN?) is using them.

(These were originally (back in May) made private so that users didn't choke their processors by reaching for lossless subscriptions to liberally. That's not really the way private methods are supposed to be used, and in any case that imagined scenario has not been a problem.)

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -251,17 +251,21 @@ class RunEngine:
             'wait_for': self._wait_for,
         }
 
-        # public dispatcher for callbacks processed on the main thread
+        # public dispatcher for callbacks
+        # The Dispatcher's public methods are exposed through the
+        # RunEngine for user convenience.
         self.dispatcher = Dispatcher()
         self.ignore_callback_exceptions = True
         self.event_timeout = 0.1
         self.subscribe = self.dispatcher.subscribe
         self.unsubscribe = self.dispatcher.unsubscribe
 
-        # private registry of "critical" callbacks
+        # private dispatcher of critical callbacks
         # which get a lossless Event stream and abort a run if they raise
-        self._lossless_cb_registry = CallbackRegistry(
-            allowed_sigs=DocumentNames)
+        # Ths subscribe/unsubscribe are exposed through the RunEngine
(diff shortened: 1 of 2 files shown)


danielballan:
Ooof. @tacaswell, Travis fails on

```python
ValueError: Allowed signals are <enum 'DocumentNames'>
```

*Really* regretting our use of a enum instead of strings. Time to walk that back (for next cycle)?

danielballan:
Update: We can postpone paying down the technical debt of DocumentNames enum for another day -- the Dispatcher handles it.

For reasons that no longer apply, the *lossless* subscriptions were handled directly to a CallbackRegistry, not laundered through a Dispatcher (which adds nice validation and convenience methods). Now in this PR we just use two Dispatchers.

Test coverage on this stuff is solid -- I'm not worried about this going in this cycle.

bluesky/run_engine.py
```diff
@@ -251,23 +251,31 @@ def __init__(self, md=None, *, md_validator=None):
...
+        # to provide special docstrings.
+        self._lossless_dispatcher = Dispatcher()
 
         self.verbose = False
 
         loop.call_soon(self._check_for_trouble)
         loop.call_soon(self._check_for_signals)
 
+        # aliases for back-compatibility
+        self._unsubscribe_lossless = self.unsubscribe_lossless
```
tacaswell: These should be done at class level as `_unsubscribe_lossless = unsubscribe_lossless`

tacaswell:
What were the reasons that no longer apply?

bluesky/run_engine.py
danielballan: Good call. Done.

danielballan:
@tacaswell The Dispatcher used to be less configurable than it is today.

klauer:
Looks good

PR #313, opened 2016-01-31
ENH: 'baseline_read' for baseline and end readings

danielballan:
For all plans, this feature provides any easy to way to collect baseline readings of objects (perhaps all objects on the beamline) at the beginning and the end of the scan.

```python
plan = Count(..., bracket_read=[det1, motor1])   # or equivalently, plan.bracket_read = [...]
```

The objects in `plan.bracket_read` may or may not also be in `plan.detectors` or `plan.motors`. The readings are taken immediately after the run is opened and immediately before the run is closed. They will be stored as a separate event stream with two events.


danielballan:
TODO:

* tests
* integration with global state

tacaswell:
:+1:

how about `'baseline_read'`  instead of `'bracket_read'` as my first reaction was "are the reads gonig to be competative?".

danielballan:
It turns out that addressing the simple request, "I want to record all my motor positions at the beginning and end of my scan" has far-reaching consequences. I found a dragon at the bottom of a rabbit hole.

Currently, most plans in production produce one event descriptor. When this baseline PR goes into use, almost every plan will produce two event descriptors. On the analysis side, this means that `get_table` and other step-scan tools will become less convenient.

To ameliorate this, I would like to introduce a new concept on an experimental basis: a descriptor 'name'. This would be optional (at least for now) and it would not break any existing code. I think of it as one more way to filter data coming out of the broker. The plans in plans.py would mark normal readings, say, as 'main' and baseline readings as 'baseline' by naming their respective descriptors. I also think of it as the name for a given "event stream," which is maybe more obviously useful than a name for a descriptor itself. When we generate multiple descriptors with the same data keys, they may (but need not) share a name.

There would be no special vocabulary for these names and, again, no hard requirement that they even exist. Plans would use the names 'main' and 'baseline' consistently, but user-defined plans could use any names they want, and the broker will know nothing about these names.

As a reminder, we currently support:

```
get_table(header)  # would return a wide table of all readings
get_table(header, fields=['det', 'motor'])  # whitelists data keys
```

With named descriptors, we could also support:
```
get_table(header, name='baseline')  # all events from descriptor(s) named 'baseline'
get_table(header, name='main')  # all events from descriptor(s) named 'main'
get_table(header, name='baseline', fields=['det'])  # only readings of 'det' from 'baseline' descriptor
```

We also have the datamuxer, but muxing is computationally expensive and not actually necessary unless you want to *combine* data from different event streams. This is a way of filtering data from a certain event stream. Within one stream, data is simply tabular, making computation and user interaction simpler. We shouldn't require use of the muxer just to "show me my data as as table." That, basically, is why I think we need these names.

danielballan:
Accept optional name in `kickoff` msg.

bluesky/run_engine.py
```diff
@@ -883,12 +891,16 @@ def _save(self, msg):
             raise IllegalMessageSequence("A 'create' message must be sent, to "
                                          "open an event bundle, before that "
                                          "bundle can be saved with 'save'.")
+        # Short-circuit if nothing has been read. (Do not create empty Events.)
+        if not self._objs_read:
+            self._bundling = False
```
danielballan: set bundle_name to None here in an abundance of caution

bluesky/run_engine.py
```diff
@@ -905,12 +917,12 @@ def _save(self, msg):
...
             yield from self.emit(DocumentNames.descriptor, doc)
             self.debug("*** Emitted Event Descriptor:\n%s" % doc)
-            self._descriptor_uids[objs_read] = descriptor_uid
+            self._descriptors[(self._bundle_name, objs_read)] = descriptor_uid
         else:
-            descriptor_uid = self._descriptor_uids[objs_read]
+            descriptor_uid = self._descriptors[(self._bundle_name, objs_read)]
         # This is a separate check because it can be reset on resume.
         if objs_read not in self._sequence_counters:
             self._sequence_counters[objs_read] = count(1)
```
danielballan: set bundle_name to None here in an abundance of caution

danielballan:
Add injection points immediately after open_run and before close_run. Rename _pre and _post to _before_run and _after_run.

danielballan:
Superceded and split into #316 and #317

PR #315, opened 2016-02-01
WIP: a sketch of a simulation mode

danielballan:
One of the attendees to a recent tutorial session picked up on the fact that bluesky/ophyd does not yet have a "simulation mode", as SPEC does.

There, I fixed it.

```python
RE(dryrun(plan))
```

Several months ago there was discussion of doing this on the RunEngine side like so:
 
```python
RE.unregister_command('set'); RE.register_command('set', fake_set)
```

But I think editing the messages on the way in is better. It avoid having to write `fake_*` methods that reimplement the RunEngine's state management.

(By the way, I notice that this will get a little more more complex if we change the Msg API to drop the required `obj`.)

### DIFF
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -1,0 +1,25 @@
+from bluesky.examples import Base, Mover, Flyer
+
+
+def dryrun(gen):
+    """
+    Replace all hardware objects with mocked objects.
+    
+    This could catch mistakes that lead to IllegalMessageSequence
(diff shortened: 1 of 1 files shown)


danielballan:
Just as good:

```python
from bluesky.plan_tools import print_summary
print_summary(plan)
```

danielballan:
cc @ambarb You expressed interest in a syntax-check "simulation" run. See code example above.

PR #317, opened 2016-02-01
Make convenient injection points for baseline reads.

danielballan:
See #313. This cherry-picks the parts that affects plans. Needs work.


danielballan:
It will be easier to rewrite this to to refactor it over #345. Closing.

PR #321, opened 2016-02-02
API: Subs is always a dict of empty lists, including all valid keys.

danielballan:
As a reminder, when attaching subscriptions to a plan, we accept `plan.subs = cb` as a shortcut for subscribing to all documents. (The explicit way is `plan.subs = {'all': [cb]}`.) The shortcut is convenient but confusing in a couple ways. It is accomplished using a descriptor that normalizes the input.

Given the confusion this has caused, removing the shortcut altogether is under discussion. But that's not what this PR does. This PR merely makes it a little more clear what is happening with subs. Maybe it goes far enough and we will not need to to remove the shortcut?

```python
In [1]: from bluesky.simple_scans import AbsScan

In [2]: ascan = AbsScan()
```

Before:

```python
In [3]: ascan.subs
Out[3]: {}
```

After:

```python
In [6]: ascan.subs
Out[6]: {'all': [], 'descriptor': [], 'event': [], 'start': [], 'stop': []}
```

Thus, `ascan.subs['event'].append(cb)` will now always work, no matter the state of `plan.subs`. This applies to both SPEC-style "simple scans" and Plans.

### DIFF
--- a/bluesky/tests/test_callbacks.py
+++ b/bluesky/tests/test_callbacks.py
@@ -78,11 +78,11 @@ def test_subs_input():
     # Test input normalization on OO plans
     obj_ascan = AbsScanPlan([det], motor, 1, 5, 4)
     obj_ascan.subs = cb1
-    assert obj_ascan.subs == {'all': [cb1]}
+    assert obj_ascan.subs == {'all': [cb1], 'start': [], 'stop': [], 'descriptor': [], 'event': []}
     obj_ascan.subs.update({'start': [cb2]})
-    assert obj_ascan.subs == {'all': [cb1], 'start': [cb2]}
+    assert obj_ascan.subs == {'all': [cb1], 'start': [cb2], 'stop': [], 'descriptor': [], 'event': []}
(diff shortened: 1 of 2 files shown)


tacaswell:
:+1: but worried this will break @klauer

klauer:
@tacaswell It's okay, I'm already broken. Anyway, this is relatively minor considering the refactoring I have left to do, so don't worry about it.

danielballan:
Cue Ken, singing:

I broke the API
To see if I could feel
I focus on the pain
The only thing that's real.

What has this become
My sweetest friend
Everything I know
Goes away in the end.

On Wed, Feb 3, 2016 at 4:14 AM -0800, "K Lauer" <notifications@github.com<mailto:notifications@github.com>> wrote:


@tacaswell<https://github.com/tacaswell> It's okay, I'm already broken. Anyway, this is relatively minor considering the refactoring I have left to do, so don't worry about it.

klauer:
It's... it's... beautiful :sob:

PR #322, opened 2016-02-03
ENH: almost clone of register_mds

tacaswell:
but it registers metadataclient

This is dumb, but effective.


danielballan:
We talked about moving `register_mds.py` to another package, but I'm not sure there's consensus on where that would be (I don't have any strong opinion yet). So this is fine.

danielballan:
cc @heroux this is progress on the metadataserver/client

PR #323, opened 2016-02-03
Enh more plan tools

tacaswell:


bluesky/plan_tools.py
```diff
@@ -149,11 +180,57 @@ def trigger_and_read(det_list):
...
+
+    plan = stage_wrapper(repeater(n, trigger_and_read, dets),
+                         dets)
+    plan_with_subs = subscription_wrapper(plan, subs)
+
+    ret = yield from plan_with_subs
+    return ret
+
+
+def bind_to_run_engine(RE, gen_func, name):
```
danielballan: Huge +1 for this. I will elaborate on it a little bit after merging.

PR #324, opened 2016-02-03
FIX: Avoid staging objects redundantly.

danielballan:
As discovered by @ambarb, if you include your 'motor' as one of your 'detectors' (which is a reasonable thing to do) `Plan` tries to stage it twice and blows up. This is a simple fix.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -92,9 +92,12 @@ class PlanBase(Struct):
         # itself as its argument and returns a generator of messages
         if self.pre_run is not None:
             yield from self.pre_run(self)
+        staged = []  # to track duplicates
         for objs in self._objects.values():
             for obj in objs:
-                yield Msg('stage', obj)
+                if obj not in staged:
+                    yield Msg('stage', obj)
+                    staged.append(obj)
 
     def _post(self):
         """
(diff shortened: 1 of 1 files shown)


bluesky/plans.py
```diff
@@ -92,9 +92,12 @@ def _pre(self):
         # itself as its argument and returns a generator of messages
         if self.pre_run is not None:
             yield from self.pre_run(self)
+        staged = []  # to track duplicates
```
danielballan: "But Dan," you ask, "why aren't you just using `set()`?" Because it is good to stage each category of objects in a deterministic order. There is no `OrderedSet` in the standard lib.

tacaswell: But this is a local variable that gets discarded.

tacaswell: and for small numbers it is not clear that sets are always a performance win.

danielballan: Yes. It is a local variable that gets discarded. But first, it tracks duplicates.

tacaswell: so you could use a set because this never gets iterated over.

danielballan: Ah. We were talking past each other. My comment about set was referring the possibility of using `self._objects.values()` -> `set(self._objects.values())` and calling it a day, not the possibility of making `staged` a set.

PR #325, opened 2016-02-06
Enh better mdstar handling

tacaswell:


danielballan:
I'm surprised the line count balance isn't more negative than it is.

PR #326, opened 2016-02-09
FIX: Do not include image data in LiveSpecFile.

danielballan:
attn @ericdill


danielballan:
poke @ericdill -- we had better get this in before we get bug reports from CHX

ericdill:
Looks good. Thanks!

PR #329, opened 2016-02-10
MNT: Clean up event document handling of SpecCallback

ericdill:


bluesky/callbacks/core.py
```diff
@@ -693,9 +693,8 @@ def descriptor(self, doc):
 
     def event(self, doc):
         """Write each event out"""""
-#         print(doc)
         data = doc['data']
-        values = [str(data[k]) for k in self._read_fields if k != self._motor]
```
tacaswell: Why make this change?

ericdill: because self._read_fields, as defined on line 681 does not have self._motor. Don't need to check again

PR #330, opened 2016-02-10
FIX: correct inserting of configuration data

tacaswell:
The data comes back from ophyd as a dict of dict keyed on 'value',
'timestamp'.  Previously this was inserting the stings 'value',
'timestamp' into the descritor, rather than the information.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -279,7 +279,6 @@ class RunEngine:
         # to provide special docstrings.
         self._lossless_dispatcher = Dispatcher()
 
-
         loop.call_soon(self._check_for_trouble)
         loop.call_soon(self._check_for_signals)
 
@@ -906,9 +905,9 @@ class RunEngine:
         "Read the object's configuration and cache it."
         config_values = {}
         config_ts = {}
-        for key, (val, ts) in obj.read_configuration().items():
-            config_values[key] = val
-            config_ts[key] = ts
+        for key, val in obj.read_configuration().items():
+            config_values[key] = val['value']
+            config_ts[key] = val['timestamp']
         self._config_values_cache[obj] = config_values
         self._config_ts_cache[obj] = config_ts
 


danielballan:
Oops. As we've said before (and I recently discussed with @dchabot) this is one more indication that we should eventually move Device, Signal, and friends into a separate package that does not import pyepics. Don't worry, @klauer: if we do this, I would advocate for supporting our current API forever using shim imports.

danielballan:
So, does *anything* return a `(val, timestamp)` tuple anymore? I thought I remembered that Signals still did this, but a quick look at the codebase tells me I'm wrong about that. Are we using `{name: {value: ..., timestamp: ...}}` everywhere?

klauer:
@danielballan Why is this an indicator and why is that a good idea?

danielballan:
I should have given that more context:

This bug occurred because the test/example devices in bluesky are not a complete reimplementation of the Device in ophyd. It seems like a good idea to test on the real thing -- not just in some proposed "vertical integration" suite but in the bluesky unit tests. It would also be nice if the motors and detectors used as toy examples were true ophyd soft devices/signals.

There are two ways to do this: make it possible to import ophyd.device/signal without pyepics or move these epics-agnostic bits to a separate package. Counterintuitively, the latter is actually less distributive from an API point of view (because ophyd imports many of its pieces to the top level).

Also, by moving Device and Signal to a separate package, we can lower the barrier for someone to reuse these abstractions on top of a different protocol. I don't think *our* group would or should ever do this, but it's neat that someone *could*, say, use pyvisa to talk to LabView through a Device.

On Thu, Feb 11, 2016 at 4:35 AM -0800, "K Lauer" <notifications@github.com<mailto:notifications@github.com>> wrote:


@danielballan<https://github.com/danielballan> Why is this an indicator and why is that a good idea?

klauer:
Problem: "we have vertical integration issues" Solution: "stack it higher!"

I think your assessment is fair working under the assumption that "someone" out there is going to be using our stuff with a completely different setup than we have. In my opinion, which you're probably already aware, we should wait until that "someone" pops up before we start creating even more unnecessary package-related maintenance for ourselves. I find it to be very painful working with 10 different packages, broken apart for reasons unrelated to the original development intent or our actual usage, for just one project.

Also, pyepics is a pure Python library. Installed and left unused it causes no harm and doesn't even try to find `libca` until you start poking at it. If someone were to come along and say "we want to use bluesky and ophyd Devices and your other stuff but we don't want EPICS", do you really think that the one objection they will have is to installing pyepics versus the other tens of packages we depend on?

Chances are this change will happen regardless of what I think in the name of "cleanliness of abstractions", but at least I've recorded my opinion.

ericdill:
> Problem: "we have vertical integration issues" Solution: "stack it higher!"

This made me lol.

> I find it to be very painful working with 10 different packages, broken apart for reasons unrelated to the original development intent or our actual usage, for just one project.

tl;dr I have nothing useful to add.

I agree that in many regards working with 10 packages over one package is more difficult.  One very real advantage that outweighs my annoyance at working with more small packages is that each package becomes responsible for less and so becomes much easier to test. That being said, ensuring that all of its components work together would be the job of that one monolithic package, but in the scenario where we have 10 packages, it is the job of the vertical integration testing which we have been continuously punting on (for obvious reasons like we dont have enough people). my 0.02 * :dollar:

danielballan:
@klauer I agree this shouldn't be done merely for the sake of one idea of abstract elegance -- splitting packages up has advantages and disadvantages. But we do have some use cases on the horizon with outside users in the 1-3 months time horizon. I leave it to @tacaswell to elaborate.

tacaswell:
I want to be able to use `Device` objects in bluesky tests without having to install pyepics.  As a practical matter, we are using conda packages which correctly has `epics-base` as a dependancy of `pyepics` so to do as you suggest would require gymnastics. 

@sbillinge wants to be able to create a simulation enviroment for XPD users to 'dry run' their experiments before they show up at the beam line. 

I am sure that this bit of refactoring can be done in a 100% back compatible way.   If we do not want to pick ophyd apart, it needs to be importable with out pyepics and we change the conda recipe to not require it.

sbillinge:
@tacaswell, we are currenlty making our simulation extremely lightweight so that it doesn't even import bluesky.  This means that this is not urgent for us right now.

We tried the bluesky route but there were too many dependencies.  In the future we may revisit if there is a bluesky conda package and it will all install together with a ``conda install bluesky``.

Our goal is not to simulate the entire xpd data acquisition process, but rather to allow users to create their sample and experiment metadata objects at home, then write them to yaml files to bring to the beamline. So our bluesky objects don't have to do anything, but just have to exist so that all the imports work in the xpdacq modules.

tacaswell:
@sbillinge you are currently importing ophyd (https://github.com/xpdAcq/xpdSim/blob/master/simulator/areadetector.py) which mean your sim project has a hard `pyepics` dependency, and hence if you are using our conda packages, a hard epics-base dependency. 

At any rate, this discussion has gotten way off the rails.  What level of testing would make people happy enough to merge this?

sbillinge:
Please wait on this.  This is code from Tim that I haven't reviewed yet.
There are still some things we need to sort out, and would like you in the
discussions, but this is not our priority right now.

simulator may be a bad name for what we are building....

S

On Thu, Feb 11, 2016 at 11:17 AM, Thomas A Caswell <notifications@github.com
> wrote:

> @sbillinge <https://github.com/sbillinge> you are currently importing
> ophyd (
> https://github.com/xpdAcq/xpdSim/blob/master/simulator/areadetector.py)
> which mean your sim project has a hard pyepics dependency, and hence if
> you are using our conda packages, a hard epics-base dependency.
>
> At any rate, this discussion has gotten way off the rails. What level of
> testing would make people happy enough to merge this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/NSLS-II/bluesky/pull/330#issuecomment-182939858>.
>



-- 
Prof. Simon Billinge
Applied Physics & Applied Mathematics
Columbia University
500 West 120th Street
Room 200 Mudd, MC 4701
New York, NY 10027
Tel: (212)-854-2918 (o) 851-7428 (lab)

Condensed Matter Physics and Materials Science Dept.
P.O. Box 5000
Upton, NY 11973-5000
(631)-344-5661

email: sb2896 at columbia dot edu
home: http:// <http://nirt.pa.msu.edu/>bgsite.apam.columbia.edu/

danielballan:
@sbillinge Thanks for providing some context about your "simulated beamline" use case. I think your last comment must be about something in xpdacq, but this is a bluesky PR. It's easy to lose track in email. :- )

@tacaswell I vote for a basic `read_configuration` method on `examples.Base` that returns a hard-coded dict of jibberish "configuration".

PR #340, opened 2016-02-16
Revert bad change to spec file output

ericdill:
Reverts fae6ef91983d49177d6964cf4cfc9e53cd12e9d9.

Might fix NSLS-II/Bug-Reports#81.  Need @yugangzhang to test it out


ericdill:
#81 was user error. yugang needs to rename the readback value on the motor.

PR #341, opened 2016-02-16
Add tifffile as a dependency

ericdill:
bluesky was getting tifffile through pims making tifffile a runtime dep.  
Now that pims 0.3.3 does not have tifffile as a runtime dep, we should make
tifffile a dep of bluesky.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ install:
   - conda create -n testenv nose python=$TRAVIS_PYTHON_VERSION scipy requests matplotlib jsonschema traitlets pytest
   - source activate testenv
   - pip install pytest-cov pytest-xdist
-  - conda install metadatastore databroker historydict boltons prettytable doct toolz -c lightsource2 -c conda-forge -c soft-matter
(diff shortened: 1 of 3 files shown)


tacaswell:
I would rather only put it in the travis file, but not in the yaml.  Or just move the tifffile import local to the function where it is used.  That is there as an example for XPD and they wrote their own anyway.

ericdill:
@tacaswell If XPD already has their own, why don't we delete this code then?  Less code to maintain == win IMO.

My preference is, in decreasing order:

1. delete the callback that uses tifffile, since it is for XPD and they have this working locally anyway
2. Move that callback to the top-level examples/ directory
3. Import tifffile local to the TIFF callback

tacaswell:
If we are going to delete it, then we need to do some deprecation (just to be sure that I am right no one is using it).  I would go with 3 and then 1.

ericdill:
Ok, well now it seems that we shouldn't delete it since someone *might* be using it. Sigh.  I'll just import tifffile local to the callback.

ericdill:
Thanks @tacaswell !

PR #342, opened 2016-02-17
Smoketest the broker_examples module

ericdill:


danielballan:
This is already covered, I believe: https://github.com/NSLS-II/bluesky/blob/master/bluesky/tests/test_vertical_integration.py#L48

ericdill:
I'm not so sure that it is covered.  There is a syntax error in the broker_examples module! None of the det_* objects at the bottom of https://github.com/NSLS-II/bluesky/blob/master/bluesky/broker_examples.py get created because the imports of the motors were deleted!

ericdill:
Err, maybe not "syntax" error, but the motor* objects don't exist in that module

danielballan:
I was conflating broker_callbacks with broker_examples. Might as well add a meaningful tests here -- LiveMesh is one way to test SynGauss2D.

bluesky/examples.py
```diff
@@ -21,6 +21,7 @@ def describe(self):
 
 class Base:
     def __init__(self, name, fields):
+        self._name = name
```
ericdill: @danielballan Something was expecting these objects to have a `_name` attribute.  I did not dig at all to see if the `.name` attribute is still used.  i guess the question here is should this be `_name` or `name`?

ericdill:
Tests are failing on travis because of the tifffile import. That is being fixed in #341.  Other than that, I only have a question about `_name` vs `name`, which is that in-line comment.

bluesky/examples.py
danielballan: It should be `name`, which is now part of the "bluesky interface" that all ophyd devices and signals are expected to have. Nothing should still be expecting a `_name` attribute.

danielballan: Looks like we failed to catch all the instances of `_name` in bluesky:

```
$ git grep -e "\._name"
bluesky/broker_examples.py:        self._data = {self._name: {'value': fs_uid, 'timestamp': ttime.time()}}
bluesky/broker_examples.py:        return {self._name: {'source': self._name,
doc/source/custom-plans.rst:                m=ret_m[motor._name]['value'], d=ret_d[det1._name]['value']))
```

ericdill: Ok, so I'll rename these `_name` to `name` in `broker_examples`

ericdill: And remove this change in `examples.py`. Thanks for the feedback

ericdill:
restart travis after tifffile dep has been fixed

PR #345, opened 2016-02-18
REF: Implement PlanBase in terms of plan_tools generators.

danielballan:
- Add generators for subscribing and flying.
- Start refactoring PlanBase in terms of plan_tools generators.


danielballan:
Oops -- some overlap with https://github.com/NSLS-II/bluesky/pull/323/files

I'll merge that first.

bluesky/plan_tools.py
```diff
@@ -136,21 +189,6 @@ def stage_wrapper(plan, dets):
     return ret
 
 
-def subscription_wrapper(plan, subs):
```
danielballan: I'm keeping your name, @tacaswell, but my code because I threw in more comments and a docstring. The substance is the same.

bluesky/plan_tools.py
```diff
@@ -214,14 +253,12 @@ def ct(dets, n, subs=None):
...
         subs = {'all': [LiveTable(dets)]}
-    if n is None:
+    if n != 1:
         subs['all'].append(LivePlot(scalar_heuristic(dets[0])))
 
     plan = stage_wrapper(repeater(n, trigger_and_read, dets),
                          dets)
-    plan_with_subs = subscription_wrapper(plan, subs)
-
-    ret = yield from plan_with_subs
```
danielballan: @tacaswell I am wrong in thinking that these subscriptions are being add too late to see the RunStart document?

bluesky/plan_tools.py
```diff
@@ -90,6 +90,59 @@ def print_summary(plan):
...
+            tokens.add(token)
+    try:
+        ret = yield from plan
+    finally:
+        # The RunEngine might never process these if the execution fails,
+        # but it keeps its own cache of tokens and will try to remove them
+        # itself if this plan fails to do so.
+        for token in tokens:
+            yield Msg('unsubscribe', None, token)
+
```
tacaswell: this should return `ret`

bluesky/plan_tools.py
tacaswell: no, I think you are correct.

bluesky/examples.py
```diff
@@ -49,10 +49,10 @@ def configure(self, d):
         return {}, {}
 
     def stage(self):
-        pass
+        return self
```
klauer: `stage` and `unstage` return lists of objects

bluesky/plans.py
```diff
@@ -33,93 +33,47 @@ class PlanBase(Struct):
...
-        yield Msg('close_run')
-        for token in tokens:
-            yield Msg('unsubscribe', None, token)
-        yield from self._post()
+        md = self.get_metadata()
+        md['plan_args'] = {field: repr(getattr(self, field))
+                           for field in self._fields}
+        flyers = self.flyers
+
+        main_with_flyers = fly_during(self._main_gen(), self.flyers)
```
klauer: I still have a difficult time parsing these things. I haven't fully thought it through, but just out of curiosity - have you thought if context manager blocks could be used to give this some sort of structure?

@tacaswell pointed out it's not currently possible in 3.4 (5?)

klauer: 3.5 has 'async with' syntax:
[PEP 0492](https://www.python.org/dev/peps/pep-0492/)

[stackoverflow](http://stackoverflow.com/questions/34052293/mixinig-async-context-manager-and-straight-await-in-asyncio) example which clued me in

danielballan: I totally read that PEP a couple months ago and forgot about `async with`. Awesome.

danielballan: I am pretty convinced that this only works with strict coroutines, not generators, which means any plan that used this could not be list-ified. This issue is likely to come up a lot as we develop fancier wrappers, so perhaps we just need to commit to coroutines and provide better way to list-ify plans.

klauer: Just a thought - I like the cleanliness that could (potentially) come with it, at least.

klauer:
What's the status on this? I'm hoping to use this as a basis for some scan logic rewrites in hxntools.

bluesky/run_engine.py
```diff
@@ -111,7 +111,7 @@ class RunEngine:
 
     _loop = loop  # just a convenient way to inspect the global event loop
     state = LoggingPropertyMachine(RunEngineStateMachine)
-    _UNCACHEABLE_COMMANDS = ['pause', 'subscribe', 'unsubscribe']
+    _UNCACHEABLE_COMMANDS = ['pause', 'subscribe', 'unsubscribe', 'stage']
```
danielballan: This fixes one case:

- checkpoint
- stage
- INTERRUPTION

but not another case:

- checkpoint
- stage
- unstage
- INTERRUPTION

This second case affects other messages as well and should be addressed in a future PR. The first case is more common, so I'd like to address it here.

bluesky/simple_scans.py
```diff
@@ -1,506 +0,0 @@
...
-    return LiveTable([scan.motor] + gs.TABLE_COLS + [gs.PLOT_Y])
-
-
-def table_gs_only(scan):
-    "Setup a LiveTable by inspecting a scan and gs."
-    # no motors
-    return LiveTable(gs.TABLE_COLS + [gs.PLOT_Y])
-
-
-def _figure_name(base_name):
```
danielballan: Don't worry, @tacaswell: this code (and related functions below) were not deleted; they were just moved. This file `simple_scans.py`, home of "bundled scans," is no longer needed.

danielballan:
This is ready for a round of view. Tests pass (we'll see if Travis agrees...).

Plan logic has been moved into generators. The class-based implementations and SPEC-like implementations have been refactored accordingly. Everybody has docstrings. The only changes to the test code relate to removed functionality and the change to how SPEC scans behave, all of which is documented in the release notes.

The "plan tools" snippets still need docstrings and maybe a little more cleanup.

bluesky/run_engine.py
tacaswell: Do not cache `'unstage'` either?

bluesky/plan_tools.py
```diff
@@ -112,43 +176,86 @@ def run_wrapper(plan, md=None, **kwargs):
...
+    """
+    Interpret set messages on objects as relative to current position.
+    """
+    initial_positions = {}
+    ret = None
+    while True:
+        msg = plan.send(ret)
+        if msg.command == 'set' and msg.obj in objects:
+            if msg.obj not in initial_positions:
+                pos = scalar_heuristic(msg.obj)
```
tacaswell: just use `position`

bluesky/plan_tools.py
```diff
@@ -112,43 +176,86 @@ def run_wrapper(plan, md=None, **kwargs):
...
+def put_back(plan, objects):
+    initial_positions = OrderedDict()
+    ret = None
     try:
-        ret = yield from plan
+        while True:
+            msg = plan.send(ret)
+            if msg.command == 'set' and msg.obj in objects:
+                if msg.obj not in initial_positions:
+                    pos = scalar_heuristic(msg.obj)
```
tacaswell: ditto on using `position`

bluesky/plan_tools.py
```diff
@@ -205,23 +317,225 @@ def caching_repeater(n, plan):
         yield from (m for m in lst_plan)
 
 
-@wrap_with_decorator(run_wrapper)
-def ct(dets, n, subs=None):
-    """Count
+def table_and_plot(plan):
+    subs = {'all': [LiveTable(detectors)]}
```
tacaswell: This looks broken?

danielballan: Ah, it's unused. Deleting.

bluesky/run_engine.py
danielballan: Redundant unstaging does no harm, so it doesn't *have* to go in this list. Let's defer the discussion of whether it should to a better solution (one that addressed Case 2 above) in a separate PR.

bluesky/plan_tools.py
danielballan: Done. `scalar_heuristic` has been removed everywhere is occurred.

klauer: 👍

bluesky/plan_tools.py
```diff
@@ -90,6 +97,63 @@ def print_summary(plan):
...
+        for token in tokens:
+            yield Msg('unsubscribe', None, token)
+    return ret
+
+
+def fly_during(plan, flyers):
+    if flyers:
+        for flyer in flyers:
+            yield Msg('kickoff', flyer, block_group='_flyers')
+        yield Msg('wait', None, '_flyers')
```
tacaswell: Won't this prevent the rest of the scan from running?

bluesky/plan_tools.py
```diff
@@ -90,6 +97,63 @@ def print_summary(plan):
...
+    ret = yield from plan
+    if flyers:
+        for flyer in flyers:
+            yield Msg('collect', flyer, block_group='_flyers')
+        yield Msg('wait', None, '_flyers')
```
tacaswell: This wait should be above the `for` loop here.

bluesky/plan_tools.py
```diff
@@ -112,43 +176,94 @@ def run_wrapper(plan, md=None, **kwargs):
...
+    This is a preprocessor that inserts 'stage' Messages.
+
+    The first time an object is read, set, triggered (or, for flyers,
+    the first time is it told to "kickoff") a 'stage' Msg is inserted first.
+    It stages the the object's ultimate parent, pointed to be its `root`
+    property.
+
+    At the end, an 'unstage' Message issued for every 'stage' Message.
+    """
+    COMMANDS = ['read', 'set', 'trigger', 'kickoff']
```
tacaswell: make this a set :bike: :house:

danielballan:
Unless there are objections, I'm going with what I consider to be the lightest-touch approach for handling the cache issue. I'm very open to revisiting this in the future.

In other news, I refactored the SPEC-like scans to use context managers. I think this is ready for one last review.

bluesky/plans.py
```diff
@@ -620,6 +620,30 @@ def reset():
     return (yield from finalize(plan_mutator(plan, insert_set), reset()))
 
 
+@contextmanager
+def baseline_context(plan_stack, devices, name='baseline'):
```
klauer: 👍 for the functionality, and an extra 👍 for being so easy to read.

bluesky/plans.py
```diff
@@ -620,6 +620,30 @@ def reset():
...
+    """
+    Read every device once upon entering and exiting the context.
+
+    The readings are designated for a separate event stream named 'baseline'
+    by default.
+
+    Parameters
+    ----------
+    plan_stack : iterable
```
klauer: Nitpicking, but `iterable` isn't really right here. You could say `'list-like'` or something along those lines.

klauer:
I can't find anything else to complain about. :+1:

danielballan:
image

klauer:
Merging for mood improvement. Fixes and tweaks to follow in separate PRs.

danielballan:
What with the picnic *and* the merge, I've rarely been in a better mood.

PR #347, opened 2016-02-19
ENH/API: Add raise_if_interrupted option and new Exception

danielballan:
If a run is interrupted but not exceptions are raised (which IIRC
can only occur because the RunEngine is paused or stopped) users
can use this flag to raise a special exception, aimed at interrupting
user scripts or functions that call the RunEngine.

- Add public raise_if_interrupted flag and private _interrupted flag.
- Add `RunEngineInterrupted(Exception)`.
- Remove `RunInterrupted(KeyboardInterrupted)` which was never used.
- Test.

### DIFF
--- a/bluesky/__init__.py
+++ b/bluesky/__init__.py
@@ -4,7 +4,7 @@ logger = logging.getLogger(__name__)
 from .run_engine import Msg
 from .run_engine import RunEngine
 from .run_engine import PanicError
-from .run_engine import RunInterrupt
+from .run_engine import RunEngineInterrupted
 from .run_engine import IllegalMessageSequence
 
 from .plans import PlanBase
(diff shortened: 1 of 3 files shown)


danielballan:
Edit: Per @tacaswell's offline suggestion, I refactored to make `raise_if_interrupted` a kwarg, not a stateful flag. This way the default behavior is always the "good" behavior that we will -- no nasty tracebacks -- but it is *possible* to get this behavior inside a script. That fact that doing so is annoyingly verbose may nudge our users toward refactoring their scripts into nice plans.

tacaswell:
attn @imrealkaren @cmazzoli The next time you get an updated bluesky please start to use this flag when you call `gs.RE` from inside a function.

PR #349, opened 2016-02-19
FIX/REF: Fix suspension during event bundling.

danielballan:
- We had fixed this issue for *pausing* by adding `self._bundling = False`
  to the `resume` method, but we (I) missed it in `request_suspend`. To avoid
  this error in the future, I refactored the common code into `_rewind`.


tacaswell:
@imrealkaren this fixes the suspender bug.

danielballan:
I just rebased, @tacaswell . Thanks for catching this one, @imrealkaren!

PR #350, opened 2016-02-22
ENH: look at success state of status objects

tacaswell:
This is a proof-of-concept commit.

This relies on @klauer refactoring of the status objects.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1006,6 +1006,9 @@ class RunEngine:
             p_event = asyncio.Event()
 
             def done_callback():
+                if not ret.success:
+                    return loop.call_soon_threadsafe(self._failed_ophyd_status,
+                                                     ret)
                 loop.call_soon_threadsafe(p_event.set)
 
             ret.finished_cb = done_callback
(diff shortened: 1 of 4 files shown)


danielballan:
This effectively special-cases failure to move over other kinds of exceptions that can bubble up from ophyd. I'm not 100% convinced that it needs to be special-cased.

tacaswell:
This catches async failures, the rest of them come up synchronously in the co-routines.   I think this is the only way to make sure that the exceptions BS needs come up on the correct threads / co-routines.

There are some epi-cycles with `self._task.cancel()` that I do not fully have pinned down.

Not sure what you mean by special casing failure to move, this also deals with trigger and kickoff failures (in the same way).

danielballan:
I meant that is special cases a certain kind of failure coming up from the device layer. My impulse would be to just use exceptions. I see the `success` bit as something that debuggers will want and plans *might* use (to decide if it's important that something failed to move) but that the RunEngine doesn't need to know about.

danielballan:
> I think this is the only way to make sure that the exceptions BS needs come up on the correct threads / co-routines.

If that's true though, my point is moot.

bluesky/run_engine.py
```diff
@@ -1113,6 +1128,12 @@ def _wait(self, msg):
         if objs:
             yield from self._wait_for(Msg('wait_for', objs))
 
+    def _failed_status(self, ret):
+        '''
```
danielballan: Cool docstring, bro.

ericdill: Actually lol'd at this

PR #351, opened 2016-02-22
ENH: Keep a record of seeing the target key in LivePlot

ericdill:
Hopefully this will allow us to more quickly debug these problems in the future.  Looking specifically at nsls-ii/bug-reports#83

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -142,6 +142,7 @@ class LivePlot(CallbackBase):
         self.lines = []
         self.legend = None
         self.legend_title = " :: ".join([name for name in self.legend_keys])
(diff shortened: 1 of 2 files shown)


ericdill:
re: discussion with @tacaswell

bluesky/callbacks/core.py
```diff
@@ -176,6 +179,9 @@ def event(self, doc):
         ax.autoscale_view(tight=True)
         ax.figure.canvas.draw()
 
+    def stop(self, doc):
+        if not self.got_event_with_target_y:
```
tacaswell: You can just check if `self.x_data` and `self.y_data` are empty lists.

PR #352, opened 2016-02-23
MNT: Fix spec callback.

ericdill:
The spacings on the #L line needs to be hyper specific or pymca and other spec-file-reading utilities do not understand how to parse the information. 

This allows the LiveSpecFile to output correctly formatted spec documents

cc @yugangzhang @lwiegart

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -574,6 +574,7 @@ _SPEC_START_TEMPLATE = env.from_string("""
 #T {{ acq_time }} (Seconds)
 #P0 {{ positioner_positions | join(' ')}}""")
 
+# It is critical that the spacing on the #L line remain exactly like this!
 _SPEC_DESCRIPTOR_TEMPLATE = env.from_string("""
 #N {{ length }}
 #L {{ motor_name }}    Epoch  Seconds  {{ data_keys | join('  ') }}\n""")
@@ -585,7 +586,6 @@ class LiveSpecFile(CallbackBase):
     """Callback to export scalar values to a spec file for viewing
 
     Expect:
-        `
     1. a descriptor named 'baseline'
     2. an event for that descriptor
     3. a descriptor named 'main'
@@ -691,13 +691,11 @@ class LiveSpecFile(CallbackBase):
         """Write the header for the actual scan data"""
         # List all scalar fields, excluding the motor (x variable).
         self._read_fields = sorted([k for k, v in doc['data_keys'].items()
-                                    if (v['object_name'] != self._motor
-                                        and not v['shape'])])
+                                    if k != self._motor and not v['shape']])
+        # Remove the motor key. It should only be in the list once!
         content = dict(motor_name=self._motor,
-                       acq_time=self._acq_time,
-                       unix_time=self._unix_time,
                        length=3 + len(self._read_fields),
-                       data_keys=doc['data_keys'].keys())
+                       data_keys=self._read_fields)
         with open(self.specpath, 'a') as f:
             f.write(_SPEC_START_TEMPLATE.render(self._start_content))
             f.write(_SPEC_DESCRIPTOR_TEMPLATE.render(content))
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -724,10 +724,14 @@ class LiveSpecFile(CallbackBase):
         data = doc['data']
         values = [str(data[k]) for k in self._read_fields]
         if self._motor == "Count":
-            doc['data']['Count'] = -1
+            motor_position = -1
+        elif self._motor = "seq_num":
+            motor_position = data['seq_num']
+        else:
+            motor_position = doc[self._motor]
         content = dict(acq_time=self._acq_time,
                        unix_time=doc['time'],
-                       motor_position=data[self._motor],
+                       motor_position=motor_position,
                        values=values)
         with open(self.specpath, 'a') as f:
             f.write(_SPEC_EVENT_TEMPLATE.render(content))
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -594,6 +594,12 @@ class LiveSpecFile(CallbackBase):
     Other documents can be issues before, between, and after, but
     these must be issued and in this order.
 
+    Notes
+    -----
+    `Reference <https://github.com/certified-spec/specPy/blob/master/doc/specformat.rst>`_
+    for the spec file format.
+    Nice find @danielballan!
+
     Example
     -------
     It is suggested to put this in the ipython profile:
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -648,44 +648,43 @@ class LiveSpecFile(CallbackBase):
         # and parsing any existing contents.
         plan_type = doc['plan_type']
         plan_args = doc['plan_args']
-        if plan_type in _PLAN_TO_SPEC_MAPPING.keys():
-            # Some of these are used in other methods too -- stash them.
-            self._unix_time = doc['time']
-            self._acq_time = plan_args.get('time', -1)
-            content = dict(scan_type=_PLAN_TO_SPEC_MAPPING[doc['plan_type']],
-                           acq_time=self._acq_time)
-            if plan_type == 'Count':
-                # count has no motor. Have to fake one.
-                self._motor = ['Count']
-            else:
-                self._motor = doc['motors']
-                content['start'] = plan_args['start']
-                content['stop'] = plan_args['stop']
-                content['strides'] = int(plan_args['num']) - 1,
-            # We only support a single scanning motor right now.
-            if len(self._motor) > 1:
-                raise NotImplementedError(
-                    "Your scan has %s scanning motors. They are %s. SpecCallback"
-                    " cannot handle multiple scanning motors. Please request "
-                    "this feature at https://github.com/NSLS-II/bluesky/issues" %
-                    (len(self._motor), self._motor))
-            self._motor, = self._motor
-            content['scan_motor'] = self._motor
-            command = _SPEC_1D_COMMAND_TEMPLATE.render(content)
-        else:
+        if plan_type not in _PLAN_TO_SPEC_MAPPING.keys():
             err_msg = ("Do not know how to represent %s in SPEC. If "
                        "you would like this feature, request it at "
                        "https://github.com/NSLS-II/bluesky/issues"
                        % plan_type)
             raise NotImplementedError(err_msg)
-        # write the new scan entry
-        content = dict(command=command,
-                       scan_id=doc['scan_id'],
-                       readable_time=datetime.fromtimestamp(doc['time']),
-                       acq_time=self._acq_time,
-                       positioner_positions=self.positions)
-        self._start_content = content  # can't write until after we see desc
-        self._start_doc = doc
+
+        # Some of these are used in other methods too -- stash them.
+        self._unix_time = doc['time']
+        self._acq_time = plan_args.get('time', -1)
+        content = dict(scan_type=_PLAN_TO_SPEC_MAPPING[doc['plan_type']],
+                       acq_time=self._acq_time)
+        if plan_type == 'Count':
+            # count has no motor. Have to fake one.
+            self._motor = 'Count'
+        else:
+            content['start'] = plan_args['start']
+            content['stop'] = plan_args['stop']
+            content['strides'] = int(plan_args['num']) - 1,
+            try:
+                # We only support a single scanning motor right now.
+                self._motor, = doc['motors']
+            except ValueError:
+                raise NotImplementedError(
+                    "Your scan has %s scanning motors. They are %s. SpecCallback"
+                    " cannot handle multiple scanning motors. Please request "
+                    "this feature at https://github.com/NSLS-II/bluesky/issues" %
+                    (len(self._motor), self._motor))
+        command = _SPEC_1D_COMMAND_TEMPLATE.render(content)
+        # Can't write the entry until we see the descriptor, so stash it until
+        # we get the descriptor
+        self._start_content = dict(
+            command=command,
+            scan_id=doc['scan_id'],
+            readable_time=datetime.fromtimestamp(doc['time']),
+            acq_time=self._acq_time,
+            positioner_positions=self.positions)
 
     def descriptor(self, doc):
         """Write the header for the actual scan data"""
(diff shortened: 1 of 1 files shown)


bluesky/callbacks/core.py
```diff
@@ -686,13 +686,11 @@ def descriptor(self, doc):
         """Write the header for the actual scan data"""
         # List all scalar fields, excluding the motor (x variable).
         self._read_fields = sorted([k for k, v in doc['data_keys'].items()
-                                    if (v['object_name'] != self._motor
```
danielballan: Why this change? I think `object_name` is correct because the motor, extracted `doc['motors']` in Line 657 above, is an object name, not a data key. We want to filter out any data keys associated with `motor` (e.g., `motor_user_readback`, `motor_user_setpoint`).

danielballan:
I thought you already ran this through pymca right before we merged it. Are there more tolerant modes of pymca or differing versions out there? Am I just misremembering?

ericdill:
I opened it with pymca, but did not check anything other than that. It opens, but it does not actually plot data correctly. That was my mistake for not being thorough

bluesky/callbacks/core.py
ericdill: Why do we want to filter out the other motor data keys?

danielballan:
Oh, OK. I wasn't trying to call you out -- just wanted to make sure I understood what broke. As I've said before, it's cool that we do this but this integration "extra credit" from my point of view.

ericdill:
No I understand, it was a valid question for sure!

bluesky/callbacks/core.py
danielballan: There is no guarantee that *any* data key will match `self._motor`. In some circumstances, `motor.user_readback.name == 'motor'`, but in other cases it might not. The only general way to handle this to come object names with object names, not object names with data keys.

danielballan: ^ edited for clarity

ericdill: Before I can fully respond, I need some information. Where do the names in `start_document['motors']` come from?

danielballan: Those are object names. They come from calling `.name` on every motor. They are implemented in plans.py.

ericdill: So the values in start_document['motors'] are **guaranteed** to be in the descriptor?

danielballan: Yes. To be precise:

```python
set(start['motors']).issubset(set([dk['object_name'] for dk in descriptor['data_keys']]))
```

ericdill:
Ok, so here are my updated changes to the LiveSpecFile callback.  Maybe we should just discuss this in person, but I'll give it a shot via github first.

`start()`
No functionality change. Simplified code by removing a large chunk of code from a nested if block.  Now there are not nested if/else. 

`descriptor()`
1. See if `self._motor` is a valid key for `descriptor['data_keys']`. If it is not, append `_user_readback` to `self._motor` and check to see if *now* it is a valid key in `descriptor['data_keys']`.  If it is still not a valid entry, then assert that we do not know what the motor name and fall back to just using the event `seq_num` for the motor position.
2. Dump all scalar keys into the spec file because (a) it is easier than parsing out motor names and (b) I cannot think of a good reason not to include scalar keys.  For a point of reference, IXS is using spec and is spitting out #L lines that have a very large number of keys, so it seems that spec users are used to long lists of "measured" data, even though they only care about 7 of those numbers. An IXS scan's #L line looks like this:
`#L Time  Epoch  Seconds  Monitor  PD12  PD13  PD14  PD21  PD22  PD23  PD24  PD31  PD32  PD33  PD34  PD41  PD42  PD43  PD44  PD51  PD52  TD1  TD2  TD3  TD4  TD5  TD6  SC1  SRcur  PD11  Detector`

`event()`
special case the motor value to be `-1` for a count, because those are not super helpful in spec anyway, `seq_num` if we cannot guess the motor name and `event_doc['data'][self._motor]` if we do have a valid value for the motor.

I have tried this with the following notebook and it is generating data that can be read by pymca https://gist.github.com/ericdill/2056d44123627f7853f1  

I am not sure how to construct a synthetic motor that has a name but no read signals with that same name.  I'm not even sure if I'm using the right language.  One way to make sure that this works is going to be to get it in front of CHX and run a scan with one of their motors whose `user_readback` was not renamed back to be the same as the motor name.

Anyway, sorry for the lengthy post, but hopefully this is sensible.  Thoughts @danielballan ?

danielballan:
Wow. I'll take a fresh look in the morning. Looks like a lot of detailed attention here -- good work for a Tuesday night.

bluesky/callbacks/core.py
```diff
@@ -570,9 +570,10 @@ def event(self, doc):
 #T {{ acq_time }} (Seconds)
 #P0 {{ positioner_positions | join(' ')}}""")
 
+# It is critical that the spacing on the #L line remain exactly like this!
```
tacaswell: Is this documented someplace?

ericdill: I have no idea

ericdill: I got this information from @lwiegard

ericdill:
Need to add the list of positioners back in to the spec header. Can  probably do this with ophyd.commands._normalize_positioners in the init of the spec callback.

yugangzhang:
@ericdill 
We just a got a error:

```
TypeError                                 Traceback (most recent call last)
/home/xf11id/.ipython/profile_collection/startup/99-bluesky.py in <module>()
     52 from bluesky.callbacks.core import LiveSpecFile
     53 from bluesky.spec_api import ascan, dscan, ct  # this may already be in your profile
---> 54 spec_cb = LiveSpecFile('/home/xf11id/specfiles/spec0.spec')
     55 
     56 for spec_scan in [ascan, dscan, ct]:

/home/xf11id/conda_envs/collection/lib/python3.5/site-packages/bluesky/callbacks/core.py in __init__(self, specpath)
    601         """
    602         from ophyd.commands import _normalize_positioners
--> 603         positioners = _normalize_positioners()
    604         self.specpath = specpath
    605         self.pos_names = ["No", "Positioners", "Were", "Given"]

TypeError: _normalize_positioners() missing 1 required positional argument: 'positioners'
```

ericdill:
@yugangzhang try it again

danielballan:
Remember that calling ophyd commands inside a bluesky callback means it cannot be used later. There is another way to do this. Is it needed immediately?

On Thu, Feb 25, 2016 at 7:55 AM -0800, "Eric Dill" <notifications@github.com<mailto:notifications@github.com>> wrote:


@yugangzhang<https://github.com/yugangzhang> try it again

yugangzhang:
@ericdill It works now. Thanks!

danielballan:
@ericdill What is deployed at CHX? Are they running source-installed bluesky with an unmerged fix?

danielballan:
> Is this documented someplace?

I did a little Googling -- it's a little tricky to Google for a "spec spec" but it only took a minute. The file spec is documented [here](https://github.com/certified-spec/specPy/blob/master/doc/specformat.rst), and it does include the remark, "The delimiteer between labels is a set of two blank spaces." @ericdill Could you include this link in the docstring or a comment?

ericdill:
@danielballan I was editing the source of the conda installed files directly.

This is the diff from current master of what is runnning at chx

```diff
@@ -560,7 +549,7 @@ _SPEC_HEADER_TEMPLATE = env.from_string("""#F {{ filepath }}
 #C {{ owner }}  User = {{ owner }}
 #O0 {{ positioners | join(' ') }}""")
 
-_SPEC_1D_COMMAND_TEMPLATE = env.from_string("{{ scan_type }} {{ start }} {{ stop }} {{ stides }} {{ time }}")
+_SPEC_1D_COMMAND_TEMPLATE = env.from_string("{{ scan_type }} {{ scan_motor }} {{ start }} {{ stop }} {{ strides }} {{ time }}")
 
 _PLAN_TO_SPEC_MAPPING = {'AbsScanPlan': 'ascan',
                          'DeltaScanPlan': 'dscan',
@@ -576,7 +565,7 @@ _SPEC_START_TEMPLATE = env.from_string("""
 
 _SPEC_DESCRIPTOR_TEMPLATE = env.from_string("""
 #N {{ length }}
-#L {{ motor_name }} Epoch Seconds {{ data_keys | join(' ') }}\n""")
+#L {{ motor_name }}    Epoch  Seconds  {{ data_keys | join('  ') }}\n""")
 
 _SPEC_EVENT_TEMPLATE = env.from_string(
     """{{ motor_position }}  {{ unix_time }} {{ acq_time }} {{ values | join(' ') }}\n""")
@@ -610,6 +599,8 @@ class LiveSpecFile(CallbackBase):
         specpath : str
             The location on disk where you want the specfile to be written
         """
+        from ophyd.commands import _normalize_positioners
+        # positioners = _normalize_positioners(None)
         self.specpath = specpath
         self.pos_names = ["No", "Positioners", "Were", "Given"]
         self.positions = ["-inf", "-inf", "-inf", "-inf"]
@@ -670,6 +661,7 @@ class LiveSpecFile(CallbackBase):
                     "this feature at https://github.com/NSLS-II/bluesky/issues" %
                     (len(self._motor), self._motor))
             self._motor, = self._motor
+            content['scan_motor'] = self._motor
             command = _SPEC_1D_COMMAND_TEMPLATE.render(content)
         else:
             err_msg = ("Do not know how to represent %s in SPEC. If "
@@ -690,13 +682,17 @@ class LiveSpecFile(CallbackBase):
         """Write the header for the actual scan data"""
         # List all scalar fields, excluding the motor (x variable).
         self._read_fields = sorted([k for k, v in doc['data_keys'].items()
-                                    if (v['object_name'] != self._motor
-                                        and not v['shape'])])
+                                    if k != self._motor and not v['shape']])
+        # Remove the motor key. It should only be in the list once!
+        #data_keys = list(doc['data_keys'].keys())
+        #data_keys.remove(self._motor)
+        #pdb.set_trace()
         content = dict(motor_name=self._motor,
-                       acq_time=self._acq_time,
-                       unix_time=self._unix_time,
+                       #acq_time=self._acq_time,
+                       #unix_time=self._unix_time,
                        length=3 + len(self._read_fields),
-                       data_keys=doc['data_keys'].keys())
+                       data_keys=self._read_fields)
+        #pdb.set_trace()
         with open(self.specpath, 'a') as f:
             f.write(_SPEC_START_TEMPLATE.render(self._start_content))
             f.write(_SPEC_DESCRIPTOR_TEMPLATE.render(content))
```

danielballan:
Could you make the diff in the form of a commit on this PR so we don't lose track of it?

danielballan:
Sorry, I shouldn't have been so nagging about that. I just don't want CHX to get away from us -- this SPEC stuff is feeling high-maintenance right now. I would really like to see contributions from the beamline staff on this feature.

lwiegart:
Dan, this is ‘high maintenance stuff’ because things under the hood that are out of our control were/keep changing. I sat with Eric at the end of the last run to implement this and hands down, Eric did the programming. However, it took only couple of hours of his time to get all features we dreamed of implemented. Thanks again Eric, awesome job!
So what happened is that some ‘upgrade’ (I don’t question its justification in some long-run) caused a lot of problems with a lot of things that were working perfectly fine, among them the live spec file. I realize that your resources are limited, however, this is equally true for the beamline stuff and we certainly already spent way more time on just getting things back into working order  than we had ever budgeted. Please note that we have to finish beamline design, build test, install and commission components, on top of making user experiments work, getting the beamline ready to pass the ERS and develop a science program. Constantly running into problems while doing things that worked perfectly fine a few weeks earlier is a complete nuisance and already caused postponed user experiments, ‘paused’ user experiments and the disability to perform certain key measurements during a user experiment.
Bottom line, if you decide to change things ‘in the architecture’ that causes tremors throughout all the code written and used at a beamline, you’ll have to budget resources to fix the resulting problems. In the middle of a run and with 16h days, don’t expect too much ‘beamline contribution’.
L

danielballan:
Dear Lutz,

I am sorry for the derogatory connotation of "high maintenance." It was a poor choice of words. I appreciate the time and effort you put into helping Eric draft the SPEC file writer. I do try to be an advocate in our group for "building bridges, not walls," to borrow a cliché phrase. I think inter-operability with the SPEC ecosystem is a good thing. Over the last couple weeks, I have put a couple hours of work into the SPEC-writer myself, and I have added features to bluesky specifically to make the SPEC-writer more informative and powerful. For example, you can now run it retroactively on saved data.

At the same time, I worry about the DAMA group investing much time in utilities like the SPEC file-writer at this time. Yes, we need to provide operational beamlines with the tools they need to function. The more beamline staff can learn to use the existing tools (with ever-improving documentation) the more time we have for tasks that beamline staff *really* do not want to worry about -- the low-level details of making sure everything just works and that nice features like the SPEC-writer can operate safely and reliably.

danielballan:
The [patch above](https://github.com/NSLS-II/bluesky/pull/352#issuecomment-188905691) is missing some lines from the header and is not usable. (The `patch` command dies). I recreated it and made a separate PR.

ericdill:
I am unclear why you decided to make a separate PR...? All of that code is already in this PR

ericdill:
> At the same time, I worry about the DAMA group investing much time in utilities like the SPEC file-writer at this time.

I have a hard time taking comments like this seriously... The entire point of the spec file writer is to allow beamline staff to make use of tools like PyMCA that are great for data visualization, which saves them time because they already know how to use that tool and saves us time because I do not think that we want to get into the business of re-creating the functionality that PyMCA already provides

danielballan:
> I am unclear why you decided to make a separate PR...? All of that code is already in this PR

I was not completely sure that the contents of your pasted diff matched the contents of this PR, and I knew that the diff worked at CHX. I was not sure this PR worked.

bluesky/callbacks/core.py
```diff
@@ -590,6 +590,12 @@ class LiveSpecFile(CallbackBase):
     Other documents can be issues before, between, and after, but
     these must be issued and in this order.
 
+    Notes
+    -----
+    `Reference <https://github.com/certified-spec/specPy/blob/master/doc/specformat.rst>`_
+    for the spec file format.
+    Nice find @danielballan!
```
tacaswell: We do not need this line.

ericdill:
@danielballan This PR does three things.

(1) Fix the format of the spec file (#368)
(2) Attempt to make this callback friendlier by trying to guess what the user intends the x axis to be and fall back to the seq_num from the event document if we cannot guess the motor name
(3) Minor clean up of the code

Without (1), the urgency of this PR is lost and therefore (2) can wait to be properly done (or not at all: make the users type in the correct x axis or we immediately bail to using the seq_num) and (3) was only done because I was in here mucking with the code anyway, but it is not important at all.  Closing this PR is definitely on the table. Thoughts?

danielballan:
Let's hold on this until we hear back from Mark and Vivek.

tacaswell:
At a minimum can you please rebase this so we can see what changes are here?

ericdill:
@tacaswell rebased

ericdill:
Will try again with the current state of bluesky.

PR #357, opened 2016-02-24
Enh snake

tacaswell:
attn @imrealkaren

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -534,13 +534,17 @@ class LiveRaster(CallbackBase):
         self.im = im
         self.ax.set_title('scan {uid} [{sid}]'.format(sid=doc['scan_id'],
                                                       uid=doc['uid'][:6]))
+        self.snaking = doc.get('snaking', (False, False))
+
         cb = self.fig.colorbar(im)
         cb.set_label(self.I)
 
     def event(self, doc):
         seq_num = doc['seq_num'] - 1
-        pos = np.unravel_index(seq_num, self.raster_shape)
-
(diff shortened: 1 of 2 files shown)


bluesky/callbacks/core.py
```diff
@@ -534,13 +534,17 @@ def start(self, doc):
         self.im = im
         self.ax.set_title('scan {uid} [{sid}]'.format(sid=doc['scan_id'],
                                                       uid=doc['uid'][:6]))
+        self.snaking = doc.get('snaking', (False, False))
```
danielballan: Should this be `doc.get('plan_args', {}).get('snaking', (False, False))`?

tacaswell: I think `snaking` goes in both places.

danielballan: I'm 60% sure you are wrong. Code link?

danielballan: ...Or, you know, a test :imp: ... but I would accept a link to the relevant line in plans.py where this happens.

tacaswell: https://github.com/NSLS-II/bluesky/blob/master/bluesky/plans.py#L782

tacaswell: Even better...with blame https://github.com/NSLS-II/bluesky/blame/master/bluesky/plans.py#L782 :smiling_imp:

danielballan: :100: 

When I write code that I can't even recall, I know I'm on my way to @tacaswell levels of glory. (Thanks for obliging me by digging this up.)

PR #363, opened 2016-02-25
ENH: Name the 'main' event stream in built-in plans.

danielballan:
Recall the "baseline readings" use case. Building baseline readings
in by default may wait until the next cycle, but this lays the
groundwork by making the primary event stream addressable by name.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -177,7 +177,7 @@ class Count(PlanBase):
         # over it in different ways in _gen, above.
         dets = self.detectors
         yield Msg('checkpoint')
-        yield Msg('create')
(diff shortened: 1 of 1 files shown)


tacaswell:
Do we want to bikeshed the name 'main' a bit?

 - core
 - primary
 - leader
 - principle
 - head
 - central

(I may have used a thesaurus)

danielballan:
I don't love 'main' but not of these jump out at me either. I guess I like 'primary'. No strong opinion (for once).

danielballan:
Opinions @ericdill @dchabot @klauer? I'd like to use this tomorrow.

klauer:
`main`/`primary` sound fine.

PR #369, opened 2016-03-02
FIX: Stage ultimate parent device (root).

danielballan:
Notice that we are doing this is plans, so it is still *possible* tell the RunEngine to stage a subdevice (not its root) if for some reason we ever want to do this.


klauer:
Looks good to me :+1:

Issue #285, opened 2016-01-21
Consider moving away from PrettyTable


ericdill:
prettytable seems to be abandonware.  The code was originally developed on google code.  The author made the switch to github a few days before google code went read only, as [this] (https://code.google.com/p/prettytable/issues/detail?id=67) issue on google code shows.  `prettytable` can now be found [here] (https://github.com/lmaurits/prettytable) on github, but there has been no work done on the project since July of 2014.  There appears to be an active fork of this project, called [PTable] (https://github.com/kxxoling/PTable) that claims to be backwards compatible with prettytable.  This is not that big of an issue and is only relevant if we discover that prettytable has some bugs or we want better features out of our magic ascii table formatter

PR #370, opened 2016-03-03
PRF/API: refactor LiveTable

tacaswell:
- avoids 'line painters' O(N^2) scaling
 - removes prettytable dep
 - consult descriptor to get precision of values
 - Changed/removed named inputs to LiveTable:
    rowwise -> [removed]
    max_post_decimal -> [removed]
    max_pre_decimal -> dflt_prec
    data_field_width -> min_width
 - changed expected signature of `logbook` to just take a string
 - slightly changed tests to always use formatting given by descriptor
 - added uid/scan_id line at bottom of table


bluesky/callbacks/core.py
```diff
@@ -552,6 +372,142 @@ def event(self, doc):
...
+    fields : list
+         List of fields to add to the table.
+
+    print_header_interval : int, optional
+         Reprint the header every this many lines, defaults to 50
+
+    min_width : int, optional
+         The minimum width is spaces of the data columns.  Defaults to 12
+
+    dflt_prec : int, optional
```
klauer: (bikeshed) Almost no other abbreviations, why such an ugly one here?

klauer:
* `avoids 'line painters' O(N^2) scaling` :+1: thumb fully extended for this portion of the refactor.

bluesky/callbacks/core.py
```diff
@@ -552,6 +372,142 @@ def event(self, doc):
...
+    def descriptor(self, doc):
+        def patch_up_precision(p):
+            try:
+                return int(p)
+            except (TypeError, ValueError):
+                return self._dflt_prec
+
+        dk = doc['data_keys']
+        self._format_info.update(
+            ((k, self._fm_sty(
```
klauer: A dictionary update using a tuple of every kv pair with a namedtuple using positional arguments and some unexplained math + magic numbers... 

I'm genuinely curious (and not in a sarcastic way). Do you seriously find the PR code to be more readable than something like this?
```python
for k in self._fields:
    width = max(self._min_width, 
                len(k) + 2, 
                self._dflt_prec + 1 + 2 * self._pad_len)
    dk_entry = dk[k]
    fmt = self._fm_sty(width=width,
                prec=patch_up_precision(dk_entry.get('precision', self._dflt_prec)),
                dtype=self._FMT_MAP[dk_entry['dtype']]))
             )
    self._format_info[k] = fmt
```

ericdill: Why use a for loop when you can use a 6 line generator comprehension? :smiling_imp:

ericdill:
Out of curiousity, what was the problem with using prettytable?

bluesky/callbacks/core.py
danielballan: Ditto.

danielballan:
Prettytable is not optimized for appending. It needed to redraw the entire table each time a row was added -- leading to the N^2 performance mentioned above. Context: https://en.wikipedia.org/wiki/Joel_Spolsky#Schlemiel_the_Painter.27s_algorithm

bluesky/callbacks/core.py
tacaswell: The first version of this comprehension was much simpler which (almost) fit on one line and then it grew....

tacaswell:
Also #285

ericdill:
@danielballan thanks for the clear explanation of the printing issue.  Are we seeing performance degradation?

@tacaswell Uh, yeah, good point... definitely forgot that I created that issue!

tacaswell:
Yes, at SRX printing out the next line of the table was completely dominating (by ~order of magnitude) per-point time in mid-sized raster scans.

klauer:
@ericdill This was an issue at the HXN ages ago for which I had a work-around (+ a scrapped PR to bluesky). I'll be able to revert to using LiveTable after this.

ericdill:
@tacaswell @klauer thanks for the info

PR #374, opened 2016-03-03
FIX: also do root-checking for unstage

tacaswell:
@danielballan


danielballan:
I simultanously discovered this at CHX. Sorry, guys.

PR #375, opened 2016-03-03
MNT: Allow sample to be dict or string.

danielballan:
We tried to encourage people to use dictionaries to record their sample information by requiring that the field called 'sample' be a dict. This has turned out to be more annoying that helpful. Sometimes you just want sample to be a string. The error is off-putting and more likely causes users to simply avoid the metadata feature than to discover the joy of dictionaries.

This would require a corresponding change in metadataservice and the (so far unused) event-model repo. Is this OK with you @arkilic?

This one's for you, @lwiegart. (But I have come to agree.)

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1416,10 +1416,13 @@ def _default_md_validator(md):
         if field not in md:
             raise KeyError("The field '{0}' was not specified as is "
                            "required.".format(field))
-        if 'sample' in md and not hasattr(md['sample'], 'keys'):
+        if 'sample' in md and not (hasattr(md['sample'], 'keys')
+                                   or isinstance(md['sample'], str)):
             raise ValueError(
                 "You specified 'sample' metadata. We give this field special "
                 "significance in order to make your data easily searchable. "
-                "Therefore, you must make 'sample' a dictionary, like so: "
+                "Therefore, you must make 'sample' a string or a  "
+                "dictionary, like so: "
+                "GOOD: sample='dirt' "
                 "GOOD: sample={'color': 'red', 'number': 5} "
-                "BAD: sample='red_5' ")
+                "BAD: sample=[1, 2] ")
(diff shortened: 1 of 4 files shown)


tacaswell:
:+1:

arkilic:
Yes, if somebody wants to use a dict for sample information, we have amostra and sample uids.

ericdill:
Any reason this cannot be merged?

tacaswell:
Just being cautious and making sure that we have buy in from everyone.

arkilic:
IMO, this is good to go

ericdill:
Great.  Seems like this is a non controversial change.  Does metadataservice/metadataclient/event-model need to be updated before this change goes in?

arkilic:
Yes, here https://github.com/NSLS-II/metadataservice/blob/master/metadataservice/schema/run_start.json#L12 and here https://github.com/NSLS-II/event-model/blob/master/event_model/schemas/run_start.json#L8

ericdill:
I would imagine that metadataservice.schema should be replaced by the event-model project?  Same with bluesky.schema?

arkilic:
Yes, @danielballan and I had this in mind but have not completed the integration yet. Also, we should apply this change to metadatastore as well

danielballan:
I don't think this change affects metadatastore. Does it?

Let's aim to do the schema refactor during downtime.

Unless I am confused, this can go into bluesky at any point and nothing currently-deployed (metadatastore) will break.

On Wed, Mar 9, 2016 at 5:58 AM -0800, "Arman Arkilic" <notifications@github.com<mailto:notifications@github.com>> wrote:


Yes, @danielballan<https://github.com/danielballan> and I had this in mind but have not completed the integration yet. Also, we should apply this change to metadatastore as well

arkilic:
You are right, mds does not need any changes: https://github.com/NSLS-II/metadatastore/blob/master/metadatastore/core.py#L488

arkilic:
I changed it in metadataservice https://github.com/NSLS-II/metadataservice/pull/5.

Once this is merged, it is good to go @tacaswell or @hhslepicka

PR #379, opened 2016-03-07
Api kb interupts

tacaswell:
Closes https://github.com/NSLS-II/wishlist/issues/105

This makes the `SignalHandler` hold onto control of the signal until the context block has exited.  Previously the handler would clean it's self up after the first signal it received which means hitting C-c three or more times could result an a KeyboardInterupt being raised deep in the code.  If this happens in the `finally` block it can leave the RE (and the hardware) in an inconsistent state.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -741,7 +741,6 @@ class RunEngine:
             logger.error("%s", err)
             raise err
         finally:
-            self.state = 'idle'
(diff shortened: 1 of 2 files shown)


danielballan:
Thanks for this careful work.

tacaswell:
There is almost a 1:2 hour:LoC changed on this issue

PR #380, opened 2016-03-09
WIP: Fuzz testing!

ericdill:
This PR is adding a module that can be used for fuzz testing (thanks for the i
term @cowanml!).  This module is attempting to break the RunEngine by throwing 
random messages at it and making sure it can recover into a usable state.  This
PR is still WIP because I am not using all possible message types yet.  I also
need to throw keyboard interrupts at random points too.

Things left to do in this PR:
- [x] Make sure I am throwing all (or at least most) of the messages at the RunEngine
- [x] Weave in full scans from `bluesky.examples` and `bluesky.plans` with the random messages and random interrupts
- [ ] Throw complete garbage messages at the RunEngine just to see what happens.

Messages that are passed in to the RunEngine:

- [x] create
- [x] save
- [x] read
- [x] null
- [x] set
- [x] trigger
- [x] sleep
- [ ] wait
- [x] checkpoint
- [x] clear_checkpoint
- [x] pause
- [x] collect
- [x] kickoff
- [x] configure
- [x] stage
- [x] unstage
- [ ] subscribe
- [ ] unsubscribe
- [x] open_run
- [x] close_run
- [ ] wait_for

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -384,6 +384,7 @@ class FlyMagic(Base):
         self._scan_points = scan_points
         self._time = None
         self._fly_count = 0
+        self.ready = False
 
     def read_configuration(self):
         return OrderedDict()
@@ -395,8 +396,10 @@ class FlyMagic(Base):
         self._fly_count = 0
 
     def kickoff(self):
+        self.ready = False
         self._time = ttime.time()
         self._fly_count += 1
+        self.ready = True
         return self
 
     def describe(self):
@@ -405,7 +408,7 @@ class FlyMagic(Base):
                 {self._det2: {'source': self.name, 'dtype': 'number'}}]
 
     def collect(self):
-        if self._time is None:
+        if self._time is None or not self.ready:
             raise RuntimeError("Must kick off flyscan before you collect")
 
         dtheta = (np.pi / 10) * self._fly_count
@@ -434,6 +437,7 @@ class FlyMagic(Base):
 
     def stop(self):
         pass
+
 
 motor = Mover('motor', ['motor'])
 motor1 = Mover('motor1', ['motor1'], sleep_time=.1)
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -177,6 +177,7 @@ class SynGauss(Reader):
         self.sigma = sigma
         self.noise = noise
         self.noise_multiplier = noise_multiplier
+        self._data = {self.name: {'value': 0, 'timestamp': ttime.time()}}
         if noise not in ('poisson', 'uniform', None):
             raise ValueError("noise must be one of 'poisson', 'uniform', None")
 
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -467,7 +467,7 @@ def conditional_break(det, motor, threshold):
         yield Msg('set', motor, i)
         yield Msg('trigger', det)
         reading = yield Msg('read', det)
-        if reading['det']['value'] < threshold:
+        if reading[det.name]['value'] < threshold:
             print('DONE')
             yield Msg('close_run')
             break
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -405,7 +405,7 @@ motor = Mover('motor', ['motor'])
 motor1 = Mover('motor1', ['motor1'], sleep_time=.1)
 motor2 = Mover('motor2', ['motor2'], sleep_time=.2)
 motor3 = Mover('motor3', ['motor3'], sleep_time=.5)
-noisy_det = SynGauss('det', motor, 'motor', center=0, Imax=1,
+noisy_det = SynGauss('noisy_det', motor, 'motor', center=0, Imax=1,
                      noise='uniform', sigma=1)
 det = SynGauss('det', motor, 'motor', center=0, Imax=1, sigma=1)
 det1 = SynGauss('det1', motor1, 'motor1', center=0, Imax=5, sigma=0.5)
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -243,7 +243,7 @@ class Syn2DGauss(Reader):
                  noise=None, noise_multiplier=1):
         super().__init__(name, [name, ])
         self.ready = True
-        self._motor0 = motor0
+        self._motor = motor0
         self._motor_field0 = motor_field0
         self._motor1 = motor1
         self._motor_field1 = motor_field1
@@ -257,7 +257,7 @@ class Syn2DGauss(Reader):
 
     def trigger(self, *, block_group=True):
         self.ready = False
-        x = self._motor0.read()[self._motor_field0]['value']
+        x = self._motor.read()[self._motor_field0]['value']
         y = self._motor1.read()[self._motor_field1]['value']
         m = np.array([x, y])
         v = self.Imax * np.exp(
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -204,11 +204,32 @@ class Syn2DGauss(Reader):
 
     Parameters
     ----------
+    name : str
+        The name of the detector
+    motor0 : `Mover`
+        The 'x' coordinate of the 2-D gaussian blob
+    motor_field0 : str
+        The name field of the motor. Should be the key in motor0.describe()
+    motor1 : `Mover`
+        The 'y' coordinate of the 2-D gaussian blob
+    motor_field1 : str
+        The name field of the motor. Should be the key in motor1.describe()
+    center : iterable, optional
+        The center of the gaussian blob
+        Defaults to (0,0)
+    Imax : float, optional
+        The intensity at `center`
+        Defaults to 1
+    sigma : float, optional
+        Standard deviation for gaussian blob
+        Defaults to 1
     noise : {'poisson', 'uniform', None}
-        Add noise to the gaussian peak.
-    noise_multiplier : float
+        Add noise to the gaussian peak..
+        Defaults to None
+    noise_multiplier : float, optional
         Only relevant for 'uniform' noise. Multiply the random amount of
         noise by 'noise_multiplier'
+        Defaults to 1
 
     Example
     -------
@@ -218,7 +239,7 @@ class Syn2DGauss(Reader):
     _klass = 'reader'
 
     def __init__(self, name, motor0, motor_field0, motor1, motor_field1,
-                 center, Imax, sigma=1,
+                 center=(0,0), Imax=1, sigma=1,
                  noise=None, noise_multiplier=1):
         super().__init__(name, [name, ])
         self.ready = True
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -1,6 +1,6 @@
 import asyncio
 import time as ttime
-from collections import deque
+from collections import deque, OrderedDict
 import numpy as np
 from .run_engine import Msg
 
@@ -294,11 +294,17 @@ class MockFlyer:
     def done(self):
         return self.ready
 
+    def read_configuration(self):
+        return OrderedDict()
+
+    def describe_configuration(self):
+        return OrderedDict()
+
     def describe(self):
         dd = dict()
-        dd.update(self._mot.describe())
-        dd.update(self._detector.describe())
-        return [dd, ]
+        for k in [self._mot, self._detector]:
+            dd.update(k.describe())
+        return dd
 
     def kickoff(self, start, stop, steps):
         self.success = True
@@ -378,6 +384,12 @@ class FlyMagic(Base):
         self._time = None
         self._fly_count = 0
 
+    def read_configuration(self):
+        return OrderedDict()
+
+    def describe_configuration(self):
+        return OrderedDict()
+
     def reset(self):
         self._fly_count = 0
 
@@ -387,9 +399,10 @@ class FlyMagic(Base):
         return self
 
     def describe(self):
-        return [{k: {'source': self.name, 'dtype': 'number'}
-                 for k in [self._motor, self._det]},
-                {self._det2: {'source': self.name, 'dtype': 'number'}}]
+        dd = dict()
+        for k in [self._motor, self._det, self._det2]:
+            dd.update(k.describe())
+        return dd
 
     def collect(self):
         if self._time is None:
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -252,6 +252,7 @@ class Syn2DGauss(Reader):
         self.sigma = sigma
         self.noise = noise
         self.noise_multiplier = noise_multiplier
+        self._data = {self.name: {'value': None, 'timestamp': None}}
         if noise not in ('poisson', 'uniform', None):
             raise ValueError("noise must be one of 'poisson', 'uniform', None")
 
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -302,9 +302,9 @@ class MockFlyer:
 
     def describe(self):
         dd = dict()
-        for k in [self._mot, self._detector]:
-            dd.update(k.describe())
-        return dd
+        dd.update(self._mot.describe())
+        dd.update(self._detector.describe())
+        return [dd, ]
 
     def kickoff(self, start, stop, steps):
         self.success = True
@@ -399,10 +399,9 @@ class FlyMagic(Base):
         return self
 
     def describe(self):
-        dd = dict()
-        for k in [self._motor, self._det, self._det2]:
-            dd.update(k.describe())
-        return dd
+        return [{k: {'source': self.name, 'dtype': 'number'}
+                 for k in [self._motor, self._det]},
+                {self._det2: {'source': self.name, 'dtype': 'number'}}]
 
     def collect(self):
         if self._time is None:
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1249,6 +1249,15 @@ class RunEngine:
 
     @asyncio.coroutine
     def _pause(self, msg):
+        """Request the run engine to pause
+
+        Expected message signature is:
+
+            Msg('pause', defer=False, name=None, callback=None)
+
+        See RunEngine.request_pause() docstring for explanation of the three
+        keyword arugments in the `Msg` signature
+        """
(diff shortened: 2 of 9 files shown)


cowanml:
Did you look at any of the python fuzzers out there?
Looks like there's a handful:  Fusil, Fuzzing, Sulley, ...

ericdill:
@cowanml I did not look at any fuzzing frameworks.  Truthfully, I never even thought about looking for fuzzing frameworks

ericdill:
Is it the desired behavior that sending an `'unstage'` message to the RunEngine with an object that has not been staged will raise a `KeyError`?

Minimum reproducible example:
```python
from bluesky.examples import motor
from bluesky import Msg
from bluesky.tests.utils import setup_test_run_engine
RE = setup_test_run_engine()
RE([Msg('unstage', motor)])
```

```
KeyError                                  Traceback (most recent call last)
<ipython-input-1-2ed377aa2d7f> in <module>()
      3 from bluesky.tests.utils import setup_test_run_engine
      4 RE = setup_test_run_engine()
----> 5 RE([Msg('unstage', motor)])

/home/edill/dev/python/bluesky/bluesky/run_engine.py in __call__(self, plan, subs, raise_if_interrupted, **metadata_kw)
    567                 exc = self._task.exception()
    568                 if exc is not None:
--> 569                     raise exc
    570             if raise_if_interrupted and self._interrupted:
    571                 raise RunEngineInterrupted("RunEngine was interrupted.")

/home/edill/miniconda/envs/bs35/lib/python3.5/asyncio/tasks.py in _step(***failed resolving arguments***)
    237                 # We use the `send` method directly, because coroutines
    238                 # don't have `__iter__` and `__next__` methods.
--> 239                 result = coro.send(None)
    240             else:
    241                 result = coro.throw(exc)

/home/edill/dev/python/bluesky/bluesky/run_engine.py in _run(self)
    739             logger.error("Run aborted")
    740             logger.error("%s", err)
--> 741             raise err
    742         finally:
    743             # call stop() on every movable object we ever set() or kickoff()

/home/edill/dev/python/bluesky/bluesky/run_engine.py in _run(self)
    713                         raise
    714                     except Exception as e:
--> 715                         self._genstack[-1].throw(e)
    716                     self.log.debug("Response: %r", response)
    717                 except KeyboardInterrupt:

/home/edill/dev/python/bluesky/bluesky/run_engine.py in <genexpr>(.0)
    558         if not isinstance(gen, types.GeneratorType):
    559             # If plan does not support .send, we must wrap it in a generator.
--> 560             gen = (msg for msg in gen)
    561         self._genstack.append(gen)
    562         self._new_gen = True

/home/edill/dev/python/bluesky/bluesky/run_engine.py in _run(self)
    709                         coro = self._command_registry[msg.command]
    710                         self.log.debug("Processing %r", msg)
--> 711                         response = yield from coro(msg)
    712                     except KeyboardInterrupt:
    713                         raise

/home/edill/miniconda/envs/bs35/lib/python3.5/asyncio/coroutines.py in coro(*args, **kw)
    204         @functools.wraps(func)
    205         def coro(*args, **kw):
--> 206             res = func(*args, **kw)
    207             if isinstance(res, futures.Future) or inspect.isgenerator(res):
    208                 res = yield from res

/home/edill/dev/python/bluesky/bluesky/run_engine.py in _unstage(self, msg)
   1247             return
   1248         result = obj.unstage()
-> 1249         self._staged.remove(obj)
   1250         return result
   1251 

KeyError: mover: motor
```

ericdill:
Messages left to hit:

{'collect',
 'configure',
 'kickoff',
 'null',
 'pause',
 'sleep',
 'subscribe',
 'unsubscribe',
 'wait',
 'wait_for'}

ericdill:
Also todo: Include example scans in `bluesky.examples`  in the list of messages that I can throw at the RunEngine

ericdill:
Also todo: discuss and probably back out this change: https://github.com/NSLS-II/bluesky/pull/380/files#diff-918281cac085be7995578c4c2c57cd4eR1325.

bluesky/run_engine.py
```diff
@@ -1212,12 +1309,21 @@ def _stage(self, msg):
...
+
+        Expected message object is:
+
+            Msg('unstage', object)
+        """
         _, obj, args, kwargs = msg
         # If an object has no 'unstage' method, assume there is nothing to do.
         if not hasattr(obj, 'unstage'):
             return
         result = obj.unstage()
```
danielballan: Good catch with this issue -- a win for your fuzz testing efforts.

In ophyd, the rule with staging/unstaging is: If an object is staged, you have to unstage it before you can stage it again. But if an object is unstaged, it's OK to unstage it again (nothing will happen). To reflect this, I think change `remove` to `discard`, which does nothing if the element to be discarded is not present.

bluesky/run_engine.py
```diff
@@ -1182,6 +1263,16 @@ def _clear_checkpoint(self, msg):
...
     def _configure(self, msg):
+        """Configure an object
+
+        Expected message object is:
+
+            Msg('configure', object, *args, **kwargs)
+
+        which results in this call:
+
+            object.configure(*args, **kwargs
```
danielballan: Missing )

bluesky/run_engine.py
```diff
@@ -876,6 +891,13 @@ def _close_run(self, msg):
 
     @asyncio.coroutine
     def _create(self, msg):
+        """Trigger the run engine to start collecting data for an event
```
danielballan: Pardon a suggestion about the wording here. This is fine, but since elsewhere in the codebase we use the word "bundling" it might help to include that here -- "...to bundle future readings into an Event document" for example.

bluesky/run_engine.py
```diff
@@ -876,6 +891,13 @@ def _close_run(self, msg):
...
+
+        Expected message object is:
+
+            Msg('create', None, name='primary')
+            Msg('create')
```
danielballan: Perhaps also note something here that is *completely* undocumented -- "name" will go into the EventDescriptor document associated with this Event.

danielballan:
These docstrings are very welcome! I expect they will be helpful to @hhslepicka and others who joined this party after the RunEngine methods had been written.

fuzz/fuzz.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def kill_func():
+    pid = os.getpid()
+    os.kill(pid, signal.SIGINT)
+    ttime.sleep(0.05)
+    os.kill(pid, signal.SIGINT)
+
+
+def spam_sigint():
```
danielballan: :- D

bluesky/examples.py
```diff
@@ -364,9 +398,10 @@ def kickoff(self):
         return self
 
     def describe(self):
-        return [{k: {'source': self.name, 'dtype': 'number'}
-                 for k in [self._motor, self._det]},
-                {self._det2: {'source': self.name, 'dtype': 'number'}}]
+        dd = dict()
+        for k in [self._motor, self._det, self._det2]:
+            dd.update(k.describe())
+        return dd
```
danielballan: While the return value from `reader.describe()` or `motor.describe()` is expected to be a dict, notice in the deleted lines here that `flyer.describe()` returns a *list* of dicts. That's because flyers can return data for multiple event streams (multiple descriptors). Each dict in the list corresponds to one set of data keys (one descriptor).

This has also caught @klauer, @tacaswell, and myself by surprise at various times. More evidence that `describe` on flyers probably needs a distinct name. But that's a separate issue. Fixing this to return a list should un-break Travis.

bluesky/run_engine.py
ericdill: Where in the event descriptor does this 'name' go?

ericdill: Seems to be literally the "name" field. So essentially `descriptor.name = msg.kwargs['name']` if `msg.kwargs['name']` exists?

danielballan: Exactly. The reason this was introduced is explained [here](https://github.com/NSLS-II/bluesky/pull/313#issuecomment-177662029).

bluesky/run_engine.py
```diff
@@ -423,33 +423,37 @@ def request_pause(self, defer=False, name=None, callback=None):
...
+        self._interrupted = True
+        print("Pausing...")
+        self.state = 'paused'
+        if not self.resumable:
+            # cannot resume, so we cannot pause.  Abort the scan
+            print("No checkpoint; cannot pause. Aborting...")
+            self._exception = FailedPause()
+            self._task.cancel()
+            return
+        # stop processing the event loop
```
danielballan: As long as you're adding a comment here, might as well clarify: "stop putting accepting new tasks in the event loop (existing tasks will still be processed)".

ericdill: Also note that this rewrite of `request_pause` is *totally* up for debate.  I have a habit of trying to un-nest nested if statements whenever I see them.  If you are :-1: on this, I will happily revert the commit.

ericdill: Oh, hah, disregard... I definitely did not see your comments on that commit. Derp.

bluesky/run_engine.py
ericdill: Is it worth warning the user if the object is not in the set?  Something like:

```python
        if obj not in self._staged:
            self.log.warn('obj: %s is not in the set of staged objects', obj)
        self._staged.discard(obj)
```
or
```python
        try:
            self._staged.discard(obj)
        except KeyError as ke:
            self.log.warn('obj: %s is not in the set of staged objects', obj)
```

or just silently eat the discard:

```python
self._staged.discard(obj)
```

Reason for not just silently eating the discard would be if someone writing a custom scan wanted to eke out a little more performance by not calling unstage on objects that were never staged, but I'm not even convinced that is a use-case worth supporting.  Probably simpler to just do the third option above and leave it at that. Thoughts?

danielballan: Yeah, interesting thought, but I vote for #3. The perf is negligible.

ericdill: > Yeah, interesting thought, but I vote for #3. 

Alright. Works for me.

> The perf is negligible.

Is `_stage/_unstage` called at the beginning/end of the scan only?

danielballan: Yes.

ericdill: Ah ok then yes, it is definitely not worth warning the user.  I was not sure if it was called before/after every trigger or something like that.

ericdill:
@danielballan @klauer @tacaswell Fuzzing found another way to get a user unfriendly exception out of the RunEngine.  If you have an open run and call collect before kickoff on a flyer:

```python
from bluesky.examples import motor, det, MockFlyer
from bluesky.tests.utils import setup_test_run_engine
from bluesky import Msg
flyer = MockFlyer(motor, det)
RE = setup_test_run_engine()
RE([Msg('open_run'), Msg('collect', flyer)])
```

produces

```
KeyError                                  Traceback (most recent call last)
<ipython-input-8-2bbedb405a50> in <module>()
----> 1 RE([Msg('open_run'), Msg('collect', flyer)])

/home/edill/dev/python/bluesky/bluesky/run_engine.py in __call__(self, plan, subs, raise_if_interrupted, **metadata_kw)
    569                 exc = self._task.exception()
    570                 if exc is not None:
--> 571                     raise exc
    572             if raise_if_interrupted and self._interrupted:
    573                 raise RunEngineInterrupted("RunEngine was interrupted.")

/home/edill/miniconda/envs/bs35/lib/python3.5/asyncio/tasks.py in _step(***failed resolving arguments***)
    237                 # We use the `send` method directly, because coroutines
    238                 # don't have `__iter__` and `__next__` methods.
--> 239                 result = coro.send(None)
    240             else:
    241                 result = coro.throw(exc)

/home/edill/dev/python/bluesky/bluesky/run_engine.py in _run(self)
    741             logger.error("Run aborted")
    742             logger.error("%s", err)
--> 743             raise err
    744         finally:
    745             # call stop() on every movable object we ever set() or kickoff()

/home/edill/dev/python/bluesky/bluesky/run_engine.py in _run(self)
    715                         raise
    716                     except Exception as e:
--> 717                         self._genstack[-1].throw(e)
    718                     self.log.debug("Response: %r", response)
    719                 except KeyboardInterrupt:

/home/edill/dev/python/bluesky/bluesky/run_engine.py in <genexpr>(.0)
    560         if not isinstance(gen, types.GeneratorType):
    561             # If plan does not support .send, we must wrap it in a generator.
--> 562             gen = (msg for msg in gen)
    563         self._genstack.append(gen)
    564         self._new_gen = True

/home/edill/dev/python/bluesky/bluesky/run_engine.py in _run(self)
    711                         coro = self._command_registry[msg.command]
    712                         self.log.debug("Processing %r", msg)
--> 713                         response = yield from coro(msg)
    714                     except KeyboardInterrupt:
    715                         raise

/home/edill/dev/python/bluesky/bluesky/run_engine.py in _collect(self, msg)
   1100                                          "run is open.")
   1101         obj = msg.obj
-> 1102         self._uncollected.remove(obj)
   1103         stream_name = self._flyer_stream_names.pop(obj)
   1104 

KeyError: <bluesky.examples.MockFlyer object at 0x7f9a69322ac8>
```

Anyone object to catching this KeyError and raising an `IllegalMessageSequence` instead?

ericdill:
Is there any built-in protection from reading a detector before it has been triggered.  If not does it make sense to add such functionality?  I guess this assumes that there is always a 1:1 correspondence between `trigger` and `read`. Is that true?

I guess I'd propose something like this:

```python
def __init__(self, ...):
    ...
    self._triggered = set()
    ...

def _trigger(self, msg):
    ...
    self._triggered.add(msg.obj)
    ...

def _read(self, msg):
    ...
    try:
        self._triggered.remove(msg.obj)
    except KeyError:
        raise IllegalMessageSequence("'read' message received before 'trigger' for object: {}".format(msg.obj))
    ...
```

 I am confident that I do not understand all of the consequences of something like this, or if it is even useful.  Thoughts @danielballan?

danielballan:
I think we can leave it up to the object to set its own trigger/read rules and keep this complexity out of the RunEngine. For example, motors don't need to be (and can't be) triggered.

But I do think that the KeyError you spotted with "collect before kickoff" should be an IllegalMessageSequence.

ericdill:
> I think we can leave it up to the object to set its own trigger/read rules and keep this complexity out of the RunEngine. For example, motors don't need to be (and can't be) triggered.

:+1: for less complexity. Thanks.

> But I do think that the KeyError you spotted with "collect before kickoff" should be an IllegalMessageSequence.

Alright, updated in https://github.com/NSLS-II/bluesky/pull/380/commits/5ddbc2c42abf4222eddb57a26eb84ddf313a08e3

Thanks for the quick response :smile:

tacaswell:
I am 50/50 on the collect `IllegalMessageSequence`  I could imagine a flyer that looks at some existing ring buffer which we might not need to actually kick off.  I suspect all of that logic is also going to get a bit of thrashing when we look into partial collection of flyers.

ericdill:
> I am 50/50 on the collect IllegalMessageSequence

The *only* problem that this is solving is transforming KeyError -> IllegalMessageSequence.  Did you look at the change [here] (https://github.com/NSLS-II/bluesky/pull/380/commits/5ddbc2c42abf4222eddb57a26eb84ddf313a08e3)?

danielballan:
I think @tacaswell is suggesting that we use `discard` instead of `remove`, as we did previously for unstage, to allow users to collect from flyers that may have been kicked off previously (and that the RunEngine doesn't know about). I would rather not allow that until someone needs it, but if you forsee that use case as likely, then we should accommodate it now.

ericdill:
> I think @tacaswell is suggesting that we use discard instead of remove

Ohhh ok, thanks. Sorry I misunderstood.  So IIUC the use case is reading a flyer multiple times?

```python
Msg('open_run')
Msg('kickoff', flyer)
data1 = Msg('collect', flyer)
data2 = Msg('collect', flyer)
```

What should happen if collect is called before kickoff?

tacaswell:
(this was sitting in my browser editor) -> That could also use `discard` and allow it to pass silently. 

As with trigger/read that complexity should be shipped to the `Flyer` objects.

ericdill:
> What should happen if collect is called before kickoff?

Basically the concern is `_collect()` returning data from the last scan. Sorry for not making that clear in my last post.  Is this something that should be taken care of by the stage/unstage methods on the flyers? IIRC all objects get staged/unstaged at the beginning/end of a scan?

ericdill:
And by "something" I mean stage initializes a ring buffer and unstage clears it (or something like that)

tacaswell:
In all of the nicely packaged plan that is the case, but may not always be.

In general I think we should reserve `IllegalMessageSequence` for cases where the RE can not properly interpret the messages (ex 'save' without a paired 'create') and leave issues about detector state to the plan or object level.

The various caches are there to ensure that in the case of a non-graceful exit we can return  everything to a 'safe' state.

ericdill:
Ok, thanks.  I appreciate you guys being willing to bring me up to speed with the design decisions in the RunEngine.  I'll update the _collect to use `set.discard()` and not raise an exception.  We will leave dealing with collect-before-kickoff to the flyer objects to deal with.

ericdill:
This is ready for a review.  I am currently not sending four messages to the run engine: {'unsubscribe', 'subscribe', 'wait', 'wait_for'}.  One other thing that *could* be done is to just send total garbage messages at the run engine, but I'm not sure how useful that is.

danielballan:
Looks great! I think this is complete enough to merge. We can revisit later if inspired to add more coverage.

I took three careful passes through all the changes to `run_engine.py` and a look through the fuzz code. @tacaswell has signed off on this also.

Looks like it needs a rebase, then good to merge.

ericdill:
> Looks great! 

Thanks!

> I think this is complete enough to merge. We can revisit later if inspired to add more coverage.

Sounds good to me.

> Looks like it needs a rebase, then good to merge.

Rebased due to a merge conflict from fa817f0e6e5c70c6a37ec5601c9700d9bb7df791

Also got annoyed that my tests were failing because i didn't have ophyd/pyepics/epics installed and xray_vision.  The last three commits on this PR are pytest related things.  Happy to make that a separate PR if you want.

Should be good to go. Merge away

PR #381, opened 2016-03-09
ENH: Add PLOT_Y to LiveTable by default.

hhslepicka:
The idea is to insert by default the PLOT_Y to the LiveTable when using the SPEC-like API.

### DIFF
--- a/bluesky/simple_scans.py
+++ b/bluesky/simple_scans.py
@@ -39,19 +39,19 @@ from itertools import filterfalse, chain, count
 def table_from_motors(scan):
     "Setup a LiveTable by inspecting a scan and gs."
     # > 1 motor
-    return LiveTable(list(scan.motors) + gs.TABLE_COLS)
+    return LiveTable(list(scan.motors) + gs.TABLE_COLS + [gs.PLOT_Y])
 
 
 def table_from_motor(scan):
     "Setup a LiveTable by inspecting a scan and gs."
     # 1 motor
-    return LiveTable([scan.motor] + gs.TABLE_COLS)
+    return LiveTable([scan.motor] + gs.TABLE_COLS + [gs.PLOT_Y])
 
 
 def table_gs_only(scan):
     "Setup a LiveTable by inspecting a scan and gs."
     # no motors
-    return LiveTable(gs.TABLE_COLS)
+    return LiveTable(gs.TABLE_COLS + [gs.PLOT_Y])
 
 
 def _figure_name(base_name):


danielballan:
Does this behave itself if gs.PLOT_Y is already in gs.TABLE_COLS?

tacaswell:
The new version of live table will happily print the same column more than once.

hhslepicka:
@danielballan 
Based on my tests, yes it will.

```python
In [1]: gs.PLOT_Y
Out[1]: 'det'

In [2]: gs.TABLE_COLS
Out[2]: []

In [3]: gs.TABLE_COLS = ['det']

In [4]: ascan(motor, 0, 10, 10)
Scan ID: 25
Unique ID: e9e417a0-b0b5-47fa-aa11-b3f521c09f38
|   seq_num |       time |      motor |        det |
|         1 | 15:26:51.8 |      0.000 |      1.000 |
|         2 | 15:26:51.9 |      1.000 |      0.607 |
|         3 | 15:26:52.0 |      2.000 |      0.135 |
|         4 | 15:26:52.1 |      3.000 |      0.011 |
|         5 | 15:26:52.2 |      4.000 |      0.000 |
|         6 | 15:26:52.3 |      5.000 |      0.000 |
|         7 | 15:26:52.4 |      6.000 |      0.000 |
|         8 | 15:26:52.4 |      7.000 |      0.000 |
|         9 | 15:26:52.5 |      8.000 |      0.000 |
|        10 | 15:26:52.6 |      9.000 |      0.000 |
|        11 | 15:26:52.7 |     10.000 |      0.000 |
AbsScanPlan ['e9e417'] (scan num: 25)
Out[4]: ['e9e417a0-b0b5-47fa-aa11-b3f521c09f38']

In [5]: ophyd.__version__
Out[5]: '0.2.1+12.gbf8ac9e'

In [6]: bluesky.__version__
Out[6]: '0.4.3+13.gf34bca8'

In [7]:
```

danielballan:
I thought I remembered that LiveTable avoided redundancies. Looks like it still does.

hhslepicka:
What about return a set here to avoid this repeated fields issue:
https://github.com/NSLS-II/bluesky/blob/master/bluesky/callbacks/core.py#L212

danielballan:
I don't think a set is exactly right because users might prefer that we respect the order that they choose.

hhslepicka:
So, in this case, what about add a filter before adding to the list of string_fields.

danielballan:
Yes, that sounds right.

PR #382, opened 2016-03-09
FIX: Avoid reading device and its subdevice.

danielballan:
As discovered at CSX, putting a device and a subdevice into the list of (detectors + motors) causes the RunEngine to error when it see the same reading twice. In this PR, we inspect (sub)device ancestry to make sure a device and its ancestor are not both read.

In each plan, we:
* Determine a unique list of all devices to be triggered and waited on -- that's detectors only.
* Determine a (potentially different) unique list of all devices to read -- that's detectors and motor(s).

As an aside, notice that using the convenience functions introduced in this PR, the `root` attribute of object could potentially be factored out and ultimately removed from ophyd if we want. One problem at a time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ before_install:
 
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
-  - conda update conda
+  - conda install "conda<4"
   - conda create -n testenv nose python=$TRAVIS_PYTHON_VERSION scipy requests matplotlib jsonschema traitlets pytest coverage
   - source activate testenv
   - conda install metadatastore databroker ophyd historydict boltons doct toolz tifffile pymongo pcaspy pyepics readline super_state_machine cycler xray-vision lmfit -c lightsource2 -c conda-forge -c soft-matter
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ install:
   - conda update conda
   - conda create -n testenv nose python=$TRAVIS_PYTHON_VERSION scipy requests matplotlib jsonschema traitlets pytest coverage
   - source activate testenv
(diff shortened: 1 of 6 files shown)


danielballan:
Added tests. I think this is ready for review/merge.

danielballan:
Hmm. Tests pass locally. Travis logs show that conda install ophyd but, inexplicably, an ImportError is raised during the tests, with no 'ophyd' package found. Are our public ophyd packages broken in a weird way?

ericdill:
> inexplicably, an ImportError is raised during the tests, with no 'ophyd' package found.
Quite explicably: https://travis-ci.org/NSLS-II/bluesky/jobs/115153564#L613

ericdill:
Now, why ^ that is happening is quite a mystery

.travis.yml
```diff
@@ -33,7 +33,7 @@ install:
   - conda update conda
   - conda create -n testenv nose python=$TRAVIS_PYTHON_VERSION scipy requests matplotlib jsonschema traitlets pytest coverage
   - source activate testenv
-  - conda install metadatastore databroker historydict boltons doct toolz tifffile pymongo pcaspy pyepics readline super_state_machine cycler xray-vision lmfit -c lightsource2 -c conda-forge -c soft-matter
+  - conda install metadatastore databroker ophyd historydict boltons doct toolz tifffile pymongo pcaspy pyepics readline super_state_machine cycler xray-vision lmfit -c lightsource2 -c conda-forge -c soft-matter
   - conda remove metadatastore databroker filestore --yes
```
ericdill: Or actually forget this. Deleting this comment and the previous one

.travis.yml
```diff
@@ -33,7 +33,7 @@ install:
   - conda update conda
```
ericdill: change this to `conda install conda<4`.  That will fix this "helpful" behavior of conda 4.0.2 where it is doing things like attempting to remove databroker when you do `$ conda remove metadatastore`

Issue #126, opened 2015-06-28
Add an Event stream that records interruptions (pause, resume, suspend)


danielballan:
Two ways to do this:

1. Make one Event Descriptor with a data key for each kind of record ('pause', 'resume', 'suspend', 'unsuspend')

2. Make several Event Descriptors, one per kind of record we want to store. Give it no data keys. Thus, its Events have no data -- only their Event time. Mark the Event Descriptors as being "pause" or "resume" records in some new field like "kind."

danielballan:
I am leaning toward approach (1) -- a single EventDescriptor. This is, roughly speaking, a timestamped history of the state of the RunEngine. There's nothing asynchronous about that, so it should be a single Event stream.

@tacaswell?

tacaswell:
I also prefer 1.

Having a second optional event stream of messages would also be useful.

danielballan:
Closed by #463

PR #387, opened 2016-03-14
ENH: Emit event stream that logs messages before processing them.

danielballan:
This breaks some tests that are not expecting the new event stream that it produces. Once people review the implementation I will go back and fix the tests.

This addresses one half of #126.


tacaswell:
Look, real _meta_ data.

Bit concerned that the down-stream tooling is not good enough yet to cope with always having 2 event streams.

danielballan:
Yeah this is for next cycle.

On Mon, Mar 14, 2016 at 4:06 PM -0700, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


Look, real meta data.

Bit concerned that the down-stream tooling is not good enough yet to cope with always having 2 event streams.

danielballan:
@tacaswell I think you came down against this because of performance concerns. What's your current feeling about this?

tacaswell:
I am +0.  If it is easy to turn off we can address the performance issues when we see them.

danielballan:
I think logging message streams will help us perform forensics on poorly-reported bugs at the beamlines. This might be especially important as we test out the plans refactor and see more frequent use of asynchronous event stream.

I'll fix the tests tomorrow....

danielballan:
This can be done on the plan side with some `log_messages` preprocessor: `RE(log_messages(plan)))`. The RunEngine does not need to be involved, and involving it would add complexity in a couple places, plus a performance hit. Closing.

PR #388, opened 2016-03-15
ENH: Add monitor/unmonitor messages.

danielballan:
WIP, needs tests

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -224,6 +224,7 @@ class RunEngine:
         self._config_values_cache = dict()  # " obj.read_configuration() values
         self._config_ts_cache = dict()  # " obj.read_configuration() timestamps
         self._descriptors = dict()  # cache of {(name, objs_frozen_set): uid}
+        self._monitor_cbs = dict()  # cache of {obj: cb}
         self._sequence_counters = dict()  # a seq_num counter per Descriptor
         self._teed_sequence_counters = dict()  # for if we redo datapoints
         self._pause_requests = dict()  # holding {<name>: callable}
@@ -240,6 +241,8 @@ class RunEngine:
             'create': self._create,
             'save': self._save,
             'read': self._read,
+            'monitor': self._monitor,
+            'unmonitor': self._unmonitor,
             'null': self._null,
             'set': self._set,
             'trigger': self._trigger,
@@ -430,6 +433,8 @@ class RunEngine:
                 self.state = 'paused'
                 if self.resumable:
                     loop.stop()
+                    for obj, cb in list(self._monitor_cbs.items()):
+                        obj.clear_sub(cb)
                     # During pause, all motors should be stopped.
                     for obj in self._movable_objs_touched:
                         try:
@@ -605,6 +610,8 @@ class RunEngine:
 
         self._interrupted = False
         self._rewind()
+        for obj, cb in self._monitor_cbs.items():
+            obj.subscribe(cb)
         self._resume_event_loop()
         return self._run_start_uids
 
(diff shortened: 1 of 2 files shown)


bluesky/run_engine.py
```diff
@@ -924,6 +944,55 @@ def _cache_config(self, obj):
...
+                   configuration=config, name=None,
+                   object_keys=object_keys)
+        yield from self.emit(DocumentNames.descriptor, doc)
+        self.log.debug("Emitted Event Descriptor")
+
+        def emit_event(*args, **kwargs):
+            # Ignore the inputs. Use this call as a signal to call read on the
+            # object, a crude way to be sure we get all the info we need.
+
+            data, timestamps = _rearrange_into_parallel_dicts(obj.read())
```
dchabot: Thoughts on using the data returned in a monitor subscription, rather than re-reading possibly stale info?

danielballan: Absolutely. I'm not clear on what I can count on getting in the args. Do you know off the top of your head what the implementation would have to be?

bluesky/run_engine.py
```diff
@@ -924,6 +944,55 @@ def _cache_config(self, obj):
...
+            # Unlike the RunEngine.emit coroutine, here both dispatchers
+            # get a lossless stream of all documents. Monitors are already
+            # "lossy"; we will not mix in our own lossiness here. If
+            # monitors are generating too much data, they should be
+            # implemented as flyers.
+            self.dispatcher.process(name, doc)
+
+        self._monitor_cbs[obj] = emit_event
+
+        obj.subscribe(emit_event)
```
dchabot: Possibility of subscriptions on events other than `Signal._default_sub`?

bluesky/run_engine.py
dchabot: This is the monster [kwargs](http://pyepics.github.io/pyepics/pv.html#user-supplied-callback-functions) from pyepics. There will also, confusingly, be an `obj` kwarg, a reference to the object on which the subscription is registered.

`value` and `timestamp` are obviously and immediately useful.

danielballan: I guess the real subtlety here, missed in this initial implementation, is that `Device` also has a `subscribe`. How to support that?

dchabot: I was naively assuming `monitor_attrs` would be the filter here.

klauer: @dchabot That's what I keep going back to as well...

danielballan: Does everything in `monitor_attrs` have to be a signal? It can't be a subdevice?

danielballan: ...Or should we just walk the `monitor_attrs` and subscribe to all the signal leaves? That's fine by me -- this implementation was just trying to get the conversation started. As I told @klauer offline, I'll delete it wholesale if it's taking things in the wrong direction. This is maybe 1.5 total hours of writing and discussion up till now. Disposable.

bluesky/run_engine.py
tacaswell: :+1:

danielballan: So, an optional kwarg to `Msg('monitor', ...)`?

danielballan:
Updated after discussion with @klauer and @dchabot. Simple test included.

* The RunEgnine calls `obj.subscribe` directly -- no changes to ophyd are required. Thus this only works on objects that have subscriptions -- mainly signals and positioners for now. (Subscribing to other Devices raises a ValueError out of ophyd.)
* Message args and kwargs are passed through to subscribe. None are required.

bluesky/run_engine.py
```diff
@@ -924,6 +944,54 @@ def _cache_config(self, obj):
...
     @asyncio.coroutine
+    def _monitor(self, msg):
+        obj = msg.obj
+        if not self._run_is_open:
+            raise IllegalMessageSequence("A 'monitor' message was sent but no "
+                                         "run is open.")
+        if obj in self._monitor_cbs:
+            raise IllegalMessageSequence("A 'monitor' message was sent for {}"
+                                         "which is already monitored".format(
+                                             obj))
```
dchabot: One `Msg('monitor', ...)` per object, for now, right?
I guess this will be revisited in a post-alpha implementation?
Thinking about a unification/generalization of a Bluesky interface to EPICS subscriptions...

danielballan: Yes to all of the above.

tacaswell:
@dchabot 'looks like a good start, but do more before merging' or 'looks like a good start, merge away'?

Issue #377, opened 2016-03-05
LiveImage callback can cause python seg fault


gjwillms:
The liveimage plugin (mapplotlib?) is causing occasional seg faults during scans.  The exact fatal error is not consistent.  It has also reported a QWidget/recursive paint fault before dumping ipython.

+-----------+------------+------------+------------+--------------------+------------------------+------------------------+
|   seq_num |       time | hf_stage_x | hf_stage_y | current_preamp_ch2 | pixi_stats1_centroid_x | pixi_stats1_centroid_y |
+-----------+------------+------------+------------+--------------------+------------------------+------------------------+
|         1 | 10:09:06.7 |  29.860010 |  23.470000 |        0.000000001 |                  233.2 |                  171.9 |
|         2 | 10:09:09.1 |  29.882220 |  23.470000 |        0.000000001 |                  233.1 |                  171.9 |
|         3 | 10:09:11.5 |  29.904450 |  23.470000 |        0.000000001 |                  233.1 |                  171.8 |
python3: Fatal IO error 11 (Resource temporarily unavailable) on X server :0.0.

tacaswell:
well, that is greatly exciting....  

How reliable is this?  There is no x-forwarding involved here correct?

I suspect the blitting is to blame.

gjwillms:
Happens about 1 time in 3 attempts.  No forwarding.  Here's a new version of the error:
|        71 | 10:37:41.2 |  29.909985 |  23.677780 |        0.000000001 |                13.4 |                20.9 |
Xlib: sequence lost (0x10fa0 > 0xfa3) in reply type 0xe!
[xcb] Unknown request in queue while dequeuing
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
python3: ../../src/xcb_io.c:179: dequeue_pending_request: Assertion `!xcb_xlib_unknown_req_in_deq' failed.

tacaswell:
Ah, I see the problem.  The event callbacks run an a separate thread, but the `LiveImage` callback is reaching in and forcing GUI loop to run which is a big no-no from not the main thread.  Try using this class instead of the one imported from bluesky:

```python

class LiveImage(CallbackBase):
    """
    Stream 2D images in a cross-section viewer.

    Parameters
    ----------
    field : string
        name of data field in an Event

    Note
    ----
    Requires a matplotlib fix that is not released as of this writing. The
    relevant commit is a951b7.
    """
[...]

PR #390, opened 2016-03-17
FIX: Allow calling LiveImage from another thread.

danielballan:
See https://github.com/NSLS-II/bluesky/issues/377


tacaswell:
Ah, sorry for letting this get lost!

PR #393, opened 2016-03-23
Restream convenience callbacks

danielballan:
Related to https://github.com/NSLS-II/ophyd/pull/307

### DIFF
--- a/bluesky/broker_callbacks.py
+++ b/bluesky/broker_callbacks.py
@@ -1,6 +1,6 @@
 import os
 import time as ttime
-from databroker import DataBroker as db, get_events
+from databroker import DataBroker as db, get_events, restream
 from databroker.databroker import fill_event
 import filestore.api as fsapi
 from metadatastore.commands import run_start_given_uid, descriptors_by_start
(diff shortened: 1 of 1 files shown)


bluesky/broker_callbacks.py
```diff
@@ -81,20 +81,37 @@ def f(name, stop_doc):
...
-        # For convenience, I'll rely on the broker to get Events.
-        header = db[uid]
-        events = get_events(header)
-        callback('start', start)
-        for d in descriptors:
-            callback('descriptor', d)
-        for e in events:
-            callback('event', e)
-        callback('stop', stop_doc)
+        return restream(db[uid], callback)
```
tacaswell: I think the test for this is not hooked up to insert into mds.

danielballan:
Revisiting this now, the new function, `make_restreamer` is so dumb it probably shouldn't even exist. But the refactor of  `post_run` is still worth doing. Thoughts?

tacaswell:
It might be a useful shim layer to keep around  Could also be used as a decorator.

tacaswell:
any reason not to merge this?

danielballan:
Worth merging, I guess. Might save someone ~10 minutes of thinking in the future.

PR #395, opened 2016-03-23
Running callbacks on a separate host or process using 0MQ

danielballan:
One or more RunEngines, running on one or more hosts, publish documents to a central hub -- in 0MQ jargon, a [forwarder device](https://learning-0mq-with-pyzmq.readthedocs.org/en/latest/pyzmq/devices/forwarder.html). Then, document consumers on any host subscribe to the central hub and receive a stream of documents.

Demo:

**Updated 8 August with new API**

1. Start the central hub:
    ```
    # uses config in /etc/zmq_forwarder_device.yml or ~/.config/zmq_forwarder_device/connection.yml
    $ python bluesky/examples/forwarder_device.py
    ```

2. Start a callback that receives documents from the hub.

    ```python
    In [1]: from bluesky.callbacks.zmqsub import RemoteDispatcher

    In [2]: d = RemoteDispatcher(<hostname>, <port>)

    In [3]: d.subscribe('all', print)

    In [4]: d.start()  # runs event loop forever
    ```

3. Hook up a subscription to publish documents to the hub. Finally, generate some documents with a simple plan.
    ```python
    In [2]: from bluesky.callbacks.zmqpub import Publisher
    In [6]: Publisher(RE, <hostname>, <port>)

    In [7]: RE([Msg('open_run'), Msg('close_run')])
    ```

Meanwhile, the callback from (2) prints:
```python
start
stop
```

The connection between the publisher and the subscriber is lossless. (Messages are cached on the publisher side if the subscriber is slow.) But if a subscriber starts *after* publication has begun, it would be nice if it could check the broker or some other cache to get itself caught up with any documents from the current run that is missed.


ericdill:
> But if a subscriber starts after publication has begun, it would be nice if it could check the broker or some other cache

What is the rationale for not making this hub the thing that caches the documents?

danielballan:
This morning, that seemed like the obvious solution to me, too. Then I realized it is much harder to implement than it seems at first. I still think it's possible, but the jury is still out on whether is it the best approach at hand.

On Wed, Mar 23, 2016 at 4:41 PM -0700, "Eric Dill" <notifications@github.com<mailto:notifications@github.com>> wrote:


But if a subscriber starts after publication has begun, it would be nice if it could check the broker or some other cache

What is the rationale for not making this hub the thing that caches the documents?

cowanml:
For big docs with a couple subscribers, or even small docs but many
subscribers, using multicast would also make it *way* more efficient.
Would be especially cool if the sender could cycle through different
multicast groups in some optimised pattern (different groups for
different nodes/segments in a stream processing cluster but central
storage/backup/cloud transfer nodes members of all groups).

That's what I've dreamed of for the Eiger.  But even if Dectris didn't
deliver a black box, I'm not sure the filewriter and streaming
interfaces can run simultaneously? (filewriter for handling missed imgs)


On Wed, 23 Mar 2016, Dan Allan wrote:
...
> The connection between the publisher and the subscriber is
> losses. (Messages are cached on the publisher side if the subscriber
> is slow.) But if a subscriber starts after publication has begun, it
> would be nice if it could check the broker or some other cache to
> get itself caught up with any documents from the current run that is
> missed.

lossless?  What zmq pattern are you looking at?  Doesn't necessarily
have to "catch up" by getting and processing any missed docs, just has
to do whatever the right thing to do is to be able to jump in and
start being productive.  For big parallel stuff, nodes should be able
to come and go very dynamically.  Ideally a node wouldn't need to know
anything about a run (zero synchronization overhead), more
realistically we could aim for fetching a minimal amount of required
current run metadata, but not doing any catch up processing.
Something needing to catch up seems like either something wasn't built
right or there was a catastrophic failure.

danielballan:
The 0MQ pattern used here is publish/subscribe. There are other technologies that could do this; I am aiming for a simple implementation using a technology that the DAMA group has at least some exposure to.

Once the subscriber is active, it is guaranteed to receive every document published by the publisher ("lossless"). In the pub/sub model, it never sees documents that passed by before it subscribed. Our document model demands that subscribers sees an entire run of documents at a time. So, subscribers must either wait for the beginning of the next run or have some means of "catching up" on documents they missed in the current run. This would not be shoehorned into pub/sub; it would be achieved through a separate client/server request either through a separate port or through the databroker directly.

cowanml:
On Thu, 24 Mar 2016, Dan Allan wrote:
> Once the subscriber is active, it is guaranteed to receive every
> document published by the publisher ("lossless").

That's not quite as simple as it might seem :(

http://zguide.zeromq.org/page:all#High-Water-Marks

'You should appreciate that making "reliable" messaging is complex
enough that you always need to ask, "Do we actually need this?" before
jumping into it. If you can get away with unreliable or with "good
enough" reliability, you can make a huge win in terms of cost and
complexity.'
http://zguide.zeromq.org/page:all#Adding-the-Binary-Star-Pattern-for-Reliability

'ZeroMQ guarantees to deliver all the parts (one or more) for a message, or none of them.'


> In the pub/sub model, it never sees documents that passed by before
> it subscribed. Our document model demands that subscribers sees an
> entire run of documents at a time.

Sounds like that would preclude streaming?  Although maybe that's really
a different use case?

Seems better for a late subscriber to only need to fetch the minimal
metadata needed to do whatever the subscriber needs to do with the
current document.

Or maybe by "entire run" you meant start/stop/descriptors (~"minimal
metadata") but not necessarily every doc (other events, unneeded meta)
in the run?


> So, subscribers must either wait for the beginning of the next run
> or have some means of "catching up" on documents they missed in the

Maybe "missed but needed"?  Hopefully shouldn't need everything they
missed, else an active peer or failover subscriber might be more
optimal.  Or you could drop+cleanup, where a subscriber seeing a
partial set of the docs it would need just drops them and separate
cleanup subscribers look for and handle processing stuff that falls
through the cracks; that seems less streamy though.


> current run. This would not be shoehorned into pub/sub; it would be
> achieved through a separate client/server request either through a
> separate port or through the databroker directly.

yup.

danielballan:
This is 99 lines of code including docstrings. I'm happy to throw it into the docs instead if people have concerns about the long-term support burden of this approach. I think an EPICS-based approach would also be useful internally. But I do think *some* approach using standard web technologies will need to be supported: when we push data to external computing facilities, we are unlikely to choose EPICS.

cowanml:
The recommendations from the hdrmx workshop last week should be out
soon, should touch on standardized streaming apis and processing
pipelines.

http://medsbio.org/meetings/BNL_May16_HDRMX_Meeting.html

-matt

danielballan:
To demo LivePlot working remotely...

Make new config file, as with mds an fs:
```
#~/.config/zmq_forwarder_device.yml
{'frontend_port': 5577
'backend_port': 5578
'host': 'localhost'}  # optional
```

Terminal 1:
```bash
python bluesky/examples/forwarder_device.py
```

Terminal 2:
```
import matplotlib
matplotlib.use('Qt4Agg')

plt.ion()

from bluesky.callbacks.zmqsub import RemoteDispatcher

install_qt_kicker()
import bluesky.utils
assert bluesky.utils._QT_KICKER_INSTALLED is not None

d = RemoteDispatcher('localhost', 5578)
d.subscribe('all', LivePlot('det', 'motor'))
d.start()
```

Terminal 3:
```python
from bluesky.callbacks.zmqpub import Publisher
RE = RunEngine({})
import bluesky.plans as bp
from bluesky.examples import motor, det
motor._fake_sleep = 0.5
p = Publisher(RE, 'localhost', 5577)
RE(bp.scan([det], motor, 1, 10, 100))
```

bluesky/callbacks/zmqpub.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Example
+    -------
+    >>> Publisher(RE, '127.0.0.1', 5567)
+    """
+    def __init__(self, RE, host, port):
+        self._host = host
+        self._port = int(port)
+        self._RE = RE
+        url = "tcp://%s:%d" % (host, port)
```
klauer: Should use `self._port` here (`port` could be a `str` here but `self._port` is an int)

bluesky/callbacks/zmqpub.py
```diff
@@ -0,0 +1,54 @@
...
+        self._socket.connect(url)
+        self._subscription_token = RE.subscribe('all', self)
+
+    @property
+    def host(self):
+        return self._host
+
+    @property
+    def port(self):
+        return self._post
```
klauer: `post`

danielballan: Yeah, I wasted a Travis cycle to discover that typo. Womp womp.

bluesky/callbacks/zmqsub.py
```diff
@@ -0,0 +1,102 @@
...
+import zmq
+import zmq.asyncio
+import asyncio
+import time
+from ..utils import expiring_function
+from ..run_engine import Dispatcher, DocumentNames
+
+
+_loop = zmq.asyncio.ZMQEventLoop()
+asyncio.set_event_loop(_loop)
```
klauer: Is it a good idea to have `set_event_loop` on import?

bluesky/callbacks/zmqsub.py
```diff
@@ -0,0 +1,102 @@
...
+    @property
+    def host(self):
+        return self._host
+
+    @property
+    def port(self):
+        return self._port
+
+    @asyncio.coroutine
+    def _poll(self):
```
klauer: Shouldn't this just be a coroutine with an endless loop?

Where does `self._loop` get set?

bluesky/callbacks/zmqsub.py
danielballan: This was the most confusing aspect of asyncio + zmq. From my experiments this morning, I think the correct (ZMQ-supporting) loop has to be set up at coroutine *definition* time. I can't think of a good reason why this should be true, but it does seem to be true.

bluesky/callbacks/zmqsub.py
klauer: Distracted incomplete comment... I see that it's now using the global `_loop` (my brain inserted `self.`)

danielballan: `self._loop` was a typo/holdover. I pushed an update; now it's `_loop`. By making this a coroutine that schedules itself again and also schedules the processing step, we leave it up to the Selector how to intersperse processing and msg pulling.

bluesky/callbacks/zmqsub.py
```diff
@@ -0,0 +1,102 @@
...
+        message = yield from self._socket.recv()
+        hostname, pid, RE_id, name, doc = message.decode().split(' ', 4)
+        doc = ast.literal_eval(doc)
+        self._loop.create_task(self._poll())
+        start_time = time.time()  # for timeout, if used below
+        yield from asyncio.sleep(0)  # Give scheduler second chance to switch.
+        if self._is_our_message(hostname, pid, RE_id):
```
klauer: As I mentioned in person, I think it will be important to come up with a standardized way to allow access to bundled messages (or the whole queue). 

For example, if there are 10 points to be added to a 2D `imshow`/`LiveImage`, the plot does not have to be redrawn for each point, but only once with the full set. It would not be appropriate to drop those points nor send them out of order just because the routine takes too long to redraw.

danielballan:
Codecov is confused because most of the code lines are only hit in a separate process. The coverage is upward of 90%.

bluesky/callbacks/zmqsub.py
klauer: So long as you have a `yield from`, shouldn't that suffice? Looking at the `zmq.examples.asyncio.coroutines` example [here](https://github.com/zeromq/pyzmq/blob/master/examples/asyncio/coroutines.py), the idle `ping` routine gets scheduled without issue alongside a readable, endless coroutine receiver loop.

klauer:
Good start overall, @danielballan!

klauer:
👍

danielballan:
@tacaswell Ready for your review/merge. Travis is happy, and codecov is (still) just confused by subprocesses. The LivePlot demo (in my comment above) still works.

@klauer and I came to a consensus that we would remove the timeout logic and push queue management in general to the next layer or some different implementation of a subscriber, TBD, with expert input from people who understand message queues.

danielballan:
Tom says, "If you and Ken are happy, merge it." Merging.

PR #396, opened 2016-03-25
rewrite suspenders

tacaswell:


bluesky/suspenders.py
```diff
@@ -27,37 +22,44 @@ class PVSuspenderBase(metaclass=ABCMeta):
...
+            The run engine instance this should work on
+
+        event_type : str, optional
+            The event type (subscription type) to watch
         '''
-        if self._indx is not None:
-            self.remove()
-        self._indx = self._pv.add_callback(self)
+        self.RE = RE
+        self._loop = RE.loop
```
danielballan: Below, you get `loop` by accessing `self.RE.loop`. I think you can remove this private attribute entirely. I can't think of a reason that `self._loop` should be anything but the loop that `self.RE` is using.

bluesky/suspenders.py
```diff
@@ -97,35 +99,42 @@ def _should_resume(self, value):
...
+                # clear that we have an event
+                self._ev = None
 
-                    def local():
-                        self._loop.call_later(sleep, ev)
-                    self._loop.call_soon_threadsafe(local)
-                    # clear that we have an event
-                    self._ev = None
+    @property
+    def alarm(self):
```
danielballan: Since "alarm" has a specific meaning in EPICS, I would rather called this something like `suspended` or `active`.

tacaswell: The name collision was intentional

klauer: As we discussed yesterday, unless you plan on following the minor/major/etc alarm schemes of EPICS, it may be better to rename these.

bluesky/suspenders.py
```diff
@@ -97,35 +99,42 @@ def _should_resume(self, value):
...
+                self._alarm = False
+                if self._ev:
+
                     ev = self._ev.set
                     sleep = self._sleep
+                    if self.RE is not None:
+                        loop = self.RE.loop
+
+                        def local():
+                            loop.call_later(sleep, ev)
```
klauer: This made me look into `asyncio.Event` to see if it was callable before I saw what you were doing. For readability, please make this `ev.set`

tacaswell: I have no memory of why I did it this way...

tacaswell:
@klauer @danielballan This is ready for a preliminary review.

Still need to add a registry of suspenders to the RE to ask for permission to start a `__call__`

tacaswell:
attn @imrealkaren @gjwillms

bluesky/suspenders.py
```diff
@@ -97,40 +106,69 @@ def _should_resume(self, value):
...
+                        self._ev.wait(),
+                        pre_plan=self._pre_plan,
+                        post_plan=self._post_plan)
+
+                    loop.call_soon_threadsafe(cb)
+            elif self._should_resume(value):
+                self._tripped = False
+                self.__set_event()
+
+    def __make_future(self):
```
ericdill: why annoyingly private instead of just private? Haven't seen you use this pattern before, so I am curious

tacaswell: because this is something that subclasses should not be messing with.  I used it in the mds/fs objects as well to hide the 'direct' collection objects. 

This code is a bit fiddly and the class in intended to be sub-classed, so hide the guts.

ericdill:
Regarding the choice of `install` and `remove`.  I would not expect these two functions to be a pair as install/uninstall or add/remove are the terms that are typically paired.  Could you elaborate on why install/remove was chosen?

bluesky/suspenders.py
danielballan: This signals, "No, really, don't touch this." And it is harder to access: http://www.diveintopython.net/object_oriented_framework/private_functions.html

ericdill: Thanks for the explanation @tacaswell

bluesky/run_engine.py
```diff
@@ -552,15 +549,18 @@ def __call__(self, plan, subs=None, *, raise_if_interrupted=False,
...
 
-        gen = iter(plan)  # no-op on generators; needed for classes
-        if not isinstance(gen, types.GeneratorType):
-            # If plan does not support .send, we must wrap it in a generator.
-            gen = (msg for msg in gen)
-        self._genstack.append(gen)
-        self._new_gen = True
+        gen = ensure_generator(plan)
+
+        self._plan_stack.append(gen)
```
danielballan: Test this.

bluesky/run_engine.py
```diff
@@ -729,16 +770,16 @@ def _run(self):
...
                         coro = self._command_registry[msg.command]
                         self.log.debug("Processing %r", msg)
                         response = yield from coro(msg)
+                        self._response_stack.append(response)
                     except KeyboardInterrupt:
                         raise
                     except Exception as e:
-                        msg = self._genstack[-1].throw(e)
-                        self._genstack.append((m for m in [msg]))
+                        msg = self._plan_stack[-1].throw(e)
```
danielballan: `self._response_stack.append(None)`

danielballan:
:+1:

bluesky/run_engine.py
klauer: @tacaswell did you test this?

tacaswell:
And I got back 2% coverage woo!

klauer:
@tacaswell I think you deserve the honors of hitting the button with this big rewrite. But you won't self-merge, so @danielballan?

bluesky/run_engine.py
tacaswell: Yes, https://github.com/NSLS-II/bluesky/pull/396/files#diff-d329e108757b258661c50410aab37659R89

danielballan:
It was an honor and a pleasure, sir.

PR #402, opened 2016-04-05
MNT: Remove spec callback from bluesky callbacks

ericdill:
It is going into suitcase via NSLS-II/suitcase#11


ericdill:
Thanks @danielballan !  

attn @yugangzhang: please do not update bluesky to the latest dev version until I have the suitcase PR done!

danielballan:
Yeah, I was going to say: running LiveSpecCallback should be part of the CHX UAT (user-acceptance test). I expect their IPython profile will break on import until it is updated, so this will be easy to catch.

PR #406, opened 2016-04-07
ENH: add MAGICs

tacaswell:
Add a helper function that registers an IPython magics to run call
the run engine.

This function can/should be called in the beamline configuration
once the 'canonical' runengine has been set up.

This is support work for a coming PR from @danielballan

### DIFF
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -599,12 +599,23 @@ def update_sub_lists(out, inp):
             out[k] = list(v)
 
 
-def register_transform():
-    '''Register < transform
+def register_transform(RE, *, prefix='<'):
+    '''Register RunEngine convenience transform
+
+    Assuming the default parameters
 
     This maps `< stuff(*args, **kwargs)` -> `RE(stuff(*args, **kwargs))`
 
     RE is assumed to be available in the global namespace
+
+    Parameters
+    ----------
+    RE : str
+        The name of a valid RunEngine instance in the global IPython namespace
+
+    prefix : str, optional
+        The prefix to trigger this transform on.  If this collides with
+        valid python syntax or an existing transform you are on your own.
     '''
     import IPython
     from IPython.core.inputtransformer import StatelessInputTransformer
@@ -613,7 +624,7 @@ def register_transform():
     def tr_re(line):
         if line.startswith('<'):
             line = line[1:].strip()
-            return 'RE({})'.format(line)
+            return '{}({})'.format(RE, line)
         return line
     ip = IPython.get_ipython()
     ip.input_splitter.logical_line_transforms.append(tr_re())
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -597,3 +597,24 @@ def update_sub_lists(out, inp):
             out[k].extend(v)
         except KeyError:
             out[k] = list(v)
+
+
+def register_transform():
+    '''Register < transform
+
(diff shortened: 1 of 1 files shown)


bluesky/utils.py
```diff
@@ -585,3 +585,38 @@ def _qt_kicker():
...
+            # You must call the parent constructor
+            super().__init__(shell)
+            self.RE = RE
+
+        @line_magic
+        def RE(self, line):
+            "my line magic"
+            if line:
+                inp = eval(line, self.shell.user_ns)
+                self.RE(inp)
```
klauer: `ret = `?

tacaswell: yup

klauer:
My alternative suggestion:

```python
import IPython
from IPython.core.inputtransformer import StatelessInputTransformer


@StatelessInputTransformer.wrap
def tr_run_engine(line):
    if line.startswith('>>'):
        line = line[2:].strip()
        return 'RE({})'.format(line)
    return line


def RE(value):
    print('runengine!', value)


ip = IPython.get_ipython()
ip.input_splitter.logical_line_transforms.append(tr_run_engine())
ip.input_transformer_manager.logical_line_transforms.append(tr_run_engine())
```

Result:

```python
In [4]: >> 'test'
runengine! test

In [5]: foo = 'a'

In [6]: >> foo
runengine! a
```

klauer:
This default input transformer is also rather interesting:
https://github.com/ipython/ipython/blob/master/IPython/core/inputtransformer.py#L39

```python
In [1]: def fcn(*args, **kwargs):
   ....:     print('args is', args, 'kwargs is', kwargs)

In [2]: /fcn 1 2 3 a=4 b=5 c=6
args is (1, 2, 3) kwargs is {'b': 5, 'c': 6, 'a': 4}
```

danielballan:
I am skeptical that this is actually better than convincing people to type `RE(...)` in the end, but I think it is a worthwhile experiment. I am in favor of merging it and seeing how people like it.

danielballan:
Why not just a single `<`? There is still no ambiguity:

```python
In [1]: < 1
  File "<ipython-input-1-732f6b4d3ec1>", line 1
    < 1
    ^
SyntaxError: invalid syntax
```

ericdill:
>  am skeptical that this is actually better than convincing people to type RE(...) in the end

I totally agree.  @yugangzhang seemed pretty happy to just throw plans at the run engine via the RE() syntax, which feels like big praise to the notion of plans.

> but I think it is a worthwhile experiment. I am in favor of merging it and seeing how people like it

I'm pretty concerned that this experiment is going to result in bug reports like:

```python
def foo():
    < ct()
    < ct()
    < ct()
```

and essentially cause more work in the short term.  That being said, experiments are always a good thing when we are trying to converge on the right user interface here and I think that doing this experiment will ultimately result in a better user interface.

Interestingly, I initially started this comment strongly in the "don't do this. this will result in more work" camp, but through the process of typing out my reasoning, I have convinced myself that it is worth doing because even though I think it will ultimately fail, we will learn important things about what the users actually want and is thus very worth doing.

danielballan:
Thanks for the comments, @ericdill. Now that we've all slept on this for awhile and the related SPEC API change is about ready to go, it's time to make a decision. Where is everyone standing on this?

danielballan:
Signs already point to this being maybe a bad idea, but I think we all agree that it might be worth trying. I'm merging it so we have it handy if/when this comes up.

klauer:
I'm just glad `git blame` points to @tacaswell. 👍

PR #413, opened 2016-04-22
Add monitor_context (and -- unrelated -- sundry RunEngine docstrings)

danielballan:
This PR is based on #345 which should be merged first.

See docstring of `monitor_context` for usage and details.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -625,6 +625,49 @@ def baseline_context(plan_stack, devices, name='baseline'):
         plan_stack.append(trigger_and_read(devices))
 
 
+@contextmanager
+def monitor_context(plan_stack, signals):
+    """
+    Asynchronously monitor signals, generating separate event streams.
+
+    Upon exiting the context, stop monitoring.
+
+    Parameters
+    ----------
+    plan_stack : list-like
+        appendable collection of generators that yield messages (`Msg` objects)
+    signals : dict or list
+        either a dict mapping Signals to event stream names or simply a list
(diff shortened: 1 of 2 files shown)


bluesky/plans.py
```diff
@@ -625,6 +625,49 @@ def baseline_context(plan_stack, devices, name='baseline'):
...
+    With no event stream names
+    >>> with monitor_context(plan_stack, [sig1, sig2]):
+            ...
+    """
+    if hasattr(signals, 'items'):
+        # interpret input as dict of signals mapped to event stream names
+        pass
+    else:
+        # interpet input as list of signals
+        signals = {sig: None for sig in signals}
```
klauer: `RE._monitor` has `name` as required, but if you pass a list to `monitor_context`, name=`None` is assumed. I think I'd prefer it to be explicit here.

klauer: Scratch that, it's the same. But what does `None` mean as a stream name?

danielballan: In this context, `None` means "unspecified." If no stream name is specified, then you can't look up the data by its stream name. It's a optional convenience, a "tag."

danielballan: Would it be better to say, if name is not specified, don't give the event descriptor a name at all? Instead of `name=None`, leave the `name` key off entirely? I would be happy either way.

klauer: As-is is fine for now, the long-awaited function docstrings are too enticing to not merge this.

Issue #407, opened 2016-04-11
A newly-running run engine should be in 'checkpointed' state.


danielballan:

tacaswell:
*grumble* dangerous *grumble*

I'll implement this.

ericdill:
What is the rationale for this?

danielballan:
@ericdill Sorry I missed this question. This was actually suggested by @tacaswell, so I'm confused by his "grumble" above. Anyway, if a scan happens to be interrupted right at the beginning, before it has had a chance to process a 'checkpoint', it dies. If instead it treats the beginning of the plan as a de-facto checkpoint, it can always recover -- no matter when the interruption happens. Users can always issue a 'clear_checkpoint' message if they want to make the scan die upon interruption.

danielballan:
@tacaswell I did this in #407 -- a two-line change plus a bug-fix I found while looking through this.

danielballan:
Closed by #407

PR #414, opened 2016-04-22
RunEngine starts in a resumable state

danielballan:
Previous, if interrupted before the first 'checkpoint' message, it died. Now, unless it sees a 'clear_checkpoint' message, it rewinds to the beginning if interrupted.


danielballan:
Closes #407

tacaswell:
attn @imrealkaren 

This will fix (even with out the suspender refactor) some of your issues.

PR #417, opened 2016-04-23
FIX: Cache kwargs for subscribe.

danielballan:
Current situation:

1. A 'monitor' Msg is processed -> `obj.subscribe(cb, *args, **kwargs)`
2. Plan is paused; monitoring is correctly suspended -> `obj.clear_sub(cb)`
3. Plan is resumed; monitoring is reapplied but *the original args and kwargs are dropped* -> `obj.subscribe(cb)`

This PR stashed those and args and kwargs and correctly reapplies them upon resuming.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -203,7 +203,7 @@ class RunEngine:
         self._config_values_cache = dict()  # " obj.read_configuration() values
         self._config_ts_cache = dict()  # " obj.read_configuration() timestamps
         self._descriptors = dict()  # cache of {(name, objs_frozen_set): uid}
-        self._monitor_params = dict()  # cache of {obj: (cb, args, kwargs)}
+        self._monitor_params = dict()  # cache of {obj: (cb, kwargs)}
         self._sequence_counters = dict()  # a seq_num counter per Descriptor
         self._teed_sequence_counters = dict()  # for if we redo datapoints
         self._pause_requests = dict()  # holding {<name>: callable}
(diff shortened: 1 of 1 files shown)


tacaswell:
Can we make this only pass through `kwargs`?  It looks like `ophydobject.subscribe` has 2 explicit optional kwargs (and does not take `*args`/`**kwargs`).  I would also like to make those keyword only on the ophyd side.

danielballan:
Sure, that make sense. Changes pushed.

bluesky/run_engine.py
```diff
@@ -640,10 +640,9 @@ def request_suspend(self, fut):
             self._exception = FailedPause()
         else:
             print("Suspending....To get prompt hit Ctrl-C to pause the scan")
-            objs_monitored = list(self._monitor_cbs)
```
tacaswell: this line should not be removed

bluesky/run_engine.py
```diff
@@ -640,10 +640,9 @@ def request_suspend(self, fut):
...
-            for obj, cb in list(self._monitor_cbs.items()):
+            for obj, (cb, kwargs) in list(self._monitor_params.items()):
                 obj.clear_sub(cb)
-                del self._monitor_cbs[obj]
+                del self._monitor_params[obj]
             wait_msg = Msg('wait_for', [fut, ])
             for obj in objs_monitored:
                 msg = Msg('monitor', obj)
```
tacaswell: and the kwarg needs to go in here.

danielballan: Ugh. Told you I was tired. Fix pushed.

PR #418, opened 2016-04-23
ENH: Add a stub plan for every command in the command_registry.

danielballan:
This enables usage like:

```python
@planify
def plan():
    return [ct(), sleep(5), ct(), rel_set(motor, 2), ct()]
```

It's just an experiment.

One big benefit is that this provides de-facto API documentation for every kind of Msg, which has till recently been completely absent or buried in the RunEngine private method docstrings.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1043,11 +1043,9 @@ def baseline_context(plan_stack, devices, name='baseline'):
     name : string, optional
         name for event stream; by default, 'baseline'
     """
-    with event_context(plan_stack, name=name):
-        plan_stack.append(trigger_and_read(devices))
+    plan_stack.append(trigger_and_read(devices), name=name)
     yield
-    with event_context(plan_stack, name=name):
-        plan_stack.append(trigger_and_read(devices))
+    plan_stack.append(trigger_and_read(devices), name=name)
 
 
 @contextmanager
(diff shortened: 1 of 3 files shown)


klauer:
👍

danielballan:
Why is codecov putting huge comments in PRs now?

klauer:
Their pretty, unlabeled graph should definitely be larger. (cc @tacaswell for additional sarcasm detection algorithm input)

bluesky/plans.py
```diff
@@ -221,6 +221,431 @@ def single_gen(msg):
...
+    ------
+    msg : Msg
+        Msg('read', obj)
+    """
+    yield Msg('read', obj)
+
+
+def monitor(obj, *args, name=None, **kwargs):
+    """
+    Asynchornously monitor for new values and emit Event documents.
```
klauer: asynch _ro_ nously

bluesky/plans.py
```diff
@@ -221,6 +221,431 @@ def single_gen(msg):
...
+    ------
+    msg : Msg
+        Msg('monitor', obj, *args, **kwargs)
+    """
+    yield Msg('monitor', obj, *args, **kwargs)
+
+
+def unmonitor(obj):
+    """
+    Stop mointoring.
```
klauer: mo _ni_ toring

bluesky/plans.py
```diff
@@ -221,6 +221,431 @@ def single_gen(msg):
...
+    args :
+        passed to obj.set()
+    kwargs :
+        passed to obj.set()
+
+    Yields
+    ------
+    msg : Msg
+    """
+    yield Msg('set', obj, *args, group=group, **kwargs)
```
klauer: What if `group=None` and `wait=True` - should it be auto-generated?

danielballan: That's a question for a separate PR, because currently the RunEngine auto-generates. (`None` is a valid group.)

danielballan: Background: we accept 'None' as a group so that the simple case is easy:

```
RE(chain([abs_set(motor1, 0), abs_set(motor2, 0), wait()]))

File: doc/source/plans.rst [hunk 4]
```diff
@@ -138,43 +181,147 @@ Here are a couple more useful recipes:
...
+------------------------------
+
+These provide a different way of using the standard plans. The plan becomes
+a reusable object, whose parameters can be adjusted interactively between
+uses.
+
+.. ipython:: python
+
+    from bluesky.plans import Scan
+    from bluesky.examples import motor, det3
```
tacaswell: this should be just `det`

PR #419, opened 2016-04-25
ENH: Support optional 'pause' and 'resume' methods on Devices.

### LINKED ISSUE #415: Proposal: optional pause and resume methods on Devices
On pausing and resuming, the RunEngine could call `device.pause()` and `device.resume()` on all objects that it knows about. This would give objects like flyers or complicated detectors a way to clean themselves up if needed.

I am proposing that this be an optional part of the bluesky interface, not a required method.

Thoughts?

danielballan:
Closes #415

I'm not doing anything with this, but now we have the hook when we need it / want to play with it. I don't think the Devices need to distinguish between 'pause' and 'suspend', so both actions call `device.pause()`.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -927,6 +927,10 @@ class RunEngine:
                                          "a 'checkpoint' message after the "
                                          "'close_run' message.")
         self.log.debug("Stopping run %s", self._run_start_uid)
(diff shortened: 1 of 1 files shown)


danielballan:
Tests should pass once #416 is merged.

PR #423, opened 2016-04-26
CLN: Consolidate examples.

danielballan:


tacaswell:
As we instantiate a couple of the syn2D detectors at the bottom of examples, all of the local imports of filestore are for naught.

ericdill:
Should this also coincide with a removal of broker_examples.py?

PR #428, opened 2016-04-27
FIX: Stage Plan.flyers.

danielballan:
Bug: In this scenario, the `flyer` is never staged.

```python
c = Count[])
c.flyers = [flyer]
```

Perhaps no one has encountered this bug yet because no one had written a flyer that *needed* to be staged until today. As a reminder, this will not affect any existing flyers that don't implement stage/unstage. The RunEngine checks to see if `msg.obj` has a stage/unstage method before attempting to call it.

The second commit may be fixing a nonexistent problem, but it feels like the right thing to do. Preventative medicine.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1948,10 +1948,11 @@ class Plan(Struct):
             setattr(self, key, val)
         try:
             plan_stack = deque()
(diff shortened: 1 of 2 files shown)


klauer:
Looks good to me

PR #429, opened 2016-04-27
ENH: Add option to emit events from RunEngine._collect in streaming fashion.

danielballan:
See commit msgs. This might never get used, but it seemed like a useful option to have.


tacaswell:
definitely want to keep the middle commit.

@klauer ?

klauer:
👍

PR #430, opened 2016-05-02
Fix mutator exception handling

tacaswell:


tacaswell:
I have done interactive testing and this seems to work as expected.  I want to push of adding through tests of this to next week though.

danielballan:
Works for me. I think we all have a lot of test-writing to do after Friday's feature-freeze.

klauer:
👍

PR #431, opened 2016-05-02
Add a new 'input' Msg that takes raw_input while running event loop in background

danielballan:
Testing raw_input in py.test is a can of worms I don't want to open this week, but this code works:

```python
from bluesky.examples import motor, det
from bluesky.plans import tweak
RE = RunEngine({})
def plan():
    p = yield Msg('input', prompt='where now?')
    print('moving to', p)
    yield Msg('set', motor, float(p))
RE(plan())
RE(tweak(det, 'det', motor, 1))
```

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1881,6 +1881,7 @@ def relative_inner_product_scan(detectors, num, *args, per_step=None, md=None):
     return [plan]
 
 
+@planify
 def tweak(detector, target_field, motor, step, *, md=None):
     """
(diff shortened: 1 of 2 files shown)


danielballan:
@klauer I think I'd rather put my energy into doing this the thorough way (with pexpect) in my inevitable test-writing binge later this cycle.

klauer:
@danielballan works for me.

tacaswell:
attn @chiahaoliu This is what you should use to get user input.

PR #432, opened 2016-05-03
MNT: Make 'stop' an optional piece of the API.

danielballan:
To discuss before merging.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -438,11 +438,7 @@ class RunEngine:
         for obj, (cb, kwargs) in list(self._monitor_params.items()):
             obj.clear_sub(cb)
         # During pause, all motors should be stopped.
-        for obj in self._movable_objs_touched:
-            try:
-                obj.stop()
-            except Exception:
-                logger.error("Failed to stop %r", obj)
+        self._stop_movable_objects()
         # Notify Devices of the pause in case they want to clean up.
         for obj in self._objs_seen:
             if hasattr(obj, 'pause'):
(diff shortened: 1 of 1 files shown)


bluesky/run_engine.py
```diff
@@ -1730,3 +1724,15 @@ def _default_md_validator(md):
             "GOOD: sample='dirt' "
             "GOOD: sample={'color': 'red', 'number': 5} "
             "BAD: sample=[1, 2] ")
+
+def _try_stopping(obj):
```
tacaswell: can we call this `_stop_me_maybe`?

tacaswell:
This seems reasonable :+1:

bluesky/run_engine.py
klauer: Oh, @tacaswell...

bluesky/run_engine.py
```diff
@@ -1730,3 +1724,15 @@ def _default_md_validator(md):
...
+    "Call obj.stop() and log any exceptions."
+    try:
+        stop = obj.stop
+    except AttributeError:
+        logger.debug("No 'stop' method available on %r", obj)
+    else:
+        try:
+            stop()
+        except Exception:
+            logger.error("Failed to stop %r", obj)
```
klauer: How do we feel about the lack of `exc_info` here?

tacaswell: it's no worse that it was before.

bluesky/run_engine.py
```diff
@@ -439,10 +439,7 @@ def request_pause(self, defer=False, name=None, callback=None):
             obj.clear_sub(cb)
         # During pause, all motors should be stopped.
         for obj in self._movable_objs_touched:
-            try:
-                obj.stop()
-            except Exception:
-                logger.error("Failed to stop %r", obj)
+            _try_stopping(obj)
```
tacaswell: excetpions in `stop` here probably should kill the scan

bluesky/run_engine.py
klauer: That doesn't really answer the question though - shouldn't we know why it failed to stop?

danielballan:
Force-pushed changes:

* Made this a private method so it can use the RunEngine's `log` attribute.
* Included exception in log message, per @klauer's comment.

PR #434, opened 2016-05-03
ENH: Add msg_hook

danielballan:


tacaswell:
:+1:

This also gives a path to do things like 

```python

def hook(msg):
    if msg.obj is not None and msg.obj.name == 'forbidden':
        raise NO()
```

Maybe too cute

danielballan:
@klauer and @tacaswell are here, self-merging

danielballan:
nevermind; @tacaswell is a purist :-P

PR #435, opened 2016-05-03
FIX: spec_api.mesh correct subs

klauer:


danielballan:
Ha. I hope you didn't lose too much time tracking this one down.

PR #436, opened 2016-05-04
API/DOC: Don't panic: Remove the 'panic' concept from the RunEngine.

danielballan:
- This feature was, to the best of my knowledge, never used in the
  wild.
- The idea was to make it possible to "lock" the RunEngine until
  the user called RE.all_is_well().
- But as of the (soon-to-be-merged) suspenders re-write, there is
  a more flexible way to do this using suspenders, rendering this code
  just dead weight and more API to learn.


tacaswell:
:+1:

klauer:
👍 from me too.

tacaswell:
Merging with the promise that this will be documented.

PR #437, opened 2016-05-04
MNT: Make log messages more useful.

danielballan:
* Remove the two DEBUG-level log messages per Msg processed. The flood this produced is not often useful, and it can always be re-created using `RE.msg_hook = lambda msg: self.log.debug(repr(msg))`
* Make log entries about emitted documents more detailed, including uid and data keys where applicable.


klauer:
Looks good to me - much prefer these messages.

PR #438, opened 2016-05-04
ENH/TST: add spiral and fermat spiral scans

klauer:
This was a lot simpler when it was hidden in hxntools...

### DIFF
--- a/bluesky/__init__.py
+++ b/bluesky/__init__.py
@@ -24,10 +24,6 @@ from .plans import InnerProductDeltaScanPlan
 from .plans import OuterProductAbsScanPlan
 from .plans import OuterProductDeltaScanPlan
 from .plans import Tweak
-from .plans import SpiralScan
-from .plans import SpiralFermatScan
-from .plans import RelativeSpiralScan
-from .plans import RelativeSpiralFermatScan
 
 from ._version import get_versions
 __version__ = get_versions()['version']
--- a/bluesky/__init__.py
+++ b/bluesky/__init__.py
@@ -24,6 +24,10 @@ from .plans import InnerProductDeltaScanPlan
 from .plans import OuterProductAbsScanPlan
 from .plans import OuterProductDeltaScanPlan
 from .plans import Tweak
+from .plans import SpiralScan
+from .plans import SpiralFermatScan
+from .plans import RelativeSpiralScan
+from .plans import RelativeSpiralFermatScan
 
 from ._version import get_versions
 __version__ = get_versions()['version']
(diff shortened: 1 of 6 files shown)


tacaswell:
Why was it simpler hidden in hxntools?

bluesky/spec_api.py
```diff
@@ -357,3 +355,162 @@ def tw(motor, step, time=None, *, md=None):
...
+        any 'setable' object (motor, temp controller, etc.)
+    y_motor : object
+        any 'setable' object (motor, temp controller, etc.)
+    x_range : float
+        X range, in engineering units
+    y_range : float
+        Y range, in engineering units
+    dr : float
+        Delta radius, in engineering units
+    nth : float
```
tacaswell: nth is not passed through?

klauer:
@tacaswell no class wrappers, no 'spec api'/gs-aware separation, no tests... and no PR means no bikeshedding 😄

tacaswell:
We could just drop the spec api / gs....

bluesky/__init__.py
```diff
@@ -24,6 +24,10 @@
 from .plans import OuterProductAbsScanPlan
 from .plans import OuterProductDeltaScanPlan
 from .plans import Tweak
+from .plans import SpiralScan
```
danielballan: The names imported here are the *old* names, imported for back-compat. (One of the PRs I'm working on adds a comment to this effect.) I don't think we need to import the new names at top-level, but if you feel differently I have no problem with it.

bluesky/plans.py
```diff
@@ -1950,6 +1950,204 @@ def core():
...
+    per_step : callable, optional
+        hook for cutomizing action of inner loop (messages per step)
+        See docstring of bluesky.plans.one_nd_step (the default) for
+        details.
+    md : dict, optional
+        metadata
+
+    See Also
+    --------
+    `bluesky.plans.spiral`
```
danielballan: Nice. I see that "See Also" is about to sweep through our codebase.

bluesky/__init__.py
klauer: Removed

bluesky/spec_api.py
klauer: I think you meant to leave this on the next function down - it was named factor, should have been nth. (correct me if I'm wrong)

klauer:
@tacaswell up to you as far as dropping the gs-aware one.

tacaswell:
I was being sarcastic about dropping gs

tacaswell:
attn @gjwillms @imrealkaren

PR #439, opened 2016-05-04
FIX: an pre-existing bug in deferred pause exposed by our new scheme

danielballan:


danielballan:
See commit msgs. Notice also that deferred pause was not doing any of the nice cleanup that a hard pause does. (Oof!) Now they share the same code path.

PR #440, opened 2016-05-05
Fix and fortify

danielballan:
This mostly adds test coverage to run_engine.py, but some minor enhancements and API changes are included. See commit msgs.


tacaswell:
Do empty events increment the sequence number (despite not being emitted)?

danielballan:
No. Remember there is one sequence number counter per descriptor, and the event with an empty set of data keys has no descriptor.

tacaswell:
ah, very good.  That makes perfect sense.

bluesky/tests/test_devices.py
```diff
@@ -62,17 +62,47 @@ def test_separate_devices():
...
+    docs = []
+    def collect(name, doc):
+        docs.append(doc)
+
+    a = A('')
+    def plan():
+        yield Msg('open_run')
+        yield Msg('monitor', a.s1)
+        yield Msg('checkpoint')
+        a.s1._run_subs(sub_type='value')
```
tacaswell: Shouldn't this process synchronously?  I think this should work with only `null` message yielded instead of a sleep?

bluesky/tests/test_devices.py
```diff
@@ -62,17 +62,47 @@ def test_separate_devices():
...
+        yield Msg('sleep', None, 0.2)  # give time for monitor to process
+        yield Msg('pause')
+        a.s1._run_subs(sub_type='value')
+        yield Msg('close_run')
+    fresh_RE(plan(), collect)
+    assert len(docs) == 3  # RunStart, EventDescriptor, one Event
+    # All but one of these will be ignored. Why is one not ignored, you ask?
+    # Beacuse ophyd runs subscriptions when they are (re-)subscriped.
```
tacaswell: there is a kwargs to subscribe that will disable that, but I think in this case this is the behavior we want.

bluesky/tests/test_devices.py
danielballan: Does it? Yes, 'null' works -- this was just caution because I was not sure while writing this that _run_subs ran synchronously. Will remove.

bluesky/run_engine.py
```diff
@@ -1343,13 +1343,16 @@ def _collect(self, msg):
         self._uncollected.remove(obj)
         stream_name = self._flyer_stream_names.pop(obj)
 
-        # TODO Since Flyer.describe() return a *list* of data_keys it should
-        # probably have a different method name (describe_all?) *and* have some
-        # way of injecting a 'name' like the create msg does. For now, the name
-        # is hard-coded to None.
-        data_keys_list = obj.describe()
+        description = obj.describe_collect()  # a dict or a list
+        if hasattr(description, 'items'):
```
danielballan: Below, we actually *call* `description.items()` so there is zero harm in performing this check.

bluesky/run_engine.py
```diff
@@ -1343,13 +1343,16 @@ def _collect(self, msg):
...
+            # e.g., {name_for_desc1: data_keys_for_desc1,
+            #        name for_desc2: data_keys_for_desc2, ...}
+            named_data_keys = description
+        else:
+            # e.g., [data_keys_for_desc1, data_keys_for_desc2, ...]
+            named_data_keys = {None: dk for dk in description}
```
tacaswell: this will only keep the last one.  I would just drop this altogether.  We are breaking the `describe` api might as well clean it up too.

danielballan:
@tacaswell This is ready.

bluesky/plans.py
```diff
@@ -859,6 +859,32 @@ def event_context(plan_stack, name='primary'):
...
+    Parameters
+    ----------
+    flyers : collection
+        objects that support the flyer interface
+
+    Yields
+    ------
+    msg : Msg
+        'kickoff', 'wait', 'complete, 'wait', 'collect' messages
+    md : dict, optional
```
tacaswell: doc needs to be moved.

tacaswell: I will fix in follow-up PR

PR #442, opened 2016-05-09
FIX: fly_during block group + misc

klauer:
Maybe more coming...


klauer:
Class-based plan_name/plan_type are useless for the class-based `Plan`s:
```
 'plan_name': '__call__',
 'plan_type': 'generator',
```

PR #444, opened 2016-05-09
TST: Check plan_name and plan_type.

danielballan:
Do not merge. This failing test reproduces a bug noticed by @klauer. `Plan.__iter__` has a nice `__name__` but `Plan.__call__` does not. The solution is not obvious to me. Punting for now.


tacaswell:
meta-class?

tacaswell:
I think that this was covered to a large degree in #457

PR #447, opened 2016-05-10
API/FIX: clear msg cache on un-cacheable msg

tacaswell:
This fixes a subtle bug where there are some messages which are
not (always) safe to replace due to being part of a context pair.

This needs testing / another set of eyes.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -2492,7 +2492,6 @@ class Plan(Struct):
                 with subs_context(plan_stack, subs):
                     plan = self._gen()
                     plan_stack.append(fly_during(plan, flyers))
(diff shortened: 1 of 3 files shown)


danielballan:
Just spelling out my thoughts from the train for the record:

* I think it's important to get this merged for the upcoming cycle (and possibly for the CSX deployment...) because it is easy to get into a state where ^C kills instead of pauses.

* I liked the idea of being able to roll back through multiple runs, which this PR makes impossible, but I grant that that abuses the limiting expressivity of the checkpoint mechanism. I hope we can get a working example of how a *plan* to orchestrate a rollback like that soon.

danielballan:
Tested in the wild at the CSX deployment today. Merging.

PR #449, opened 2016-05-11
ENH: Improve plan md.

danielballan:
- Use ChainMap to give md from kwargs precedence over hard-coded md.
- Add 'plan_name' to everything.
- Add a couple 'plan_args' accidentally left out in the first pass.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -5,7 +5,7 @@ import itertools
 import functools
 import operator
 from contextlib import contextmanager
-from collections import OrderedDict, Iterable, defaultdict, deque
+from collections import OrderedDict, Iterable, defaultdict, deque, ChainMap
 
 import numpy as np
 from cycler import cycler
@@ -1503,8 +1503,11 @@ def count(detectors, num=1, delay=None, *, md=None):
     """
     if md is None:
         md = {}
-    md.update({'detectors': [det.name for det in detectors]})
-    md['plan_args'] = {'detectors': list(map(repr, detectors)), 'num': num}
+    md = ChainMap(
+        md, 
+        {'detectors': [det.name for det in detectors],
+         'plan_args':{'detectors': list(map(repr, detectors)), 'num': num},
+         'plan_name': 'count'})
 
     if num is None:
         counter = itertools.count()  # run forever, until interrupted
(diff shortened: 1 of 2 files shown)


bluesky/plans.py
```diff
@@ -1790,6 +1809,9 @@ def relative_log_scan(detectors, motor, start, stop, num, *, per_step=None,
     `bluesky.plans.log_scan`
     """
     # TODO read initial positions (redundantly) so they can be put in md here
+    if md is None:
+        md = {}
+    md = ChainMap(md, {'plan_name': 'relative_log_scan'})
```
tacaswell: This is not going to propagate correctly

bluesky/plans.py
```diff
@@ -1833,6 +1855,22 @@ def adaptive_scan(detectors, target_field, motor, start, stop,
...
+         'motors': [motor.name],
+         'plan_args':{'detectors': list(map(repr, detectors)),
+                      'motor': repr(motor),
+                      'start': start,
+                      'stop': stop,
+                      'min_step': min_step,
+                      'max_step': max_step,
+                      'target_delta': target_delta,
+                      'backstep': backstep,
+                      'threshhold': threshold},
```
klauer: typo

bluesky/plans.py
```diff
@@ -1985,10 +2026,14 @@ def scan_nd(detectors, cycler, *, per_step=None, md=None):
     """
     if md is None:
         md = {}
-    md.update({'detectors': [det.name for det in detectors],
-               'motors': [motor.name for motor in cycler.keys]})
-    md['plan_args'] = {'detectors': list(map(repr, detectors)),
-                       'cycler': repr(cycler)}
+    md = ChainMap(
+        md,
+        {'detectors': [det.name for det in detectors],
```
klauer: Elsewhere `repr()` is used on detectors, why `name` here? (or why `repr` elsewhere...)

bluesky/plans.py
```diff
@@ -1624,6 +1630,9 @@ def relative_list_scan(detectors, motor, steps, *, per_step=None, md=None):
     `bluesky.plans.list_scan`
     """
     # TODO read initial positions (redundantly) so they can be put in md here
+    if md is None:
+        md = {}
+    md = ChaimMap(md, {'plan_name': 'relative_list_scan'})
```
klauer: Chaim

bluesky/plans.py
klauer: I blame it on lack of sleep. I get it now: top-level metadata gets the name, plan_args gets the repr

bluesky/plans.py
danielballan: It is; in an offline-conversion, @tacaswell established that the precedence order is the opposite of what he was thinking here.

PR #450, opened 2016-05-11
Fix sanitize np

danielballan:
It turns out that `np.ndarray` is not a subclass of `np.generic`. We need the latter to catch scalars and the former, obviously, to catch arrays.

While I'm here, I moved some stuff to utils.py.


danielballan:
Oh, also, we weren't sanitizing the values from `obj.read_configuration()`. This does that well.

This change is source-installed at CSX. It sounds like ISS could use it as well.

klauer:
Looks good to me.

PR #452, opened 2016-05-12
ENH: print out warning when suspenders prevent RE

tacaswell:


bluesky/run_engine.py
```diff
@@ -508,7 +508,13 @@ def __call__(self, plan, subs=None, *, raise_if_interrupted=False,
         if not self.state.is_idle:
             raise RuntimeError("The RunEngine is in a %s state" % self.state)
 
-        futs = [f for sup in self.suspenders for f in sup.get_futures()]
+        futs = []
+        for sup in self.suspenders:
+            f_lst, msg = sup.get_futures()
+            if f_lst:
+                futs.extend(f_lst)
+                print(msg)
```
danielballan: @klauer @tacaswell Is it intention that this message does not print when the suspender trips mid-scan -- only when it trips at the start? Also, should the message indicate that the plan will still run if/when the suspender becomes un-tripped?

klauer: Good catch - we would like to see the message when it trips mid-scan. @tacaswell just didn't get around to implementing that when this PR went up, if I recall correctly.

tacaswell: Correct.  We already get _any_ message when that trips.  The we were getting _nothing_ (just an apparent hang) when starting the scan.

PR #453, opened 2016-05-12
FIX: sanitize variable arguments in outer product scans

klauer:


klauer:
Force pushed fix for spiral scans as well

PR #454, opened 2016-05-13
MNT: add num_steps to metadata

klauer:
(for discussion, do not merge)

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1505,7 +1505,7 @@ def count(detectors, num=1, delay=None, *, md=None):
     if md is None:
         md = {}
     md = ChainMap(
-        md, 
+        md,
         {'detectors': [det.name for det in detectors],
          'plan_args':{'detectors': list(map(repr, detectors)), 'num': num},
          'plan_name': 'count'})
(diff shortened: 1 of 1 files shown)


bluesky/plans.py
```diff
@@ -2352,10 +2376,11 @@ def spiral_fermat(detectors, x_motor, y_motor, x_start, y_start, x_range,
...
     md = ChainMap(
         md,
-        {'detectors': [detector.name for detector in detectors],
-         'motors': [motor.name for motor in [x_motor, y_motor]],
+        {# 'detectors': inserted by scan_nd
+         # 'motors': inserted by scan_nd
+         # 'num_steps': inserted by scan_nd
          'plan_args': {'detectors': list(map(repr, detectors)),
-                       'x_motor': x_motor, 'y_motor': y_motor,
+                       'x_motor': repr(x_motor), 'y_motor': repr(y_motor),
```
danielballan: :+1: The existing implementation was a mistake; this is more consistent and more useful.

bluesky/plans.py
```diff
@@ -2260,6 +2283,7 @@ def tweak(detector, target_field, motor, step, *, md=None):
         md,
         {'detectors': [detector.name],
          'motors': [motor.name],
+         # 'num_steps': 'adaptive',
```
danielballan: I think it's better to leave this out (as your hedged `#` seems to suggest). IMO, it's easier to test `if 'num_steps' in header.start` than `if isinstance(header.start.get('num_steps'), six.string_types)`.

bluesky/plans.py
```diff
@@ -2078,10 +2086,17 @@ def inner_product_scan(detectors, num, *args, per_step=None, md=None):
     """
     if md is None:
         md = {}
+
+    # mds can't handle ophyd devices, so all replace them with reprs where
+    # necessary:
+    mds_args = [repr(arg) if hasattr(arg, 'name') else arg
```
danielballan: :bike: :house: I would prefer not inferring this from `.name` and instead building `mds_args` as part of the loop over `chunked(args, 3)` below, where we know for sure which pieces are which.

danielballan:
Annnnd I'm back!

I think `num_steps` is a good addition to the top-level metadata. See line comments.

klauer:
@danielballan welcome back! (*phew* :relieved: )

Great. I'll fix up a few of those things on Monday.

bluesky/plans.py
klauer: Agreed, this was a "get PR up for discussion" set of lines. Will fix.

danielballan:
Superceded by #467

PR #457, opened 2016-05-17
FIX: correct all spec names

tacaswell:
For all of the plans is spec_api make the `plan_name` value in the
run_start metadata match the plan name.

attn @ericdill

### DIFF
--- a/bluesky/spec_api.py
+++ b/bluesky/spec_api.py
@@ -14,14 +14,13 @@ http://www.certif.com/downloads/css_docs/spec_man.pdf
 """
 from collections import deque
 import matplotlib.pyplot as plt
-from bluesky import plans, Msg
+from bluesky import plans
 from bluesky.callbacks import LiveTable, LivePlot, LiveRaster
 from bluesky.callbacks.scientific import PeakStats
 from boltons.iterutils import chunked
 from bluesky.global_state import gs
-from bluesky.utils import (normalize_subs_input, Subs, DefaultSubs,
-                           first_key_heuristic, apply_sub_factories,
-                           update_sub_lists)
+from bluesky.utils import (first_key_heuristic)
+
 from bluesky.plans import (subs_context, count, scan,
                            relative_scan, relative_inner_product_scan,
                            outer_product_scan, inner_product_scan,
--- a/bluesky/spec_api.py
+++ b/bluesky/spec_api.py
@@ -29,6 +29,7 @@ from bluesky.plans import (subs_context, count, scan,
                            baseline_mutator)
 import itertools
 from itertools import chain
+from collections import ChainMap
 
 ### Factory functions for generating callbacks
 
@@ -98,6 +99,9 @@ def ct(num=1, delay=None, time=None, *, md=None):
     md : dict, optional
         metadata
     """
+    if md is None:
+        md = {}
+    md = ChainMap(md, {'plan_name': 'ct'})
     subs = {'all': [LiveTable(gs.TABLE_COLS + [gs.PLOT_Y])]}
     if num is not None and num > 1:
         subs['all'].append(setup_plot([]))
@@ -132,6 +136,9 @@ def ascan(motor, start, finish, intervals, time=None, *, md=None):
     md : dict, optional
         metadata
     """
+    if md is None:
+        md = {}
+    md = ChainMap(md, {'plan_name': 'ascan'})
     subs = {'all': [LiveTable([motor] + gs.TABLE_COLS + [gs.PLOT_Y]),
                     setup_plot([motor]),
                     setup_peakstats([motor])]}
@@ -168,7 +175,9 @@ def dscan(motor, start, finish, intervals, time=None, *, md=None):
     subs = {'all': [LiveTable([motor] + gs.TABLE_COLS + [gs.PLOT_Y]),
                     setup_plot([motor]),
                     setup_peakstats([motor])]}
-
+    if md is None:
+        md = {}
+    md = ChainMap(md, {'plan_name': 'dscan'})
     plan_stack = deque()
     with subs_context(plan_stack, subs):
         plan = relative_scan(gs.DETS, motor, start, finish, 1 + intervals,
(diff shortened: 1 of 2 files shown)


ericdill:
LGTM.  The only three that I care about are dscan, ascan and ct.  All the others are too complicated to represent in a spec format, I think.

ericdill:
If you agree that the others are too complicated, then just let me know here and I'll merge this then update my spec PR

tacaswell:
I intend to add tests but am happy to add those in a new PR if you want to merge this to get the suitcase PR testing properly.

ericdill:
Oh, well I'll let you add tests.  I'll update the suitcase PR with these names though.  Thanks!

tacaswell:
@edill this is ready for review.

ericdill:
Will do first thing tomorrow

bluesky/spec_api.py
```diff
@@ -132,6 +135,9 @@ def ascan(motor, start, finish, intervals, time=None, *, md=None):
     md : dict, optional
         metadata
     """
+    if md is None:
+        md = {}
+    md = ChainMap(md, {'plan_name': 'ascan'})
```
ericdill: Why use ChainMap instead of 
```
md.update({'plan_name': 'ascan'})
```

I am positive that I just missed this discussion. Sorry for making you repeat it

bluesky/spec_api.py
```diff
@@ -168,7 +174,10 @@ def dscan(motor, start, finish, intervals, time=None, *, md=None):
     subs = {'all': [LiveTable([motor] + gs.TABLE_COLS + [gs.PLOT_Y]),
                     setup_plot([motor]),
                     setup_peakstats([motor])]}
-
+    if md is None:
+        md = {}
+    md = ChainMap(md, {'plan_name': 'dscan'})
+    print(list(md.items()))
```
ericdill: Is this debug info?

bluesky/spec_api.py
```diff
@@ -267,8 +282,14 @@ def a2scan(*args, time=None, md=None):
         plan_stack.append(plan)
     return plan_stack
 
+
 # This implementation works for *all* dimensions, but we follow SPEC naming.
-a3scan = a2scan
+def a3scan(*args, time=None, md=None):
```
ericdill: Why is this one not planified?

ericdill: Oh, probably because it is yielding from a function that *is* planified

bluesky/tests/test_simple_api.py
```diff
@@ -1,56 +1,70 @@
...
+    (d2scan, 'd2scan', (motor, 1, 2, 2), {}),
+    (d3scan, 'd3scan', (motor, 1, 2, 2), {}),
+    (mesh, 'mesh', (motor1, 1, 2, 2, motor2, 1, 2, 3), {}),
+    (th2th, 'th2th', (1, 2, 2), {}),
+    (aspiral, 'aspiral', (motor1, motor2, 0.0, 0.0, 0.1, 0.1, 0.05, 1.0), {}),
+    (spiral, 'spiral', (motor1, motor2, 0.1, 0.1, 0.05, 1.0), {}),
+    (afermat, 'afermat', (motor1, motor2, 0.0, 0.0, 0.1, 0.1, 0.05, 1.0), {}),
+    (fermat, 'fermat', (motor1, motor2, 0.1, 0.1, 0.05, 1.0), {}),
+    # with count time specified as keyword arg
+    (ct, 'ct', (), {}),
```
ericdill: This is a duplicate of line 10

bluesky/tests/test_simple_api.py
```diff
@@ -1,56 +1,70 @@
...
-    RE(mesh(motor1, 1, 2, 2, motor2, 1, 2, 3, time=0.1))
-    RE(th2th(1, 2, 2, time=0.1))
-    RE(aspiral(motor1, motor2, 0.0, 0.0, 0.1, 0.1, 0.05, 1.0, time=0.1))
-    RE(spiral(motor1, motor2, 0.1, 0.1, 0.05, 1.0, time=0.1))
-    RE(afermat(motor1, motor2, 0.0, 0.0, 0.1, 0.1, 0.05, 1.0, time=0.1))
-    RE(fermat(motor1, motor2, 0.1, 0.1, 0.05, 1.0, time=0.1))
+    run_start = None
+
+    def capture_run_start(name, doc):
+        nonlocal run_start
```
ericdill: neat usage of nonlocal

ericdill:
I'd consider replacing all these code snippets:

```python
def some_scan(...):
    if md is None:
        md = {}
    md = ChainMap(md, {'plan_name': PLAN_NAME})
```

with:

or 
```python
md = ChainMap(md or {}, {'plan_name': PLAN_NAME})
```
Probably worth pointing out that these are not *strictly* equivalent since `md or {}` will evaluate to `{}` for things that are falsey (`''`, `False`, `[]`, `set()`, etc...) in addition to `None`, but they're close enough IMO

ericdill:
Two minor nits, but LGTM other than that

bluesky/spec_api.py
tacaswell: md has precedence over the `{}` added here so passed in values override default values so we would want to use `setdefault` instead of `update`.

It is also rude to mutate input.  Could have done this as

```python
local_md = {'plan_name': 'ascan'}
if md is not None:
    local_md.update(md)
md = local_md
```

but `ChainMap` is a bit more readable.

ericdill: Great. Thanks for the clear explanation

tacaswell:
fixed and forcepushed.

tacaswell:
the in-line `or` is compact, but would allow false-non-mapping types to be valid input for `md` which is probably not something we want to do.

bluesky/spec_api.py
tacaswell: planify is a work-around to make context managers work.  That they do is make a list of plans that should be run in sequence, `planify` just call `pchain` in the return value of the wrapped function.

ericdill:
> but would allow false-non-mapping types to be valid input for md which is probably not something we want to do.

Oh yeah that's a fair point.

EDill:
great job, team!

ericdill:
> great job, team!

Thanks!

Sorry that we seem to ping you with some regularity D-:

EDill:
No worries--it's been a really long time!

klauer:
I'm a bit late, but 👍 from me. And edill, want to come work with us?

EDill:
Well, I have always wanted to work for ~*Science*~...

PR #458, opened 2016-05-18
ENH: pass mappings into logbook_cb_factory

tacaswell:
This allows customization based on the 'plan_name' field of the
RunStart.

attn @NSLS-II-CHX


danielballan:
Add a smoke test?

klauer:
:+1: Somewhat related while I'm thinking about it: I'd like to see some way to keep a logbook entry open for appending during a run. @tacaswell and I wanted to see if we could get support from Kunal or someone else (?) to do so.

tacaswell:
Getting any test coverage on this is on my todo list for the weekend.

danielballan:
I added tests. The actual enhancement code is Tom's, so I will merge.

PR #459, opened 2016-05-19
ENH: add time to md for spec api

klauer:


klauer:
Added test from @tacaswell. Will make a decision later as to allowing key customization or fixing it throughout (may be preferable).

danielballan:
Hahahah. `gs.MD_TIME_KEY` seems to be code for, "I do not want to participate in a facility-wide debate about what we are going to call this."

klauer:
@danielballan Am I so transparent? :blush:

PR #461, opened 2016-05-21
API/ENH: More verbose suspenders

danielballan:


bluesky/suspenders.py
```diff
@@ -162,13 +163,10 @@ def get_futures(self):
         -------
         futs : list
             List of futures to wait on
-
-        msg : str
-            String explaining why the suspender is tripped
         '''
         if not self.tripped:
-            return [], ''
-        return [self.__make_event().wait()], self.get_justification()
```
danielballan: Here the API is reverted to what it was. I gather that this was a quick fix and not an especially intentional API choice. But feel free to argue why we shouldn't just call `get_justification()` separately, @tacaswell.

tacaswell:
:-1:  The hacked API here is `get_justification` which the RE does not know about on master.

Getting the futures and the message needs to be atomic (as the reasons the suspender might report it is suspending may be different by the time we call `get_justification()`. )

bluesky/run_engine.py
```diff
@@ -640,7 +643,8 @@ def remove_suspender(self, suspender):
             suspender.remove()
         self._suspenders.discard(suspender)
 
-    def request_suspend(self, fut, *, pre_plan=None, post_plan=None):
+    def request_suspend(self, fut, *, pre_plan=None, post_plan=None,
+                        justification=None):
```
tacaswell: :+1: on this API

bluesky/run_engine.py
```diff
@@ -653,15 +657,20 @@ def request_suspend(self, fut, *, pre_plan=None, post_plan=None):
...
+        justification : str, optional
+            explanation of why the suspension has been requested
         """
         if not self.resumable:
             print("No checkpoint; cannot suspend. Aborting...")
             self._exception = FailedPause()
         else:
             print("Suspending....To get prompt hit Ctrl-C twice to pause.")
+            if justification is None:
+                print("No justification for this suspension was given.")
```
tacaswell: if it is optional do not include passive aggressive commentary :wink:

danielballan:
> Getting the futures and the message needs to be atomic

Got it. That makes sense. I reverted by changed to `get_futures` and force-pushed.

PR #462, opened 2016-05-21
MNT: Ensure per-run state is always cleared.

danielballan:
Edge case: When the RunEngine cleans up from a bad exit, it should clear the per-run caches.


bluesky/run_engine.py
```diff
@@ -843,6 +844,7 @@ def _run(self):
                                    self._run_start_uid, exc)
                     # Exceptions from the callbacks should be re-raised.
                     # Close the loop first.
+                    self._clear_run_cache()
```
tacaswell: We did not do this initially to leave some state for debugging.

PR #463, opened 2016-05-21
ENH: Optionally, record interruptions in a special event stream.

danielballan:
An old branch resurrected for discussion. Needs tests.

attn @klauer


bluesky/run_engine.py
```diff
@@ -977,6 +1005,17 @@ def _open_run(self, msg):
         self.log.debug("Emitted RunStart (uid=%r)", doc['uid'])
         yield from self._reset_checkpoint_state()
 
+        # Emit an Event Descriptor for recording any interruptions as Events.
+        if self.record_interruptions:
+            self._interruptions_desc_uid = new_uid()
+            dk = {'dtype': 'boolean', 'shape': None, 'source': 'RunEngine'}
```
klauer: dtype incorrect here

bluesky/run_engine.py
```diff
@@ -1528,7 +1567,9 @@ def _pause(self, msg):
         See RunEngine.request_pause() docstring for explanation of the three
         keyword arugments in the `Msg` signature
         """
+        print("about to call request_pause")
```
klauer: Debug prints here and below

klauer:
Looks good

PR #464, opened 2016-05-22
DOC: Document `baseline` and bike-shed the name (with back-compat)

danielballan:
- Other preprocessors (fly_during, relative_set, reset_positions, etc.) do
  not have _mutator in the name -- only the mutator utils do.


klauer:
Too soon for back-compatibility. Just change it.

danielballan:
Works for me. Done and force-pushed.

PR #465, opened 2016-05-22
MNT: Remove too-cute repr for examples objs.

danielballan:
This repr dates back to the original implementation demo, where the RunEngine
printed every message is saw for the purposes of explaining its operation.

The raw repr looks like more like the ones from ophyd and makes a less
confusing example nowadays.

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -73,9 +73,6 @@ class Base:
                     'precision': self.precision}
                 for k in self._fields}
 
-    def __repr__(self):
-        return '{}: {}'.format(self._klass, self.name)
-
     def read_configuration(self):
         return {'some_configuration':
                    {'value': 'there is a sandwich in the vacuum chamber',


danielballan:
In case this seems neurotic and unimportant, it came up while writing another batch of examples for NSLS-II/docs. Does that help? :sheep:

klauer:
Not neurotic. I ran into it before with the "too-cute ':' " not quite working with `eval()` (*cough*).

danielballan:
Believe it or not, IIRC this particular repr was a rare effort from @tacaswell to make things *human*-readable. I hate to discourage the impulse, you know? But, yes, in the end it seems it was not the right choice.

tacaswell:
Lesson learned: reading is only for computers.

Probably should replace this with a round-trippable repr

klauer:
`__str__` != `__repr__`

To further waste our time on this: I'd recommend the original `__repr__` becomes `__str__` for readability, and make a new round-trippable `__repr__` as @tacaswell mentions.

danielballan:
Discussed offline -- we will take a different approach, but not soon.

PR #466, opened 2016-05-23
ENH: add plan patterns module

klauer:
Allows motor trajectories to be determined purely from scan metadata.

@tacaswell, was the dictionary truly necessary in `plan_patterns`? I can't seem to recall my (our?) reasoning at this point.

Supercedes #454, which will require the tweaks mentioned in that PR - after we agree on this implementation.


klauer:
Github problems today... closing this one.

PR #467, opened 2016-05-23
ENH: plan patterns

klauer:
Allows motor trajectories to be determined purely from scan metadata.

Supercedes #454, which will require the tweaks mentioned in that PR - after we agree on this implementation.

### DIFF
--- a/bluesky/plan_patterns.py
+++ b/bluesky/plan_patterns.py
@@ -1,0 +1,213 @@
+import functools
+import operator
+
+import numpy as np
+from cycler import cycler
+from boltons.iterutils import chunked
+
+from .utils import snake_cyclers
+
+
+def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth):
+    '''Spiral scan, centered around (x_start, y_start)
+
+    Parameters
+    ----------
+    x_motor : object, optional
+        any 'setable' object (motor, temp controller, etc.)
+    y_motor : object, optional
+        any 'setable' object (motor, temp controller, etc.)
+    x_start : float
+        x center
+    y_start : float
+        y center
+    x_range : float
+        x width of spiral
+    y_range : float
+        y width of spiral
+    dr : float
+        Delta radius
+    nth : float
+        Number of theta steps
+
+    Returns
+    -------
+    cyc : cycler
+    '''
+    half_x = x_range / 2
+    half_y = y_range / 2
+
+    r_max = np.sqrt(half_x ** 2 + half_y ** 2)
+    num_ring = 1 + int(r_max / dr)
+
+    x_points = []
+    y_points = []
+    for i_ring in range(1, num_ring + 2):
+        radius = i_ring * dr
+        angle_step = 2. * np.pi / (i_ring * nth)
+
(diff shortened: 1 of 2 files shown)


klauer:
Sorry for PR confusion - for the record, there were a bunch of github-related issues today. 

Latest commit should fix inner/outer product args without resorting to checking for a name attribute.

bluesky/plan_patterns.py
```diff
@@ -0,0 +1,213 @@
...
+    (motor, start, stop, num, snake)
+    '''
+    args = list(args)
+    # The first (slowest) axis is never "snaked." Insert False to
+    # make it easy to iterate over the chunks or args..
+    args.insert(4, False)
+    if len(args) % 5 != 0:
+        raise ValueError("wrong number of positional arguments")
+
+    yield from chunked(args, 5)
```
danielballan: As much as I love generators, I think I would rather `return list(chunked(args, 5))` for consistency with the rest of these.

danielballan: Nevermind, I'm reading this to hastily and totally mistook what this is for.

bluesky/plans.py
```diff
@@ -1631,10 +1629,19 @@ def list_scan(detectors, motor, steps, *, per_step=None, md=None):
...
          'motors': [motor.name],
+         'num_steps': len(steps),
          'plan_args': {'detectors': list(map(repr, detectors)),
                        'motor': repr(motor), 'steps': steps,
                        'per_step': repr(per_step)},
-                       'plan_name': 'list_scan'})
+         'plan_name': 'list_scan',
+
+         # TODO what to do in this case?
+         'plan_pattern': 'list',
```
danielballan: 'list' seems fine

bluesky/plans.py
```diff
@@ -1631,10 +1629,19 @@ def list_scan(detectors, motor, steps, *, per_step=None, md=None):
...
+         'plan_pattern_module': '__builtin__',
+         }
+    )
+
+    md['plan_pattern_args'] = [steps]
```
danielballan: Shouldn't `[steps]` be `steps`?

bluesky/plans.py
klauer: This is the only place where the kwarg-only `plan_pattern_args` fails. Is it ok to special-case this? What's your preference?

bluesky/plans.py
klauer: For the most generality, it would seem that a tuple of (args, kwargs) would be preferable. I was aiming for simplicity by only allowing kwargs, but that may have been the wrong way to go. Which would you prefer?

danielballan: Let's go with kwargs only (labels are good...) and side-step the problem with `list` by using `np.array` instead:

```python
In [4]: np.array(object=[1,2,3])
Out[4]: array([1, 2, 3])
```

so we can do `md['plan_pattern_args'] = dict(object=steps)`

klauer: Good idea. Will switch to that.

bluesky/plans.py
danielballan: See comment below for a suggestion.

danielballan:
:+1:

PR #469, opened 2016-05-23
ENH: add tilt angle option for all spiral scans

klauer:
cc @xiaojinghuang

### DIFF
--- a/bluesky/plan_patterns.py
+++ b/bluesky/plan_patterns.py
@@ -8,7 +8,8 @@ from boltons.iterutils import chunked
 from .utils import snake_cyclers
 
 
-def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth):
+def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
+           tilt=0.0):
     '''Spiral scan, centered around (x_start, y_start)
 
     Parameters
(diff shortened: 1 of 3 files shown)


bluesky/plan_patterns.py
```diff
@@ -29,6 +30,8 @@ def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth):
         Delta radius
     nth : float
         Number of theta steps
+    tilt : float, optional
+        Tilt angle in degrees, default 0.0
```
danielballan: Standard practice in scipy is to make the interfaces radians, not degrees. Is that acceptable?

klauer: Switched to radians.

danielballan:
:+1:

bluesky/plan_patterns.py
```diff
@@ -105,7 +109,8 @@ def spiral_fermat(x_motor, y_motor, x_start, y_start, x_range, y_range, dr,
         x = radius * np.cos(angle)
         y = radius * np.sin(angle)
 
-        if abs(x) <= half_x and abs(y) <= half_y:
+        if (abs(x - y / tilt_tan) <= half_x and
```
tacaswell: Can you add `()` here just to be sure of what the order of operations is?

klauer: Added

PR #470, opened 2016-05-24
FIX: plan pattern args motor repr

klauer:
Yeah, at least the second time around with these motor reprs; this time in the plan pattern args.

Added bson encode test in test_simple_api. This should not add a bson dependency as it is pulled in through metadatastore (which I assume is in `nslsii_dev_configuration`)

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -2149,9 +2149,9 @@ def inner_product_scan(detectors, num, *args, per_step=None, md=None):
          }
     )
 
+    md['plan_pattern_args'] = dict(num=num, args=md_args)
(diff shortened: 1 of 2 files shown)


danielballan:
Why not set the `plan_pattern_args` as part of the `ChainMap` where all the other md is set? It's a minor readability/consistency improvement, but more importantly it makes that key overridable if the caller knows something that the hard-coded function does not.

tacaswell:
I think doing it this way is better, the caller _can not_ know more about this than we know locally (you should not be able to lie about which `plan_pattern_args` where used as this is the final authority on what was used).

PR #471, opened 2016-05-24
Subs wrapper

danielballan:
* Rename all the generator instance preprocessors to end with `_wrapper`.
* Add a `make_decorator` utility function that converts a generator instance wrapper to a generator function decorator.
* Ship a `*_decorator` variant of each `*_wrapper`. These are not yet documented, but they will be once I have the time to explain them carefully.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -20,6 +20,34 @@ def _short_uid(label, truncate=6):
     return '-'.join([label, str(uuid.uuid4())[:truncate]])
 
 
+def make_decorator(wrapper, *args, **kwargs):
+    """
+    The functions named *_wrapper accept a generator instance and return
+    a mutated generator instance.
+
+    Example of a 'wrapper':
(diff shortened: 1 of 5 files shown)


klauer:
👍

bluesky/plans.py
```diff
@@ -779,6 +808,67 @@ def unsubscribe(token):
...
+
+    def insert_subscribe(msg):
+        nonlocal first_msg
+        if first_msg:
+            first_msg = False
+            return pchain(_subscribe(), single_gen(msg)), None
+        else:
+            return None, None
+
+    return (yield from finalize_wrapper(plan_mutator(plan, insert_subscribe),
```
tacaswell: Why use the mutator here?  Will

```python
finalize_wrapper(pchain(_subscribe(), plan), _unsubscribe())
```
work?

tacaswell:
:+1: Other than my one comment.

PR #474, opened 2016-05-27
FIX/TST: Test pausing from suspended state; fix bug in deferred pause logic

danielballan:
A little due-diligence testing uncovered a mistake in the deferred pause processing. This probably did not  and will not affect production; it only goes awry when the `RunEngine.loop` is not the default event loop.


danielballan:
anaconda.org seems to be struggling today. Travis gets: `HTTPError: 503 Server Error: Service Unavailable: Back-end server is at capacity for url`

ericdill:
I'm hoping to get some feedback from some continuum folks on this problem. I posted an issue on conda-forge, let's see what sort of traction it can get https://github.com/conda/conda/issues/2561#issuecomment-222141827

PR #475, opened 2016-05-27
API: changes to count time in global state

danielballan:
- Remove unused gs.COUNT_TIME.
- Change default of gs.MD_TIME_KEY to a name that does not collide with any
  official AreaDetector jargon.

### DIFF
--- a/bluesky/global_state.py
+++ b/bluesky/global_state.py
@@ -123,8 +123,7 @@ class GlobalState(HasTraits):
     TABLE_COLS = List()
     PLOT_Y = Unicode()
     OVERPLOT = Bool(True)
-    COUNT_TIME = Float(1.0)
-    MD_TIME_KEY = Unicode('exposure_time')
+    MD_TIME_KEY = Unicode('count_time')
 
 
 gs = GlobalState()  # a singleton instance


tacaswell:
@klauer has to sign off on this (you can not escape the :bike: :house: )

klauer:
1 is a DNC
2 I already modified HXN's config for backward compatibility, expecting this PR to be merged: `gs.MD_TIME_KEY = 'exposure_time'`

tacaswell:
2 is just going to cause us pain on the analysis side down the road when we want to look at data from HXN + anywhere else.

danielballan:
Well, the pain of mixing count time across beamlines is unavoidable, because there each beamline -- each device! -- can have its own private definition of that heuristic term.

The important numbers are the real AD attrs, which should be going into configuration_attrs. Those are well defined and comparable across the facility.

On Fri, May 27, 2016 at 6:45 PM -0700, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


2 is just going to cause us pain on the analysis side down the road when we want to look at data from HXN + anywhere else.

PR #477, opened 2016-06-01
ENH: add skeleton of interactive plan creation

tacaswell:
- Adds a base-class to handle the mpl related work
 - Add one example of specializing to an outer_product_plan

attn @licode

### DIFF
--- a/bluesky/interactive.py
+++ b/bluesky/interactive.py
@@ -1,0 +1,130 @@
+import matplotlib.widgets as mwidgets
+from bluesky.plans import outer_product_scan
+from collections import ChainMap
+
+
+class ROIPlanCreator:
+    def __init__(self, ax):
+        '''Base class for interactively creating plans in a 2D bounding box.
+
+        Parameters
+        ----------
+        ax : matplotlib.axes.Axes
+            The axes to install the widget on
+        '''
+        self.ax = ax
+        self.widget = mwidgets.RectangleSelector(
+            self.ax, self._onselect, useblit=True, interactive=True)
+        self._pt1 = self._pt2 = None
+
+    def _onselect(self, pt1, pt2):
+        self._pt1 = pt1
+        self._pt2 = pt2
+
+    @property
+    def last_rect(self):
+        '''data coordinates of the first and second corners.
+
+        Coordinates are lower left and upper right.
+
+        Returns
+        -------
+        points : tuple
+           ((x1, y1), (x2, y2))
+        '''
+        # TODO deal with ordering issues
+        x1, y1 = self._pt1.xdata, self._pt1.ydata
+        x2, y2 = self._pt2.xdata, self._pt2.ydata
+        return (x1, y1), (x2, y2)
+
+    @property
+    def last_plan(self):
+        '''Plan constructed from the last rectangle
+
+        Returns
+        -------
+        plan : iterable
+        '''
+        return self._plan(*self.last_rect)
+
+    def _plan(self, pt1, pt2):
+        '''Construct a plan given the ROI
+
+        Sub-classes must override this
+
+        Returns
+        -------
+        plan
+        '''
+        raise NotImplementedError()
+
+    @property
+    def active(self):
+        'If the widget should be active'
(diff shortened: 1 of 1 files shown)


bluesky/interactive.py
```diff
@@ -0,0 +1,130 @@
...
+        'If the widget should be active'
+        return self.widget.get_active()
+
+    @active.setter
+    def active(self, active):
+        return self.widget.set_active(active)
+
+
+class OuterProductWidget(ROIPlanCreator):
+    def __init__(self, ax, dets, m1, m2, numsteps1, numsteps2=None,
```
klauer: Why don't these arg names match what bluesky docstrings call them?

bluesky/interactive.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def _plan(self, p1, p2):
+        x1, y1 = p1
+        x2, y2 = p2
+
+        local_md = {'interactive': {
+            'created_by': self.__class__.__name__,
+            'rect_points': {'p1': p1, 'p2': p2}}}
+
+        md = ChainMap(self.md, local_md)
```
licode: what's the difference between self.md and local_md?

bluesky/interactive.py
```diff
@@ -0,0 +1,130 @@
...
+        '''data coordinates of the first and second corners.
+
+        Coordinates are lower left and upper right.
+
+        Returns
+        -------
+        points : tuple
+           ((x1, y1), (x2, y2))
+        '''
+        # TODO deal with ordering issues
```
licode: Can you elaborate more on ordering issues?

licode:
examples or tests on this function?
Just left some simple questions. 
LGTM 👍

bluesky/interactive.py
tacaswell: `self.md` hold the metadata that the user passed in on init, `local_md` is metadata we generated as part of this call.

bluesky/interactive.py
tacaswell: that comment should not have survived.  I was not sure if mpl return the points in the order you drew them or (bottom left, top right) and it turns out it always is (BL, UR).

bluesky/interactive.py
```diff
@@ -0,0 +1,130 @@
...
+        x1, y1 = p1
+        x2, y2 = p2
+
+        local_md = {'interactive': {
+            'created_by': self.__class__.__name__,
+            'rect_points': {'p1': p1, 'p2': p2}}}
+
+        md = ChainMap(self.md, local_md)
+
+        return outer_product_scan(self.dets,
```
klauer: Having access to `plan_args` before creating the plan would be nice, too.

bluesky/interactive.py
```diff
@@ -0,0 +1,130 @@
...
+        if md is None:
+            md = {}
+        self.md = md
+
+    def _plan(self, p1, p2):
+        x1, y1 = p1
+        x2, y2 = p2
+
+        local_md = {'interactive': {
+            'created_by': self.__class__.__name__,
```
klauer: `created_by_widget`?

bluesky/interactive.py
tacaswell: Not sure what you mean, the `plan_args` should be put into the md by the underlying plan.

klauer: I mean `widget.plan_args` would give back that information without creating a plan.

tacaswell: What is the use case for that?  I would rather keep the 'required' implementation as simple as possible (`__init__` + `_plan`).

klauer: Re-use / tweak without rewriting - pretty much what I'm always pushing for

tacaswell:
I rearranged the meta-data capture to make `_plan` extremely thin.

bluesky/interactive.py
danielballan: Can you delete that comment, then?

danielballan:
Rather than spin Travis again, I'll just merge as is and cleanup that potentially-confusing comment separately....

PR #480, opened 2016-06-04
Fix describe cache

tacaswell:


klauer:
HXN ran into this issue tonight.

I definitely wouldn't have been able to fix this having seen the solution. :+1: @tacaswell!

tacaswell:
The trigger for this bug was reading out a sub-device in one event stream and the full device in the other event stream because the describe cache is shared across all of the event streams and was rejecting any collisions between any existing return from `describe`.

This changes the logic to no longer check for collisions in the describe cache, but only against already read objects only if we are bundling.  This is nominally more expensive, but I doubt it matters.  It does, we can pre-compute collision lists when populating the descriptor cache.

The alternative to this fix would be to have the RE know about how to traverse the object hierarchy and accept collisions. However this has the down side of teaching the RE about the details of how we are composing objects and would still fail in the case where you had the same PV in two objects (hence they can have the same name).  They should not be allowed to be read in the same event, but they should be allowed to be read in separate event streams.

A down side of this is that it is now possible (through pathological device set up) to construct event streams in a single run with identical datakeys.  This can be confusing, but I think that this foot-cannon is worth the simplicity and flexibility of this solution.

There is also a minor behavior change in that `describe` is now called after the first read (not before), but if the things that describe hits change due to a read we have bigger problems so I think this is ok.  The previous order could be restored, it would just require an extra if block or unconditionally caching the return

danielballan:
IIRC we made this assumption of global name unquieness intentionally at the beginning, and we said we'd break it when we found a use case for doing so. (I could have sworn there was a code comment to that effect, but I don't see it here....)

I think there may be some assumptions in the broker that this PR breaks. Let's take a careful look at that next week.

tacaswell:
Baseline already broke those assumptions in broker.

Issue #483, opened 2016-06-06
`step` is unused in `spec_api.tw`...


ericdill:
Turns out that the `tw()` function isn't actually tested in the bluesky test suite. I'm not entirely sure how to test this function or I'd just add a test and submit a PR.  Note specifically that the `step` positional argument is not used within the function.

klauer:
I believe it's used here:
https://github.com/NSLS-II/bluesky/blob/master/bluesky/plans.py#L2470

User input overrides the given default step. The user input portion is what makes this difficult to test, although overriding the `'input'` command in the RE would work well enough in my opinion.

ericdill:
But the `step` argument from `spec_api.tw` never gets passed to the `plans.tweak` function: https://github.com/NSLS-II/bluesky/blob/master/bluesky/spec_api.py#L385

klauer:
Good catch; you're right.

PR #484, opened 2016-06-07
FIX: Pass args through to tweak.

danielballan:
The tweak/tw plan was basically written to prove the point that, yes, SPEC's concept of "tweak" is within the scope of the Python language. When I see someone actually use it at a beamline, I'll take the trouble to write a test. ;- )

Closes https://github.com/NSLS-II/bluesky/issues/483


danielballan:
P.S. @klauer Your suggestion in #483 is reasonable, but I think the most interesting/important thing to test is the actual behavior of `Msg('input')`.

PR #486, opened 2016-06-09
ENH: add flag to control replay on resume

tacaswell:
Do this by controlling if messages are cached or not.

Still needs tests.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1692,6 +1692,7 @@ finalize_decorator = make_decorator(finalize_wrapper)
 lazily_stage_decorator = make_decorator(lazily_stage_wrapper)
 fly_during_decorator = make_decorator(fly_during_wrapper)
 inject_md_decorator = make_decorator(inject_md_wrapper)
+run_decorator = make_decorator(run_wrapper)
 
 
 @planify
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -806,7 +806,7 @@ def unsubscribe(token):
     return (yield Msg('unsubscribe', token=token))
 
 
-def run_wrapper(plan, md=None):
+def run_wrapper(plan, *, md=None):
     yield from open_run(md)
     yield from plan
     yield from close_run()
@@ -1020,7 +1020,7 @@ def subs_context(plan_stack, subs):
 
 
 @contextmanager
-def run_context(plan_stack, md=None):
+def run_context(plan_stack, *, md=None):
     """Enclose in 'open_run' and 'close_run' messages.
 
     Parameters
@@ -1765,7 +1765,7 @@ def count(detectors, num=1, delay=None, *, md=None):
 
     plan_stack = deque()
     with stage_context(plan_stack, detectors):
-        with run_context(plan_stack, md):
+        with run_context(plan_stack, md=md):
             plan_stack.append(plan)
     return plan_stack
 
@@ -1835,7 +1835,7 @@ def list_scan(detectors, motor, steps, *, per_step=None, md=None):
 
     plan_stack = deque()
     with stage_context(plan_stack, list(detectors) + [motor]):
-        with run_context(plan_stack, md):
+        with run_context(plan_stack, md=md):
             for step in steps:
                 plan_stack.append(per_step(detectors, motor, step))
     return plan_stack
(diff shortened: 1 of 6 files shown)


tacaswell:
next steps are:

 1. add messages for this
 2. add 'rewindable' flag to `trigger_and_read`

danielballan:
Alternate idea: it might be possible to solve the underlying problem (detectors that can't be rewound) without additional state and without customizing plans to know about un-rewindable detectors.

When a RunEngine goes down for pause or suspend, it calls `pause()` on all objects it has seen. At that time, any detectors that have been caught in an un-rewindable state could raise a special Exception, causing the RunEngine to clear its msg cache.

tacaswell:
That is trading state for increased mandatory interface.  With this state
we might be able to get rid of the pause and resume api.

danielballan:
It's true -- this is not a reduction in overall complexity, but my initial reaction (60/40) is that I prefer the complexity to go on the problematic devices than to infect every plan. FWIW, this is not new *mandatory* interface: Devices that have no special pause/resume behavior need not implement the method.

tacaswell:
If it should be caching / trying to rewind is an internal detail of the RE.  Device should be able to expose to plan if it is safe to rewind or not.

More code coming.

tacaswell:
This is ready for a review, still need to update some tests

bluesky/plans.py
```diff
@@ -808,6 +809,12 @@ def unsubscribe(token):
     return (yield from single_gen(Msg('unsubscribe', token=token)))
 
 
+def run_wrapper(plan, md=None):
```
danielballan: Find the codeblock in this file where we make a corresponding `*_decorator` for every `*_wrapper` and make one for this.

bluesky/plans.py
```diff
@@ -1038,6 +1045,42 @@ def event_context(plan_stack, name='primary'):
...
+    ----------
+    plan_stack : list-like
+        appendable collection of generators that yield messages (`Msg` objects)
+    rewindable : bool
+
+    '''
+    initial_rewindable = None
+
+    def capture_rewindable_state():
+        nonlocal initial_rewindable
```
danielballan: :100:

klauer:
Why is CI failing here?

klauer:
👍 Looks good to me otherwise.

tacaswell:
because it is seeing an old version of FS

tacaswell:
This is ready for review, I will put up a separate PR(s) implementing @danielballan 's scheme for doing this.  They are going to be completely orthogonal.

Should I squash this?

bluesky/plans.py
```diff
@@ -805,6 +806,12 @@ def unsubscribe(token):
     return (yield Msg('unsubscribe', token=token))
 
 
+def run_wrapper(plan, md=None):
```
klauer: `md` is kwarg-only in many places - not sure how it should be here

bluesky/run_engine.py
```diff
@@ -268,6 +270,16 @@ def __init__(self, md=None, *, loop=None, md_validator=None):
         self.loop.call_soon(self._check_for_signals)
 
     @property
+    def rewindable(self):
+        return self._rewindable_flag
+
+    @rewindable.setter
+    def rewindable(self, v):
+        self._rewindable_flag = bool(v)
```
klauer: If I set `rewindable = True` twice (accidentally or otherwise), is it desirable to reset in this situation?

bluesky/run_engine.py
```diff
@@ -1668,6 +1681,9 @@ def _checkpoint(self, msg):
 
     @asyncio.coroutine
     def _reset_checkpoint_state(self):
+        self._reset_checkpoint_state_meth()
+
+    def _reset_checkpoint_state_meth(self):
```
klauer: Could you make this the main method, then just wrap with `asyncio.coroutine` to create the other one?
Bike-sheddy, but it feels a bit more natural to me:

```python
    def _reset_checkpoint_state(self):
        if self._msg_cache is None:
            return
        ...

    _reset_checkpoint_state_coroutine = asyncio.coroutine(_reset_checkpoint_state)
```

Almost seems like the coroutines should have a common prefix or something instead... but I've exceeded my picky comment limit, so ignore this.

bluesky/run_engine.py
tacaswell: good catch.  I do not think so.

klauer:
I'm 👍 except for the double-rewindable setting.

tacaswell:
The missing lines in the patch are in `tweak`.

danielballan:
I think so.

PR #487, opened 2016-06-09
ENH: Add inject_md_wrapper and inject_md_decorator.

danielballan:
Example of where this is useful:

```python
inject_md_wrapper(multi_run_plan, {'last_dark_frame_uid': '...'})
```

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1068,6 +1068,29 @@ def fly(flyers, *, md=None):
     yield from close_run()
 
 
+def inject_md_wrapper(plan, md):
+    """
+    Inject additional metadata into a run.
+
+    This takes precedences over the original metadata dict in the event of
+    overlapping keys, but it does not mutate the original metadata dict.
+    (It uses ChainMap.)
+
+    Parameters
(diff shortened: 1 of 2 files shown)


tacaswell:
Should we be using this all over the place in the default plans?

danielballan:
Interesting suggestion, but I don't think so. This is a tool for injecting md into a plan from some third party that may or may not be exposing an md kwarg. But it is good that bluesky plans all have an md kwarg.

On Thu, Jun 9, 2016 at 8:21 PM -0400, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


Should we be using this all over the place in the default plans?

danielballan:
Actually I think you are right, @tacaswell. Notice that the built-in plans are mostly not using the stub plans (e.g., we use `Msg('set', ...)` instead of `abs_set(...)`) because I wrote the stubs last. If they turn out to be a good idea, we can do the refactor and test that the message sequences are exactly the same.

tacaswell:
Will merge this now, refactor later.

PR #492, opened 2016-06-20
Fix motor reset

tacaswell:
- do we want to try to reset motors on abort? The last commit makes it so that we do, but we might want to replace `CanceledError` with a custom one to catch a la `GeneratorExit` in `finalize_wrapper` to by-pass the clean up
 - we probably want to re-think how pausing works.  Instead of stopping the loop in `request_pause` we stash an exception in `self._exception` and then once it bounces through the plans and then stop from with in the `_run` co-routine 
 - `_exception` can now get clobbered a couple of ways, should maybe make it a stack?

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -922,8 +922,15 @@ class RunEngine:
 
             for task in asyncio.Task.all_tasks(self.loop):
                 task.cancel()
+            for p in self._plan_stack:
+                try:
+                    p.close()
+                except RuntimeError as e:
+                    print('The plan {!r} tried to yield a value on close.  '
+                          'Please fix your plan.')
             self.loop.stop()
             self.state = 'idle'
+        print('exited run loop')
 
     def _check_for_signals(self):
         # Check for pause requests from keyboard.
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -802,7 +802,7 @@ class RunEngine:
                     # get new message but don't process it yet.
                     try:
                         if self._exception is not None:
-                            # throw the message at the current plan
+                            # throw the exception at the current plan
                             try:
                                 msg = self._plan_stack[-1].throw(
                                     self._exception)
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -859,6 +859,15 @@ class RunEngine:
                           "a hard pause instead.")
                     self.loop.call_soon(self.request_pause, False)
                     print(PAUSE_MSG)
+                except asyncio.CancelledError as e:
+                    # if we are handling this twice, raise and leave the plans
+                    # alone
+                    if self._exception is e:
+                        raise e
+                    # the case where FailedPause, RequestAbort or a coro
+                    # raised error is not already stashed in _exception
+                    if self._exception is None:
+                        self._exception = e
         except (StopIteration, RequestStop):
             self._exit_status = 'success'
             # TODO Is the sleep here necessasry?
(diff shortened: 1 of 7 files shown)


bluesky/run_engine.py
```diff
@@ -900,8 +929,15 @@ def _run(self):
...
                 task.cancel()
+            for p in self._plan_stack:
+                try:
+                    p.close()
+                except RuntimeError as e:
+                    print('The plan {!r} tried to yield a value on close.  '
+                          'Please fix your plan.')
             self.loop.stop()
             self.state = 'idle'
+        print('exited run loop')
```
danielballan: rogue print

danielballan:
* Yes.
* Good idea. I think it will make this logic easier to follow as well as give the plans an opportunity to know about / catch a pause.
* Makes sense.

tacaswell:
I think we need feedback from BL staff about the 'reset on abort' question.

tacaswell:
On a bit more consideration, using a pause exception will not work because it will kill the plans if they do not handle it.

klauer:
Since it's extremely important to get `_run` perfect, can you try to aim for 100% coverage of it? There are a few 'else' clauses not currently being covered.

tacaswell:
I spent a while last night trying to get those last few lines and they are only hit in surprisingly pathological cases.  I think I can manufacture them with suspenders?

bluesky/run_engine.py
```diff
@@ -900,8 +929,15 @@ def _run(self):
 
             for task in asyncio.Task.all_tasks(self.loop):
                 task.cancel()
+            for p in self._plan_stack:
+                try:
+                    p.close()
+                except RuntimeError as e:
+                    print('The plan {!r} tried to yield a value on close.  '
```
klauer: Missing format argument

arkilic: @klauer hawk-like vision

bluesky/run_engine.py
```diff
@@ -798,20 +798,43 @@ def _run(self):
...
                     yield from asyncio.sleep(0.0001, loop=self.loop)
-                    if self._exception is not None:
-                        raise self._exception
                     # Send last response;
                     # get new message but don't process it yet.
                     try:
-                        resp = self._response_stack.pop()
-                        msg = self._plan_stack[-1].send(resp)
+                        if self._exception is not None:
+                            # throw the message at the current plan
```
danielballan: throw the ~~message~~ exception at the current plan

bluesky/run_engine.py
```diff
@@ -839,6 +859,15 @@ def _run(self):
...
                     print(PAUSE_MSG)
+                except asyncio.CancelledError as e:
+                    # if we are handling this twice, raise and leave the plans
+                    # alone
+                    if self._exception is e:
+                        raise e
+                    # the case where FailedPause, RequestAbort or a coro
+                    # raised error is not already stashed in _exception
+                    if self._exception is None:
+                        self._exception = e
```
danielballan: This line still needs coverage according to the Coverage.io report.

danielballan:
Looks good. Aside from the one line I just marked and the logger lines in 902 and 909 (which log when an exception from `device.unstage()` or `device.stop()` is swallowed) all lines in `_run` now have coverage.

I'll cover the logger lines in a later PR. Better cover the line I marked before we merge though.

tacaswell:
@klauer I have covered your lines

klauer:
Excellent!

PR #493, opened 2016-06-20
CI: remove conda restriction

tacaswell:


tacaswell:
This un-breaks the CI, but there seem to be two legit failing tests now :cry:

tacaswell:
If this passes I will self merge and rebase my other branches.

PR #499, opened 2016-06-23
ENH: signal devices are not rewindable via pause

tacaswell:
When the pause method is called on a device it may raise a
`NoReplayAllowed` exception which will indicate to the RE that it
is not safe to re-issues any of the commands to it, thus we must
reset the checkpoint cache (if any).

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -25,6 +25,10 @@ from .plan_tools import ensure_generator
 from .plans import single_gen
 
 
+class NoReplayAllowed(Exception):
+    pass
+
+
 def expiring_function(func, loop, *args, **kwargs):
     """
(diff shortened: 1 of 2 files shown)


tacaswell:
er, force-push coming on this

tacaswell:
Other signaling methods would be on the return value (`True` vs `False`, `None` vs `not None`, make our own sentinal?).

danielballan:
I think an Exception is best, but I could live with any of the alternatives you listed if others feel differently.

klauer:
I like the exception, not some magic value

tacaswell:
so we are not going to :bike: :house_with_garden: that then.

Any thoughts on the name?

klauer:
It's good enough

danielballan:
Agreed.

PR #500, opened 2016-06-23
FIX: resume RE if msg in fuzz is 'pause'

tacaswell:


danielballan:
We should make running these part of the pre-tag process.

PR #503, opened 2016-06-23
TST: tests are marginally faster by removing delay

tacaswell:
- The SynGauss had a 0.05s sleep in it's trigger method, set to 0
 - use fresh_RE on more tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ before_install:
 
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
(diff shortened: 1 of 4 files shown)


tacaswell:
@ericdill do you know where conda stashes the un-packed tarballs?  By adding that to the lists of files to cache we might get some speed ups due to only doing one unpacking (the way the caching works is the tar up those directories and dump it on to s3)  and one DL step.

Also might protect us a bit from anaconda.org being flaky?   Only need to grab the md to verify we have the proper versions cached.

tacaswell:
Well, fine, the branch that was meant to work on exception handling turned into test go-fast-enating

ericdill:
@tacaswell conda dumps to `~/install_dir/pkgs`. You can get a lot of extra info with `conda --verbose install <pkg>`

ericdill:
@tacaswell Also have a look here: https://github.com/conda-forge/staged-recipes/issues/610. conda-forge folks are discussing the exact same thing

PR #507, opened 2016-06-24
WIP: Make ensure_generator accept a single message

danielballan:
Make this valid: `plan = pchain(count(...), Msg(...))`

This is broken. I have to move code around avoid circular imports. Leaving this here so I remember it.


danielballan:
Superceded by #508

PR #504, opened 2016-06-24
Enh re exception handling

tacaswell:
This still needs 

  - more evil tests
  - There is now a path to `halt` and for a plan to kill the RE dead as `GeneratorExit` is not a sub-class of `Exception` for exactly this reason.

PR #508, opened 2016-06-24
Enh single msg ensure generator

tacaswell:
Merge this after #504 as it includes all of the commits from that PR.

Pushing this up so we don't lose it.

### DIFF
--- a/bluesky/__init__.py
+++ b/bluesky/__init__.py
@@ -1,23 +1,12 @@
-from collections import namedtuple
-
 import logging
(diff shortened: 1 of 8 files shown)


danielballan:
Re-running tests after #504 went in

danielballan:
@tacaswell This 'halt' code is still showing in the diff. Is there git funniness going on here?

tacaswell:
@danielballan rebased, it was just gh funnyness because the same commits were in two PRs

PR #512, opened 2016-06-30
DOC: draft of through prose documentation of RE

tacaswell:
Still in .org for nice emacs python-mode integration, this will be
converted to rst before the PR is merged.


danielballan:
These commits should be in master. via one of my self-pushed docs commits from last week...not sure why github didn't close this.

PR #513, opened 2016-07-01
Remove conda recipe. lightsource2-recipes/recipes-dev is the new canonical source

ericdill:


ericdill:
This travis failure does not appear to be related to the conda recipe removal. Is that true @tacaswell ?

PR #515, opened 2016-07-01
TST: always bind Event to correct loop

tacaswell:


tacaswell:
This should fix the intermittent failures on travis.

PR #516, opened 2016-07-01
ENH: return values from co-routines/generators

tacaswell:
- return runstart uid from `open_run`
 - bounce returned values through single_gen
 - planify returns a tuple of collected results

This needs some tests and maybe a re-think of what `planify` returns.

### DIFF
--- a/bluesky/spec_api.py
+++ b/bluesky/spec_api.py
@@ -20,7 +20,8 @@ from bluesky.callbacks import LiveTable, LivePlot, LiveRaster
 from bluesky.callbacks.scientific import PeakStats
 from boltons.iterutils import chunked
 from bluesky.global_state import gs
-from bluesky.utils import (first_key_heuristic)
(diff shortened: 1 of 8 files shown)


danielballan:
I think returning run start uids is good and should go in ASAP. But, yes, the return value from `planify` is problematic. I sketched out a couple ideas and none seem workable. Maybe `planify` is hiding too much and should not be used in the built-in plans....

bluesky/spec_api.py
```diff
@@ -71,19 +74,176 @@ def setup_plot(motors):
...
+
+    for factory in factories:
+        # TODO, support positional injection
+        sig = signature(factory)
+        fact_kwargs = {}
+        missing_kwargs = set()
+        missing_args = set()
+        for k, v in sig.parameters.items():
+            if v.kind is Parameter.POSITIONAL_ONLY:
+                if v.defaut is Parameter.empty:
```
klauer: `defaut`

PR #520, opened 2016-07-07
ENH/TST/DOC: Add `monitor_wrapper`

danielballan:
- Add `monitor_wrapper` and `monitor_decorator`.
- Document `make_decorator` and `run_wrapper`.
- Document the difference between _wrapper and _decorator funcs.
- Rename `fly_during_wrapper` -> `fly_wrapper` but keep back-compat.
- Bike-shed order functions are listed in API docs.
- Change `make_decorator` to return proper decorators.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -20,7 +20,7 @@ def _short_uid(label, truncate=6):
     return '-'.join([label, str(uuid.uuid4())[:truncate]])
 
 
-def make_decorator(wrapper, *args, **kwargs):
+def make_decorator(wrapper):
     """
     The functions named *_wrapper accept a generator instance and return
     a mutated generator instance.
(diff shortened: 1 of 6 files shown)


danielballan:
attn @hhslepicka 

While I was in here making other changes, @tacaswell pointed out that my implementation of `make_decorator` wasn't quite right. The usage in current master is:

```
f = some_decorator(plan, *args, **kwargs)
```

but it should be

```
f = some_decorator(*args, **kwargs)(plan)
```

so that the normal decorator usage works:

```
@some_decorator(*args, **kwargs):
def plan(...):
    ...
````

which it currently does not.

This PR makes normal decorator usage work, and that may break the code that you wrote today when you upgrade to v0.5.4. Just a heads up.

hhslepicka:
@danielballan I see... When do you think that will be a good time to update?

danielballan:
Probably wait until next cycle, unless users and encountering problems with pause/resume, which have been cleaned up since 0.5.3, the current deployment.

You might notice that we have been better about recording API changes (and there are not many of them) in the docs, so get upgrade process should be more organized that in the past.

On Thu, Jul 7, 2016 at 10:35 PM -0400, "Hugo Slepicka" <notifications@github.com<mailto:notifications@github.com>> wrote:


@danielballan<https://github.com/danielballan> I see... When do you think that will be a good time to update?

tacaswell:
I am still a bit unhappy about the name change.  These wrappers take in plans which are assumed to have open/close messages in them (which is driven by t

he fact that many of all of the non-trivial plans do).  However it is possible that we will refactor those to _not_ include the open/close (more like `trigger_and_read`) and in that case it will no longer be clear that `*_wrapper` assumes that there is an open/close.

danielballan:
OK, that's a fair point. I reverted the name change and force-pushed. We should merge that to fix the `make_decorator` API before more people use it.

danielballan:
All tests pass locally. On Travis, 3.4 is stalling on the vertical integration test (which also just happened on master). 3.5 is randomly failing on one fermat spiral test (but all others pass). The clock is ticking and no one is watching GChat, so I'm going to self-merge this. I acutely feel all attendant shame, etc.

PR #521, opened 2016-07-19
FIX/API: rework exception handling in plan mutator

tacaswell:
Mirror the same code in the RunEngine.  Unfortunately the exact
exception handling is different so abstracting the shared code out is
difficult.


klauer:
Given the similarity to the RE implementation, should we just trust @tacaswell here or do we need another code-flow explanation meeting?

tacaswell:
In either case, this still needs some more tests.

tacaswell:
For future reference this was superseded by #625

PR #523, opened 2016-07-19
PERF: asyncio.sleep(0) is enough to relinquish control.

### LINKED ISSUE #506: Replace the asyncio.sleep in the _run loop with something more efficient

danielballan:
Closes #506

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -844,7 +844,7 @@ class RunEngine:
                     # This sleep has to be inside of this try block so
                     # that any of the 'async' exceptions get thrown in the
                     # correct place
-                    yield from asyncio.sleep(0.0001, loop=self.loop)
+                    yield from asyncio.sleep(0, loop=self.loop)
                     # The case where we have a stashed exception
                     if self._exception is not None:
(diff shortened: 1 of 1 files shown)


tacaswell:
huh, looks like the thread executor does not execute correctly with a 0 sleep...

klauer:
Hmm, only one test fails on 3.4 and 6 fail on 3.5. 

Is the issue report in python/asyncio not accurate or did I/we misinterpret it?

arkilic:
Seems like bunch of asserts failing, I think your interpretation is correct

tacaswell:
Restarted now that we have removed the thread pool.

klauer:
Passes! 👍

danielballan:
Huh. I don't have a precise understanding of why. Do you?

tacaswell:
because although the sleep(0) is enough for asyncio to context switch, it is not enough to context switch the threads.  Putting a non-zero sleep in is enough to (almost always) ensure that the thread context switch happens and the event callbacks to process.

danielballan:
I see. I'm satisfied. Merging.

PR #525, opened 2016-07-27
ENH: Add optional progress bar in _wait and _sleep.

danielballan:
WIP

attn @klauer It turns out you were right again. `.position` and `.target` and `.device` were all good ideas.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -22,7 +22,7 @@ from .utils import (CallbackRegistry, SignalHandler, normalize_subs_input,
                     RequestAbort, RequestStop,  RunEngineInterrupted,
                     IllegalMessageSequence, FailedPause, FailedStatus,
                     InvalidCommand, PlanHalt, Msg, ensure_generator,
-                    single_gen, short_uid, status_progress_bar)
+                    single_gen, short_uid, async_status_progress_bar)
 
 
 class RunEngineStateMachine(StateMachine):
@@ -1654,7 +1654,7 @@ class RunEngine:
         statuses, futs = list(zip(*self._groups.pop(group)))
         if msg.kwargs.get('progress'):
             # Poll the status objecst and show a progress bar (uses tqdm).
-            yield from status_progress_bar(*statuses, loop=self.loop)
+            yield from async_status_progress_bar(*statuses, loop=self.loop)
 
             # If some status has been marked failed, it will be handled
             # just below via the usual route, in the closure given to
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1,5 +1,6 @@
 import asyncio
 import time as ttime
+import math
 import sys
 import logging
 from warnings import warn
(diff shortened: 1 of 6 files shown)


tacaswell:
why isn't travis running?

danielballan:
While I'm in the business of changing `_wait`, I made it illegal to wait on a group that isn't in the cache. (You'll get a `KeyError`.) This change exposed a bug in the `adaptive_*` plans, which is a point in its favor I'd say.

danielballan:
This has decent coverage (68% patch coverage, compared to the 83% coverage in bluesky at large) and does not change in default behavior. I'd like to merge it so that we can start playing with it in master. There will definitely be tweaks.

bluesky/run_engine.py
```diff
@@ -1688,15 +1690,22 @@ def _wait(self, msg):
...
+            Msg('wait', group=<GROUP>, *, progress=False)
 
         where ``<GROUP>`` is any hashable key.
         """
         if msg.args:
             group, = msg.args
         else:
             group = msg.kwargs['group']
-        futs = list(self._groups.pop(group, []))
+        statuses, futs = list(zip(*self._groups.pop(group)))
```
tacaswell: This lost a default value on the pop, was that intentional?

bluesky/utils.py
```diff
@@ -787,3 +788,33 @@ def sanitize_np(val):
...
+
+
+def status_progress_bar(*status_objects, loop):
+    from tqdm import tqdm
+    N = 2  # max decades to divide steps into
+    pbars = {}
+    for st in status_objects:
+        pbar = tqdm()
+        if hasattr(st, 'device'):
+            initial = status.device.position
```
tacaswell: we maybe want to consider shoving more information into the status objects.  This will start the progress bar from where it is _now_, not where the move started, do we care?

bluesky/utils.py
```diff
@@ -787,3 +788,33 @@ def sanitize_np(val):
...
+            if hasattr(st, 'target'):
+                dx = abs(st.target - initial)
+                pbar.total = dx / 10**(int(floor(log10(dx))) + (1 - N))
+        pbars[st] = pbar
+    sleep_times = itertools.chain((10**i for i in range(-6, -1)),
+                                  itertools.repeat(0.1))
+    while True:
+        # Poll the status objects.
+        if all(st.done for st in status_objects):
```
tacaswell: can this conditional go in the `while` clause instead of infinite loop?

bluesky/utils.py
```diff
@@ -787,3 +788,33 @@ def sanitize_np(val):
...
+            break
+        for st, pbar in pbars.items():
```
tacaswell: should there we a version of this outside of this loop to finish all of the progress bars?

bluesky/utils.py
```diff
@@ -787,3 +788,33 @@ def sanitize_np(val):
             return val.item()
         return val.tolist()
     return val
+
+
+def status_progress_bar(*status_objects, loop):
```
tacaswell: can you tack 'async' on this name?  I think we want a synchronous version too

bluesky/utils.py
danielballan: Good point. Let's see how well this works before we worry about a coordinated PR with ophyd. There might be yet other things we want to add to Status.

bluesky/run_engine.py
danielballan: Yes; see [this comment](https://github.com/NSLS-II/bluesky/pull/525#issuecomment-236270694) above.

danielballan:
We will revisit this. Tagging @awalter-bnl so he can get updates.

danielballan:
We really ought to try to get this in. As much as I love tqdm, I think the progress bar implementation should be pluggable from the RunEngine's point of view.

danielballan:
I have a better idea for this that is less invasive, so I'm going to close my own PR.

PR #527, opened 2016-07-27
FIX: allow `None` to be a valid group for waiting

tacaswell:
This makes the code work the way that @danielballan thought it
worked.

All objects that are set/trigger/kickoff/collect will have their status
objects hooked up to a an `asyncio.Event` and if the Status object
reports that it failed will cause an exception to be raised.

Before this, if there was not a non-None group, the RunEngine would not
check the success or failure of the action.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1479,18 +1479,17 @@ class RunEngine:
 
         ret = obj.kickoff(*msg.args, **msg.kwargs)
 
-        if group:
-            p_event = asyncio.Event(loop=self.loop)
+        p_event = asyncio.Event(loop=self.loop)
 
-            def done_callback():
-                if not ret.success:
-                    task = self.loop.call_soon_threadsafe(self._failed_status,
-                                                          ret)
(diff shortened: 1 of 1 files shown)


danielballan:
Much to my dismay, this bug goes [all the way back to the beginning](https://github.com/tacaswell/bluesky/blob/c6f925ca1f4a5dc0157d7abc3bb99f38f90f3468/bluesky/run_engine.py#L528). Sometimes weekend hacks become production code.

danielballan:
I'll give @klauer a chance to glance at this, but it looks good.

klauer:
Looks good to me.

tacaswell:
Hold off on merging this until I get some docs with this.

danielballan:
As in release notes docs? Let's get this in.

tacaswell:
Yes, mostly want to document that you used to be able to escape all of the status object stuff, but now you can not.

danielballan:
Reviewed with @klauer and @tacaswell

PR #528, opened 2016-07-28
When LiveTable skips an unrecognized dtype, warn

danielballan:
Example: If a user adds a field with an 'array' dtype, LiveTable silently skips it. This confused me for a minute earlier this week.


tacaswell:
Can you fake up a test?

danielballan:
Done.

PR #530, opened 2016-08-01
SNIP: remove lossy subscriptions

klauer:
After a discussion with @tacaswell. Anything that's slow should be subscribed on a separate process somehow (outside of the scope of this PR).

Proof-of-concept / seeing what this looks like / seeing what tests fail

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -358,6 +358,9 @@ class LiveRaster(CallbackBase):
         cb.set_label(self.I)
 
     def event(self, doc):
+        if self.I not in doc['data']:
+            return
+
         seq_num = doc['seq_num'] - 1
         pos = list(np.unravel_index(seq_num, self.raster_shape))
         if self.snaking[1] and (pos[0] % 2):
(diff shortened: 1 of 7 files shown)


danielballan:
Interesting. I assumed that we would maintain this long term. Running in one process does make many things easier. Why remove the option?

danielballan:
I like the lossiness mechanism. It doesn't necessarily need to live in the RunEngine, but I hope it survives.

tacaswell:
I think if you want lossy events you subscribe a callback which handles that / farming out to an executor pool

tacaswell:
This change also means we can start to safely use blitting in callbacks :)

klauer:
It doesn't have to be merged - we can discuss this at some point.

That said, removal should probably be paired with the addition of a standardized interprocess subscription method (e.g., zmq)

danielballan:
Definitely.

danielballan:
For the record, I have been persuaded that this does not solve a problem that we currently have, so there is no reason to wait for multi-process and other dispatch mechanism before merging it.

PR #531, opened 2016-08-01
MNT: Tweak default log templates.

danielballan:
- Use 'plan_name', often more descriptive than 'plan_type'.
- Only assume that plan has metadata that is ensured by RE. (Do not assume
  it has 'plan_args'.)

Solves issue where user makes custom plan, omits plan_args metadata, and see that olog callback is not run.

### DIFF
--- a/bluesky/callbacks/olog.py
+++ b/bluesky/callbacks/olog.py
@@ -8,14 +8,16 @@ logger = logging.getLogger(__name__)
 
 TEMPLATES = {}
 TEMPLATES['long'] = """
-{{- start.plan_type }} ['{{ start.uid[:6] }}'] (scan num: {{ start.scan_id }})
+{{- start.plan_name }} ['{{ start.uid[:6] }}'] (scan num: {{ start.scan_id }})
 
 Scan Plan
 ---------
-{{ start.plan_type }}
-{%- for k, v in start.plan_args | dictsort %}
-    {{ k }}: {{ v }}
-{%-  endfor %}
+{{ start.plan_name }}
+{% if 'plan_args' in start %}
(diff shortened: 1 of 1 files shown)


klauer:
👍  much better using `'plan_name'` instead.

danielballan:
attn @Brudhu

bluesky/callbacks/olog.py
```diff
@@ -141,7 +143,7 @@ def lbcb(name, doc):
 def call_str(start, call_template=None):
     """Given a start document generate an evalable call scring
 
-    The default template assumes that `plan_args` and `plan_type`
+    The default template assumes that `plan_args` and `plan_name`
```
tacaswell: No longer make strict assumption about `'plan_args'`

danielballan: Hmm. How can we write a call string without args? This template is not used by default as far as I can tell, so a hard assumption here does not break the default logging config.

On Mon, Aug 1, 2016 at 5:49 PM -0400, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


In bluesky/callbacks/olog.py<https://github.com/NSLS-II/bluesky/pull/531#discussion_r73059470>:

> @@ -141,7 +143,7 @@ def lbcb(name, doc):
>  def call_str(start, call_template=None):
>      """Given a start document generate an evalable call scring
>
> -    The default template assumes that `plan_args` and `plan_type`
> +    The default template assumes that `plan_args` and `plan_name`


No longer make strict assumption about 'plan_args'

tacaswell: nm :sheep:

PR #536, opened 2016-08-03
TST: non-rewindable pause while bundling

tacaswell:


tacaswell:
And this PR shows why line coverage is a bad measure of test coverage :disappointed:

klauer:
Assuredly a single metric is enough to tell you that your code is perfect. Right, guys? ... guys?

danielballan:
Clearly the answer is to the failure of annoying github integrations is more annoying github integrations.

PR #541, opened 2016-08-12
Fix bunch of bugs

tacaswell:
This still needs work, but broad strokes are correct.


bluesky/run_engine.py
```diff
@@ -690,23 +692,88 @@ def request_suspend(self, fut, *, pre_plan=None, post_plan=None,
...
+                    continue
+                # normal use, if it runs cleanly, stash the response and
+                # go to the top of the loop
+                else:
+                    ret = response
+                    continue
+            except StopIteration:
+                break
+            finally:
+                print('BAILED')
```
danielballan: rogue print

danielballan:
@tacaswell I think most of this was addressed by #579 #601 and #625 but that 60e4590 might need to be cherry-picked. Does that look right to you?

tacaswell:
#625 took over #521 but nothing from this one

danielballan:
The title of this PR suggests it would be good to revisit it soon.

danielballan:
Soon it will be time to celebrate the 1-year anniversary of this PR.

danielballan:
5+ years

tacaswell:
I still think this is worth considering.  The problem that this was trying to fix was suspenders firing while in the pre/post plan of a suspender which at SRX was causing the shutter to flicker so running the pre/post plans in a more isolated task would make those "un suspendable".

This is also the first step to making a "run this (sub) plan and tell me what it is done" message which is something I think we do want (but has not been a high priority).

If there is a reason why this is a bad idea or we see a better way to solve incorrectly re-entrant pre/post plans on suspenders then we should close this.  Just being "old" is not a reason to close something.

PR #542, opened 2016-08-15
MNT: monitor fixes

tacaswell:


bluesky/run_engine.py
```diff
@@ -1303,7 +1303,9 @@ def _monitor(self, msg):
         if msg.args:
             raise ValueError("The 'monitor' Msg does not accept positional "
                              "arguments.")
-        name = msg.kwargs.get('name')
+        kwargs = dict(msg.kwargs)
+        name = kwargs.pop('name', new_uid()[:6] + '_monitor')
```
klauer: I thought we had like a `new_short_uid` somewhere...

tacaswell: we probably do...

bluesky/plans.py
```diff
@@ -1164,7 +1160,8 @@ def monitor_during_wrapper(plan, signals):
     --------
     `bluesky.plans.fly_during_wrapper`
     """
```
danielballan: Agree that my scheme for defaulting name was not configurable enough. Could we keep `sig.name + '-monitor` as a default?

bluesky/run_engine.py
```diff
@@ -1248,7 +1248,9 @@ def _monitor(self, msg):
         if msg.args:
             raise ValueError("The 'monitor' Msg does not accept positional "
                              "arguments.")
-        name = msg.kwargs.get('name')
+        kwargs = dict(msg.kwargs)
+        name = kwargs.pop('name', short_uid('monitor'))
```
danielballan: Good catch.

bluesky/plans.py
danielballan: I read the diff backward. Carry on!

danielballan:
👍

PR #545, opened 2016-08-22
FIX: add `aspect` kwarg to LiveRaster

### LINKED ISSUE #544: liveraster unuseful on high-aspect ratio scans
We need to perform scans that are of order 10 mm x 10 microns in scale.  LiveRaster insists on a fixed aspect ratio pixel, but as my scan has an aspect ratio of order 1000, the resulting image is useless.  Subclassing LiveRaster and attempting to use ax.set_aspect() creates a plot with the correct size, which is overwritten with a fixed-aspect ratio plot when it is populated with data.  I presume that this means that ax.set_aspect() should be issued between populating the plot and drawing it, rather than where it is currently issued at initialization.  I believe that exposing an aspect ratio option through the LiveRaster interface would be beneficial.

At the moment, I can't plot such scans on the fly.

tacaswell:
fixes #544

Also adds missing documentation

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -321,10 +321,19 @@ class LiveRaster(CallbackBase):
 
     cmap : str or colormap, optional
        The color map to use
+
+    xlabel, ylabel : str, optional
+       Labels for the x and y axis
+
+    extent : scalars (left, right, bottom, top), optional
+       Passed through to `Axes.imshow`
+
+    aspect : str or float, optional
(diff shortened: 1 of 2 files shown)


klauer:
Can we have someone who knows matplotlib better than @tacaswell verify this?

... kidding. Looks good to me. Waiting on travis.

danielballan:
Probalby everything in `examples/` should just land in the docs. Separate PR, obviously.

danielballan:
Travis stalled on a conda network error. Kicked.

PR #546, opened 2016-08-22
FIX: pass through tilt angle in spec api

klauer:
Reminder to fix on the current master


klauer:
Also includes fix to the tilt calculations themselves 

cc @xiaojinghuang

PR #548, opened 2016-08-23
FIX: ct should have a plot if num=None.

danielballan:


klauer:
Unrelated test failure 👍

PR #551, opened 2016-08-26
Plot cbs take ax

danielballan:


danielballan:
I'm not sure why the example deletions ended up on this branch, but in any case they are intentional (having been superceded by documentation already in master).

bluesky/callbacks/core.py
```diff
@@ -347,20 +351,25 @@ class LiveRaster(CallbackBase):
...
     :class:`bluesky.callbacks.LiveMesh`.
     """
     def __init__(self, raster_shape, I, *,
                  clim=None, cmap='viridis',
-                 xlabel='x', ylabel='y', extent=None, aspect='equal'):
-        fig, ax = plt.subplots()
+                 xlabel='x', ylabel='y', extent=None, aspect='equal',
+                 ax=None):
+        if ax is None:
+            ax = plt.gca()
```
tacaswell: This should still do `fix, ax = plt.subplots` if an axes is not passed in. This will re-use the last axes the use clicked on.

bluesky/callbacks/core.py
```diff
@@ -266,13 +266,18 @@ class LiveMesh(CallbackBase):
...
     See Also
     --------
     :class:`bluesky.callbacks.LiveRaster`.
     """
     def __init__(self, x, y, I, *, xlim=None, ylim=None,
-                 clim=None, cmap='viridis'):
-        fig, ax = plt.subplots()
+                 clim=None, cmap='viridis', ax=None):
+        if ax is None:
+            ax = plt.gca()
```
tacaswell: some comment about `fig, ax = plt.subplots()`

bluesky/callbacks/core.py
```diff
@@ -126,8 +127,7 @@ def __init__(self, y, x=None, legend_keys=None, xlim=None, ylim=None,
         else:
             self.x = None
         self.y, *others = _get_obj_fields([y])
-        self.fig = fig
-        self.ax = fig.gca()
+        self.ax = plt.gca()
```
tacaswell: I think we should do some `isinstance` checking here to see if it is a figure and raise a warning.  I am 100% sure that this kwarg is being used on the floor.

danielballan: I addressed this by leaving the `fig` kwarg in, with a deprecation warning, and adding `ax`. Using both `fig` and `ax` raises since that would lead to confusing behavior. I don't think we need `isinstance`; we are two separate kwargs.

danielballan:
Ready for another round of review.

bluesky/callbacks/core.py
```diff
@@ -112,11 +114,20 @@ class LivePlot(CallbackBase):
     >>> RE(my_scan, my_plotter)
     """
     def __init__(self, y, x=None, legend_keys=None, xlim=None, ylim=None,
-                 fig=None, **kwargs):
+                 ax=None, fig=None, **kwargs):
```
tacaswell: can you make everything after `x` kwarg only?

danielballan:
Travis is happy; codecov patch is complaining that we don't test all the input-validation cutouts. OK to merge as is?

PR #553, opened 2016-08-29
Overhaul examples

danielballan:
Two motivations for this change:

1. It is now easy to make simulated hardware that returns realistic data for testing analysis workflows when the beam is down. See docstrings for suggestive examples.

2. The `examples.py` has built up a lot of cruft as did not serve as a nice entry point for developers trying to understand the bluesky interface.

Still to do (in a later PR): unify the three different implementations of simulated flyers.

### DIFF
--- a/bluesky/broker_examples.py
+++ b/bluesky/broker_examples.py
@@ -12,40 +12,22 @@ class SynGauss2D(Reader):
 
     Example
     -------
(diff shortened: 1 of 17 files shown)


danielballan:
Tests pass; coverage is mad merely because I deleted code.

klauer:
👍 👍

PR #555, opened 2016-08-30
DOC: tighten document intro

danielballan:
Is this an improvement? It basically flips the order of the two intro paragraphs and deletes some words. I'm not sure that flipping the order is an improvement on the original.


danielballan:
Superceded

PR #561, opened 2016-09-29
ENH: toy detector which writes to filestore

### LINKED ISSUE #560: Help Wanted: subclass of Reader that uses filestore
@CJ-Wright

```python
class ReaderWithFileStore(Reader):

    def __init__(self, *args, fs, **kwargs):  # fs is a required keyword arg
        """
        ... copy docstring from Reader...
        fs : FileStore
            FileStore object that supports inserting resource and datum documents
        """
        super().__init__(*args, **kwargs)
        self.fs = fs
        self._resource_id = None

    def stage(self):
        self._resource_id = self.fs.insert_resource(...)

    def read(self):
        result = {}
        for name, val in super().read().items():
            # save val (probably an array) to a file somewhere, like ~/.cache/fs_cache
            # make a datum_id and figure out datum_kwargs
            self.fs.insert_datum(...)
            result[name] = datum_id
        return result

    def unstage(self):
        self._resource_id = None
```

tacaswell:
closes #560

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -364,6 +364,99 @@ class Syn2DGauss(Reader):
         return super().trigger()
 
 
+class Syn2DGaussArea(Reader):
+    """
+    Evaluate a point on a Gaussian based on the value of a motor.
+
+    Parameters
+    ----------
+    name : str
+        The name of the detector
+    motor0 : `Mover`
+        The 'x' coordinate of the 2-D gaussian blob
+    motor_field0 : str
+        The name field of the motor. Should be the key in motor0.describe()
+    motor1 : `Mover`
+        The 'y' coordinate of the 2-D gaussian blob
+    motor_field1 : str
+        The name field of the motor. Should be the key in motor1.describe()
+    center : iterable, optional
+        The center of the gaussian blob
+        Defaults to (0,0)
+    Imax : float, optional
+        The intensity at `center`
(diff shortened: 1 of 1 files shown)


tacaswell:
This needs #573 to be merged and then be re-based to use the async wait.

danielballan:
Superceded by #574. @tacaswell, feel free to reopen if you think we need both. I like that #574 exposes the calls to `insert_*`.

tacaswell:
I like that this one is much simpler.

danielballan:
OK, last question then: @CJ-Wright's variant writes to /tmp instead of the XDG cache location. For a "toy" application like this, I think I prefer that (at least as a default). I can see the argument for leaving it as is; I just want to raise the point.

Also, can we give the built-in instance a more descriptive name than `det6`? Maybe `areadet`?

danielballan:
@dchabot Yep, thanks for chiming in though.

I'm imaging tutorial attendees (the primary users of this `bluesky.examples` stuff, at least right now) filling up their `~/.cache/filestore` directory and having no idea why their hard drive is getting full. This is why `/tmp` seems like a good default.

danielballan:
The simulated devices have moved into ophyd, so this PR is not actionable. I think we are pretty happy with the simulated AD classes in `ophyd.sim` at this point, but feel free to re-open this over there if you like.

PR #564, opened 2016-10-04
MNT: stop all movable objects as successful

tacaswell:
These relies on changes to ophyd to allow a `success` kwarg in stop.


danielballan:
Tests passed for all builds, but one hung after success. Merging.

danielballan:
Cross-referencing ophyd API change needed by this PR: https://github.com/NSLS-II/ophyd/pull/362

PR #565, opened 2016-10-07
Make example hardware more realistic

danielballan:
* Add a setpoint to the motor instances. This makes it easier to see the difference between `.name` and `.describe().keys()`. It's confusing when there is only one key.
* Refactor how read vs. conf works to make it more like ophyd.

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -179,7 +179,7 @@ class Reader:
                         'dtype': 'number',
                         'shape': [],
                         'precision': 2}
-                for field in self._fields if field in self.read_attrs}
+                for field in self.read_attrs}
 
     def read_configuration(self):
         """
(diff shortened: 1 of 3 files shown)


bluesky/examples.py
```diff
@@ -170,21 +179,22 @@ def describe(self):
                         'dtype': 'number',
                         'shape': [],
                         'precision': 2}
-                for field in self._read_fields}
+                for field in self._fields if field in self.read_attrs}
```
tacaswell: wouldn't this be simpler as a loop over `self.read_attrs`?

danielballan: Yes.

danielballan: Fix pushed.

PR #573, opened 2016-10-14
Fix sigint again

tacaswell:


bluesky/tests/conftest.py
```diff
@@ -10,6 +10,15 @@ def fresh_RE(request):
     loop.set_debug(True)
     RE = RunEngine({}, loop=loop)
     RE.ignore_callback_exceptions = False
+
+    def clean_al():
```
danielballan: Is this supposed say `clean_all`? Maybe `al` stands for something that I'm not seeing; if so, I'll leave this :sheep: here preemptively.

tacaswell: It should be `clean_el` (for event loop).

bluesky/tests/test_devices.py
```diff
@@ -8,6 +8,7 @@
 
 if ophyd:
     from ophyd import Component as Cpt, Device, Signal
+    ophyd.setup_ophyd()
```
danielballan: This probably explains some of our intermittent failures. Good catch.

bluesky/examples.py
```diff
@@ -154,8 +153,14 @@ def __getstate__(self):
         return (self.name, self._fields, self.read_attrs, self.conf_attrs,
                 self._monitor_intervals)
 
-    def trigger(self):
-        "No-op: returns a status object immediately marked 'done'."
+    def trigger(self, *, delay_time=0):
```
danielballan: Bad choice of words. I agree that ophyd is the reference implementation. With the recent changes, `Reader` is a useful public class in its own right, not just a base class. Is there a good reason to have it present a non-standard API? This `delay_time` seems like a init-time decision ("how should this fake device behave on trigger") and not something that should be passed through a Msg into the RE into the device at trigger-time.

tacaswell: to avoid picking a magic attribute name and spending too much time walking the class tree.

tacaswell:
This should be merged and a more agressive re-working of sig-int handling will come as a follow on

bluesky/tests/conftest.py
danielballan: Plan to fix this?

danielballan: Nevermind; I merged locally and added a commit.

PR #574, opened 2016-10-17
Fs reader

CJ-Wright:
@danielballan Here is the filestore reader you asked for. Next step is to write the handler, and maybe the more general version as well. More than happy to get an initial review though.

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -618,6 +618,73 @@ class MockFlyer:
         pass
 
 
+class GeneralReaderWithFileStore(Reader):
+    """
+
+    Parameters
+    ----------
+    name : string
+    read_fields : dict
+        Mapping field names to functions that return simulated data. The
+        function will be passed no arguments.
+    conf_fields : dict, optional
+        Like `read_fields`, but providing slow-changing configuration data.
+        If `None`, the configuration will simply be an empty dict.
+    monitor_intervals : list, optional
+        iterable of numbers, specifying the spacing in time of updates from the
+        device (this applies only if the ``subscribe`` method is used)
+    loop : asyncio.EventLoop, optional
+        used for ``subscribe`` updates; uses ``asyncio.get_event_loop()`` if
+        unspecified
+    fs : FileStore
+        FileStore object that supports inserting resource and datum documents
+
+    """
+
+    def __init__(self, *args, fs, save_path=None, save_func=np.save,
+                 save_spec='RWFS_NPY', save_ext='.npy',
+                 **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fs = fs
+        self.save_func = save_func
+        self.save_ext = save_ext
+        self._resource_id = None
+        if save_path is None:
+            self.save_path = mkdtemp()
+        else:
+            self.save_path = save_path
+        self.filestore_spec = save_spec  # spec name stored in resource doc
+
+        self._file_stem = None
+        self._path_stem = None
+        self._result = None
+
+    def stage(self):
+        self._file_stem = short_uid()
+        self._path_stem = os.path.join(self.save_path, self._file_stem)
+        self._resource_id = self.fs.insert_resource(self.filestore_spec,
+                                                    self._path_stem, {})
+
+    def trigger(self):
+        # save file stash file name
+        self._result = {}
+        for idx, (name, val) in enumerate(super().read().items()):
+            self.save_func('{}_{}.{}'.format(self._path_stem, idx,
+                                             self.save_ext), val)
+            datum_id = str(uuid4())
+            self.fs.insert_datum(self._resource_id, datum_id,
+                                 dict(index=idx))
+            self._result[name] = datum_id
+
+    def read(self):
+        return self._result
+
+    def unstage(self):
+        self._resource_id = None
+        self._file_stem = None
+        self._path_stem = None
+
+
 class ReaderWithFSHandler(HandlerBase):
     specs = {'RWFS_NPY'} | HandlerBase.specs
 
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -418,6 +418,60 @@ class Syn2DGauss(Reader):
         super().__init__(name, {name: func})
 
 
+class ReaderWithFileStore(Reader):
+    """
+
+    Parameters
+    ----------
+    name : string
+    read_fields : dict
+        Mapping field names to functions that return simulated data. The
+        function will be passed no arguments.
+    conf_fields : dict, optional
+        Like `read_fields`, but providing slow-changing configuration data.
+        If `None`, the configuration will simply be an empty dict.
+    monitor_intervals : list, optional
+        iterable of numbers, specifying the spacing in time of updates from the
+        device (this applies only if the ``subscribe`` method is used)
+    loop : asyncio.EventLoop, optional
+        used for ``subscribe`` updates; uses ``asyncio.get_event_loop()`` if
+        unspecified
+    fs :FileStore
+        FileStore object that supports inserting resource and datum documents
+
+    Examples
+    --------
+    A detector that always returns 5.
+    >>> det = Readable('det', {'intensity': lambda: 5})
+
+    A detector that is coupled to a motor, such that measured insensity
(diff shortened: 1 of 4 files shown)


CJ-Wright:
Closes #560 ?

bluesky/examples.py
```diff
@@ -4,6 +4,19 @@
 from threading import RLock
 import numpy as np
 from .run_engine import Msg
+from uuid import uuid4
+import uuid
+from tempfile import mkdtemp
+import os
+
+def new_uid():
```
danielballan: Use `new_uid` and `short_uid` from `bluesky.utils`

bluesky/examples.py
```diff
@@ -393,8 +411,77 @@ def trigger(self):
...
+    conf_fields : dict, optional
+        Like `read_fields`, but providing slow-changing configuration data.
+        If `None`, the configuration will simply be an empty dict.
+    monitor_intervals : list, optional
+        iterable of numbers, specifying the spacing in time of updates from the
+        device (this applies only if the ``subscribe`` method is used)
+    loop : asyncio.EventLoop, optional
+        used for ``subscribe`` updates; uses ``asyncio.get_event_loop()`` if
+        unspecified
+    fs :FileStore
```
danielballan: missing whitespace after `:`

bluesky/examples.py
```diff
@@ -393,8 +411,77 @@ def trigger(self):
...
+        FileStore object that supports inserting resource and datum documents
+
+    Examples
+    --------
+    A detector that always returns 5.
```
danielballan: Examples need to be updated or deleted -- either is fine.

bluesky/examples.py
```diff
@@ -393,8 +411,77 @@ def trigger(self):
...
+
+    def stage(self):
+        self._resource_id = self.fs.insert_resource(self.filestore_spec,
+                                                    self.path_stem, {})
+
+    def trigger(self):
+        # save file stash file name
+        self.result = {}
+        for idx, name, val in enumerate(super().read().items()):
+            np.save('{}_{}.{}'.format(self.path_stem, idx, '.tif'), val)
```
danielballan: This names something `*..tif` You might as well hard-code the `.tif` in the format string since it's not configurable anyway.

bluesky/examples.py
```diff
@@ -393,8 +411,77 @@ def trigger(self):
...
+        super().__init__(*args, **kwargs)
+        self.fs = fs
+        self._resource_id = None
+        if save_path is None:
+            self.save_path = mkdtemp()
+        else:
+            self.save_path = save_path
+        self.filestore_spec = 'RWFS_TIFF'  # spec name stored in resource doc
+
+        self.file_stem = new_short_uid()
```
danielballan: I would prefer to make a new file stem on stage, not at init.

bluesky/examples.py
```diff
@@ -393,8 +411,77 @@ def trigger(self):
...
+            datum_id = str(uuid4())
+            self.fs.insert_datum(self._resource_id, datum_id,
+                                 dict(index=idx))
+            self.result[name] = datum_id
+
+    def read(self):
+        return self.result
+
+    def unstage(self):
+        self._resource_id = None
```
danielballan: Reset result (and file_stem and path_step) to None here as well.

bluesky/examples.py
```diff
@@ -393,8 +411,77 @@ def trigger(self):
...
+        if save_path is None:
+            self.save_path = mkdtemp()
+        else:
+            self.save_path = save_path
+        self.filestore_spec = 'RWFS_TIFF'  # spec name stored in resource doc
+
+        self.file_stem = new_short_uid()
+        self.path_stem = os.path.join(self.save_path, self.file_stem)
+        self.result = None
```
danielballan: Can all three of three (file_stem, path_step, result) be renamed to _file_stem, etc.?

bluesky/examples.py
```diff
@@ -4,6 +4,12 @@
 from threading import RLock
 import numpy as np
 from .run_engine import Msg
+from uuid import uuid4
+import uuid
+from tempfile import mkdtemp
+import os
+from bluesky.utils import new_uid, short_uid
+from filestore.handlers_base import HandlerBase
```
danielballan: Bluesky does not have a hard dependency on filestore. You probably don't need anything in `HandlerBase`, anyway, right?

bluesky/examples.py
```diff
@@ -393,8 +404,71 @@ def trigger(self):
...
+        self._path_stem = os.path.join(self.save_path, self._file_stem)
+        self._resource_id = self.fs.insert_resource(self.filestore_spec,
+                                                    self._path_stem, {})
+
+    def trigger(self):
+        # save file stash file name
+        self._result = {}
+        for idx, (name, val) in enumerate(super().read().items()):
+            np.save('{}_{}.npy'.format(self._path_stem, idx), val)
+            datum_id = str(uuid4())
```
danielballan: `new_uid`?

bluesky/examples.py
```diff
@@ -516,16 +591,93 @@ def stop(self):
...
+        self._resource_id = self.fs.insert_resource(self.filestore_spec,
+                                                    self._path_stem, {})
+
+    def trigger(self):
+        # save file stash file name
+        self._result = {}
+        for idx, (name, val) in enumerate(super().read().items()):
+            self.save_func('{}_{}.{}'.format(self._path_stem, idx,
+                                             self.save_ext), val)
+            datum_id = str(uuid4())
```
danielballan: `new_uid`?

bluesky/examples.py
```diff
@@ -516,16 +591,93 @@ def stop(self):
...
+        self._resource_id = None
+        if save_path is None:
+            self.save_path = mkdtemp()
+        else:
+            self.save_path = save_path
+        self.filestore_spec = save_spec  # spec name stored in resource doc
+
+        self._file_stem = None
+        self._path_stem = None
+        self._result = None
```
danielballan: Initialize this to an empty dict so that `self.read()` returns a valid dict-like result at all times.

bluesky/examples.py
```diff
@@ -516,16 +591,93 @@ def stop(self):
...
+                                 dict(index=idx))
+            self._result[name] = datum_id
+
+    def read(self):
+        return self._result
+
+    def unstage(self):
+        self._resource_id = None
+        self._file_stem = None
+        self._path_stem = None
```
danielballan: self._result.clear()

bluesky/examples.py
```diff
@@ -516,16 +591,93 @@ def stop(self):
...
+
+    def stage(self):
+        self._file_stem = short_uid()
+        self._path_stem = os.path.join(self.save_path, self._file_stem)
+        self._resource_id = self.fs.insert_resource(self.filestore_spec,
+                                                    self._path_stem, {})
+
+    def trigger(self):
+        # save file stash file name
+        self._result = {}
```
danielballan: It's better to use `self.result.clear()` so that anything (say, in a subclass) that refers to `self._result` does not lose its reference.

danielballan:
Also, update docstrings to include new kwargs.

CJ-Wright:
@danielballan I rebased onto master, ready for more review. Should I add the new readers to the test imports?

danielballan:
OK, looks good.

Could you copy just [this function](https://github.com/NSLS-II/filestore/blob/master/filestore/test/conftest.py#L56) into bluesky's conftest.py and then make a proper test using filestore?

bluesky/tests/conftest.py
```diff
@@ -2,7 +2,8 @@
 from bluesky.run_engine import RunEngine
 from bluesky.examples import Mover, SynGauss
 import pytest
-
```
danielballan: As with the mds imports, please move these inside the db fixture so they fail at runtime, not import time

danielballan:
The 3.5 tests passed, and the 3.4 failure is a known intermittent failure unrelated to this PR (temp_ramp). Squashing and merging.

danielballan:
Thanks, @CJ-Wright!

CJ-Wright:
Thank you @danielballan

PR #575, opened 2016-10-17
ENH: add print_summary_wrapper

tacaswell:
Print a nice summary of the plan as it goes by.

### DIFF
--- a/bluesky/plan_tools.py
+++ b/bluesky/plan_tools.py
@@ -67,6 +67,25 @@ def print_summary(plan):
     plan : iterable
         Must yield `Msg` objects
     """
+    for msg in print_summary_wrapper(plan):
+        ...
+
+
+def print_summary_wrapper(plan):
+    """Print summary of plan as it goes by
+
+    Prints a minimal version of the plan, showing only moves and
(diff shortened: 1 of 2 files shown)


danielballan:
Could you slap a smoke test on this?

tacaswell:
huh, I thought print_summary already had a test.

danielballan:
Seems like it does, but my codecoverage browser plugin in show that this *wrapper* function does not. Also, please add a entry to the block of `make_decorator` calls (also in plans.py) and add an entry to the table of Stub Plans (doc/source/plan.rst).

tacaswell:
But the existing `print_summary` is implemented in terms of the wrapper

danielballan:
```
$ git grep bluesky/tests/ -e "print_summary"
$
```

danielballan:
I'll leave this un-documented for now because we probably want to add other variants and I don't want to lock the naming in hard just yet. Also, it's not clear to me if `plan_tools.py` vs `plans.py` is a useful distinction or if we should just move this stuff into `plans.py`. (There isn't much in `plan_tools.py`.)

Issue #134, opened 2015-07-02
Revisit schemas to make validation stricter.


danielballan:
Our jsonschemas had never been actually tested on our real Documents either on the ophyd side or the metadatastore side, and I took some liberties to get things going because less-than-perfectly-strict validation is better than no validation.

Here's what I remember removing:
- Validate the type of `shape` to be an array or null. If I understood correctly, it seems that tuples don't get interpreted as arrays. @tacaswell 
- Validate the details of some of the nested Dynamic Documents.

There might be others that I forgot, so to be thorough we should diff these against the documented ones. I also updated a bunch of names. Our docs should be updated there.
#128

tacaswell:
This should include ripping the schema out into a stand-alone repo.

klauer:
[bump] Schema is in its own repo now. What's the status on the validation, @arkilic? (this is assigned to you for whatever reason)

arkilic:
bluesky and mds informally uses this validation scheme via json validation. @danielballan and I are planned to be using event-model by next downtime in bs and mds

danielballan:
@arkilic I don't think @klauer was asking what the event-model repo is about, I think he's wondering if any work has been done on making the schemas stricter. AFAIK no work has been done. Would you be willing to take a careful look at the schemas and see if we can be doing more to validate, @arkilic?

arkilic:
@klauer, @tacaswell has done some work making the schema stricter by enforcing data keys and ensuring one run_stop per run_start during the migration. I need more details on `Validate the details of some of the nested Dynamic Documents`. This is something I can take care of if I have more details. Let's discuss what needs to be tackled tmrw. 
Welcome back @danielballan :octocat:

danielballan:
Are you in today? We could touch base on this.

arkilic:
Wfh Wednesday (also down with @tacaswell plague) are you in tomorrow?

danielballan:
The goal as I recall is to understand jsonschema and its limitations and capabilities and more detail and then read our schemes carefully and make them more specific where appropriate.

PR #577, opened 2016-10-19
FIX: _monitor was packing configuration wrong.

danielballan:
Stop-stopping bug in the RunEngine monitoring code organized the configuration into the descriptor with the wrong layout. The currently-deployed code acquired data without complaint but the databroker errors (rightly) when it can't find the keys it expects.

In addition to this PR, which should probably be merged and deployed ASAP, we should:

* add more tests that exercise configuration (I will add at least one to this PR later today, but I have to teach first.)
* make the event_model stricter (mentioned way back in https://github.com/NSLS-II/bluesky/issues/134) -- this should have produced an error at acquisition time!

Can you look into this next, @licode? Reference for how the Event Descriptor is organized: http://nsls-ii.github.io/bluesky/event_descriptors.html

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1260,7 +1260,7 @@ class RunEngine:
         data_keys = obj.describe()
         config = {obj.name: {'data': {}, 'timestamps': {}}}
         config[obj.name]['data_keys'] = obj.describe_configuration()
-        for key, val in list(obj.read_configuration().items()):
+        for key, val in obj.read_configuration().items():
             config[obj.name]['data'][key] = sanitize_np(val['value'])
             config[obj.name]['timestamps'][key] = val['timestamp']
         object_keys = {obj.name: list(data_keys)}
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1258,9 +1258,11 @@ class RunEngine:
                                              obj))
         descriptor_uid = new_uid()
         data_keys = obj.describe()
-        config = obj.read_configuration()
-        for key, val in list(config.items()):
-            val['value'] = sanitize_np(val['value'])
+        config = {obj.name: {'data': {}, 'timestamps': {}}}
+        config[obj.name]['data_keys'] = obj.describe_configuration()
+        for key, val in list(obj.read_configuration().items()):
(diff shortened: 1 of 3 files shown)


bluesky/run_engine.py
```diff
@@ -1258,9 +1258,11 @@ def _monitor(self, msg):
...
         descriptor_uid = new_uid()
         data_keys = obj.describe()
-        config = obj.read_configuration()
-        for key, val in list(config.items()):
-            val['value'] = sanitize_np(val['value'])
+        config = {obj.name: {'data': {}, 'timestamps': {}}}
+        config[obj.name]['data_keys'] = obj.describe_configuration()
+        for key, val in list(obj.read_configuration().items()):
+            config[obj.name]['data'][key] = sanitize_np(val['value'])
+            config[obj.name]['data'][key] = val['timestamp']
```
tacaswell: I think the second one should be `'timestamps'`?

danielballan: No, the structure of `read()` and `read_configuration()` is `{<data_key>: {'value': ..., 'timestamp': ...}, ...}`

danielballan: :sheep:.

danielballan:
Addressed line comment and added test. Anaconda.org connectivity is killing our Travis builds more than half the time today, so odds are Travis will fail. Tests pass locally.

licode:
Busy schedule today. Sure, I will catch up.

bluesky/run_engine.py
```diff
@@ -1258,9 +1258,11 @@ def _monitor(self, msg):
                                              obj))
         descriptor_uid = new_uid()
         data_keys = obj.describe()
-        config = obj.read_configuration()
-        for key, val in list(config.items()):
-            val['value'] = sanitize_np(val['value'])
+        config = {obj.name: {'data': {}, 'timestamps': {}}}
+        config[obj.name]['data_keys'] = obj.describe_configuration()
+        for key, val in list(obj.read_configuration().items()):
```
tacaswell: minor :bike: :house_with_garden: , we do not need the `list` here.

danielballan: Rookie move there... what was I thinking?

PR #579, opened 2016-10-20
FIX: do not mutate the message

tacaswell:
The Msg is immutable, but the contents are.  Do not pop things off the
kwargs as it can be destructive if the Msg comes back via the msg_cache.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1414,9 +1414,10 @@ class RunEngine:
                                          "run is open.")
         _, obj, args, kwargs = msg
         self._uncollected.add(obj)
-        group = msg.kwargs.pop('group', None)
+        kwargs = dict(msg.kwargs)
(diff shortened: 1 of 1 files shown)


danielballan:
To be safe should we use Raymond Hettiger's `ImmutableDict` for `Msg.kwargs`? I'm pretty sure there a couple places in the codebase (e.g., in the mutators) that I have done this. Might be good to be sure we've caught them all.

http://stackoverflow.com/a/9997519

danielballan:
Meh. I guess it's fine to mutate `msg.kwargs` outside the RunEngine. We only really need to worry about mutating it after it's gone into `msg_cache`. Ignore my last comment.

danielballan:
Tests passed on 3.4, failed on 3.5 due to the usual anaconda.org connectivity issue. Merging.

PR #580, opened 2016-10-24
ENH: add pause_for_debug flag to finalize_wrapper

tacaswell:
If the flag is true, yield a pause() message before finishing the clean up.

Does this seem useful @NSLS-II/dama ?


arkilic:
@tacaswell very

PR #581, opened 2016-10-24
FIX: complete should *not* wait by default

danielballan:
- The docstring states that it does not wait by default. This changes makes the
  docstring correct.
- All other plans that have a `wait` keyword set to False by default, except
  `mv` which is a convenience.


tacaswell:
Can you add an API change note?

danielballan:
Done.

PR #582, opened 2016-10-24
CI: tweak ramp plan test timings

tacaswell:
Maybe this will be more reliable on travis?


danielballan:
To save you some time: both builds failed on `test_ramp_plan.py::test_timeout`

danielballan:
We have six open PRs that have failed Travis because of this test. Should we just known-fail it until we can sort it out?

tacaswell:
Ah, that is a different failure!

danielballan:
🎉

PR #590, opened 2016-10-25
FIX: monitor seq_num should start 1 for consistency

danielballan:
For better or worse, the convention we have adopted (I think in deference to SPEC?) is that seq_num starts at 1. Callbacks like LiveRaster rely on this.


hhslepicka:
As talked with Dan, everything is good! 👍

PR #591, opened 2016-10-26
Rename 'sleep' -> 'nap' and 'input' -> 'async_input'. (Old names warn, for now.)

### LINKED ISSUE #578: Deprecate `sleep` and `input` in favor of non-ambiguous names.
Although the [standard config](http://nsls-ii.github.io/fresh-installation.html#create-new-ipython-profile) does not import `bluesky.plans.sleep` or `bluesky.plans.input` into the global namespace, users seem to do it themselves (probably via `from bluesky.plans import *`). We can and should discourage bulk-importing, but I don't think this is enough to remove the potential ambiguity.

How about `sleep` -> `delay` (thanks @hhslepicka) and `input` -> `plan_input`? I would advocate deprecating the existing names very slowly, leaving them in (with warnings) for the next several cycles.

danielballan:
Closes #578


danielballan:
3.5 passed; 3.4 failed on the infamous ramp_plan test.

danielballan:
To record Tom's verbal counterargument to this renaming:

> There is precedent for having functions that share the same name as common built-ins: for example, `asyncio.sleep`.

I'm reasonably convinced by that. The advantage of having the same name is, it correctly suggests that `bluesky.plans.sleep` is like `time.sleep` but for plans. Likewise for `input`. The disadvantage, of course, is that people aren't used to thinking about scope can get confused.

For now we will *not* make this change, but we'll revisit if people keep shooting themselves in the foot with this. In the meantime, scientists can always alias `nap = bluesky.plans.sleep` in their startup files.

danielballan:
To record Tom's verbal counterargument to this renaming:

> There is precedent for having functions that share the same name as common built-ins: for example, `asyncio.sleep`.

I'm reasonably convinced by that. The advantage of having the same name is, it correctly suggests that `bluesky.plans.sleep` is like `time.sleep` but for plans. Likewise for `input`. The disadvantage, of course, is that people aren't used to thinking about scope can get confused.

For now we will *not* make this change, but we'll revisit if people keep shooting themselves in the foot with this. In the meantime, scientists can always alias `nap = bluesky.plans.sleep` in their startup files.

PR #592, opened 2016-10-26
ENH: Add a stop command and a stop plan.

danielballan:


danielballan:
Incredibly, we never added a command for this.

PR #596, opened 2016-10-27
WIP: Fix FS-integrated Reader example.

danielballan:
@CJ-Wright This is a partial fix. The other FS Reader will have to fixed in a similar way, and we have to think about configuration_attrs as well. Would you push it across the finish line?

I discovered the problem writing this example (which works on this branch but does not work on master). https://gist.github.com/danielballan/e7474490fab2139ed2981b6ca156597d

cc @chiahaoliu

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -489,9 +489,16 @@ class ReaderWithFileStore(Reader):
             # a reference to FileStore.
             reading['value'] = datum_id
             self._result[name] = reading
+        return NullStatus()
 
     def read(self):
         return self._result
+
+    def describe(self):
+        res = super().describe()
+        for key in res:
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@danielballan Is there any chance you can give some more insight into the error? (I am thinking of writing a test, you think it is prudent)

danielballan:
>  Is there any chance you can give some more insight into the error?

The Travis failure was just random. The 3.4 build passed; the 3.5 build failed.

> I am thinking of writing a test, you think it is prudent.

We can skip this for now because there is still some internal discussion about whether bluesky should test against broker in this repo at all.

CJ-Wright:
ah, ok

Issue #595, opened 2016-10-27
Implement ensure_uid and use it where needed in callbacks


danielballan:
This will fix errors where LiveTable and friends don't work on documents from the databroker, which are referenced.

danielballan:
Closed by #597

PR #597, opened 2016-10-28
FIX: Various fixes to make callbacks work on saved data.

danielballan:
Documents from the databroker come back with their reference fields (e.g., `event['descriptor']`) dereferenced, which is different that how they come out of bluesky. The new `ensure_uid` utility func guards against errors due to that distinction.


danielballan:
Intentionally removed test_broker_examples which required the old DataBroker (the singleton configured with files). Left in test_vertical_integration, which uses a nice `db` fixture and portable-mds.

danielballan:
Closes #595

PR #598, opened 2016-10-29
REF: Make callbacks call super(); split plot drawing into update().

danielballan:
For the plotting callbacks, here `event` just performs the document munging and calls `update(...)` to update the caches and line objects do the redraw. The arguments to `update` are just numbers, and this provides a good hook for subclasses that want to alter those numbers without mutating the document.

Also, all callbacks that inherit from `CallbackBase` now call super() to play well with multiple inheritance. We don't actually *use* MI anywhere in the package, but now we play nicer with user MI code.

### DIFF
--- a/bluesky/callbacks/broker.py
+++ b/bluesky/callbacks/broker.py
@@ -14,7 +14,7 @@ class LiveImage(CallbackBase):
     field : string
         name of data field in an Event
     """
-    def __init__(self, field):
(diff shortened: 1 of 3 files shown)


danielballan:
Travis is happy. Codecov is mad because previously-untested codepaths are now one line longer by the addition of `super()...`. Would like to punt adding test coverage to that callbacks to another day...

PR #600, opened 2016-10-30
ENH: Add preprocessors to the RunEngine.

### LINKED ISSUE #568: RunEngine should have a list preprocessors
The RunEngine should have a list of preprocessors that it applies to every plan. This would be a way run beamline-specific cleanup, for example. The list would be empty by default.

This is an old idea I mentioned near the beginning of the generator refactor.

danielballan:
This PR adds an attribute to the RunEngine, `preprocessors`, that holds a list of functions that will be applied to the plan before execution. It's a hook that could be used to add things like beamline-specific cleanup that should always be performed.

Since preprocessors see all of the messages, they can do conditional cleanup like, "If we see detector X in this plan, do the following to 'cleanup' for detector X."

I'm pretty sure this is a good idea, but just in case we discover it's bad idea and want to take it away, let's hold off on documenting it for now. The PR includes a test and a docstring update but no prose documentation.

Closes #568

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -88,7 +88,8 @@ class RunEngine:
                              'unstage', 'monitor', 'unmonitor', 'open_run',
                              'close_run']
 
-    def __init__(self, md=None, *, loop=None, md_validator=None):
+    def __init__(self, md=None, *, loop=None, preprocessors=None,
+                 md_validator=None):
         """The Run Engine execute messages and emits Documents.
 
         Parameters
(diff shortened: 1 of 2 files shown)


tacaswell:
The current ways to achieve this are:

 - sub-class the RE
 - remember to wrap all of your plans with your decorator

-----

stream of thought notes follow (read at your own peril):

The bad-idea-smell of this is that it is starting to push very specific knowledge about a BL into the RE where as it (maybe?) should live in the plans.

On the other hand, this is doing it in a consistent way which seems more graceful than subclassing the RE.

Also seems more reliable than requiring users to remember to wrap _every_ plan.  On the other-other hand, the wrapping could (should?) happen in the BL config so that is not a super high burden

On the 5th hand, if this is a bad idea then the way suspenders is done is _also_ a bad idea.

If we merge this we clearly have to add a message to manage the pre-processors so plans can install functions that mutate future plans in the current `__call__` context :smiling_imp: .

danielballan:
Notes on an in-person conversation for posterity / the rest of the team:

I agree all around. The problem with asking BLs to wrap all their plans with custom decorators (say, decorators that perform BL-specific cleanup) is that then they can't use any bluesky plans out of the box. After time we add a plan, they have to remember to wrap it.

Like suspenders and some metadata, the stateful workflow makes sense here.

@tacaswell wants to me note that he is persuaded by that. We will probably merge this, but debate is still open if any DAMA members or ex-pats want to drop in and comment. :ghost:

arkilic:
@irawaluyo I think this is something you are interested in

PR #601, opened 2016-10-31
FIX: do not mutate the set message

tacaswell:
Missed this when fixing the other cases


danielballan:
Tests passed on 3.4 and then Travis hung. Connection problems plagued 3.5.

PR #603, opened 2016-11-01
API: Change default monitor name.

danielballan:
Change the default stream name for monitors from `<signal_name>-monitor` to just `<signal_name>`. The original default was both unnecessarily verbose and not a valid Python identifier (due to the `-`), which is fine but less convenient to access within a document / document stream because it is not a valid attribute name.

Since monitoring was completely broken in the last tagged release, this is a cost-free API break. We just need to notify CSX and SRX where fixed versions of monitoring are source-installed. They can keep the old names (it's configurable through the 'monitor' plans) or transition to the new ones at their preference.

attn @cmazzoli @ambarb @gjwillms @imrealkaren

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1273,7 +1273,7 @@ def monitor_during_wrapper(plan, signals):
     --------
     :func:`bluesky.plans.fly_during_wrapper`
     """
-    monitor_msgs = [Msg('monitor', sig, name=sig.name + '-monitor')
+    monitor_msgs = [Msg('monitor', sig, name=sig.name + '_monitor')
                     for sig in signals]
     unmonitor_msgs = [Msg('unmonitor', sig) for sig in signals]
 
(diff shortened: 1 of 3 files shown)


tacaswell:
I disagree it is overly verbose.  The same signal may also be on one or more other streams.  Would prefer to go to `_` instead of `-`.

I assume this is to make the dot-access work.  Should probably take the pandas approach here and just not expose streams that are not valid python identifiers through this API.

danielballan:
How does "_monitor" solve the problem of one signal being part of multiple streams?

On Mon, Oct 31, 2016 at 10:28 PM -0400, "Thomas A Caswell" <notifications@github.com<mailto:notifications@github.com>> wrote:


I disagree it is overly verbose. The same signal may also be on one or more other streams. Would prefer to go to _ instead of -.

I assume this is to make the dot-access work. Should probably take the pandas approach here and just not expose streams that are not valid python identifiers through this API.

danielballan:
On second thought, no argument that leaving the suffix the more conservative choice, so fair enough. I force-pushed a new commit that merely makes the change `-` -> `_`.

danielballan:
@tacaswell :+1: the change I ultimately made above. Self-merging.

PR #604, opened 2016-11-01
LiveFit fixes

danielballan:
* Test a multivariate model with LiveFit.
* Fix bugs exposed by that test.
* Document LiveFitPlot and add more examples.


hhslepicka:
LGTM, 👍

PR #605, opened 2016-11-04
Improve LiveFit examples (use noisy data) and add new Cookbook section

danielballan:
There a couple ways that projects automate generating examples. I looked at a couple and didn't see a clear winner, so I'd like to stick to a basic manual approach until it's more clear which solution best fits bluesky.


danielballan:
P.S. This includes some actual code changes as well as docs changes, so it need actual review.

danielballan:
@tacaswell +1'ed this over gchat, modulo the change just pushed. Merging.

PR #608, opened 2016-11-06
ENH: add hard-kill option to SignalHandler

tacaswell:
If the handler has accumulated 10 sigints without it being cleared by
the RunEngine, something is going very wrong.  Let python raise a
KeyboardInterupt.

This needs a bit of consideration to make sure that we are properly resetting the count from the RE.

### DIFF
--- a/bluesky/tests/test_run_engine.py
+++ b/bluesky/tests/test_run_engine.py
@@ -502,7 +502,7 @@ def test_sigint_three_hits(fresh_RE, motor_det):
     assert done_cleanup_time - end_time > 0.3
 
 
-def test_sigint_many_hits(fresh_RE):
+def test_sigint_many_hits_pln(fresh_RE):
     RE = fresh_RE
     pid = os.getpid()
 
@@ -527,6 +527,18 @@ def test_sigint_many_hits(fresh_RE):
     assert RE.state == 'paused'
     RE.abort()
 
+
+def test_sigint_many_hits_cb(fresh_RE):
+    RE = fresh_RE
+    pid = os.getpid()
+
+    def sim_kill(n=1):
+        for j in range(n):
+            print('KILL')
+            ttime.sleep(0.05)
+            os.kill(pid, signal.SIGINT)
+
+    @bp.run_decorator()
     def infinite_plan():
         while True:
             yield Msg('null')
@@ -537,12 +549,14 @@ def test_sigint_many_hits(fresh_RE):
     start_time = ttime.time()
     timer = threading.Timer(0.2, sim_kill, (11,))
     timer.start()
-    RE(infinite_plan())
+    RE(infinite_plan(), hanging_callback)
     # Check that hammering SIGINT escaped from that 10-second sleep.
     assert ttime.time() - start_time < 2
     # The KeyboardInterrupt will have been converted to a hard pause.
     assert RE.state == 'paused'
     RE.abort()
+    # Check that hammering SIGINT escaped from that 10-second sleep.
+    assert ttime.time() - start_time < 2
 
 
 def _make_plan_marker():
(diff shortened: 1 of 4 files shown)


danielballan:
Assuming the "very wrong" state occurs in running mode, the `KeyboardInterrupt` will pass through the finally block. We should set `self._sigint_handler.count = 0` there. FYI, the only other place we update the count is [in the 10-second timeout](https://github.com/NSLS-II/bluesky/blob/master/bluesky/run_engine.py#L1076).

tacaswell:
That is OK, because in either of the other cases you are going through the pause path which (eventually) tears down our sigint handler to give the user back the prompt.  A new one is created when we resume.

danielballan:
Great. Then I would say add add that line for the finally block of the _run coroutine and then this is good to go.

danielballan:
This seems impossible to test because if we knew how to get the RE into the proverbial "very wrong" state we would fix it. :- D

tacaswell:
Put a long sleep in a callback or plan

danielballan:
Bump -- Add line to finally block as mentioned above and add test.

tacaswell:
the test will require threads :disappointed:

danielballan:
Offline, @tacaswell asked if I would take this over. I rebased, added test, and force-pushed to this branch. This is my first time using the power to push to another person's branch. It feels a bit sketchy, but I appreciate that it reduces duplicate/superceding PRs.

danielballan:
@licode It would be nice to get this in before 0.7.1 is tagged. I think this is the last thing we need for that release, but I'd like @tacaswell to confirm that before we tag it.

tacaswell:
Can you also add a test that hangs in a callback?  Not sure it is worth holding the PR on that.

Not sure what is going wrong with the 3.4 travis.  I don't think there were any changes to signal handling recently...

licode:
The upgrade will start at week of Jan 3rd, so we have time on this.

tacaswell:
We only have a few working days between now and then though ;)

danielballan:
There are now two tests:

* interrupting a hanging plan
* interrupting a hanging callback

In both cases, notice that the RunEngine is catching the `KeyboardInterrupt` originating from Python's default SIGINT handler and going through the code path that we created to defend oursevles from third-party libraries generating rogue KeyboardInterrupt exceptions. The user will see:

```
An unknown external library has improperly raised KeyboardInterrupt. Intercepting and triggering a hard pause instead.
```

and the RunEngine will land in a paused state, as these tests now verify.

If Travis is happy, this is good to merge.

arkilic:
@danielballan I read that error with your voice in my head

danielballan:
We also need to update the test `test_rogue_sigint`. While there, might as well update/remove the comment which has been rendered incorrect.

bluesky/tests/test_run_engine.py
```diff
@@ -501,6 +502,61 @@ def sim_kill(n=1):
...
+    def hanging_callback(name, doc):
+        ttime.sleep(10)
+
+    start_time = ttime.time()
+    timer = threading.Timer(0.2, sim_kill, (11,))
+    timer.start()
+    RE(infinite_plan(), {'start': hanging_callback})
+    # Check that hammering SIGINT escaped from that 10-second sleep.
+    assert ttime.time() - start_time < 2
+    # The KeyboardInterrupt will have been converted to a hard pause.
```
danielballan: This comment (and similar above) should be deleted or updated.

danielballan:
Commit message is wrong (4 <-> 5) but otherwise looks good. Thanks for grappling with that.

tacaswell:
force-pushed to fix title, but code is identtcal

Issue #588, opened 2016-10-25
Clean up FailedStatus errors in 'finally' of run loop


danielballan:

danielballan:
Closed by #630

PR #610, opened 2016-11-10
FIX: Ensure tasks that would raise FailedStatus are canceled on exit.

danielballan:
Meant to address #588 but I'm not sure this will do the trick. Needs test.


tacaswell:
I think we need to add a check in the internal `done_callback` which is registered onto the `Status` objects that generates those tasks to not create the tasks if we want to ignore the failures.

I think the `stop` not reporting failure will also help with this in many cases.

danielballan:
Superceded by #630

PR #611, opened 2016-11-11
TST: Test that event_model forbids dots in key names.

danielballan:
attn @licode


danielballan:
cc @chiahaoliu This should address the problem wherein portable-mds allowed dots but production metadatastore did not. Now they are universally prohibited by bluesky via event_model. Thanks for your report.

licode:
👍  LGTM.

PR #612, opened 2016-11-11
MNT: Specify a root in FileStore resource generated by example.

danielballan:
Need this to test `db.export` and it's good to advertise this best-practice anyway.

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -473,7 +473,8 @@ class ReaderWithFileStore(Reader):
         self._file_stem = short_uid()
         self._path_stem = os.path.join(self.save_path, self._file_stem)
         self._resource_id = self.fs.insert_resource(self.filestore_spec,
-                                                    self._path_stem, {})
+                                                    self._path_stem, {},
+                                                    root=self.save_path)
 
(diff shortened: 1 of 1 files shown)


bluesky/examples.py
```diff
@@ -473,7 +473,8 @@ def stage(self):
         self._file_stem = short_uid()
         self._path_stem = os.path.join(self.save_path, self._file_stem)
```
tacaswell: you need to not do the join here.

danielballan: I agree that this could be factored a little differently, but I wanted to give the author some space to breathe. Meanwhile, what about the diff in this PR? It is a prerequisite to getting export tests working.

tacaswell: No, I mean that my adding 'root' to the insert you need to _not_ do this join as root is joined to the path in filestore and you will end up with `'/save/path/save/path/file/stem'`

danielballan: Oh, I see. The handler was ignoring root as well, so this "worked" as is but would not have worked with a new root. Fixed in latest commit.

PR #614, opened 2016-11-24
DOC: Add grid-in-grid example.

danielballan:
An example inspried by Melinda's use case at XPD
grid_in_grid-1

Don't merge quite yet -- there is a sphinx problem.


danielballan:
Sphinx issues fixed. Self-merging this docs-only PR.

@tacaswell You mentioned you made an improved version of this using Poisson noise for your APS talk. I failed to easily reverse-engineer that so please submit the patch if you still have it.

PR #619, opened 2016-12-08
Simplify md composition

danielballan:
We got carried away with our love of `ChainMap` and made the plans code more complicated than it needs to be. We only need `ChainMap` in places where we want to avoid mutating the lowest-precedence dict. So, we *do* need it in the RunEngine, where we want to compose with `RE.md` without mutating it. But we *don't* need it inside plans, where we are composing with a hard-coded, single-use local var that we can always safely mutate.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -948,9 +948,7 @@ def open_run(md=None):
     --------
     :func:`bluesky.plans.close_run`
     """
-    if md is None:
-        md = {}
-    return (yield Msg('open_run', **md))
+    return (yield Msg('open_run', **(md or {})))
 
 
 def close_run():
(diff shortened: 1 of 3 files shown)


bluesky/plans.py
```diff
@@ -3033,8 +2972,7 @@ def ramp_plan(go_plan,
 
         In seconds.
     '''
-    if md is None:
-        md = {}
+    md = {'plan_name': 'ramp_plan'}.update(md or {})
```
klauer: The one line pattern here is pretty unsightly... Consider splitting it into two:

```python
md = {'plan_name': 'ramp_plan'}
md.update(md or {})
```

danielballan: That's reasonable. I chose the one liner because the two liner requires a new variable and more changes...

```python
_md = {'plan_name': 'ramp_plan'}
_md.update(md or {})
...
yield from blerg(md=_md)
```

but you convinced me it's worth it.

danielballan:
This is now passing and can be merged before the tagging.

bluesky/plans.py
```diff
@@ -3014,13 +2965,12 @@ def relative_spiral(detectors, x_motor, y_motor, x_range, y_range, dr, nth,
     :func:`bluesky.plans.spiral`
     :func:`bluesky.plans.spiral_fermat`
     '''
-    if md is None:
-        md = {}
-    md = ChainMap(md, {'plan_name': 'relative_spiral_fermat'})
+    _md = {'plan_name': 'relative_spiral_fermat'}
```
tacaswell: plan name is wrong as well.

tacaswell: and there are extra documented arguments.

PR #620, opened 2016-12-11
Move cookbook to top-level docs

danielballan:
It would be better to present examples like these in the top-level docs, where we can integrate databroker, ophyd, and others. I also like the Problem / Approach / Example Solution structure we adopted there, inspired by some O'Reilly books. I have applied it to these examples while moving them.

They are already pushed and published here: http://nsls-ii.github.io/examples/index.html

Self-merging this since it only affects docs. Making the PR to leave a record of the change and the reason for it.


hhslepicka:
Nice work! Just one point here, the outputs at example 2 looks like broken.

Well done! \o

PR #621, opened 2016-12-15
DOC: put API specification first

tacaswell:


danielballan:
Would you mind removing the whitespace changes to the ChainMap section? They would create needless merge conflicts with #619.

tacaswell:
But this one passes tests, #619 does not :smiling_imp:

PR #625, opened 2016-12-16
Fix return uids

tacaswell:


tacaswell:
This supersedes #521

bluesky/tests/test_run_engine.py
```diff
@@ -565,7 +565,7 @@ def test_finalizer_closeable():
     plan.close()
 
 
-def test_invalid_generator(fresh_RE, motor_det):
+def test_invalid_generator(fresh_RE, motor_det, capsys):
```
danielballan: pytest never ceases to amaze

PR #626, opened 2016-12-20
ENH: first draft at qt plan builder

tacaswell:
Type on the right, nested messages show up on the left

Still hacky.

To test, in IPython

```python
%matplotlib qt
plt.plot(range(5))
```
to make sure qApp and everything is setup, then run

```python
plan_writer()
```

type a plan and set the iterable equal to `target`, ex

```python
import bluesky.plans as bp
from bluesky.examples import det

target = bp.count([det])
```

The code is evaluated against the current globals, but the result is shoved into a local dict.


danielballan:
Superseded by work in bluesky-widgets and related repos.

PR #630, opened 2016-12-21
FIX: Block old failures from polluting the event loop like some sort of zombie

danielballan:


danielballan:
FYI, @wen-hu @ambarb @cmazzoli this corrects the bug that you have encountered. It will be deployed before the start of the upcoming cycle.

bluesky/run_engine.py
```diff
@@ -1041,6 +1044,10 @@ def _run(self):
                 except RuntimeError as e:
                     print('The plan {!r} tried to yield a value on close.  '
                           'Please fix your plan.'.format(p))
+            # Some done_callbacks may still be alive in other threads. Mutate
+            # this dict to block them from creating new 'failed status' tasks
+            # on the loop.
+            self._pardon_failures['state'] = True
```
tacaswell: Put this at the top of the finally to be safer?

danielballan: Good idea; thanks.

danielballan:
3.5 tests pass; 3.4 tests are slow / probably hanging on conda.

bluesky/run_engine.py
```diff
@@ -1489,9 +1494,10 @@ def _kickoff(self, msg):
         ret = obj.kickoff(*msg.args, **kwargs)
 
         p_event = asyncio.Event(loop=self.loop)
+        pardon_failures = self._pardon_failures
 
         def done_callback():
-            if not ret.success:
+            if not ret.success and not pardon_failures.is_set():
```
tacaswell: I think this needs to be passed into `call_soon_threadsafe` and this check done in `RunEngine._failed_status` as this callback is run in the pyepics thread.  I think we might want to refactor this to push all of the logic of this callback into a `RunEngine._status_object_completed` callback so that the `p_event.set` call can be run in the event loop.

Tracking all of the status-related tasks that end up on the loop is probably a good idea, right now we are losing track of the tasks to set the `p_event`.

danielballan: That makes sense.

bluesky/run_engine.py
```diff
@@ -1708,7 +1696,15 @@ def _wait(self, msg):
         if futs:
             yield from self._wait_for(Msg('wait_for', None, futs))
 
-    def _failed_status(self, ret):
+    def _status_object_completed(self, ret, p_event, pardon_failures):
+        if not ret.success:
+            task = self.loop.call_soon_threadsafe(self._failed_status,
+                                                  ret, pardon_failures)
+            self._failed_status_tasks.append(task)
+            self.loop.call_soon_threadsafe(p_event.set)
```
tacaswell: this needs to be not indented

danielballan: There it is! I was going crazy looking for this.

tacaswell:
I had in my head the done callback doing 

```python
loop.call_soon_threadsafe(self._complete_status_object, ret, pardon)
```

and the changes that would require.  However I do not see any advantage to that over this (1 less task in the case of failure?) other than 'good code is code _I_ write'.

danielballan:
As written, there are separate tasks for failed status and `p_event.set`. The former is added to `_failed_status_tasks`. Wouldn't refactoring it into one task make it impossible to capture the failed_status_task?

tacaswell:
I think we want to capture _all_ tasks we throw on, doing both the set and the fail in the same task saves book keeping ?

danielballan:
That makes sense. New commit pushed refactors in that way. Now we track `_status_tasks` instead of `_failed_status_tasks`, and the `_failed_status` method can be scrapped, with that logic incorporated directly in `_status_object_completed`.

tacaswell:
:+1: Will merge once it goes green.

PR #631, opened 2017-01-03
DOC: Update release notes.

danielballan:


tacaswell:
Might be better to go to 0.8?  Did change an API and added a significant feature.

tacaswell:
attn @licode

danielballan:
As always, it's a judgment call. Bumping a minor release is always the "safer" choice, so sure. Update pushed.

Issue #634, opened 2017-01-17
Suspender usability improvements


danielballan:
* Give Suspenders a repr that includes, minimally, the signal they are watching (probably the full constructor).
* When suspension is requested, print current time alongside justification.
* When conditions return to nominal, print the current time and also the time that the suspension will be lifted (which is often some time later, after a sleep). This way users know that the suspender correctly recognized the return to nominal conditions and is merely sleeping as instructed.

PR #635, opened 2017-01-17
MNT: Add repr and additional prints to suspenders.

danielballan:
* Nice repr renders `RE.suspenders` easier to understand.
* The message when suspension is requested now includes a time.
* A new message comes out when conditions return to nominal and suspender begins sleeping, giving the time when that particular suspender will release its hold. (Other suspenders may continue to have a hold, and that will be clear from these messages.)

### DIFF
--- a/bluesky/suspenders.py
+++ b/bluesky/suspenders.py
@@ -1,4 +1,5 @@
 import asyncio
+from datetime import datetime, timedelta
 from abc import ABCMeta, abstractmethod, abstractproperty
 import operator
 from threading import Lock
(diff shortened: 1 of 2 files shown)


danielballan:
attn @cmazzoli @ambarb Your feedback on the content of these message would be great.

danielballan:
Closes #634

hhslepicka:
:+1: Sweet!

bluesky/suspenders.py
```diff
@@ -148,6 +158,11 @@ def __set_event(self):
                 loop = self.RE._loop
 
                 def local():
+                    ts = ((datetime.now() + timedelta(seconds=sleep))
+                           .strftime('%Y %m %d %H:%M:%S'))
+                    print("Suspender {!r} reports a return to nominal "
+                          "conditions. Will sleep for {!r} seconds and then "
+                          "release suspension at {!r}".format(self, ts))
```
tacaswell: There are 3 format targets here, but only 2 inputs.  This is hanging because `local` is silently raising on an epics thread and the call-later is not going it.

danielballan: I caught this on the train but didn't have time to push. Sorry for not speaking up, and thanks for looking into it anyway.

PR #639, opened 2017-02-07
FIX: actually wait for the motor to finish in adaptive scan

tacaswell:


danielballan:
Another 3.4 only failure. This failure happens during the conda setup, so again I'm not worried that we have a code issue.

PR #641, opened 2017-02-10
ENH: ReaderWithFilestore now uses its own read

CJ-Wright:


bluesky/examples.py
```diff
@@ -523,7 +523,7 @@ def trigger(self):
         return NullStatus()
 
     def read(self):
-        return self._result
+        return super().read()
```
tacaswell: Just remove this method.

CJ-Wright: That causes read to give back arrays not filestore uids.

danielballan: @CJ-Wright What? I'm not sure how this is different than just removing the method.

danielballan:
bump

CJ-Wright:
Yea, sorry that I haven't fixed this yet (although I guess this might be of interest with the configuration refactor?)

CJ-Wright:
I'm superseding this with #698

PR #643, opened 2017-02-19
TST: Add Python 3.6 to test matrix

stuartcampbell:
If we are considering moving to python 3.6 - should we start testing on it ?


stuartcampbell:
Of course, this won't work yet - need to update the dependencies first!

danielballan:
I'll leave this open so we don't forget to do it (once we're ready). Oh, I can't reopen because the branch is gone. Feel free to re-push.

PR #644, opened 2017-02-21
Fix stopped plans

tacaswell:
attn @gjwillms @ambarb 

The fix was smaller than I expected.


danielballan:
Travis passes 3.5 but fails on 3.4. Probably a fluke, and we're about to drop 3.4 support anyway.

PR #650, opened 2017-02-24
FIX: key sequence number on both stream name and objects read

tacaswell:


danielballan:
3.6 test passed; 3.4 failed on a conda error

PR #654, opened 2017-03-02
IMP to better track delay in count

danielballan:
I'm opening this PR from @cmazzoli's branch so we can have discussion in the PR rather than in the commit.

Delay in count was jiittering due to BS overhead.
Added a simple check in order to minimize the effect if specified delay is at least long enough (in finite plan only for the moment).
Left a comment for modification of delay array generation that I believe would be useful
@danielballan, @tacaswell how does it look like?

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1868,6 +1868,13 @@ def count(detectors, num=1, delay=None, *, md=None):
     _md.update(md or {})
 
     # If delay is a scalar, repeat it forever. If it is an iterable, leave it.
+    
+    # I would anyway repeat the set of delays enough to cover num-1.
+    # For example it is conceivable to provide a subset of delays like [1,2,3] to be used
+    # multiple times over for example num = 30, for statistical purposes and to remove higher order effects.
+    # Probably a declaration of the delay array generation should be provided as a warning at the beginning of the scan
+    # a general question: are messages like the proposed above intercepted in the metadata?
+    
     if not isinstance(delay, Iterable):
         delay = itertools.repeat(delay)
     else:
@@ -1877,6 +1884,7 @@ def count(detectors, num=1, delay=None, *, md=None):
     @run_decorator(md=_md)
     def finite_plan():
         for i in range(num):
+            now = time.time() # placed here to intercept the flow in its earliest moment..
             yield Msg('checkpoint')
             yield from trigger_and_read(detectors)
             try:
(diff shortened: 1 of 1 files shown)


danielballan:
First of all: this is really exciting! It's great to have not one but *two* contributions from BL scientists coming into bluesky this week.

Some notes on process: You created a "commit" on a new "branch" (`improve_delay_in_count`) on your "fork" of bluesky, and you made some comments there. So far so good. Now, for longer discussion and review, it's helpful to open a "pull request," which I have just done on your behalf. During this discussion, if you make further edits on your branch, they will be reflected here. When we are satisfied with the changes, we will merge your branch into the master branch of bluesky.

You left some remarks as code comments: generally it's better to post comments like that here in the pull request where we can discuss. I'll reproduce them here.

> I would anyway repeat the set of delays enough to cover num-1. For example it is conceivable to provide a subset of delays like [1,2,3] to be used multiple times over for example num = 30, for statistical purposes and to remove higher order effects. Probably a declaration of the delay array generation should be provided as a warning at the beginning of the scan a general question: are messages like the proposed above intercepted in the metadata?

danielballan:
You are suggesting alternative ways to handle the case where the user specifies N delay times but asks for M > N + 1 triggers. To review the current behavior:

```python
count([det], num=5, delay=1)  # 5 triggers each with 1-second delay
count([det], num=5, delay=[1,2,4,8])  # 5 triggers with variable delays
count([det], num=5, delay=[1,2,4,8,16,32])  # 5 triggers with variable delays
count([det], num=5, delay=[1,2])  # errors out once it runs out of delays to use
```

You are suggesting that we automatically repeat the delays when we reach the end. I would rather make the user be clear about what they want like this:

```python
count([det], num=5, delay=itertools.repeat([1, 2]))  # repeat entire cycle
```

I think starting from the beginning is just one of several behaviors they might want, and it's not necessarily the most common one.

You also suggest that we should check the length of the delays in advance. This is only sometimes possible -- some iterables -- don't "know" their lengths -- but, yes, we could at least try, like this:

```python
try:
    num_delays = len(delay)
except TypeError:
    # no way to tell if we have enough delays -- we'll just have to find out!
    pass
else:
    if num - 1 > delays:
        raise ValueError("num=%r but delays only provides %r "
                                       "entries" % (num, i))
```

tacaswell:
I am in favor of merging this as-is and moving the implementation of cycling through the delays to another PR.

I would lean towards expecting the users to explicitly say they would like to cycle their delays rather than assuming it.

danielballan:
Agree, that's a separate issue. I'll throw on a commit to scrub out the code comments about cycling and merge as is. Thanks, @cmazzoli!

tacaswell:
:fireworks: :tada: Thank @cmazzoli :tada: :fireworks:

PR #655, opened 2017-03-02
Update plans.py

awalter-bnl:


awalter-bnl:
Sorry forgot to add the reason for this request.

I have modified spiral and fermat_spiral plans to include an optional asymetry parameter, which passed on to the respective plan_patterns. The y axis value is multiplied by this factor to result in an elliptical spiral/ fermat spiral. A pull request for changes to plan_patterns.py is also being submitted

danielballan:
Superceded by #675

PR #656, opened 2017-03-02
Update plan_patterns.py

awalter-bnl:

### DIFF
--- a/bluesky/plan_patterns.py
+++ b/bluesky/plan_patterns.py
@@ -9,7 +9,7 @@ from .utils import snake_cyclers
 
 
 def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
-           tilt=0.0):
+           asym=1, tilt=0.0):
     '''Spiral scan, centered around (x_start, y_start)
 
     Parameters
@@ -28,6 +28,8 @@ def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
         y width of spiral
     dr : float
         Delta radius
+    asym : float
+        asymetry of the scan, giving a y axis 'radius' (dr) = asym*dr while keeping the x axis 'radius' (dr) = dr 
     nth : float
         Number of theta steps
     tilt : float, optional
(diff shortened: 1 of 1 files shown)


danielballan:
Re-running the tests....

tacaswell:
Can we either document this as an aspect ratio or as major/minor axes?

I suspect the generalization we want here is actually the full set of ellipse parameters (5 dof: x-center, y-center, major/minor axis, and a rotation angle).

awalter-bnl:
So the ‘asymetry’ term that I use is the same as the 'aspect ratio’, if it is clearer to label it ‘aspect' this could easily be changed. But in my mind ‘aspect' is defined relative to the x and y axes whereas  ‘asymetry’ is defined relative to some arbitrary axis along which the symmetry is broken. This may just be due to my training in crystallography however !-).

In fact we have all 5 dof’s.

input values related to the ellipse are x_start, y_start, dr, asym and tilt.

These are related to the 5 dof’s specified as

x-center = x_start
y-xenter = y_start
major/minor axis = dr (depending if asym is < or > 1)
minor/major axis = dr* asym (depending if asym is < or > 1)
tilt = rotation angle

The benefit of using asymmetry instead of axis for me is that asymmetry is a more logical “optional” parameter than “major/minor axis” (personal preference). In addition there is no confusion regarding which axis needs to be major and which is minor, as that is defined within the asymmetry term.

Anyway as this is really a philosophical argument about naming conventions I leave the decision up to the DAMA group.

cheers
Andrew


Andrew Walter
Assc. Physicist
Photon Sciences, NSLS 2
Building 741
Upton, New York
USA 11973
phone: office +1 631 344 4055
    beamline +1 631 344 1621
email: awalter@bnl.gov<mailto:awalter@bnl.gov>







On Mar 5, 2017, at 9:28 PM, Thomas A Caswell <notifications@github.com<mailto:notifications@github.com>> wrote:


Can we either document this as an aspect ratio or as major/minor axes?

I suspect the generalization we want here is actually the full set of ellipse parameters (5 dof: x-center, y-center, major/minor axis, and a rotation angle).

danielballan:
Superceded by #675

PR #657, opened 2017-03-02
ENH: Try to check in advance that count 'delay' compatible with 'num'

danielballan:
Carried over from discussion in #654


danielballan:
Thanks for the suggestion, @cmazzoli!

PR #663, opened 2017-03-23
Enh suspenders

tacaswell:
## Description

 - let pre/post plans be callabels
 - never cache suspender related messages
## Motivation and Context
This was found debugging some shutter + suspender related issues at SRX

## How Has This Been Tested?

Needs tests.


danielballan:
Real failures: https://travis-ci.org/NSLS-II/bluesky/jobs/214296569#L1305

danielballan:
In one build, `test_simple_fly` died on a race condition in the test that we should probably fix someday, but that's unrelated to this PR. Merged!

Issue #645, opened 2017-02-24
LiveMesh spot colours not working


awalter-bnl:
So I have found that in LiveMesh the colour of the dots is not varying with the value defined by I.  Also I am not seeing a colour legend like in LiveRaster.

As additions can we add the 'aspect' like in LiveRaster and  'marker' and 'markersize' as in LivePlot

PR #669, opened 2017-03-30
FIx up LiveRaster, LiveMesh

### LINKED ISSUE #667: Fix up LiveMesh and LiveRaster
* Allow all relevant mpl args to pass through (we only support some currently).
* Make orientation uniform.
* More obvious names (not breaking back-compat) like LiveScatter and LiveGrid? I can never keep these straight.

danielballan:
## Description

Rename:
LiveRaster -> LiveGrid
LiveMesh -> LiveScatter

The old names still work but emit warnings.

Extend LiveGrid and LiveScatter signatures to allow more plot customization. Include a colorbar for both.

## Motivation and Context

Closes #667 and #645

## How Has This Been Tested?

These are exercised in the docs build and the unit tests. Parts of this were tested in production at ESM.


bluesky/callbacks/core.py
```diff
@@ -347,17 +354,29 @@ def update(self, x, y, I):
         offsets = np.vstack([self._xdata, self._ydata]).T
         self.sc.set_offsets(offsets)
         self.sc.set_array(np.asarray(self._Idata))
+        if self.clim is None:
+            clim = np.nanmin(self._Idata), np.nanmax(self._Idata)
+            self._norm.vmin, self._norm.vmax = clim
```
tacaswell: us `sc.set_clim(*clim)`  instead of touching the norm.

PR #670, opened 2017-03-30
CI: Require the conda-installed 3.6 test to pass.

danielballan:


danielballan:
For those of you following along at home, we briefly allowed 3.6 conda-installed build to fail since our internal packaging wasn't sorted out. Now it's all good.

PR #672, opened 2017-03-30
REF: Drop boltons dependency by using toolz (already a dep).

danielballan:
## Description

Use `toolz.partition` instead of `boltons.chunked`.

Also, remove doct as a core dep (it's only use in `bluesky.callbacks.broker`) and matplotlib.

## Motivation and Context

We don't need both toolz and boltons, and it seems more likely we'll want more things from toolz in the future. I also have the impression it's more widely used than boltons.

## How Has This Been Tested?

There is good test coverage of the relevant sections of bluesky.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
   - |
     if [ $CONDA_ENV ]; then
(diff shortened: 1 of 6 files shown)


danielballan:
We can file this under, "Times when @klauer Was Out-Voted and Then Later Vindicated". We keep finding these....

PR #674, opened 2017-03-30
API: Do not 'sanitize' numpy arrays into lists.

### LINKED ISSUE #636: ENH: Run engine subscription gives listified arrays
Would it be possible for the run engine to give back the actual arrays rather than their list equivalents? This way db.restream(hdr, fill=True) and the run engine give back the same things.

danielballan:
## Description

Remove `sanitize_np` everywhere it is used in run_engine.py.

## Motivation and Context

We are currently 'sanitizing' numpy arrays returned by devices into lists (or lists of lists, etc.) so that they can go safely into pymongo (which doesn't know what to do this numpy arrays). This should done on the databroker/metadatastore side, not in bluesky, because some callbacks need numpy arrays and it's wasteful to convert to lists and then back to arrays.

**This should not be merged until there is a corresponding PR in databroker providing this functionality there.**

Closes #636 

## How Has This Been Tested?

This will definitely fail tests until databroker is updated.

### DIFF
--- a/bluesky/callbacks/zmq.py
+++ b/bluesky/callbacks/zmq.py
@@ -5,7 +5,7 @@ import os
 import socket
 import time
 from ..run_engine import Dispatcher, DocumentNames
-from ..utils import expiring_function
+from ..utils import apply_to_dict_recursively, sanitize_np
 
 
 class Publisher:
@@ -51,6 +51,7 @@ class Publisher:
         self._subscription_token = RE.subscribe(self)
 
     def __call__(self, name, doc):
+        apply_to_dict_recursively(doc, sanitize_np)
         message = self._fmt_string.format(name=name, doc=doc)
         self._socket.send_string(message)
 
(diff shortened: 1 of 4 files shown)


tacaswell:
Also, should add a test that fails!

danielballan:
Fixed for:

- [x] metadataclient https://github.com/NSLS-II/metadataclient/pull/33
- [x] metadatastore https://github.com/NSLS-II/metadatastore/pull/266
- [ ] portable-mds

danielballan:
> Also, should add a test that fails!

https://github.com/NSLS-II/databroker/pull/195

danielballan:
To check that merging this will not break the databroker, I checkout out this branch of bluesky and the master branch of databroker and I ran the databroker tests. They pass.

@arkilic and I added new bluesky test to make sure that the 0MQ tools are robust against this change as well. If Travis passes, this is good to go.

PR #653, opened 2017-03-01
Add square spiral plan

awalter-bnl:

PR #675, opened 2017-03-30
Spiral aspect

danielballan:
Combines ~~#653,~~ #655, #656. Completes the 'spiral' plans by allowing the user to control all the degrees of freedom.

@awalter-bnl Thanks for this contribution to bluesky. I know we've said it before, but it bears repeating: it's great to see beamline scientists contributing upstream!

I merged your two commits (from your `patch-1` and `patch-2` branches) into this one PR and made the following revisions on top:

* Fixed a place where the 'asym' argument was accepted but then not used.
* Renamed 'asym' to 'aspect_ratio' (as discussed [previously](https://github.com/NSLS-II/bluesky/pull/656#issuecomment-284383082)) because Tom and I agree that that is more clear to general users. In the event of ambiguity, the docstring clarifies the exact meaning.
* Addressed minor style and whitespace issues.

### DIFF
--- a/bluesky/plan_patterns.py
+++ b/bluesky/plan_patterns.py
@@ -13,7 +13,7 @@ from .utils import snake_cyclers
 
 
 def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
-           tilt=0.0):
+           dr_y=None, tilt=0.0):
     '''Spiral scan, centered around (x_start, y_start)
 
     Parameters
@@ -31,7 +31,9 @@ def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
     y_range : float
         y width of spiral
     dr : float
-        Delta radius
+        Delta radius along the minor axis of the ellipse.
+    dr_y : float, optional
+        Delta radius along the major axis of the ellipse, if not specifed defaults to dr
     nth : float
         Number of theta steps
     tilt : float, optional
@@ -41,8 +43,13 @@ def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
     -------
     cyc : cycler
     '''
+    if dr_y is None:
+        dr_aspect = 1
+    else:
+        dr_aspect = dr_y/dr
+    
     half_x = x_range / 2
-    half_y = y_range / 2
+    half_y = y_range / (2 * dr_aspect)
 
     r_max = np.sqrt(half_x ** 2 + half_y ** 2)
     num_ring = 1 + int(r_max / dr)
@@ -57,8 +64,9 @@ def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
         for i_angle in range(int(i_ring * nth)):
             angle = i_angle * angle_step
             x = radius * np.cos(angle)
-            y = radius * np.sin(angle)
-            if ((abs(x - y / tilt_tan) <= half_x) and (abs(y) <= half_y)):
+            y = radius * np.sin(angle) * dr_aspect
+            if ((abs(x - (y/dr_aspect) / tilt_tan) <= half_x) and 
+                    (abs(y/dr_aspect) <= half_y)):
                 x_points.append(x_start + x)
                 y_points.append(y_start + y)
 
@@ -201,7 +209,7 @@ def spiral_square_pattern(x_motor, y_motor, x_center, y_center,
 
 
 def spiral_fermat(x_motor, y_motor, x_start, y_start, x_range, y_range, dr,
-                  factor, *, tilt=0.0):
+                  factor, *, dr_y=None, tilt=0.0):
     '''Absolute fermat spiral scan, centered around (x_start, y_start)
 
     Parameters
@@ -219,7 +227,9 @@ def spiral_fermat(x_motor, y_motor, x_start, y_start, x_range, y_range, dr,
     y_range : float
         y width of spiral
     dr : float
-        delta radius
+        delta radius along the minor axis of the ellipse.
+    dr_y : float, optional
+        Delta radius along the major axis of the ellipse, if not specifed defaults to dr
     factor : float
         radius gets divided by this
     tilt : float, optional
@@ -229,10 +239,15 @@ def spiral_fermat(x_motor, y_motor, x_start, y_start, x_range, y_range, dr,
     -------
     cyc : cycler
     '''
+    if dr_y is None:
+        dr_aspect = 1
+    else:
+        dr_aspect = dr_y / dr
+    
     phi = 137.508 * np.pi / 180.
 
     half_x = x_range / 2
-    half_y = y_range / 2
+    half_y = y_range / (2 * dr_aspect)
     tilt_tan = np.tan(tilt + np.pi / 2.)
 
     x_points, y_points = [], []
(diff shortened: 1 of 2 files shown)


awalter-bnl:
This looks good to me, excited to see it being merged !-)

danielballan:
Since I've got some non-trivial commits in here, would someone else review this and merge?

danielballan:
@awalter-bnl On reflection, Tom and I are confused again about what this is for. It seems like it leaves the spiral specification over-determined. Could you explain again what this parameter controls and how it is different from simply adjusting (the ratio of) x_range and y_range.

awalter-bnl:
@danielballan and @tacaswell .OK so this is a subtle difference I admit, but important from a scanning point of view. x_range and y_range do indeed define a different range of the x_axis and y_axis but not a different step size in these directions, as the step size is defined by delta_r.

Therefore in order to define a different step size in both directions, for instance when the 2 axes have very different units/values, I include the aspect ratio which acts on the relative step size of each axis not the range of each axis. 

So hopefully this makes more sense, the other option of course is to define a delta_rx and delta_ry. But the implementation would be the same replacing "aspect ratio" with "delta_ry/delta_rx"

tacaswell:
So it is tuning the distribution of sample points along the path, not the shape of the path?

awalter-bnl:
This is correct, x_range and y_range change the shape, aspect_ratio changes the step size.

danielballan:
This got hung up on naming-related confusion. This meaning of `aspect_ratio` -- "tuning the distribution of sample points along the path, not the shape of the path" -- is domain-specific. Should we:

* Merge as is
* Qualify the name with another word -- something like "spacing_aspect_ratio"
* Reject this as too domain-specific for bluesky and push it into nsls2utils for an ESM-specific thing

awalter-bnl:
My thoughts are that we merge as is, but am also happy with a name change for aspect_ratio. As suggested above we could define delta_rx and delta_ry instead of delta_r and then substitute 'aspect_ratio' in the plan with 'delta_ry/delta_rx'. Additionally we could have an optionally 'delta_ry' which when used defines delta_r as 'delta_rx'.

Line 2 I think is a bad option, I believe that other beamlines will be scanning over 2 axes that don't have the same units, or where you don't want the same spatial resolution in both directions. In both cases this change would be required so I think this is something that is generally required not just ESM specific.

danielballan:
How about we keep the signature of `spiral` as is but add a separate `asym_spiral` plan with parameters `delta_x` and `delta_y`? That's nice and explicit. The `spiral` plan will become a thin wrapper around `asym_spiral`.

prjemian:
+1

awalter-bnl:
I am happy with this solution, do you want me to make the change (as I made the initial change for the pull request) or will someone from DAMA do this?

danielballan:
We can take this one. Sorry for letting this languish so long, @awalter-bnl.

awalter-bnl:
An additonal thought, for consistency perhaps we want to consider a similar solution with square_spiral ? at present it has x_num and y_num to specify the 2 so perhaps this should become 'asym_square_spiral' and a thin wrapper written for 'square_spiral'.

danielballan:
Sounds good.

awalter-bnl:
OK so there are a few issues to resolve here. 
1. Do we want to write 2 plans (asym_spiral and spiral) or keep everything in one plan (my preference).
2. What kwarg name do we want to use. 
        My suggestion here is that we should use  dr_y=None as the kwarg then if dr_y is None then set   the aspect_ratio in the plan to 1 else set it to dr_y/dr. 

We could also change aspect_ratio in the plan to a different name, but so far I don't know what makes a better name.

Thoughts? comments?

mrakitin:
The title of the PR may be a good candidate - `spiral_aspect` or `spiral_ratio`...

awalter-bnl:
I am not sure any of your options @mrakitin resolve the core issue @tacaswell  has. Which I think is that using 'aspect' or 'ratio' in the keyword argument creates confusion in regard to if this affects the range or the number of points (currently set using dr for 'delta radius'). I think the same is true for the coefficient_of_stretching_y_axis, as this again sounds like it changes the range not the number of points. The only solution that is clear enough to seperate them at present I think is 'dr_y' or perhaps dr_ratio, what does everyone think?

awalter-bnl:
I realize my last suggestion uses 'ratio' in the kwarg but it prefixes it with dr to indicate what it is a ratio of.

awalter-bnl:
@tacaswell and @mrakitin I have made some commits which I think resolve the issues you mentioned. as discussed in the meeting I have changed the kwarg to dr_y, and set it to default to dr if it is not specified. I also fixed the pep8 issues. Can the two of you review this and merge if you are happy.

mrakitin:
@awalter-bnl, please rebase on current master.

awalter-bnl:
I rebased yesterday, will do it again

awalter-bnl:
@mrakitin  it is now rebased

bluesky/plans.py
```diff
@@ -1272,7 +1274,7 @@ def spiral(detectors, x_motor, y_motor, x_start, y_start, x_range, y_range, dr,
                          'x_motor': repr(x_motor), 'y_motor': repr(y_motor),
                          'x_start': x_start, 'y_start': y_start,
                          'x_range': x_range, 'y_range': y_range,
-                         'dr': dr, 'nth': nth, 'tilt': tilt,
+                         'dr': dr,'dr_y',dr_y, 'nth': nth, 'tilt': tilt,
```
mrakitin: Should it be `'dr_y':  'dr_y`? Don't forget about blank spaces.

bluesky/plans.py
```diff
@@ -1242,7 +1242,9 @@ def spiral(detectors, x_motor, y_motor, x_start, y_start, x_range, y_range, dr,
     y_range : float
         y width of spiral
     dr : float
-        Delta radius
+        Delta radius along the minor axis of the ellipse. 
+    dr_y : float,optional
```
mrakitin: Blank space?

bluesky/plan_patterns.py
```diff
@@ -229,10 +239,15 @@ def spiral_fermat(x_motor, y_motor, x_start, y_start, x_range, y_range, dr,
     -------
     cyc : cycler
     '''
+    if dr_y is None:
+        dr_aspect = 1
+    else:
+        dr_aspect=dr_y/dr
```
mrakitin: Blanks spaces? Or some people may better like a one-liner, e.g.:
```py
dr_aspect = 1 if not dr_y else dr_y / dr
```

bluesky/plans.py
```diff
@@ -1313,7 +1317,9 @@ def rel_spiral(detectors, x_motor, y_motor, x_range, y_range, dr, nth,
...
```
mrakitin: Blank space is missed

mrakitin:
Actually, the tests caught this problem: https://travis-ci.org/NSLS-II/bluesky/jobs/376951202.

bluesky/plan_patterns.py
prjemian: Please don't write code like that - it's obfuscated.  Stick to the indented form.  It is very clear.

bluesky/plans.py
awalter-bnl: Thanks @mrakitin, I have made the changes and the Travis build is in progress. Still getting use to the required format !-).

bluesky/plan_patterns.py
mrakitin: Maybe. Either way is fine to me. List comprehensions can be unclear too, but we use them a lot, e.g. in https://github.com/NSLS-II/bluesky/blob/master/bluesky/callbacks/fitting.py. Just a thought. I'm +1 for clearness!

danielballan: It's a fair point. As a data point, I think this specific case is against Google code style (not that we strictly follow that or anything) whereas list comprehensions are fine.

PR #678, opened 2017-04-06
DOC: Add new pages showing RunEngine API docs

danielballan:


tacaswell:
Sorry, I broke this by fixing .travis in #682

danielballan:
@tacaswell Your point (offline I think, or in Slack) about rethinking the Reader and Mover API -- and also reusing more or ophyd -- is well taken. Is it worth documenting the current situation a little better anyway? I would vote "yes" unless we are confident we'll commit to a new API before the next cycle.

tacaswell:
I am against documenting the current example classes better until we sort out what we want to do with them.  End users should not be using them (other than the few instances we provide) and they seem needlessly complicated for what they do.

danielballan:
FWIW I don't feel so negatively about them and I would like to see users developing more interesting simulated hardware. Let's sort out a permanent direction for these soon.

danielballan:
> I am against documenting the current example classes better until we sort out what we want to do with them. End users should not be using them (other than the few instances we provide) and they seem needlessly complicated for what they do.

The current example classes are still not documented. Do you still want to hold off on this, @tacaswell?

CJ-Wright:
What is the status of the simulated `example` classes? We are building some infrastructure with them so it would be nice to know if they are going to still be around and/or what their replacements will be.

danielballan:
I can't envision how we could make Reader or Mover any simpler than they are now, or what improvements would justify an API break. But I'm not sure if Tom's concerns relate to those or the more specialized one-offs like SynGauss.

On Mon, Jul 10, 2017 at 6:00 PM -0400, "Christopher J. Wright" <notifications@github.com<mailto:notifications@github.com>> wrote:


What is the status of the simulated example classes? We are building some infrastructure with them so it would be nice to know if they are going to still be around and/or what their replacements will be.

CJ-Wright:
I'm mostly worried about `ReaderWithFilestore`, although `Mover` and `Reader` are critical too.

tacaswell:
I am still :-1: on documenting the examples and to @CJ-Wright building anything on top of them.

Ophyd now imports with out epics, you should use `Device` and `Signal` to build synthetic devices.

danielballan:
When one sits down the actually write these for testing or simulation, I think the sparseness of Reader(name, {field: func}) is very appealing. Perhaps we can rebuild that API on top of Device. Also remember that ophyd is not available in analysis envs, as a layer of security.

danielballan:
I removed the commits related to `bluesky.examples`, punting that discussion to a separate issue. This PR now only contains RunEngine and Dispatcher API documentation, with associated changes to the docstrings in `run_engine.py`.

bluesky/run_engine.py
```diff
@@ -288,6 +280,52 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
 
         self.loop.call_soon(self._check_for_signals)
 
+    def subscribe(self, name, func):
```
tacaswell: This should be flipped in accoradnce with #691

doc/source/run_engine_api.rst
```diff
@@ -0,0 +1,56 @@
...
+
+.. autoclass:: RunEngine
+
+    .. automethod:: __call__
+
+    The RunEngine maintains a callback registry of functions that receive any
+    :doc:`documents` generated by plan execution. These methods add and remove
+    functions from that registry.
+
+    .. automethod:: subscribe
```
tacaswell: I think using `autosummary` here would be better so each method gets it's own page and these are just tables. 

Not a blocking concern.

danielballan: I gave it a try (in a currently stashed local commit) and I don't like the results as much. I do like autosummary for something like our table built-in plans, where the functions are fairly self-describing and its docstring's top-line sentence is enough to explain its purpose. But for many RunEngine methods, the purpose requires more explanation. It's nice to put more detail up front.

bluesky/run_engine.py
danielballan: Updated and force-pushed.

doc/source/run_engine_api.rst
danielballan: Actually, sleeping on this, maybe tables _are_ better. But it seems that sphinx doesn't give methods separate pages the way it does for functions. Our choices are (1) automethod style or (2) autosummary with no links, and therefore no access to the details like "Examples" or "See also".

tacaswell: There is a way to make it do it, see http://matplotlib.org/api/axes_api.html#basic

bluesky/run_engine.py
```diff
@@ -2011,13 +2067,17 @@ def subscribe(self, func, name='all'):
         func: callable
             expecting signature like ``f(name, document)``
             where name is a string and document is a dict
-        name: {'start', 'descriptor', 'event', 'stop', 'all'}, optional
-            Defaults to 'all'
+        name: string
+            one of {'start', 'descriptor', 'event', 'stop', 'all'}
```
tacaswell: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt see section 4 the convention is to list
the valid values (and the default should go first).

doc/source/run_engine_api.rst
danielballan: OK, I'm up for that, but this feels like the beginning of a three-hour fight with sphinx that I don't have time for before Chicago. Punting to https://github.com/NSLS-II/bluesky/issues/701

bluesky/run_engine.py
danielballan: Added new commit to address

PR #679, opened 2017-04-06
TST: do not skip ophyd tests if libca is not installed

tacaswell:
This relies on a not-yet-merged change to ophyd


danielballan:
This ophyd change that @tacaswell referred to above is sneaked into the docs update, https://github.com/NSLS-II/ophyd/pull/396, and it has been merged.

PR #681, opened 2017-04-13
FIX: make sure motors are in the right order

### LINKED ISSUE #680: outer_product scan list attributes "motors" and "extents" are not ordered in the same way.
In outer_product scan the metadata lists 'start.motors' and 'start.extents' are not ordered the same so that the 'ith' item in start.motors does not match the 'ith' item in start.extents. This makes scaling of these multi-dimensional arrays difficult.

tacaswell:
closes #680

Fix bug where the motors in inner and outer scans were randomized

## Description

Added motors in order when we add extents friends to the start document

## Motivation and Context

See #680 

## How Has This Been Tested?

New tests added


tacaswell:
attn @awalter-bnl

awalter-bnl:
Great, thanks for the quick fix.

PR #682, opened 2017-04-13
CI: try to fix docs building

tacaswell:


tacaswell:
Self-merging to un-break all the other PRs.

Issue #677, opened 2017-03-31
Plans for future of "spec-like" API


danielballan:
This issue documents our intentions for unifying some features of the "spec-like" API with the "standard" built-in plans and any user-defined plans. The motivation of this change is to avoid two disjoint forks of plans development.

For background, see [differences from non-spec-like plans](http://nsls-ii.github.io/bluesky/plans.html#differences-from-non-spec-like-plans).

## 1. Smarter Dispatcher

Motivation: Make it easier to apply nice subscriptions (e.g. plots) to all plans where they apply

Enhancement: Refactor the dependency-injection scheme in `bluesky.plans.spec_api` into a  "smart" Dispatcher-like object that can apply to *all* plans (including built-in standard plans, user-defined plans, and spec-like plans) auto-subscribing relevant subscriptions based on the contents of the RunStart document and/or EventDescriptors.

Planned Release: v0.9.0 (next release)

```python
# Currrent
RE(bluesky.spec_api.ct(...))  # provides table
RE(bluesky.plans.count(...))  # provides nothing

# Planned
RE(bluesky.spec_api.ct(...))  # provides table
RE(bluesky.plans.count(...))  # provides nothing
D = bluesky.default_dispatcher()
RE.subscribe('all', D)
RE(bluesky.plans.count(...))  # now provides table
```

`D` will contain logic for mapping metadata to appropriate subscriptions, such as 1D plots for 1D scans and various 2D plots for various types of 2D scans. It will provide methods to configure its behavior interactively, including an 'off' switch. It will absorb many of the attributes that we currently hang off of `gs`, such as `PLOT_Y` and `TABLE_COLS` (maybe). It does _not_ include `gs.DETS`. See below for the fate of `gs.DETS`.

## 2. Consistent Signature and Deprecating Global State Detectors

Motivation: In an early version, we introduced the global detectors list `gs.DETS` as a way to synchronize plans and automatic subscriptions. Now that plans report richer metadata, it's not necessary, and the role of `gs.DETS` can be replaced by a normal user-defined variable. This simplifies the bluesky codebase and the usage, and it allows users to switch between multiple different lists easily.

API break: Require the detector argument for spec-like plans.

Planned Release: v0.10.0

```python
# Current
gs.DETS = [detector1, detector2]
ct()  # implicitly uses gs.DETS for detectors

# Planned
d = [detector1, detector2]
ct(d)  # where `d` is a list of detectors, akin to how some users use `gs.DETS` now
[...]

PR #683, opened 2017-05-02
Replace dependency injection scheme with "smart dispatcher"

danielballan:
See #677 for motivation. This addresses Step 1 of that.

I have tried to compensate for the down-to-the-wire timing of this PR with thorough testing -- more detailed testing that the callbacks are actually configured the way we expect than we have had in any previous iteration of bluesky.

We could certainly still make the SmartDispathcer smarter, but for now I have constrained the scope to replicating the current behavior in a new API. This new API should make it easier to experiment with smarter and perhaps BL-specific heuristics in user space.


bluesky/callbacks/core.py
```diff
@@ -1,7 +1,8 @@
 """
 Useful callbacks for the Run Engine
 """
-from itertools import count
```
danielballan: changed because the word 'count' usually means `bluesky.plans.count` in this codebase

bluesky/callbacks/core.py
```diff
@@ -862,3 +873,166 @@ def descriptor(self, doc):
...
+            if num is None or num > 1:
+                sniff_dim = 1
+        self.__sniff_dim = sniff_dim
+
+        if self.__sniff_dim == 1 and self.x is not None:
+            for y_key in self.ys:
+                fig_name = self._figure_name(
+                    'BlueSky: {} v {}'.format(y_key, self.x))
+                ax = plt.figure(fig_name).gca()
+                self.active_cbs.append(LivePlot(y=y_key, x=self.x, ax=ax))
```
tacaswell: This probably needs a cut-out for count / ct to always plot against time or seq_num

bluesky/callbacks/core.py
```diff
@@ -862,3 +873,166 @@ def descriptor(self, doc):
...
+        # callbacks will be.
+
+        self.clean()
+        self.__start_doc = doc
+        if self.muted:
+            return
+
+        # We can always set up some table.
+        motors = doc.get('motors', [])
+        table = LiveTable(list(motors) + list(self.ys) + list(self.cols))
```
tacaswell: does this assume that the motors names are the same as their primary readout?  I do not think this was an issue before because we had access to the actual motor objects.

bluesky/callbacks/core.py
tacaswell: This probably needs to be deferred until we see the descriptor like `PeakStats`

tacaswell:
I think the injector code should stay, even if we do not use it for now.

This is not really a one-for-one change and gives up a fair amount of re-usability.

`SmartDispatcher` seems too generic.

danielballan:
Given our discussion and the potentially sprawling scope of providing visualization hints in plan metadata, I think this will not be ready in time for tomorrow. I think we can send this to the floor in a safe micro release in a month or two.

danielballan:
I learned some things from this attempt, but I'm going to start from scratch. Better to prototype this at an actual beamline (as I have begun doing at SIX).

PR #685, opened 2017-06-19
WIP: live image callback (with slider, thanks browser!)

CJ-Wright:
Add a callback with a live image viewer (with a slider)

## Description
Add live image viewing to the plethora of callbacks

## Motivation and Context
It might be nice to be able to see the images as we take them.

## How Has This Been Tested?
It hasn't, yet. This is more of a testing the waters/proof of concept kind of thing (especially since this may induce the movement of various chunks of the browser to someplace else).

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -8,6 +8,7 @@ import time as ttime
 
 import matplotlib.pyplot as plt
 import matplotlib.colors as mcolors
+from databroker_browser.qt._cross_section_2d import StackViewer, CrossSection
 from datetime import datetime
 import numpy as np
 import logging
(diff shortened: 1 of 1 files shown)


CJ-Wright:
@danielballan thoughts?

tacaswell:
:-1: to the `qt_browser` dependancy

How is this different than https://github.com/NSLS-II/bluesky/blob/master/bluesky/callbacks/broker.py#L9 ?

CJ-Wright:
It may be that the `qt_browser` stuff needs to move someplace else, I think the main difference is that this has a slider bar for sifting through the images.

danielballan:
I'm glad you're interested in this problem. I think it would be fine to vendor the slider bar code (placed in such a way that it only puts an _optional_ dependency on qt5).

Either way, this should go in `callbacks.broker` because, once done correctly, it will require a databroker dep in order to get access to the image data.

CJ-Wright:
Hmm, this is interesting, would it be possible to separate out the filestore call?
I'd like to use callbacks for data processing, but will have the data in memory during this so I don't really need it.

PR #686, opened 2017-06-27
ENH: check if events are filled before asking filestore for the data

CJ-Wright:
Check if event doc is filled before trying to fill it in `LiveImage` callback

## Motivation and Context
Data analysis pipelines will hand down filled documents, this change allows data to be poured into this callback

### DIFF
--- a/bluesky/callbacks/broker.py
+++ b/bluesky/callbacks/broker.py
@@ -30,10 +30,12 @@ class LiveImage(CallbackBase):
         self.fs = fs
 
     def event(self, doc):
-        uid = doc['data'][self.field]
         if 'filled' not in doc.keys() or \
                         doc['filled'].get(self.field, False) is False:
+            uid = doc['data'][self.field]
             data = self.fs.retrieve(uid)
+        else:
+            data = doc['data'][self.field]
         self.update(data)
         super().event(doc)
 
--- a/bluesky/callbacks/broker.py
+++ b/bluesky/callbacks/broker.py
@@ -44,7 +44,7 @@ class LiveImage(CallbackBase):
         self.fs = fs
 
     def event(self, doc):
-        if doc.get('filled', {}).get(self.field):
+        if doc.get('filled', {}).get(self.field) and self.fs is not None:
             uid = doc['data'][self.field]
             data = self.fs.retrieve(uid)
         else:
--- a/bluesky/callbacks/broker.py
+++ b/bluesky/callbacks/broker.py
@@ -15,6 +15,18 @@ class LiveImage(CallbackBase):
     ----------
     field : string
         name of data field in an Event
+    fs: FileStore instance
+        The FileStore instance to pull the data from
+    cmap : str,  colormap, or None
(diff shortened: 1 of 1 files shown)


CJ-Wright:
@tacaswell @arkilic @danielballan ready for review. (the travis failure is coming from something other than this PR)

bluesky/callbacks/broker.py
```diff
@@ -15,21 +16,28 @@ class LiveImage(CallbackBase):
...
+                 limit_func, auto_redraw, interpolation)
         self.cs._fig.show()
         if fs is None:
             import filestore.api as fs
         self.fs = fs
 
     def event(self, doc):
-        uid = doc['data'][self.field]
-        data = self.fs.retrieve(uid)
+        if 'filled' not in doc.keys() or \
```
danielballan: Nik-pick this can be simplified to `if doc.get('filled', {}).get(self.field)` because `None` is false-y.

bluesky/callbacks/broker.py
```diff
@@ -15,21 +16,28 @@ class LiveImage(CallbackBase):
     field : string
         name of data field in an Event
```
danielballan: Can the docstring be updated to match the signature?

CJ-Wright: Can I just put **kwargs, and note that they are all kwargs to mpl imshow? (note that I will need to update the cross section viewer to take all the mpl stuff

tacaswell: no because we do not pass all of them through.

CJ-Wright: I was planning on editing it so that it did.

danielballan: Good idea! Go for it.

CJ-Wright:
@danielballan @tacaswell ready for more review

bluesky/callbacks/broker.py
```diff
@@ -14,22 +15,40 @@ class LiveImage(CallbackBase):
...
         if fs is None:
             import filestore.api as fs
         self.fs = fs
 
     def event(self, doc):
-        uid = doc['data'][self.field]
-        data = self.fs.retrieve(uid)
+        if doc.get('filled', {}).get(self.field):
+            uid = doc['data'][self.field]
+            data = self.fs.retrieve(uid)
```
tacaswell: check that `fs` is not `None`?

CJ-Wright: Can `self.fs` be none? We import it if it is none

CJ-Wright: https://github.com/NSLS-II/bluesky/pull/686/files#diff-55f2206880210fd5f4542f0cefe0eae2R43

CJ-Wright:
@tacaswell ready for more review, don't know where that error is coming from.

tacaswell:
Both of those fails are from somewhat flaky tests.

CJ-Wright:
Thanks

PR #687, opened 2017-06-30
Slider callback

CJ-Wright:
## Description
Add an image viewer with a slider.

## Motivation and Context
Now users can look through their stack of images as they are taking them

### DIFF
--- a/bluesky/callbacks/broker.py
+++ b/bluesky/callbacks/broker.py
@@ -30,10 +30,12 @@ class LiveImage(CallbackBase):
         self.fs = fs
 
     def event(self, doc):
-        uid = doc['data'][self.field]
         if 'filled' not in doc.keys() or \
                         doc['filled'].get(self.field, False) is False:
+            uid = doc['data'][self.field]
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@tacaswell @danielballan Is it possible to add a check button to toggle whether or not the gui updates to the latest image?

danielballan:
Sure! Looks like you're already working on that.

CJ-Wright:
Yes, and it caused a lot of problems, so I removed it pending guidance.

danielballan:
There's a reason I haven't done it yet. :- ) It's possible but it's tricky to get right.

CJ-Wright:
I'd also like to get some guidance on how to not have to keep replacing the slider every time I add an image.

CJ-Wright:
@tacaswell This was the issue I was mentioning the other day and you said I didn't have enough calls to draw idle.

CJ-Wright:
This will work nicely as a subclass of `Filler`.

PR #689, opened 2017-07-12
API: Rework 0MQ Proxy

danielballan:
- Change the API of the Publisher and the RemoteDispatcher.
- Add a Python API for starting a 0MQ proxy (formerly called 'forwarder').
- Add a CLI, bluesky-0MQ-proxy, for starting a 0MQ proxy.


danielballan:
More functionality, only one additional line of code. :- )

I'm excited about this one.

danielballan:
1/4 builds failed randomly on a ramp_plan-related test. Re-running....

bluesky/callbacks/zmq.py
```diff
@@ -0,0 +1,238 @@
...
+            return ((hostname is None or hostname == _hostname)
+                    and (pid is None or pid == _pid)
+                    and (run_engine_id is None or run_engine_id ==
+                         run_engine_id))
+        self._is_our_message = is_our_message
+
+        super().__init__()
+
+    @asyncio.coroutine
+    def _poll(self):
```
danielballan: These lines are marked as not covered by Coverage because they are executed in a subprocess. I burned two hours trying to write a separate test running them in the main process, and gave up because of bad interactions between pytest and zmq.asyncio's event loop. We do test that these lines actually execute the callback registry, so I'm comfortable leaving it alone.

bluesky/callbacks/zmq.py
```diff
@@ -0,0 +1,238 @@
...
+
+    Example
+    -------
+
+    Publish to an existing Proxy at some known address.
+
+    >>> RE = RunEngine({})
+    >>> publisher = Publisher(RE, ('localhost', 5567))
+    """
+    def __init__(self, RE, address):
```
tacaswell: Using a pattern like https://github.com/NSLS-II/GeRM/blob/master/pygerm/zmq.py#L51 can decouple this from which event loop it runs on.

danielballan: Is f67fe6f what you had in mind? I admit I'm not sure why this would be used, but I'm happy to provide the hook.

tacaswell: Yes.  It is so that you can use this with curio, asyncio, or threads (as they all integrate with zmq differently).

bluesky/callbacks/zmq.py
```diff
@@ -0,0 +1,257 @@
...
+    Publish to an existing Proxy at some known address.
+
+    >>> RE = RunEngine({})
+    >>> publisher = Publisher(RE, ('localhost', 5567))
+    """
+    def __init__(self, RE, address, *, zmq=None):
+        if zmq is None:
+            import zmq
+        if isinstance(address, str):
+            address = address.split(':')
```
tacaswell: I am feeling paranoid today, can you also include `maxsplit=1`?

bluesky/callbacks/zmq.py
```diff
@@ -0,0 +1,257 @@
...
+        self.address = (address[0], int(address[1]))
+        self.RE = RE
+        self.hostname = socket.gethostname()
+        self.pid = os.getpid()
+        url = "tcp://%s:%d" % self.address
```
tacaswell: who needs 3.5, f-strings!!!!

bluesky/callbacks/zmq.py
```diff
@@ -0,0 +1,257 @@
...
+from ..utils import expiring_function
+
+
+class Publisher:
+    """
+    A callback that publishes documents to a 0MQ proxy.
+
+    Parameters
+    ----------
+    RE : ``bluesky.RunEngine``
```
tacaswell: Is there a way we can drop this?

I don't think so because it needs to know how to un-subscribe it's self

bluesky/callbacks/zmq.py
```diff
@@ -0,0 +1,257 @@
...
+    @asyncio.coroutine
+    def _poll(self):
+        while True:
+            message = yield from self._socket.recv()
+            hostname, pid, RE_id, name, doc = message.decode().split(' ', 4)
+            if self._is_our_message(hostname, pid, RE_id):
+                doc = ast.literal_eval(doc)
+                self._loop.call_soon(self.process, DocumentNames[name], doc)
+
+    def start(self):
```
tacaswell: Slight preference for the pattern we are using in the caproto servers of these object exposing a `async def run` and have `stop` poison that loop.

doc/source/callbacks.rst
```diff
@@ -909,119 +909,64 @@ a separate process.
...
-running in the background as a service. Here is how to start one just for play:
+Alternatively, you can start the proxy using a Python API:
 
-.. code-block:: bash
+.. code-block:: python
 
-    # uses config in /etc/zmq_forwarder_device.yml
-    #  or ~/.config/zmq_forwarder_device/connection.yml
-    $ python bluesky/examples/forwarder_device.py
+    from bluesky.callbacks.zmq import Proxy
```
tacaswell: This is a pretty compelling reason not use the `async def run` pattern, but if someone gets here (rather than using the proxy cli tool above they are down the rabbit hole :dragon: .

bluesky/callbacks/zmq.py
danielballan: Correct.

bluesky/callbacks/zmq.py
danielballan: Seems like you countermanded this below, right?

tacaswell: Everything is trade offs!

PR #690, opened 2017-07-13
FIX: Compatibility with numpy 1.13

danielballan:
Subtracting booleans (via numpy.diff) raises an error in numpy 1.13:

https://docs.scipy.org/doc/numpy/release.html#deprecationwarning-to-error


stuartcampbell:
The fix looks sensible to me.

danielballan:
The build with latest numpy passed this time. There was a random failure (`test_prompt_stop`) on a different build which I am ignoring.

PR #691, opened 2017-07-13
API: Support RE.subscribe(func).

danielballan:
*The original usage, RE.subscribe(name, func), is supported with a warning.*

All tests pass, unchanged.

This is a case of the adage, "You design the most important part of your
software at the beginning, when you understand it the least." The switch on
'name' is rarely used.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1991,7 +1991,7 @@ class Dispatcher:
                  "set RunEngine.ignore_callback_exceptions = False "
                  "and run again." % (exc, name.name))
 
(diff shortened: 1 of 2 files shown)


tacaswell:
Do the test now generate oodles of warnings?

This is going to require a walk through all of the BL configs again correct?

danielballan:
Yes, the tests generate warnings. Will fix in a separate PR; it seemed worth showing as plainly as possible that they pass unchanged.

Yeah, we'll need to fix the BL configuration to remove these warnings and many others that have been there awhile now.

PR #695, opened 2017-07-13
ENH: Add DiagnosticPreprocessor.

danielballan:


danielballan:
This is intended to surface the monitoring, flying, and baseline features in a way that is more accessibly (you don't have to understand preprocessors...) but not tied to SPEC-like plans. The docstring is a bit verbose, in that it basically explains some standard features of Python lists, but some conversations with BL staff have illustrated that this is useful to include.

danielballan:
Address one aspect of https://github.com/NSLS-II/bluesky/issues/677

bluesky/plans.py
```diff
@@ -2998,6 +2998,123 @@ def polling_plan():
...
+    >>> D.baseline.append(some_detector)
+    """
+    def __init__(self, *, baseline=None, monitors=None, flyers=None):
+        if baseline is None:
+            baseline = []
+        if monitors is None:
+            monitors = []
+        if flyers is None:
+            flyers = []
+        self.baseline = baseline
```
tacaswell: Explicitly list-ify here so input of generators and tuples does the right thing.

tacaswell:
We may want to give a bit more thought to tooling around the preprocessor list, but that is out of scope for this PR.

bluesky/plans.py
danielballan: Done and force-pushed.

stuartcampbell:
Yeah, I agree about the tooling and some user facing helpers...

PR #696, opened 2017-07-14
TST: Let test_timeout take longer and decrease chance of random failure.

danielballan:


danielballan:
This test seems to fail about 1/4 of the time. Finally got around to adjusting it. It is still rigorous: it should not be possible for it to randomly _pass_ -- only less likely to randomly fail.

PR #698, opened 2017-07-18
ENH: ReaderWithFilestore uses its own read

CJ-Wright:
This way `trigger_read` can be overridden to deal with more complex readers
eg readers with shutters

## Motivation and Context
This allows for a more flexible setup, where the data can be modified from the `Reader` after the read but before the data is stashed into FS

## How Has This Been Tested?
I ran the existing tests.

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -501,7 +501,7 @@ class ReaderWithFileStore(Reader):
     def trigger(self):
         # save file stash file name
         self._result.clear()
-        for idx, (name, reading) in enumerate(super().read().items()):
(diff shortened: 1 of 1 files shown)


CJ-Wright:
@tacaswell ready for review.

CJ-Wright:
@brookeferber This was the PR I said I'd put in.

tacaswell:
I still have grave reservations about you extending these classes.

CJ-Wright:
I guess my current feeling is that if/when you get rid of bluesky examples I will move the code base to whatever setup you suggest.

PR #702, opened 2017-07-25
CI: Use doctr to publish docs.

danielballan:
As we already did on NSLS-II/docs and NSLS-II/caproto,
use doctr to publish documentation (instead of my homemade
scheme).


danielballan:
Self-merging this to get the docs building again. They haven't been building for awhile....

I hoped to only touch CI, but I had to touch documentation and docstrings too. But no library code was touched.

danielballan:
One last note for anyone following along: from now on, if the docs building/publishing system fails, Travis-CI should fail too and alert us.

PR #705, opened 2017-07-26
DOC: Draft a comparison with SPEC with @prjemian.

danielballan:
Distilling some thoughts that often come up verbally at tutorials
into a new page of the documentation.

### DIFF
--- a/doc/source/comparison-with-spec.rst
+++ b/doc/source/comparison-with-spec.rst
@@ -15,7 +15,8 @@ bluesky, including:
 * Easy hardware configuration
 * Interruption handling (Ctrl+C)
 * Integration with EPICS (and potentially other instrument control systems)
-* "Pseudomotors" and integration with reciprocal space transformation code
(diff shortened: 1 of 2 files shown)


danielballan:
@dchabot -- You can a good slide on topic this back in the day. Any comments?

danielballan:
Are there important differences in our handling of pseudomotors worth highlighting here? Or should I just mention it as a good thing about SPEC that we have tried to retain?

Good thought about a link to the data model.

prjemian:
Good catch about the event model.  Very important new design component.
Allows for collection of asynchronous, uncoordinated data.

danielballan:
I think this is good to go. I'll let @stuartcampbell pull the trigger.

danielballan:
Disregard last comment. I forgot I had solicited some feedback that just came in, which I should now incorporate.

danielballan:
OK, I think this is good to go.

PR #709, opened 2017-07-30
DOC: Document suggestions for defining IPython magics for bluesky.

danielballan:
Since last fall, we have considered introducing an IPython magic that would work save users from typing `RE()`. I first show normal usage for comparison.

```python
In  [1]: 1 + 1
Out [1]: 2

In [2]: RE(count(d))
Out [2]: ['some uid']

In [3]: %REmode
# You are now in RunEngine mode! Type plans to execute them or Ctrl+D to exit to normal IPython.

RE [3]: count(d)
Out [3]: ['some_uid']

RE [4]: 1 + 1  # raises an error -- only plans are allowed in RE mode

^D
# You are now in normal IPython.
In [5]:
```

I have been pushing discussion on this as much as anyone, but I am still lukewarm on this particular compromise. It makes key features of bluesky less accessible, and it doesn't give users much in return.  Specifically, because every input in 'RunEngine mode' is expected to be a valid plan, there isn't a natural way to capture a uid in a variable:

```python
uids = RE(count(d))  # capture the uids in a variable
```

or provide metadata in line:

```python
RE(count(d), sample='A')
```

It also places `print_summary`, `check_limits`, and other plan tools further out of reach. (You have to leave 'RE mode' to use them.)

Even having sacrificed these things, the approach doesn't actually go very far toward solving the problem it is trying to solve: providing a terse command line interface optimized for interactive beamline debugging. In the context of driving sophisticated data acquisition routines, users generally agree that the complexity of bluesky is justified and that the rich syntax of Python is helpful. In a debugging context -- doing thing like quickly moving a motor or taking a one-off reading -- validation, rich metadata, and in general the full power of Python are less compelling.

From the developer's perspective, we aspire to improve software culture in synchrotron science, which we believe to be necessary and inevitable in a future where experiments become more complex. As stated in #705, integrating with modern hardware and modern software ecosystems requires a general-purpose language that is necessarily less terse than what served in the past. Hacking the language to make it seem simpler than it is -- or making one tool look like another -- is a recipe for confusion.

From a user's perspective, if I were configuring a beamline, even with my deep understanding of bluesky and facility with Python, I can imagine getting tripped up on brackets in:

```python
d = [det]
RE(count(d), LiveTable(['det']))
```

Our plans for the next release of bluesky (see #677) will help that, reducing common cases to:

```python
d = [det]
RE(count(d))
```

Even so, I can imagine myself defining IPython magics for common beamline debugging tasks, such as `%ct` or `%mov motor 5`. This kind of hack is exactly why IPython is so valuable as a practical tool. It's not rich enough to build custom plans, but it's useful for common simple cases that might occur when my mind is focused on hardware problems. The documentation in this PR is an attempt to guide users to use IPython magics effectively and sparingly -- and not get tempted to invent a custom language.

This approach would answer a need that we have never answered to the satisfaction of our users. Additionally, it's a viable replacement for ophyd's "command API", which was un-documented a long time ago but is still frequently used at some beamlines.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -47,6 +47,7 @@ Index
    async
    debugging
    run_engine_api
+   magics
 
 .. toctree::
    :caption: Developer Documentation
--- a/doc/source/magics.rst
+++ b/doc/source/magics.rst
@@ -1,0 +1,205 @@
+*******************************
+Writing Custom IPython 'Magics'
+*******************************
+
+This section is not about bluesky itself; it highlights a feature of
+IPython.
+
+Bluesky is designed to be usable in an interactive session and also as a
+library for building higher-level tools, such as a Graphical User Interface.
+Presently, there is no officially-supported GUI for bluesky. (We may provide
+tools for building bluesky GUIs in the future.) There is, however, a way to
+build a terse command-line interface on top of bluesky using a feature of
+IPython.
+
+IPython is an interactive Python interpreter designed for and by scientists. It
+includes a feature called "magics" --- convenience commands that aren't part of
+the Python language itself. For example, ``%history`` is a magic:
+
+.. ipython:: python
+
+    a = 1
+    b = 2
+    %history
+
+The IPython documentation documents the
+`complete list of built-in magics <https://ipython.readthedocs.io/en/stable/interactive/magics.html>`_
+and, further,
+`how to define custom magics <https://ipython.readthedocs.io/en/stable/config/custommagics.html>`_.
+
+What follows are some examples for defining IPython magics for bluesky and some
+suggestions on using them effectively. We emphasize that these examples are not
+part of the official bluesky interface; they merely illustrate a reasonable
+pattern for user customization.
+
+Suppose you have imported the some plans, defined a RunEngine ``RE``, and
+created a list of detectors ``d``.
+
+.. code-block:: python
+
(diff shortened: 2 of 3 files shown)


prjemian:
I agree with the notion that user-defined magics are likely to result in more productive work at the beam line and less time spent by developers maintaining an optimized RunEngine interface that provides limited access to the Python language.

The challenge to overcome is that the magics are restricted to the ipython shell.  Full Bluesky syntax would be needed in Python scripts that call the RunEngine.

tacaswell:
Another concern from talking to @cmazzoli today is that the context switch from `RE(count(d))` to

```python
def over_night():
    yield from count(d)
```

is significant.  Adding a third (restricted) way of driving the beamline may not help end-users.

danielballan:
This may be more useful for debugging the beamline than making things easy on users. In the PR I go into some depth musing about what this kind of customization would be good for.

danielballan:
@prjemian You can actually access magics from scripts. You just have to execute them with IPython, like `ipython -i my_script.ipy`

danielballan:
FYI, we have pivoted from documenting that this _can_ be done to actually shipping some working magics in the package itself, primarily so that we can justify removing the problematic 'command' module from ophyd -- replacing that functionality with equivalent magics.

The prose in this PR has been updated and folded into #720, which adds the magics too.

PR #719, opened 2017-08-03
Fill callback

### LINKED ISSUE #717: ENH: Broker Callback base class
As we add more callback tools which require filled documents it may be nice to have a base class which figures that out.

## Expected Behavior
Callback class has filling machinery (actually we need to standardize that first since the two filled callbacks have different syntax)

CJ-Wright:
Create base class for dealing with the databroker in callbacks.
Closes #717

### DIFF
--- a/bluesky/callbacks/broker.py
+++ b/bluesky/callbacks/broker.py
@@ -7,7 +7,33 @@ import doct
 import matplotlib.pyplot as plt
 
 
-class LiveImage(CallbackBase):
+class BrokerCallbackBase(CallbackBase):
+    """
+    Base class for callbacks which need filled documents
+
+    Parameters
+    ----------
+    field: str
+        name of data field in an Event
+    fs: FileStore instance
+        The FileStore instance to pull the data from
+    """
+    def __init__(self, field, *, fs=None):
+        if fs is None:
+            import filestore.api as fs
+        self.fs = fs
+        self.field = field
+
+    def event(self, doc):
+        if doc.get('filled', {}).get(self.field):
+            data = doc['data'][self.field]
+        else:
+            uid = doc['data'][self.field]
+            data = self.fs.retrieve(uid)
+        return data
+
+
+class LiveImage(BrokerCallbackBase):
     """
     Stream 2D images in a cross-section viewer.
 
@@ -33,24 +59,15 @@ class LiveImage(CallbackBase):
                  limit_func=None, auto_redraw=True, interpolation=None):
         from xray_vision.backend.mpl.cross_section_2d import CrossSection
         import matplotlib.pyplot as plt
-        super().__init__()
-        self.field = field
+        super().__init__(field, fs=fs)
         fig = plt.figure()
         self.cs = CrossSection(fig, cmap, norm,
                  limit_func, auto_redraw, interpolation)
         self.cs._fig.show()
-        if fs is None:
-            import filestore.api as fs
-        self.fs = fs
 
     def event(self, doc):
-        if doc.get('filled', {}).get(self.field):
-            data = doc['data'][self.field]
-        else:
-            uid = doc['data'][self.field]
-            data = self.fs.retrieve(uid)
+        data = super().event(doc)
         self.update(data)
-        super().event(doc)
 
     def update(self, data):
         self.cs.update_image(data)
@@ -169,7 +186,7 @@ def verify_files_saved(name, doc, db):
         print('\x1b[1A\u2713')  # print a checkmark on the previous line
 
 
-class LiveTiffExporter(CallbackBase):
+class LiveTiffExporter(BrokerCallbackBase):
     """
     Save TIFF files.
 
@@ -184,19 +201,20 @@ class LiveTiffExporter(CallbackBase):
         the attributes of 'start', 'event', and (for image stacks) 'i',
         a sequential number.
         e.g., "dir/scan{start[scan_id]}_by_{start[experimenter]}_{i}.tiff"
-    db : Broker
-        The databroker instance to use
     dryrun : bool
         default to False; if True, do not write any files
     overwrite : bool
         default to False, raising an OSError if file exists
+    db : Broker, optional
+        The databroker instance to use, if not provided use databroker
+        singleton
 
     Attributes
     ----------
     filenames : list of filenames written in ongoing or most recent run
     """
 
-    def __init__(self, field, template, db, dryrun=False, overwrite=False):
+    def __init__(self, field, template, dryrun=False, overwrite=False,):
         try:
             import tifffile
         except ImportError:
@@ -208,11 +226,6 @@ class LiveTiffExporter(CallbackBase):
         else:
             # stash a reference so the module is accessible in self._save_image
             self._tifffile = tifffile
-
-        if db is None:
-            from databroker import DataBroker as db
-
-        self.db = db
 
         self.field = field
         self.template = template
@@ -240,8 +253,8 @@ class LiveTiffExporter(CallbackBase):
     def event(self, doc):
         if self.field not in doc['data']:
             return
-        self.db.fill_event(doc)  # modifies in place
-        image = np.asarray(doc['data'][self.field])
+        data = super().event(doc)
+        image = np.asarray(data)
         if image.ndim == 2:
             filename = self.template.format(start=self._start, event=doc)
             self._save_image(image, filename)
--- a/bluesky/callbacks/broker.py
+++ b/bluesky/callbacks/broker.py
@@ -44,11 +44,11 @@ class LiveImage(CallbackBase):
         self.fs = fs
 
     def event(self, doc):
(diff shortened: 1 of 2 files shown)


CJ-Wright:
Failure is on unrelated test

danielballan:
This is a very useful component to add to databroker. I would suggest a different interface. Callbacks, like `event`, aren't expected to return data. You can do something like this:

```python
class BrokerCallbackBase:
    def __init__(self, fields):
        # fields can be a *list* of things to fill, not just one field

    def event(self, doc):
        # mutate doc *in place*, filling the fields specified at init
        # do not return anything
```

Then subclasses like LiveImage can do:

```python
class LiveImage(BrokerCallbackBase):
    def __init__(self, field):
        super().__init__([field])

    def event(self, doc):
        super().event(doc)  # fills in place
        # do stuff
```

Mutating the documents in place used to be scary because of the possibility that callbacks would mutually interfere, but that is not as an issue since we added `filled`.

CJ-Wright:
@danielballan can you please assign this to me so I can keep track of it?

danielballan:
Done.

bluesky/callbacks/broker.py
```diff
@@ -7,7 +7,34 @@
...
+
+    Parameters
+    ----------
+    field: str
+        name of data field in an Event
+    fs: FileStore instance
+        The FileStore instance to pull the data from
+    """
+    def __init__(self, fields, *, fs=None):
+        if not isinstance(fields, (tuple, list)):
```
CJ-Wright: This could be much more elegant I know that it has to be something we can loop through, but we can loop through strings now :frowning_face:

CJ-Wright:
I might put this up in a different issue and I may have advocated for this in the past, we may consider cleaving callbacks from bluesky in the general sense. Callbacks are useful outside the context of data acquisition (I personally feel that any data analysis system down stream should support them).

danielballan:
Sure. It's been under discussion since the very beginning. Start an issue!

bluesky/callbacks/broker.py
danielballan: I would rather insist that fields be a list. The plural argument name (and of course the docstring) can make that clear.

bluesky/callbacks/broker.py
```diff
@@ -33,24 +60,15 @@ def __init__(self, field, *, fs=None, cmap=None, norm=None,
...
-            import filestore.api as fs
-        self.fs = fs
 
     def event(self, doc):
-        if doc.get('filled', {}).get(self.field) and self.fs is not None:
-            uid = doc['data'][self.field]
-            data = self.fs.retrieve(uid)
-        else:
-            data = doc['data'][self.field]
+        data = super().event(doc)
```
danielballan: Is this part still a work in progress?

CJ-Wright: Sorry I got so upset by the lackluster conditional in the base class I completely overlooked that.

bluesky/callbacks/broker.py
```diff
@@ -7,7 +7,34 @@
...
+            import filestore.api as fs
+        self.fs = fs
+        self.fields = fields
+
+    def event(self, doc):
+        for field in self.fields:
+            if doc.get('filled', {}).get(field):
+                pass
+            else:
+                doc['data'].update(self.fs.retrieve(doc['data'][field]))
```
danielballan: We should really give you a nice way to do this with a Broker instance. Users shouldn't have to juggle a filestore instance, especially since there might be multiple different ones needed to fill an event. Track https://github.com/NSLS-II/databroker/issues/213

CJ-Wright: That would be nice (now all the `don't use filestore` warnings can stop!)

bluesky/callbacks/broker.py
```diff
@@ -33,24 +58,16 @@ def __init__(self, field, *, fs=None, cmap=None, norm=None,
                  limit_func=None, auto_redraw=True, interpolation=None):
         from xray_vision.backend.mpl.cross_section_2d import CrossSection
         import matplotlib.pyplot as plt
-        super().__init__()
-        self.field = field
```
danielballan: I think it's useful for `LiveImage` to know its own field. I think it is clearer than referring to `self.fields[0]` of its parent.

bluesky/callbacks/broker.py
```diff
@@ -7,7 +7,32 @@
...
+    Parameters
+    ----------
+    fields: Iterable of str
+        Names of data field in an Event
+    fs: FileStore instance
+        The FileStore instance to pull the data from
+    """
+    def __init__(self, fields, *, fs=None):
+        if fs is None:
+            import filestore.api as fs
```
danielballan: This is going to break in the upcoming release.

danielballan: (So let's just require `fs` explicitly. It's simpler anyway.)

CJ-Wright: Is the new broker partial fill not coming down in the next release?

CJ-Wright: I think that it might be good to leave filestore optional. We only require it if the document is to be filled. The subclassed callbacks can have either pre-loaded data (coming from data analysis) or stored data (from the RunEngine) in the case of data analysis there isn't a concept of a filestore.

We may want to further separate this down the line, eg have a two node pipeline: the first loads the data in from wherever (if needed) and the second performs the callback. This way the callbacks are completely blind to the source of the data, it is always loaded by the time it gets to them.

bluesky/callbacks/broker.py
```diff
@@ -27,6 +28,9 @@ def event(self, doc):
             if doc.get('filled', {}).get(field):
                 pass
             else:
+                if self.fs is None:
+                    raise RuntimeError('Either the data must be pre-loaded or'
```
danielballan: I don't like that signature suggests that this parameter is optional (`fs=None`) but that it errors out if it's not provided. Either the parameter should be truly optional, or the signature should make it clear that the parameter is required.

CJ-Wright: Note that this only errors if we need to go looking for a filestore resource (it is not filled) and don't have a filestore instance, which I agree is an undue burden on the users.
I don't see a good way out of this one. Our options (as I see it):
1. Assume that the events are never filled and require a FileStore instance (which breaks if any of the events are filled as it tries to use a numpy array as a key, eg any `hdr.stream(fill=True)`)
1. Assume that the events are always filled and don't require a FileStore instance (which breaks all the existing callbacks at the beamlines as there is no way to fill the events before they reach this callback (at least without some sort of pipeline before the callback))
1. Assume neither and figure it out when we get the events, which may or may not require a FileStore instance. This is as implemented.
1. Split the codebase into two sets of callbacks, one that operates on filled events, the other which does not (which I imagine would become a nightmare to maintain).

Personally I like option 2. as it separates all the database/retrieval part of the code from the data visualization code, but I'm not certain that it is a good idea to break most of the callbacks as they are currently implemented at the beamlines.

bluesky/callbacks/broker.py
```diff
@@ -7,7 +7,34 @@
...
+
+    def event(self, doc):
+        for field in self.fields:
+            if doc.get('filled', {}).get(field):
+                pass
+            else:
+                if self.db is None:
+                    raise RuntimeError('Either the data must be pre-loaded or'
+                                       'a Broker instance must be provided.')
+                doc['data'][field] = self.db.fill_events([doc], field=field)
```
danielballan: `self.db.fill_events` is a generator. I think you need to do `doc['data'][field], = self.db.fill_events([doc], field=field)` (note trailing comma on left side). If I'm right, then we should be concerned that the test is passing now!

CJ-Wright: I think this was pulling from an old databroker?

danielballan:
Test fails on `ImportError: cannot import name 'ReaderWithFileStore'`

CJ-Wright:
Not it should be failing on the key filled.
```python
KeyError: 'filled'

../../../databroker/databroker/broker.py:1230: KeyError
```

CJ-Wright:
Not it should be failing on the key filled.
```python
KeyError: 'filled'

../../../databroker/databroker/broker.py:1230: KeyError
```

danielballan:
I'm not sure what you meant there. Missing a word?

CJ-Wright:
https://travis-ci.org/CJ-Wright/bluesky/jobs/264866515#L1205

danielballan:
Can you summarize? Do you have a question?

CJ-Wright:
Summary: It seems to be failing on the databroker side. 
Questoin: Are you planning on fixing it there or should I?

CJ-Wright:
I am on board with this

danielballan:
The pip-based build passes. We know the conda-based ones will fail until databroker packaging is fixed. Merging.

danielballan:
Thanks for surfing the wave of databroker internal changes to make this work, @CJ-Wright!

PR #720, opened 2017-08-03
WIP: Build BestEffortCallback.

### LINKED ISSUE #677: Plans for future of "spec-like" API

### LINKED ISSUE #712: Add 'hints' key to plan metadata
All built-in plans will get a new metadata key like:

```python
def my_scan(detectors, motor, ...):
    # Just for a dumb example, this plan specifies that the first
    # two detectors should be plotted against the motor.
    # We are generating something in the spirit of the Vega Lite spec.
    # It can't be literal Vega Lite because it needs additional info
    # from the Event Descriptors to be complete it, so it's sort of a
    # Vega Lite template.
    md = {...,
          'hints': {'vis': [{'version': 0,
                            'data': 'primary',   # the name of the stream the data will come from
                            'mark': 'point',
                            'x': motor.name,
                            'y': [d.name for d in detectors[:2]]}]}
         }
```

This tells us which *devices* might be useful in a plot. The devices themselves add a `hints()` method -- see https://github.com/NSLS-II/ophyd/issues/413 -- that put into the Event Descriptors which *fields* a.k.a "data keys" from those devices should be used, such as whether we should plot against the motor readback or the motor setpoint.

These are all just hints designed to provide nice defaults for plotting tools, hence the name 'hints', designed so that we don't have to do complex guessing on the other end.

danielballan:
This closes #677 and also closes #712.

TO DO:

- [x] Add hints to all built-in plans (where applicable). Closes #712.
- [x] Incorporate PeakStats into the BestEffortCallback.
- [x] Plot monitored signals with the BestEffortCallback.
- [x] Plot 2D data with the BestEffortCallback.
- [x] Delete `global_state.py` and write detailed release notes on migration.
- [x] Delete `spec_api.py` and write detailed release notes on migration.
- [x] Write IPython magics `%ct`, `%ascan`, etc. as experimental shortcuts.
- [ ] Add hints to ophyd devices in a companion PR (not required for merging this one -- it will just make this one *better* when it's in).


bluesky/magics.py
```diff
@@ -0,0 +1,213 @@
...
+        if len(line.split()) % 2 != 0:
+            raise TypeError("Wrong parameters. Expected: "
+                            "%mov motor position (or several pairs like that)")
+        ns = Namespace(self.shell.user_ns)
+        RE = ns['RE']
+        args = []
+        strs = []
+        for motor, pos in partition(2, line.split()):
+            args.extend([ns[motor], le(pos)])
+            strs.extend([motor, repr(le(pos))])
```
tacaswell: do not use repr, just use original string againg

doc/source/api_changes.rst
```diff
@@ -4,9 +4,181 @@ Release Notes
...
+
+        RE(ct())  # monitoring, flying, and baseline readings are added
+
+     And formerly, those settings only affected the behavior of the "SPEC-like"
+     plans, such as ``ct`` and ``ascan``. They were ignored by their
+     counterparts ``count`` and ``scan``, as well as user-defined plans. This
+     was not desirable!
+
+     This scheme has been replaced by the
+     :ref:`diagnostic preprocessor <diagnostic_preprocessor>`, which can be
```
tacaswell: :bike: :houses: 'diagnostic preprocessor' sounds like a tool you reach for when things are broken, not day-to-day operations.

 - ~housekeeping~
 - standard
 - background
 - bookkeeping
 - moosefish
 - actuarial
 - status
 - health
 - checkup
 - tuneup

danielballan: That name change would touch a lot of things unrelated to these release notes, so let's duke that out in a separate PR.

bluesky/magics.py
```diff
@@ -0,0 +1,213 @@
...
+        print('---> RE(mvr({}))'.format(', '.join(strs)))
+        plan = bp.mvr(*args)
+        return RE(plan)
+
+@magics_class
+class SPECMagics(Magics):
+
+    @line_magic
+    def ct(self, line):
+        if line.split():
```
danielballan: Yes. Agreed `strip` is clearer. Changed.

bluesky/plans.py
```diff
@@ -2090,18 +2173,19 @@ def scan(detectors, motor, start, stop, num, *, per_step=None, md=None):
...
+           'num_intervals': num - 1,
+           'plan_args': {'detectors': list(map(repr, detectors)), 'num': num,
+                         'motor': repr(motor),
+                         'start': start, 'stop': stop,
+                         'per_step': repr(per_step)},
+           'plan_name': 'scan',
+           'plan_pattern': 'linspace',
+           'plan_pattern_module': 'numpy',
+           'plan_pattern_args': dict(start=start, stop=stop, num=num),
+           'hints': {'dimensions': [('primary', motor.hints()['fields'])]},
```
danielballan: Yes. Good catch.

Issue #714, opened 2017-08-02
Revisiting the idea of a progress bar


danielballan:
Originally in https://github.com/NSLS-II/bluesky/pull/525 we started adding code to the `_wait` and `_sleep` coroutines to add progress bars. Instead it would be nice to do this in a pluggable way.

I propose exposing a method that returns the set of status objects that the RunEngine is currently waiting on. Then, it could be possible to add a periodic callback to the event loop that checks this list and shows a progress bar when there are status objects.

The status objects could also become richer and expose "heartbeat", "fraction completed" and "estimate time remaining." They would be optional but likely implemented in that order (of increasing complexity). The progress bar could be made richer depending on what's available. Even the current status objects could be used to make progress bars that display the fact that we are waiting on hardware, which devices we are waiting on, and how long we have been waiting.

This is a simple change to bluesky to provide the API, a week-long project to write the progress bar (or integrate something like tqdm), and probably a week-long project to make status objects smarter.

danielballan:
Actually, I had a better idea over the weekend: instead of asking user-defined progress bars to _poll_ some RunEngine method, the RunEngine can push to alert the user that it is waiting. It's more efficient, and it follows a pattern we already use (for `msg_hook` and `state_hook`). See #729.

danielballan:
Track https://github.com/NSLS-II/ophyd/issues/419

danielballan:
The hooks to support this are going into this release, but we won't be shipping a ready-to-use progress bar implementation in this release.

danielballan:
I'm moving the milestone back to 0.10. I think we should try to squeeze a minimal version of this into this release to validate https://github.com/NSLS-II/ophyd/pull/420. We should also incorporate it into the `%mov` magic. The function that `%mov` is supposed to be replacing (`ophyd.commands.mov`) provided a rudimentary progress bar, and I think users would not appreciate that regression.

PR #729, opened 2017-08-07
ENH: Add waiting_hook to support progress bars.

danielballan:


danielballan:
This addresses the only part of #714 that requires changes to the RunEngine. Everything else can be done in user space.

Lame demo:

```python
from bluesky.examples import motor

def progress_bar(payload):
    if payload is None:
        print("DONE!")
    else:
        print("waiting for...")
        for status in payload:
            print(vars(status))

RE = RunEngine({})
RE.waiting_hook = progress_bar
RE([Msg('set', motor, 5, group='A'), Msg('wait', group='A')])
```

Output:

```Editing... done. Executing edited code...
waiting for...
{'_cb': <function RunEngine._set.<locals>.done_callback at 0x10c539ae8>, 'done': True, 'success': True}
DONE!
```

Real Status objects in ophyd (as opposed to the dummy ones in bluesky) provide more useful information, such as which Device they pertain to, and can be used to generate more useful info. Getting an _actual_ progress bar complete with an ETA will require additions to the Status objects in ophyd.

PR #730, opened 2017-08-07
API: Rename plan_tools

### LINKED ISSUE #706: Write a plan tool that checks motor limits
Suggested by @ambarb. This is something SPEC does that we could easily do. Something like:

```python
class OutofLimits(Exception):
    ...

def check_limits(plan):
    ignore = []
    for msg in plan:
        if msg.command == 'set'
            if msg.obj in ignore:
                continue  # we have already warned about this device
            try:
                low, high = msg.obj.low_limit, msg.obj.high_limit
            except AttributeError:
                warn("Limits of {} are unknown and can't be checked.".format(msg.obj.name))
                ignore.append(msg.obj)
                continue
            if low == high:
                warn("Limits are not set on {}".format(msg.obj.name))
                ignore.append(msg.obj)
                continue
            pos, = msg.obj.args
            if not (low < pos < high):
                raise OutOfLimits("This plan would put {} at {} "
                                  "which is outside of its limits, {}."
                                  "".format(msg.obj.name, pos, (low, high)))
```

### LINKED ISSUE #721: Revisit plan_tools
The module `bluesky.plan_tools` contains `print_summary` and `plot_raster_path`. We were never really happy with the name (what's a "plan tool"?).

We should think of this module as a group of functions that *comsume* plans -- RunEngine stand-ins -- with the promise that nothing in this module actually touches hardware. They are allowed to access `.name`, `.root`, `.parent`, and `.describe()`, but not `.read()` or above all `.set()`. The name `bluesky.sims` for "simulators" seems reasonable. We can shim `bluesky.plan_tools` as an alias for awhile to avoid an API break.

Thoughts?

danielballan:
[Ugh -- my local github CLI is confused and originally published a description for an unrelated PR here. Sorry for the noise.]

This moves plan_tools -> simulators, replaces it with a shim, and adds a test to ensure the old imports work.

~~DO NOT MERGE -- pending more discussion on what to name the functions themselves.~~

Closes #721 
Closes #706


danielballan:
For people following via email: my local CLI got confused and re-published a description for an unrelated PR. See GitHub for correct PR title and description.

danielballan:
Once hints are merged (#720) we can make `plot_raster_path` smarter and N-dimensional.

PR #731, opened 2017-08-07
Wip persistify

tacaswell:
This needs some docs and clean up.

Maybe worth hacking a bit so it introspects better in the RE.

### DIFF
--- a/bluesky/plan_tools.py
+++ b/bluesky/plan_tools.py
@@ -105,3 +105,13 @@ def print_summary_wrapper(plan):
             print('  Read {}'.format(read_cache))
             read_cache = []
         yield msg
+
+
+class persistify:
+    def __init__(self, plan_func, *args, **kwargs):
+        self._f = plan_func
+        self._args = args
+        self._kwargs = kwargs
+
+    def __iter__(self):
+        return (yield from self._f(*self._args, **self._kwargs))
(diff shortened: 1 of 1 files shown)


danielballan:
Why not just advertise `functools.partial` and avoid inventing another name? A `partial` object has pretty good introspection:

```python
In [12]: partial(count, [det])
Out[12]: functools.partial(<function count at 0x1088fcae8>, [SynGauss(name=det)])

In [13]: partial(count, detectors=[det], num=3)
Out[13]: functools.partial(<function count at 0x1088fcae8>, detectors=[SynGauss(name=det)], num=3)
```

tacaswell:
Can you replace the values bound into `partial`

doing `__call__` as

```python
def __call__(self, **kwargs):
    kw = dict(self._kwargs)
    kw.update(kwrargs)
    return (yield from self._func(**kw))
```

might be useful?

Some people might also love if it did

```python
def __call__(self, **kwargs):
    self._kwargs.update(kwargs)

    return (yield from self._func(**self._kwargs))
```

danielballan:
I see the case for that but I'm personally very excited by it. It could be hard to trace back the state of each parameter through time.

Is there still active interest in this PR?

tacaswell:
`partial` is better.

PR #733, opened 2017-08-08
TST: change example code to match new Asset naming

tacaswell:
## Description

Changes to Bluesky due to changes in Databroker
## Motivation and Context

We are re-working how the internal of `Registry` (nee `FileStore`) objects work.  The bluesky examples will need to keep up.

## How Has This Been Tested?

The working branch on databroker has been changed to install from this branch which will take care of testing the interaction.

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -470,22 +470,22 @@ class ReaderWithRegistry(Reader):
     loop : asyncio.EventLoop, optional
         used for ``subscribe`` updates; uses ``asyncio.get_event_loop()`` if
         unspecified
-    fs : Registry
+    reg : Registry
         Registry object that supports inserting resource and datum documents
     save_path : str, optional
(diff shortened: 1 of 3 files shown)


bluesky/examples.py
```diff
@@ -470,8 +470,8 @@ class ReaderWithFileStore(Reader):
     loop : asyncio.EventLoop, optional
         used for ``subscribe`` updates; uses ``asyncio.get_event_loop()`` if
         unspecified
-    fs : FileStore
-        FileStore object that supports inserting resource and datum documents
+    fs : Registry
```
danielballan: fs -> registry?

tacaswell: That one is much riskier to `sed`

danielballan: But not too hard to do by hand!

Issue #710, opened 2017-08-02
Sanitize numpy array to Python list for zmq Publisher


danielballan:
Sanitize numpy array to Python list for zmq Publisher so that it is JSON seralizable, since this will be removed at the RunEngine level.

danielballan:
Closed by #674

PR #735, opened 2017-08-09
Sanitize all docs

arkilic:
Recursively sanitizes documents that are emitted by the RunEngine from np types
## Description

Basically added two functions, one sanitizes the docs from np.generic and np.ndarray and another one that applies this(or any other) function to a doc recursively.

## Motivation and Context

See https://github.com/NSLS-II/bluesky/issues/710

## How Has This Been Tested?

test_utils/test_illegal_np emulates what happened at HXN the other day.

More testing recommendations are welcome.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1230,7 +1230,7 @@ class RunEngine:
         # Use copy for some reasonable (admittedly not total) protection
         # against users mutating the md with their validator.
         self.md_validator(dict(md))
-
+        apply_to_dict_recursively(d=dict(md), f=sanitize_np)
         doc = dict(uid=self._run_start_uid, time=ttime.time(), **md)
         yield from self.emit(DocumentNames.start, doc)
         self.log.debug("Emitted RunStart (uid=%r)", doc['uid'])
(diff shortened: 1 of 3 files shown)


tacaswell:
I thought we were going push all of the sanitization to the ingest side and let the RE emit numpy arrays and such?  

#710 is to put this sanitazition into the ZMQ publisher.

danielballan:
Right. This would undo #674. The hope was to make #674 safe to merge by sanitizing on the way _in_ to the Publisher, as you have already sanitized on the way in to metadatastores.

stuartcampbell:
Looks like there's been a little misunderstanding here.  Am I correct in saying that the numpy 'sanitation' will happen for the ZMQ publisher (plus any other future publishers of a similar ilk) and the databroker 'side' ? 

Do we have an issue on the databroker side ?

tacaswell:
We had https://github.com/NSLS-II/databroker/pull/195 which had some bugs and needed a rebase.  I needed the same code for the https://github.com/NSLS-II/databroker/pull/216 so cherry-picked the commits.

danielballan:
@arkilic and I superceded this with a new commit in #674.

PR #739, opened 2017-08-11
WIP: Make Publisher usable without a RunEngine

### LINKED ISSUE #738: Stream to ZMQ from databroker
Would it be possible to stream data from the databroker though ZMQ? The current callback requires a RE instance.

CJ-Wright:
Closes #738


CJ-Wright:
@danielballan I am getting a very interesting error for this so if you could look at it when you get a chance I'd be grateful.

danielballan:
Debugging inside a subprocess is hard. Open three terminals; run each component in a separate terminal, and see what happens.

danielballan:
Commits ported to #746.

PR #746, opened 2017-08-14
ENH: Make Publisher usable without a RunEngine (2)

danielballan:
This adds some commits to #739 by @CJ-Wright.

CJ, I didn't manage to figure out where your tests failures are coming from,
but your test was more complex than it strictly needed to be to test this
change. In an effort to get this API change in for you before the v0.10.0 tag,
I'm scaling back the ambition of the test.


CJ-Wright:
Ok

PR #748, opened 2017-08-15
ENH: Include 'filled' key in events.

danielballan:
This is the final step in making a RunEngine and a Broker
(configured with a suitable prepare_hook) return identical
documents. It should remove some annoying edge cases for
consumers that want to receive from both.

This also includes some commits to update bluesky to the new
databroker API.


danielballan:
The pip-based build passed. The conda-based builds failed because they got an old version of databroker.

danielballan:
Our dev conda builder seems to be 19 hours behind, hence the test failures on the conda build.

PR #757, opened 2017-08-16
TST: destroy databroker on finish

CJ-Wright:
(for those of us with small memory/paranoia)

### DIFF
--- a/bluesky/tests/conftest.py
+++ b/bluesky/tests/conftest.py
@@ -37,6 +37,6 @@ def motor_det(request):
 def db(request):
     """Return a data broker
     """
-    from databroker.tests.utils import build_pymongo_backed_broker
-    db = build_pymongo_backed_broker(request)
+    from databroker.tests.utils import build_sqlite_backed_broker
+    db = build_sqlite_backed_broker(request)
     return db
--- a/bluesky/tests/conftest.py
+++ b/bluesky/tests/conftest.py
@@ -37,17 +37,6 @@ def motor_det(request):
 def db(request):
     """Return a data broker
     """
-    from databroker import temp_config, Broker
-    db = Broker.from_config(temp_config())
-    fs = db.fs
-    mds = db.mds
-    fs_test_conf = fs.config
-    md_test_conf = mds.config
-
-    def delete_fs_mds():
-        print("DROPPING DB")
-        fs._connection.drop_database(fs_test_conf['database'])
-        mds._connection.drop_database(md_test_conf['database'])
-
-    request.addfinalizer(delete_fs_mds)
+    from databroker.tests.utils import build_pymongo_backed_broker
+    db = build_pymongo_backed_broker(request)
     return db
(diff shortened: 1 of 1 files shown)


tacaswell:
I don't see how this is different.

CJ-Wright:
I don't think that currently there is a cleanup step on the databroker so it could persist on disk

danielballan:
I would like to make it possible to run the bluesky tests without running a mongo server, and this would be a step backward in that respect.

tacaswell:
Ah, that makes sense.  Can you add a finalizer instead that removes the files?  I agree with @danielballan running these tests without mongo is a major :+1:

CJ-Wright:
Ok, will get in tomorrow.

CJ-Wright:
@tacaswell can you please assign me?

CJ-Wright:
@danielballan we could do a try accept which checks if we can open up a mongo server, and if not falls back on sqlite.

danielballan:
@CJ-Wright That sounds like a recipe for confusion and complexity. We can do all the bluesky testing we need to do against sqlite. Many of the databroker tests use a real RunEngine and test it against all the backends. There is no need for bluesky to test that integration as well.

bluesky/tests/conftest.py
```diff
@@ -36,5 +37,6 @@ def motor_det(request):
 def db(request):
     """Return a data broker
     """
-    from databroker import temp_config, Broker
-    return Broker.from_config(temp_config())
+    from databroker.tests.utils import build_pymongo_backed_broker
+    db = build_pymongo_backed_broker(request)
```
tacaswell: Another option is to switch this to the sqlite backend factory which already has the clean up attached to it.

CJ-Wright: That is my plan now that people have asked for sqlite

tacaswell:
Tests will be fixed by https://github.com/NSLS-II/databroker/pull/278

tacaswell:
@CJ-Wright can I force-push to this branch?

CJ-Wright:
@tacaswell of course

CJ-Wright:
cycle for ci

tacaswell:
It looks like the release version of databroker has a missing ujson dep.  We should just hold off on merging this until we tag the next databroker.

CJ-Wright:
Are you going to tag databroker and then bluesky? (Is this getting in this cycle?)

CJ-Wright:
Has there been any progress on this? The lack of test file removal almost crashed my machine.

danielballan:
For simplicit, can `build_sqlite_backed_broker` use the new `Broker.from_config(temp_config())`? It can of course also include cleanup.

CJ-Wright:
Sure works for me

danielballan:
@tacaswell?

danielballan:
See https://github.com/NSLS-II/databroker/pull/294

PR #760, opened 2017-08-16
API/DOC: Rethink magics.

### LINKED ISSUE #756: Change %ct magic and remove other magics
Change `%ct` to use a private RunEngine that does not subscribe to the databroker. 

The golden rule is that magics do not write data that will be kept.

We should therefore also remove the magics that are doing real data collection.

danielballan:
Meetings with beamline staff raised concerns that IPython magics for
data-taking plans like '%ascan' were more confusing than helpful.

On the other hand, magics like '%wa' and '%mov' seem harmless.
'%wa' just reads hardware and prints to the screen. '%mov' touches
hardware, but it does not generate any documents.

We realized that the important distinction is that these magics
do not save data. The ones that save data ('%ascan', '%mesh', etc.)
are more troubling because they do not combine well and could easily
lead users astray. This PR removes them.

We have kept '%ct' but changed what it does: instead of executing
`RE(count(dets))` with the user's RunEngine, it uses a private
RunEngine that is not subscribed to the databroker; thus it does not
save data. It just prints results to the screen --- along with a
message reminding users that nothing has been saved. This seems like
a useful tool for quick diagnostics by beamline staff.

We all had some reservations about deploying this as it was, and
we generally feel better about this way of delineating good and bad
uses of magics in bluesky.

If anyone is still excited about '%ascan', we can reopen the
discussion for a future release, once we see how these are received.
And of course, clever users can always write these for themselves or
lift them from bluesky source history.

[Edit: Closes #756]

### DIFF
--- a/bluesky/magics.py
+++ b/bluesky/magics.py
@@ -1,14 +1,14 @@
-# These are experimental IPython magics for executing a plan using a syntax
-# familiar to users of SPEC.
+# These are experimental IPython magics, providing quick shortcuts for simple
+# tasks. None of these save any data.
 
 # To use, run this in an IPython shell:
 # ip = get_ipython()
 # ip.register_magics(BlueskyMagics)
-# ip.register_magics(SPECMagics)
 
 from IPython.core.magic import Magics, magics_class, line_magic
 from ast import literal_eval as le
 import bluesky.plans as bp
+from bluesky import RunEngine
 import numpy as np
 from operator import attrgetter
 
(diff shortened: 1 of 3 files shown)


bluesky/magics.py
```diff
@@ -19,162 +19,50 @@
...
-            return self.ns[key]
-        except KeyError:
-            raise KeyError("No variable named {!r} is defined in the "
-                           "user namespace. Define it and try again."
-                           "".format(key))
-
 @magics_class
 class BlueskyMagics(Magics):
 
+    RE = RunEngine({})
```
tacaswell: Probably better to pass it a new loop?  Do not want this to conflict with the main RE at all.

tacaswell:
I am liking these a lot more.

Should probably install them by default.

bluesky/magics.py
danielballan: Good idea. Fixed in new commit.

danielballan:
2/4 builds pased; 2/4 failed on `test_broker_base`, an unrelated issue.

Issue #734, opened 2017-08-09
Make it possible to enable and disable BestEffortCallback more granularly


danielballan:
A follow-up issue on #720 that needs more consideration.

(for @hhslepicka who, though he abandoned us for California, is never far from our hearts)

PR #765, opened 2017-08-17
ENH: Add 'epoch' option to LivePlot.

### LINKED ISSUE #759: Consider adding absolute time as an option for LivePlot
Consider adding `x='wallclock'` to the `LivePlot` command.

danielballan:
Closes #759

LivePlot now supports plotting against `event['time']`. By
default, it sets t=0 to the beginning of the run rather than
the UNIX epoch. This PR provides an optional to turn off
that nicety.

This approach --- adding a separate flag --- avoids squatting any names that
might show up in an event's data keys, and it also opens the possibility of
accepting other epochs or a floating point number.

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -91,7 +91,6 @@ class LivePlot(CallbackBase):
     ----------
     y : str
         the name of a data field in an Event
-
     x : str, optional
         the name of a data field in an Event, or 'seq_num' or 'time'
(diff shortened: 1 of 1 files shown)


tacaswell:
Should we wait to merge this on threading the config through BEC?

danielballan:
I'm still not sure how to thread this kind of thing through BEC without adding a bunch of specially-named methods. Let's do that in a separate PR that also addresses #734.

PR #773, opened 2017-08-18
ENH: Add a progress bar capability

### LINKED ISSUE #714: Revisiting the idea of a progress bar

danielballan:
Works with https://github.com/NSLS-II/ophyd/pull/420

Looks like:
```
mtr2  9%|███▊                                       | 0.09/1.0 [00:00<00:01,  1.21s/deg]
mtr1100%|████████████████████████████████████████████| 1.0/1.0 [00:01<00:00,  1.12s/deg]
```

Closes #714

### DIFF
--- a/bluesky/examples.py
+++ b/bluesky/examples.py
@@ -17,12 +17,19 @@ class SimpleStatus:
     It is "simple" because it does not support a timeout or a settling time.
     """
 
-    def __init__(self, *, done=False, success=False):
+    def __init__(self, *, done=False, success=False, device=None):
         super().__init__()
         self._lock = RLock()
         self._cb = None
         self.done = done
         self.success = success
+        self.device = device
+        self._watchers = []
+
+    def watch(self, func):
+        if self.device is not None:
+            self._watchers.append(func)
+            func(name=str(self.device.name))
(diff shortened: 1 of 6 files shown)


danielballan:
Travis-CI coverage is not good, but there is an interactive test script for humans to run.

PR #775, opened 2017-08-18
FIX: don't actually get the data just make sure the key is there

CJ-Wright:
Previously the field was gotten, now we just make certain the key exists. The filled part assures us that the data is loaded.

## How Has This Been Tested?
Ran on my computer.

@tacaswell @danielballan @stuartcampbell Urgent Fix

### DIFF
--- a/bluesky/callbacks/broker.py
+++ b/bluesky/callbacks/broker.py
@@ -36,7 +36,7 @@ class BrokerCallbackBase(CallbackBase):
     def event(self, doc):
         # the subset of self.fields that are (1) in the doc and (2) unfilled
         fields = [field for field in self.fields
-                  if (doc['data'].get(field) and
+                  if (field in doc['data'] and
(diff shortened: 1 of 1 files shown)


CJ-Wright:
Note without this it fails to work, as the truth value of an array is ill defined

bluesky/callbacks/broker.py
```diff
@@ -36,7 +36,7 @@ def descriptor(self, doc):
     def event(self, doc):
         # the subset of self.fields that are (1) in the doc and (2) unfilled
         fields = [field for field in self.fields
-                  if (doc['data'].get(field) and
```
tacaswell: Do we even need the first check anymore?

danielballan: Does `fill_events` just ignore fields that an event doesn't have? Anyway, this seems safer, cleaner.

PR #784, opened 2017-08-21
TST: Test Ctrl+C and magics.

danielballan:


danielballan:
Addresses the bad test coverage of #783 and adds an unrelated commit to make progress bars play nicer with py.test.

PR #788, opened 2017-08-22
FIX: RunEngine returns uids as immutable tuple

CJ-Wright:
Test has two calls to RE but they both put out the same uid. 
@danielballan this was the minimal test you were asking for, sorry for the delay.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -648,7 +648,7 @@ class RunEngine:
             if raise_if_interrupted and self._interrupted:
                 raise RunEngineInterrupted("RunEngine was interrupted.")
 
-        return self._run_start_uids
+        return tuple(self._run_start_uids)
 
     def resume(self):
         """Resume a paused plan from the last checkpoint.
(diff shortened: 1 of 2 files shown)


danielballan:
About the bug:

This was a good puzzle. The uids were in fact different, but the list they were returned in (`RunEngine._run_start_uids`) was being re-used between runs. This always worked as expected:

```python
uid1, = RE(count([]))
uid2, = RE(count([]))
assert uid1 != uid2
```

but (until the fix in this PR) this failed:
```python
uids1 = RE(count([]))
uids1 = RE(count([]))
assert uids1 != uids2
```

because `uid1s` was the reference to the mutable list `RunEngine._run_start_uids`, and it was overwritten during the second call.

About the bug report:

This became easier to debug once I reduced it to a [SSCCE](http://sscce.org/). It's natural start in your own context -- with a databroker, registry-backed detectors, and XPD-specific metadata. But you should try to peel back those elements, committing as you go, to achieve the simplest possible thing that reproduces the bug. As you can see, it was possible to make this a lot simpler! It also helps to remove jargon like `dark_uid` and `light_uid` that don't relate to the bug itself. Use dumb placeholders like foo, bar so it's clear that this aren't critical to the behavior being tested.

CJ-Wright:
That's very cute. Thank you for reducing it! I thought I'd try to put up something since I owed you the test.

PR #790, opened 2017-08-22
FIX: Do not allow the drawn attribute to be out of sync with the drawing.

danielballan:
This _might_ fix the bug we saw at ESM. Should test in the wild before
merging.

### DIFF
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -966,16 +966,18 @@ class ProgressBar:
         self.delay_draw = delay_draw
         self.drawn = False
         self.done = False
+        self.lock = threading.RLock()
 
         # If the ProgressBar is not finished before the delay_draw time but
         # never again updated after the delay_draw time, we need to draw it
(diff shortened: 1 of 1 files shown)


bluesky/utils.py
```diff
@@ -1006,30 +1007,34 @@ def update(self, pos, *,
...
-        self.meters.clear()
+        with threading.Lock():
+            self.done = True
+            if self.drawn:
+                for meter in self.meters:
+                    self.fp.write('\r')
+                    self.fp.write(' ' * self.ncols)
+                    self.fp.write('\r')
+                    self.fp.write('\n')
+                self.fp.write(_unicode(_term_move_up() * len(self.meters)))
```
danielballan: The indentation change here is the crucial change, but the additional locking doesn't hurt.

danielballan:
Force-pushed amendments made and tested at ESM.

PR #797, opened 2017-08-28
DOC: Document how to use a kicker with the RemoteDispatcher.

### LINKED ISSUE #795: MPL callbacks fail to update when subscribed to zmq
When using ZMQ, subscribing MPL based callbacks to the `RemoteDispatcher` causes them to not be updated.

## Expected Behavior
MPL based callbacks update themselves when the ZMQ feed it idle.

## Current Behavior
MPL based callbacks don't update themselves when running inside ZMQ

## Possible Solution
It seems that ZMQ's loop and MPL's `draw_idle` loop are in conflict, with MPL never getting control.

1. start up proxy
1. Run
```python
# pull from local data, not needed at beamline
import numpy as np

from bluesky.callbacks.broker import LiveImage
from shed.event_streams import istar

a = LiveImage('pe1_image')

def put_in_queue(nd):
    if nd[0] == 'event':
        nd[1]['data']['pe1_image'] = np.asarray(nd[1]['data']['pe1_image'])
    a(*nd)
    plt.pause(.1)


disp = RemoteDispatcher('127.0.0.1:5568')
disp.subscribe(istar(put_in_queue))
print("REMOTE IS READY TO START")
disp.start()
```
3. Run
```python
import numpy as np

# pull from local data, not needed at beamline
from bluesky import RunEngine
from bluesky.examples import Reader
RE = RunEngine({})
p = Publisher('127.0.0.1:5567', RE=RE)  # noqa

det = Reader('pe1_image', 
             {'pe1_image': lambda : np.random.random((10, 10))})

uid = RE(count([det]))
```
Note that at the end of this neither the side 1D plots nor the cross hairs work.

## Context
This is going to cause tons of problems at XPD when we try to live ZMQ driven visualization.

danielballan:
Closes #795

### DIFF
--- a/doc/source/callbacks.rst
+++ b/doc/source/callbacks.rst
@@ -1091,7 +1091,22 @@ simple example, just print them.
     from bluesky.callbacks.zmq import RemoteDispatcher
     d = RemoteDispatcher('localhost:5578')
     d.subscribe(print)
+
+    # when done subscribing things and ready to use:
     d.start()  # runs event loop forever
+
+As `described above <kickers>`_, if you want to use any live-updating plots,
+you will need to install a "kicker". It needs to be installed on the same
+event loop used by the RemoteDispatcher, like so, and it must be done before
+calling ``d.start()``.
+
+.. code-block:: python
+
+    from bluesky.utils import install_qt_kicker
+    install_qt_kicker(loop=d.loop)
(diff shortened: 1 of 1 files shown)


CJ-Wright:
Thank you!

PR #802, opened 2017-08-28
WIP: try to make breaking test for ZMQ

CJ-Wright:
Based on issues out at XPD (may not work, as this may be an XPD specific issue).


CJ-Wright:
@danielballan this was the ZMQ thing I was discussing earlier.

Issue #796, opened 2017-08-26
Reader example should be filled


CJ-Wright:
The `Reader` example should give its results as filled.

## Expected Behavior
Since `Reader` doesn't fill if you make an image in a reader when passed to `LiveImage` it errors as their neither a `filled` key nor a `Broker` instance which harbors the data.

danielballan:
Sorry, I don't quite follow this. Can you share a minimal example?

CJ-Wright:
Run the live image callback against a reader which generates 10x10 arrays. The callback will fail as the result is filled but the filled dict is not set. Also it doesn't have a databroker to get the info from (not that really matters in this example

danielballan:
Can you share a minimal example with code?

CJ-Wright:
Sorry, I wrote that on the train on my phone.

```python
In [4]: from bluesky.examples import Reader
   ...: from bluesky.callbacks.broker import LiveImage
   ...: import numpy as np
   ...: from bluesky.run_engine import RunEngine
   ...: from bluesky.plans import count
   ...: 
   ...: r = Reader('reader', fields={'reader': lambda: np.random.random((10, 10))})
   ...: RE = RunEngine({})
   ...: RE.subscribe(LiveImage('reader'))
   ...: RE(count([r], num=5))
```
gave
```python
/home/christopher/mc/envs/dp_dev/lib/python3.5/site-packages/matplotlib/axes/_base.py:2917: UserWarning: Attempting to set identical left==right results
in singular transformations; automatically expanding.
left=-0.5, right=-0.5
  'left=%s, right=%s') % (left, right))
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-4-85cb5411dd88> in <module>()
      8 RE = RunEngine({})
      9 RE.subscribe(LiveImage('reader'))
---> 10 RE(count([r], num=5))

/home/christopher/dev/bluesky/bluesky/run_engine.py in __call__(self, plan, subs, raise_if_interrupted, **metadata_kw)
    644                     # it (unless it is a canceled error)
    645                     if exc is not None:
--> 646                         raise exc
    647 
    648             if raise_if_interrupted and self._interrupted:

/home/christopher/mc/envs/dp_dev/lib/python3.5/asyncio/tasks.py in _step(***failed resolving arguments***)
    237                 # We use the `send` method directly, because coroutines
    238                 # don't have `__iter__` and `__next__` methods.
[...]

PR #803, opened 2017-08-28
BrokerCallback only will try to load external things

CJ-Wright:
`BrokerCallback` only will try to load external things

### DIFF
--- a/bluesky/tests/test_callbacks.py
+++ b/bluesky/tests/test_callbacks.py
@@ -577,7 +577,7 @@ def test_broker_base(fresh_RE, db):
 
 
 @pytest.mark.xfail(reason='something funny going on with 3.5, 3.6 and sqlite')
-def test_broker_base(fresh_RE, db):
+def test_broker_base_no_unpack(fresh_RE, db):
     class BrokerChecker(BrokerCallbackBase):
         def __init__(self, field, *, db=None):
             super().__init__(field, db=db)
--- a/bluesky/tests/test_callbacks.py
+++ b/bluesky/tests/test_callbacks.py
@@ -576,7 +576,6 @@ def test_broker_base(fresh_RE, db):
     RE(count([det]))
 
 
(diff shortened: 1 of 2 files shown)


CJ-Wright:
Closes #796

danielballan:
That was fast. Looks right. We'll see if Travis is happy.

bluesky/tests/test_callbacks.py
```diff
@@ -573,3 +574,21 @@ def event(self, doc):
...
+    class BrokerChecker(BrokerCallbackBase):
+        def __init__(self, field, *, db=None):
+            super().__init__(field, db=db)
+
+        def event(self, doc):
+            super().event(doc)
+            assert isinstance(doc['data'][self.fields[0]], np.ndarray)
+
+    RE = fresh_RE
+    RE.subscribe(db.insert)
```
danielballan: I would remove this (and the `db` fixture altogether). `BrokerCallbackBase` doesn't need it, right? This would simplify the test and also fix the sqlite + numpy array failure that killed Travis.

CJ-Wright: I just took the other approach using the `raises` flag in `xfail` (although I could do this too, but I applied that one to the previous callback base test, which should make that one more robust)

danielballan: Fair enough. This also needs a `@py3` decorator.

bluesky/tests/test_callbacks.py
```diff
@@ -554,7 +556,8 @@ def test_live_scatter(fresh_RE):
                     xlim=(-3, 3), ylim=(-5, 5)))
 
 
-@pytest.mark.xfail(reason='something funny going on with 3.5, 3.6 and sqlite')
+@pytest.mark.xfail(raises=InterfaceError,
```
danielballan: Good improvement!

danielballan:
Awesome. The reviewer (me) should have caught this originally.

CJ-Wright:
Thank you!

PR #805, opened 2017-08-28
TST: More realistic zmq test.

danielballan:
Now the test ensures that event and descriptor documents go
through 0MQ successfully. As far as we can tell nothing is
broken, but this is a better test for future-proofing the code.


danielballan:
I think coverage is confused.

PR #807, opened 2017-08-29
ENH: Simple install_kicker()

hhslepicka:
Usually users need to call either ```install_qt_kicker``` or ```install_nb_kicker``` in order to get the live plots and other callbacks to work properly. The idea of this PR is to add a functionality that I added to LIX while working there to make it easier for the user by just installing the correct one based on how they are running the code.

## Description

Added a method at utils.py named ```install_kicker``` which based on the environment will install either qt_kicker or nb_kicker.

## Motivation and Context

It simplifies the setup of the kicker.

## How Has This Been Tested?

Locally tested.

### DIFF
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -688,30 +688,43 @@ def get_history():
 _QT_KICKER_INSTALLED = {}
 _NB_KICKER_INSTALLED = {}
 
-def install_kicker(loop=None):
-    """Install the proper kicker (qt_kicker or nb_kicker) based on where 
-    the code is running
+def install_kicker(loop=None, update_rate=0.03):
     """
-    def is_ipython():
-        ip = True
-        if 'ipykernel' in sys.modules:
-            ip = False # Notebook
-        elif 'IPython' in sys.modules:
-            ip = True # Shell
-        return ip
+    Install a periodic callback to integrate drawing and asyncio event loops.
 
-    if is_ipython():
-        install_qt_kicker(loop=loop)
+    This dispatches to :func:`install_qt_kicker` or :func:`install_nb_kicker`
+    depending on the current matplotlib backend.
+
+    Parameters
+    ----------
+    loop : event loop, optional
+    update_rate : number
+        Seconds between periodic updates. Default is 0.03.
+    """
+    import matplotlib
+    backend = matplotlib.get_backend()
+    if backend == 'nbAgg':
+        install_nb_kicker(loop=loop, update_rate=update_rate)
+    elif backend in ('Qt4Agg', 'Qt5Agg'):
+        install_qt_kicker(loop=loop, update_rate=update_rate)
     else:
-        install_nb_kicker(loop=loop)
+        raise NotImplementedError("The matplotlib backend {} is not yet "
+                                  "supported.".format(backend))
+
 
 def install_qt_kicker(loop=None, update_rate=0.03):
-    """Install a periodic callback to integrate qt and asyncio event loops
+    """Install a periodic callback to integrate Qt and asyncio event loops.
 
-    If a version of the qt bindings are not already imported, this function
+    If a version of the Qt bindings are not already imported, this function
     will do nothing.
 
     It is safe to call this function multiple times.
+
+    Parameters
+    ----------
+    loop : event loop, optional
+    update_rate : number
+        Seconds between periodic updates. Default is 0.03.
     """
     if loop is None:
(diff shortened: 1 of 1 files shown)


danielballan:
I like this thought. Forgive me for raining on this PR with some complications -- but then one positive suggestion.

* Modern IPython recommends launching kernels with `ipykernel` but in the "old days" (~2 years ago) `ipykernel` didn't exist and kernels were launched from `IPython` directly. At NSLS-II, we have some old conda environments still in use on the floor that predate `ipykernel` and would thus take the wrong code path here.
* The `ipython console` or `ipython qtconsole` invoke `ipykernel` (I think) but use Qt plotting. Again, they would take the wrong code path.
* Worst of all, a given kernel may have multiple front-ends ("sessions") attached to it -- a notebook and a Qt console. _A stated design principle of IPython is that a kernel is not allowed to know/care what kinds of front-ends it is talking to._

But let me suggest this solution: ask matplotlib which backend it is using and use _that_ to differentiate which kicker to install. A kernel might have multiple front-ends, but matplotlib can only have one backend at a time, and that's the real determining factor as to which kicker we need.

hhslepicka:
@danielballan suggestions are always welcome! 👍 

How do you guys have old IPython running since every deploy the environments are wiped out and recreated with the new cycle?

I will take a look at your suggestion and send modifications to the PR.

danielballan:
We don't wipe out the old conda environments (anymore, anyway). We keep them around so that old user notebooks can run on their original kernels without suffering from API changes.

hhslepicka:
I see! I will work on the suggested changes...
Invoking @tacaswell so he can tell me which backends go with qt_kicker and which ones goes with nb_kicker...
I saw nbAgg which seems a bit obvious but there was one very weird when I use ```%matplotlib inline```

danielballan:
The "inline" magic isn't a real backend; it's sort of an IPython hack. It can't do live-updating plots anyway, so I don't think we have to worry about it for the purposes of the kicker.

tacaswell:
`mpl.get_backend()` is the right way to sort out how do do the dispatch.

tacaswell:
This kicker may work for all GUI backends

``python
def _kicker()
    Gcf.draw_all()
    # there is not a singleton exposed through Mpl, but all of the GUI backends have one (and 
    plt.gcf().canvas.flush_events()
    loop.call_later(0.03, _kicker)
```

but may not work for nbagg or ipympl

**Comment by danielballan** [comment]
For the record, my wife has inquired why we are kicking backends. We work in a strange profession.

**Comment by tacaswell** [comment]
Or I should just not be allowed to name things....

**Comment by danielballan** [comment]
@hhslepicka I took the liberty of rebasing and force-pushing to this branch. This is the approach I was suggesting. Can you check that it works for you before we merge?

**Comment by danielballan** [comment]
bump @hhslepicka

**Comment by hhslepicka** [comment]
@danielballan sorry for the delay. I did some tests and it worked for my using iPython and the notebook.
I believe it is good to go. 👍 Thank you for finishing this!

**Comment by jrmlhermitte** [comment]
I like this. 
We should eventually add a note in the [docs](http://nsls-ii.github.io/bluesky/callbacks.html#aside-making-plots-update-live) which is [here](https://github.com/NSLS-II/bluesky/blob/master/doc/source/callbacks.rst#aside-making-plots-update-live).

Codecov seems to complain. Is there an easy way to test something like this or should we just ignore?

Maybe add this to .coveragerc in main dir?
```
exclude_lines =
    if __name__ == '__main__':
    def install_qt_kicker
    def install_nb_kicker
    def install_kicker
```

danielballan:
This one is hard to test programmatically but quite easy to test interactively. Since we use it regularly, the risk of an unnoticed regression seems no. I would be inclined to let coverage continue to shame us (that is, not silence it by modifying `exclude_lines`) but not worry about addressing this any time soon.

jrmlhermitte:
sounds good. If someone can confirm they've tested it for both ipython and jupyter I second on merging.
(I can test a little later on as well if no one else can get to it)

danielballan:
Looks like Hugo [confirmed it](https://github.com/NSLS-II/bluesky/pull/807#issuecomment-342595300) since the most recent commit was added. Merging!

hhslepicka:
@danielballan yes I did!  Thank you for merging. 👍

jrmlhermitte:
oops yes thanks @hhslepicka ! useful addition!

PR #810, opened 2017-08-31
FIX: Make progress bar work when position is a tuple.

danielballan:
- Handle case where current, initial, total are tuples.
- Handle case when 'name' is not provided by status obj.
- Handle case where 'unit' is not provided by status obj.
- Use status.done rather than fraction to check completeness.
- Note that 'time_elapsed' and 'fraction' are not yet used by this
  implementation of a progress bar.

### DIFF
--- a/bluesky/tests/test_progress_bar.py
+++ b/bluesky/tests/test_progress_bar.py
@@ -22,6 +22,63 @@ def test_status_with_name():
     time.sleep(0.25)
     st._finished()
 
+
+def test_tuple_progress():
+
+    class StatusPlaceholder:
+        "Just enough to make ProgressBar happy. We will update manually."
+        def __init__(self):
+            self.done = False
+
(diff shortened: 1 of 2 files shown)


tacaswell:
This will allow progress bars to be re-enabled at SMI attn @mrakitin

mrakitin:
Thanks @tacaswell, we will check it tomorrow on the beamline. Should `pip install bluesky --upgrade` be sufficient? Or should I use `conda ...`?

tacaswell:
We may have another rc ready by tomorrow in which case

```
pip uninstall bluesky
conda install bluesky
```

should work.  In either case

```
pip install --upgrade git+https://github.com/nsls-ii/bluesky.git@master
```

should work.  There are two lines in your config where the wait_hook is set to `None` that will need to be removed as well.

PR #811, opened 2017-08-31
FIX: Fix issue with OrderedDict in 0MQ and add debug flag.

danielballan:
This was a subtle one exposed by @CJ-Wright.

In addition to fixing the bug, this adds a debug flag that will make
ast-related issues easier to debug in the future.


danielballan:
On second thought, this is a silly solution. Will revisit later. @CJ-Wright you can keep using the fix I did in place XPD; the new one will be transparent.

PR #815, opened 2017-09-01
Mnt bec 2 d plotting

tacaswell:
## Motivation and Context

The fixed aspect + shared axis where producing sub-optimal plots for multiple detectors plotted with BEC.

BEC now also guesses if the image should have equal aspect ratio (for example scanning two motors in the same units) or automatic aspect (which makes sense for motor vs E or something like that).

## How Has This Been Tested?

```python
import matplotlib
matplotlib.use('Qt5Agg')
plt.ion()
from bluesky.examples import motor, motor1, det, det1
from bluesky import RunEngine
import bluesky.plans as bp
install_qt_kicker()

RE = RunEngine()
bec = BestEffortCallback()
RE.subscribe(bec)

```

```python
RE(bp.outer_product_scan([det1, det], motor, -1, 1, 5, motor1, 10, 20, 3, False))
```
so

```python
RE(bp.outer_product_scan([det1, det], motor, -1, 1, 3, motor1, 0, 2, 3, False))
```
so

### DIFF
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -72,7 +72,7 @@ class BestEffortCallback(CallbackBase):
             return
 
         super().__call__(name, doc)
-    
+
     def start(self, doc):
         self.clear()
         self._start_doc = doc
@@ -101,9 +101,9 @@ class BestEffortCallback(CallbackBase):
                  "combine streams.")
         self.dim_fields = [f
                            for field, stream_name in dimensions
-                               for f in field]
-        _, self.dim_stream  = dimensions[0]
-    
+                           for f in field]
+        _, self.dim_stream = dimensions[0]
+
     def descriptor(self, doc):
         self._descriptors[doc['uid']] = doc
         stream_name = doc.get('name', 'primary')  # fall back for old docs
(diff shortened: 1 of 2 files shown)


bluesky/callbacks/best_effort.py
```diff
@@ -101,9 +101,9 @@ def start(self, doc):
                  "combine streams.")
         self.dim_fields = [f
                            for field, stream_name in dimensions
-                               for f in field]
-        _, self.dim_stream  = dimensions[0]
-    
+                           for f in field]
```
danielballan: You suggested this indentation when we pair-coded this part. You pointed out that it correctly illustrates the nested-ness of the for loop.

bluesky/callbacks/best_effort.py
```diff
@@ -115,7 +115,7 @@ def descriptor(self, doc):
 
         columns = hinted_fields(doc)
 
-        ### This deals with old documents. ### 
+        # ## This deals with old documents. ## #
```
danielballan: I suspect you let your linter off its leash here (and similar below).

bluesky/callbacks/best_effort.py
```diff
@@ -191,15 +195,20 @@ def descriptor(self, doc):
...
+                    elif ndims == 2:
+                        share_kwargs = {'sharex': ax, 'sharey': ax}
+                    else:
+                        raise NotImplementedError("we now support 3D?!")
                 else:
-                    ax = fig.add_subplot(len(columns), 1, 1 + i, sharex=ax)
-            fig.subplots_adjust()
-            fig.tight_layout()
+                    ax = fig.add_subplot(len(columns), 1, 1 + i,
+                                         **share_kwargs)
```
danielballan: This is elegantly done.

PR #816, opened 2017-09-01
API: input -> input_plan

### LINKED ISSUE #808: inclusion of 'input' as a generator in bluesky.plans is causing issues with the use of pythons 'input' .

danielballan:
Closes #808

Thanks for the feedback @kevinyager, @cmazzoli, @ambarb -- and @awalter-bnl for kicking this off.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -644,7 +644,7 @@ def deferred_pause():
     return (yield Msg('pause', None, defer=True))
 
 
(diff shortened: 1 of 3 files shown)


tacaswell:
do we want to do the same with the sleep plan while we are here?

danielballan:
No, I don't think so. See my comment in #808 on that subject.

PR #820, opened 2017-09-06
FIX: Minor fixes to BEC

danielballan:
Fixes two minor bugs caught during the CMS deployment.

(cc @ordirules)

Tested interactively.


licode:
LGTM.

PR #822, opened 2017-09-06
ENH: Add update_rate as kwarg in kicker utils

teddyrendahl:
## Description
Allow users to select the update_rate of either the `nb_kicker` or `qt_kicker` at call time. 

## Motivation and Context
Doesn't seem like it could hurt anything. Main motivation is we are experimenting with putting the RunEngine behind a PyQT UI and 0.03s between updates is too unresponsive

Have you experimented with controlling the RunEngine through Qt? We might end up just launching multiple threads to have completely separate event loops, so this may end up being moot. However, the option seems like it would be nice to have.

### DIFF
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -688,7 +688,7 @@ _QT_KICKER_INSTALLED = {}
 _NB_KICKER_INSTALLED = {}
 
 
-def install_qt_kicker(loop=None):
+def install_qt_kicker(loop=None, update_rate=0.03):
     """Install a periodic callback to integrate qt and asyncio event loops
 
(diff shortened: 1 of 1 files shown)


danielballan:
Yeah, there's no harm in making this configurable.

@Brudhu has written a Qt GUI in front of bluesky/ophyd optimized for one particular beamline (ISS) with just a little input from Tom and me.

tacaswell:
Might be worth using https://github.com/harvimt/quamash to source the eventloop you pass into RE (rather than pushing it to a thread).

Pushing it to a thread is currently not possible because there is some signal handling baked into `__call__`.  Moving that has been on the to-do list for a while now.

PR #825, opened 2017-09-07
FIX: Missed a line in #820

danielballan:


stuartcampbell:
The build errors are not connected to the changes.,

PR #826, opened 2017-09-07
ENH: Include offset (if applicable) in 'wa' output.

danielballan:


danielballan:
This adds a column to the output of `%wa` displaying the value of `user_offset`, if applicable. Tested at CMS and in addition to the new unit tests here.

PR #836, opened 2017-09-11
FIX: Bug in baseline_wrapper

danielballan:


bluesky/plans.py
```diff
@@ -60,15 +60,30 @@ def plan_mutator(plan, msg_proc):
...
-
-        * msg -> None, None (no op)
-        * msg -> gen, None (mutate and/or insert before current message;
-          last message in gen must invoke a response compatible
-          with original msg)
-        * msg -> gen, tail (same as above, but insert some messages *after*)
-        * msg -> None, tail (illegal -- raises RuntimeError)
+        This function takes in a message and specifies messages(s) to replace
+        it with. The function must account for what type of response the
+        message would prompt. For example, an 'open_run' message causes the
```
ordirules: maybe `prompt` -> `return`?

bluesky/plans.py
```diff
@@ -60,15 +60,30 @@ def plan_mutator(plan, msg_proc):
...
+        * ``(None, tail)`` Let the original message pass through and then
+        insert messages after.
+
+        The reason for returning a pair of generators instead of just one is to
+        provide a way to specify which message's response should be sent out to
+        the host plan. Again, it's the last message yielded by the first
+        generator (``head``).
 
     Yields
     ------
```
ordirules: I wrote write an example use case, this would speak more for itself than anything.

bluesky/plans.py
```diff
@@ -60,15 +60,30 @@ def plan_mutator(plan, msg_proc):
...
-        * msg -> None, None (no op)
-        * msg -> gen, None (mutate and/or insert before current message;
-          last message in gen must invoke a response compatible
-          with original msg)
-        * msg -> gen, tail (same as above, but insert some messages *after*)
-        * msg -> None, tail (illegal -- raises RuntimeError)
+        This function takes in a message and specifies messages(s) to replace
+        it with. The function must account for what type of response the
+        message would prompt. For example, an 'open_run' message causes the
+        RunEngine to send a uid string back to the plan, while a 'set' message
```
ordirules: Maybe 'yields a uid string back to the plan'. Correct me if I misunderstood this part. I feel that this is modifying the generator behaviour, and thus the RunEngine does not need to be introduced.

bluesky/plans.py
```diff
@@ -60,15 +60,30 @@ def plan_mutator(plan, msg_proc):
...
+        causes the RunEngine to send a status object back to the plan. The
+        function should return a pair of generators ``(head, tail)`` that yield
+        messages. The last message out of the ``head`` generator is the one
+        whose response will be sent back to the host plan. Therefore, that
+        message should prompt a response compatible with the message that it is
+        replacing. Any responses to all other messages will be swallowed. As
+        shorthand, either ``head`` or ``tail`` can be replaced by ``None``.
+        This means:
+
+        * ``(None, None)`` No-op. Let the original message pass through.
```
ordirules: this is very good and clear

danielballan:
Thanks for taking the time to look at this. The main conclusion I draw is that we need a larger section of prose on adaptiveness and how the RunEngine and plans use `generator.send` to pass information back and forth. I think I'l punt that for a separate PR because this one contains a hotfix, but your comments are helpful for outlining what more documentation needs to explain.

Issue #684, opened 2017-06-16
LiveFit Sequence Numbering


teddyrendahl:
Building on top of the LiveFit callback for a few applications and running into an issue if my subscription doesn't start at the first event document in the Run. Currently LiveFit calls .update_fit() if the 'seq_num' key is higher than the number of constants. 

https://github.com/NSLS-II/bluesky/blob/master/bluesky/callbacks/core.py#L758

Is there a reason that we can't just check the length of the cached array?
```python
i  = len(self.ydata)
```
I'm trying to write an averaging fit callback which is how I stumbled upon this issue, but I would imagine it would happen if you subscribed the callback mid-run as well.

tacaswell:
Checking the sequence number is less work :smiling_imp: 

We spent a fair amount of time talking about late subscriptions and came down on it is generally a bad idea  as jumping into a stream part way through means you may see documents without seeing required preceding documents (ex, events without descriptors and descriptors without starts).  That _seems_ easy enough to fix by caching and letting lat subscribers catch up, but at that point you have started to re-implement a message queue with topics (like kafka) and you should just use one of the existing message queues (and persistently subscribe the insert all documents into the queue).

danielballan:
@teddyrendahl I'd be interested to hear more about the use case for subscribing mid-run. It's probably at least loosely related to some things I am working on now.

teddyrendahl:
**TLDR:** I want to apply a set of filters on my event documents so that certain ones are not included in either my `LiveFit` or `LivePlot`. I have hacked through this by updating the `seq_num` attribute of each doc as it passes through my filters to account for dropped shots.

A little bit of background, our main motivation for using bluesky is for some up and coming automation projects at LCLS. We decided the easiest way to do this was to essentially create a plan that directly incorporated a `LiveFit` into an `adaptive_scan`. This means you give a `lmfit.Model` that you think will correctly fit your data, a motor scans the parameter space at first making naive steps but slowly gathering more data, at each checkpoint checking with the fit what the estimated move is to reach a target. 

In theory this all works very well! But the LCLS FEL is very unstable, even when using the builtin suspenders, if we are reading from a detector as fast as possible, we get some "bad shots" that mess up the fit. Since we are doing all of this live we don't have the benefit of just dropping these shots in offline analysis, they have to be excluded via filters on the ``LiveFit``. 

[...]

PR #845, opened 2017-10-11
MAINT: LiveFitPlot should reference ydata before updating

teddyrendahl:
Solution for issue #684. Reference the length of the stored `ydata` instead of the sequence numbering before updating the `lmfit` model
## Description

* Instead of referencing the sequence number we see check the `ydata` array directly
* I replaced the `RuntimeError` with a print statement. It seems to me the operating model is that even when they are misconfigured the callbacks should not raise exceptions. For example, if you enter the wrong key name in `LivePlot` it doesn't break your scan, it just doesn't display well. 

## Motivation and Context

There are a number of situations where checking the sequence number is less advantageous. For instance, if you have multiple groups being read out and your y data is not present in every emitted event document. If the first sequence number you see the y data key is document five the LiveFit would have tried to update itself.

As far as replacing the `RuntimeError`, I've run into a few places where this has been super annoying. For instance, your scan takes a single measurement  then an exception is raised, calling `stop`. Which in turns raising the `RuntimeError` on the callback because it tries to update the fit. This makes it hard to debug as the traceback just shows that the `LiveFit` raised an exception.

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -801,8 +801,8 @@ class LiveFit(CallbackBase):
     def update_fit(self):
         N = len(self.model.param_names)
         if len(self.ydata) < N:
-            print("LiveFitPlot cannot update fit until there are at least {} "
-                  "data points".format(N))
+            warnings.warn("LiveFitPlot cannot update fit until there are at least {} "
+                          "data points".format(N))
         else:
             kwargs = {}
             kwargs.update(self.independent_vars_data)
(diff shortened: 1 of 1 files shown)


danielballan:
Sounds good. I agree that misconfigured callbacks should generally not raise. Minor request: Would you mind replacing the print with a `UserWarning`? Print statements cannot be silenced, but warnings can be filtered.

danielballan:
Thanks!

PR #851, opened 2017-10-16
CI: Simplify build process by using pip only.

danielballan:
We currently run some test builds by using pip and others by using conda.
We originally used conda because it was much faster than pip. Now that
we have pip wheels, pip is as fast or faster. It is also empircally more
reliable and results in a simpler .travis.yml file.

The main reason for continuing to test with conda packages is to make
sure our public conda channels are up to date. This can be done using
a single integration test. There is no need to do it separately for
every package (bluesky, databroker, ophyd).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,42 +10,25 @@ cache:
     - $HOME/.cache/matplotlib
 
 matrix:
+  fast_finish: true
   include:
-    - python: 3.6
     - python: 3.5
-      env: CONDA_ENV=1
-    - python: 3.5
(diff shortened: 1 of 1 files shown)


teddyrendahl:
Just curious. Are you also running integration tests on Travis?

danielballan:
We created a repo for doing so but never developed it. It looks like it hasn't been touched in years. I intend to reinstate conda-based builds there shortly.

https://github.com/NSLS-II/vertical-integration

Issue #850, opened 2017-10-16
Split plans module into separate modules


danielballan:
The `plans.py` file is quite large and somewhat difficult to navigate. Typing `bluesky.plans.<TAB>` issues a very long list of suggestions, mixing high-level plans like `count` with low-level plans like `stage_wrapper`. To improve the discoverability, we propose breaking the plans module into:

* `plans.py` -> things like `scan` and `count` that create a run
* ~`plan_parts.py`~ `plan_stubs.py` -> things like `mv`, `trigger_and_read`, `fly` ("stub plans")
* `preprocessors.py` -> the wrappers and decorators that transform a plan into a different plan

For backward-compatibility, we can shim imports into `plans.py` with a warning about the new location.

Thoughts -- @teddyrendahl, @ZLLentz, @hhslepicka, @prjemian?

teddyrendahl:
Seems very reasonable. I like `plan_stubs.py`  🤷‍♂️ 

It was a pain when starting to use bluesky to remember when I needed a `run_wrapper` and when I didn't. Seems like this resolves that pretty elegantly. 

On a related note, it might be worth looking at `callbacks/core.py`. It is getting quite large as well ...

Issue #833, opened 2017-09-10
Ship 'unboxed' versions of plans that do not create runs.


danielballan:
If a user wants to combine several plans into a custom plan but save the resultant data as part of a single 'run' (i.e., getting one scan ID), they cannot easily reuse the built-in plans. The built-in plans are create one run per invocation.

Two possiblities:

* Create a counterpart for each plan, like `unboxed_count` for `count` and `unboxed_scan` for `scan`.
* Add a keyword argument like `unboxed` for all plans.

(Do we already have an issue for this? We've been aware of the issue for awhile, but I couldn't find an open issue on GH.)

tacaswell:
I like option 1 much more than option 2

danielballan:
I start with this:

```python
import bluesky.plans as bp

def custom_scan(...):
    yield from bp.scan(...)
    yield from bp.scan(...)
```

"Oh darn that made two runs and I wanted to jump that data into one run to make my analysis easier." Honestly, neither of these look great:

Option 1:

```python
import bluesky.unboxed_plans as bup
import bluesky.plans as bp

def custom_scan(...):
    # create md similar to how we do in the source of bp.scan
    @bp.run_decorator(md=md)
    def inner():
        yield from bup.scan(...)
        yield from bup.scan(...)
   uid = yield from inner()
   return uid
```

Option 2:

```python
import bluesky.plans as bp

def custom_scan(...):
    # create md similar to how we do in the source of bp.scan
    @bp.run_decorator(md=md)
    def inner():
        yield from bp.scan(..., unboxed=True)
        yield from bp.scan(..., unboxed=True)
   uid = yield from inner()
   return uid
```

danielballan:
@stuartcampbell had a wild idea: a variant of `pchain` that uses plan_mutator to rip off the open_run/close_run messages. I kind of like that.

tacaswell:
I think ripping out the start/stop messages will cause issues because the plans will not get back the expected results (the uid from the start).  I think we do not do anything other than return the uid, but I have definitely written plans for beamlines that do so.  Can see work arounds to my immediate concerns, but worry it will quickly grow more complex than we want to deal with.

danielballan:
Given the complexity of employing an "unboxed" plan (illustrated above) it's probably better _not_ to provide this intermediate layer. It will be easier for users to use `trigger_and_read` and `mv` directly.

PR #853, opened 2017-10-16
ENH: Plan preprocessor to remove run_control

teddyrendahl:
A simple plan preprocessor called `stub_wrapper` that strips `open_run`, `closed_run`, `stage` and `unstage`. 
## Description

I think of this as the inverse method for `run_wrapper`. There is a lot of logic built into the complete plans, but these are slightly awkward to stitch together. This allows you to remove all of the run control and staging functionality and just use the scanning logic.
## Motivation and Context

This opens up a simple way to stitch plans together without having to rewrite things from scratch. For instance, this implementation seems pretty intuitive

```python

list_scan([det], motor, [1,2,3,4,5], per_step=stub_wrapper(count([det], num=5, delay=3)))
```
We have been using this at SLAC in a few places so I figured I would throw it back and see if you found it useful. The nice thing is as a developer you can write a plan that nicely configures metadata e.t.c, but if someone wants to use it as part of a larger scan they are free to rip out the staging / RunEngine control.

The only concern I might have is this opens up a whole new way to put plans together that could be exploited. For instance, do we want users running a grid scan with nested calls of list_scan?

## How Has This Been Tested?
Added a test in `bluesky/tests/test_generators.py`

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1403,24 +1403,34 @@ def stub_wrapper(plan):
 
     This will remove any `open_run`, `close_run`, `stage` and `unstage` `Msg`
     objects present in the plan in order for it to be run as part of a larger
-    scan.
+    scan. Note, that any metadata from the provided plan will not be sent to
+    the RunEngine automatically.
 
     Parameters
     ----------
     plan : iterable or iterator
         A generator list or similar containing `Msg` objects
+
+    Returns
+    -------
+    md : dict
+        Metadata discovered from `open_run` Msg
     """
+    md = {}
     def _block_run_control(msg):
         """
         Block open and close run messages
         """
-        if msg.command in ['open_run', 'close_run',
-                           'stage', 'unstage']:
+        #Capture the metadata from open_run
+        if msg.command == 'open_run':
+            md.update(msg.kwargs)
+            return None
+        elif msg.command in ('close_run', 'stage', 'unstage'):
             return None
         return msg
 
-    return (yield from msg_mutator(plan, _block_run_control))
-
+    yield from msg_mutator(plan, _block_run_control)
+    return md
 
 def monitor_during_wrapper(plan, signals):
     """
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1397,6 +1397,31 @@ def inject_md_wrapper(plan, md):
     return (yield from msg_mutator(plan, _inject_md))
 
 
+def stub_wrapper(plan):
+    """
+    Remove Msg object in order to use plan as a stub
+
+    This will remove any `open_run`, `close_run`, `stage` and `unstage` `Msg`
+    objects present in the plan in order for it to be run as part of a larger
+    scan.
+
+    Parameters
+    ----------
+    plan : iterable or iterator
+        A generator list or similar containing `Msg` objects
+    """
+    def _block_run_control(msg):
+        """
+        Block open and close run messages
+        """
+        if msg.command in ['open_run', 'close_run',
+                           'stage', 'unstage']:
+            return None
+        return msg
+
+    return (yield from msg_mutator(plan, _block_run_control))
+
+
 def monitor_during_wrapper(plan, signals):
     """
     Monitor (asynchronously read) devices during runs.
@@ -2007,6 +2032,7 @@ fly_during_decorator = make_decorator(fly_during_wrapper)
 monitor_during_decorator = make_decorator(monitor_during_wrapper)
 inject_md_decorator = make_decorator(inject_md_wrapper)
 run_decorator = make_decorator(run_wrapper)
+stub_decorator = make_decorator(stub_wrapper)
 configure_count_time_decorator = make_decorator(configure_count_time_wrapper)
 
 
--- a/bluesky/tests/test_generators.py
+++ b/bluesky/tests/test_generators.py
@@ -5,7 +5,7 @@ from itertools import zip_longest
 
 from bluesky import Msg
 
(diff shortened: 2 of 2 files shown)


danielballan:
I raised something like this in #833 (not quite so nice as what you've done here) and @tacaswell noted an important consideration:

> I think ripping out the start/stop messages will cause issues because the plans will not get back the expected results (the uid from the start). I think we do not do anything other than return the uid, but I have definitely written plans for beamlines that do so. Can see work arounds to my immediate concerns, but worry it will quickly grow more complex than we want to deal with.

I have also written plans that return the `uid` to the caller, and it's important that we not break that. But in cases like the one you showed above --

```python
list_scan([det], motor, [1,2,3,4,5], per_step=stub_wrapper(count([det], num=5, delay=3)))
```

-- I think the return behavior is correct; it's what I would expect.

To your concern:

> The only concern I might have is this opens up a whole new way to put plans together that could be exploited. For instance, do we want users running a grid scan with nested calls of list_scan?

That actually sounds pretty useful to me! There is a specific use case her for exploring a grid around a sample position for a range of (arbitrarily spaced) samples.

I would advocate for merging this and marking it as experimental. Having it handy would reduce the "cliff" of complexity that I sometime falls off when a plan insists on staging things when I wish it wouldn't and my recourse is to start copy/pasting the source of `count` while the beamline scientist's eyes glaze over.

tacaswell:
I am :-1:  on this.  We discussed this in #833 and a bunch in person at BNL yesterday (sorry for not recording enough of those conversations!).

We propose splitting the current `plans.py` up into 3 files:

 - plans.py for things that create a run
 - plan_parts.py for things that yield messages but do not create a run
 - plan_mutators.py for all of the wrappers / decorators

For the use-case you show here, I think it would be much simpler to just provide a `mulit_trigger_and_read` plan_part rather than a way to add counter-acting wrappers to existing plans.

tacaswell:
I completely forgot that `msg_mutator` had the ability to drop messages all together (and I wrote it! 3dd2ef03d73fc9b614abff1724a55a06ba74374b ).  That actually solves a bunch of the complications I was thinking about above, but still think it is more trouble than it is worth.

Most of the 'boxed' plans put a fair amount of meta-data about their parameters into the start document which will now be dropped on the floor.  To use a 'list scan that does a mesh scan at every point' correctly you are going to have to do all of the argument parsing, meta-data formatting, and start/stop management your self _anyway_.  That is the messy/annoying part of the plans so adding one more layer of indirection seems more complex than just writing the nested for-loops.

stuartcampbell:
I can see the use for this, and as it is in use at SLAC already - I think getting it in the code base (and maybe marked as experimental as @danielballan suggested) would be a good thing.

danielballan:
@tacaswell Aside -- the proposal for splitting plans was recorded in https://github.com/NSLS-II/bluesky/issues/850 and @teddyrendahl gave some feedback there.

danielballan:
One could avoid throwing away the metadata by doing something like this:

```python
def stub_wrapper(plan)
     md = {}
     def _block_run_control(msg):
         """
         Block open and close run messages
         """
         if msg.command == 'open_run':
             md.update(msg.kwargs)  # capture the metadata from open_run
             return None
         if msg.command in ('close_run', 'stage', 'unstage'):
             return None
         return msg

    yield from msg_mutator(plan, _block_run_control)
    return md
```

but of course the caller receives `md` after the generator is drained -- too late to put it into its own `open_run` message at the top of the run.

I think it would be OK to find out from experience whether or not this more complex than writing everything from scratch.

stuartcampbell:
@teddyrendahl Do you think it is worth adding the `md` change as suggested by @danielballan or should we merge now.

klauer:
This actually looks really useful. 👍

teddyrendahl:
@stuartcampbell I'll add the `md` suggestion real quick and put a note that unless the user intervenes this information will be lost. My original thought was that if you are using this as a stub plan you probably don't care too much about the metadata, but admit it does seem like a waste to ignore it. Maybe there is a way to inject metadata into `RE` mid-execution but don't see anything obvious.

To be honest, I thought dropping the staging was going to be more controversial 😄 

@tacaswell @danielballan Regarding #850, in my mind this accomplishes a different goal. From my impression, that was about how to make it more explicit which plans were boxed and ready to go, and which were meant as a toolkit for larger plan creation.

We would like to use this to ease the burden on our development process. It largely means we try and write most plans as stand-alone boxed plans. That way a user can trust that the appropriate `md` will be stored and devices staged, then if a more advanced user wants to use it as part of a different plan they can use `stub_wrapper`, making sure that they include the metadata and staging that they care about. This way we don't repeat ourselves and/or maintain a boxed vs. stub library?

hhslepicka:
I am also 👍 to this PR.

tacaswell:
Do we check in the RunEngingine that all of the descriptors in an event stream actually have the same keys? [checking now]

tacaswell:
We key the descriptor cache on `(stream_name, frozenset(objs_read))` so something like


```python
@run_wrapper()
def mixed_stream():
    yield from stub_wrapper(count([det3]))
    yield from stub_wrapper(count([det2]))
```

will run result in a 'primary' event stream which has different keys in the two events which violates the idea that streams <-> to dense tables.  That should probably be fixed independent of this PR though.

danielballan:
> Maybe there is a way to inject metadata into RE mid-execution but don't see anything obvious.

You are correct: there isn't. But this makes me wonder whether there should be. This feeds into a larger discussion about better tools for large-scale plans that is out of scope for this PR thread.

@tacaswell Good point. I created #856 to address this.

Thanks again for sharing this, @teddyrendahl.

PR #857, opened 2017-10-17
API: Always raise when interrupted. Reserve no kwargs in RE.__call__.

### LINKED ISSUE #704: When state is paused, raise RuntimeError from None
A good suggested made in an APS tutorial

danielballan:
~There are couple more tests to fix before this is done, but let's kick off discussion in the meantime....~ Update: tests pass.

## Part I: Always raise when interrupted.

When the RunEngine is paused by Ctrl+C or by a plan (using `Msg('pause')`) it does not raise an exception by default. This is usually "nice" for interactive use (no ugly tracebacks!) but it is bad behavior as a library. This kind of thing is dangerous:

```python
def f():
    RE(plan1)  # if this catches Ctrl+C and goes into a paused state
    RE(plan2)  # this will immediate start running! bad!
```

Even here at NSLS-II where @tacaswell and I prowl the ring with clubs and bludgeon scientists about avoiding this, it happens all the time. It is just too obvious of thing to do, and users naturally discover it. We should continue to tell them not to. (Don't worry, @tacaswell, I'm not compromising my position on that.) But as an extra safely measure, I think the RunEngine should always raise when interrupted so that the function above will _not_ execute `plan2` if `plan1` is interrupted.

Before:

```python
In [5]: RE(count([], delay=2, num=2))
^CA 'deferred pause' has been requested. The RunEngine will pause at the next checkpoint. To pause immediately, hit Ctrl+C again in the next 10 seconds.
Deferred pause acknowledged. Continuing to checkpoint.
^C
Your RunEngine is entering a paused state. These are your options for changing
the state of the RunEngine:

RE.resume()    Resume the plan.
RE.abort()     Perform cleanup, then kill plan. Mark exit_stats='aborted'.
RE.stop()      Perform cleanup, then kill plan. Mark exit_status='success'.
RE.halt()      Emergency Stop: Do not perform cleanup --- just stop.

Pausing...
```

After:

```python
In [4]: RE(count([], delay=2, num=2))
^CA 'deferred pause' has been requested. The RunEngine will pause at the next checkpoint. To pause immediately, hit Ctrl+C again in the next 10 seconds.
Deferred pause acknowledged. Continuing to checkpoint.
^CPausing...
RunEngineInterrupted                      Traceback (most recent call last)
<ipython-input-4-764cff068f39> in <module>()
----> 1 RE(count([], delay=2, num=2))

~/Documents/Repos/bluesky/bluesky/run_engine.py in __call__(self, *args, **metadata_kw)
    650
    651             if self._interrupted:
--> 652                 raise RunEngineInterrupted(PAUSE_MSG) from None
    653
    654         return tuple(self._run_start_uids)

RunEngineInterrupted:
Your RunEngine is entering a paused state. These are your options for changing
the state of the RunEngine:

RE.resume()    Resume the plan.
RE.abort()     Perform cleanup, then kill plan. Mark exit_stats='aborted'.
RE.stop()      Perform cleanup, then kill plan. Mark exit_status='success'.
RE.halt()      Emergency Stop: Do not perform cleanup --- just stop.
```

It's a bit uglier, but the important thing is that it correctly raises an exception that will propagate up the stack!

There is a special kwargs in `RE.__call__` to optionally turn on behavior like this -- `RE(..., raise_if_interrupted=True)` -- but generally no one who needs it knows that they do or thinks to look for it. We should just remove it. Which brings me to...

## Part II: Reserve no kwargs in `RE.__call__`.

Removing the `raise_if_interrupted` keyword argument solves an unrelated problem: we can now say without reservation that _all_ keyword arguments passed to `RunEngine.__call__` are treated as user metadata. We can even avoid shadowing the terms 'plan' and 'subs' by using some tricks with `Signature` to make them positional-only arguments, reserving all keyword arguments for the user. It's a minor thing, but it's nice to be able to say "All keyword args are interpreted as metadata," without going on to qualify that with some (admittedly hard-to-hit) exceptions.

Thoughts?

Closes #704

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -4,6 +4,7 @@ import time as ttime
 import sys
 import logging
 from warnings import warn
+from inspect import Parameter, Signature
 from itertools import count, tee
 from collections import deque, defaultdict, ChainMap
 import signal
@@ -80,6 +81,14 @@ class LoggingPropertyMachine(PropertyMachine):
                      obj, old_value, value)
         if obj.state_hook is not None:
             obj.state_hook(value, old_value)
+
+
+# See RunEngine.__call__.
+_call_sig = Signature(
+    [Parameter('self', Parameter.POSITIONAL_ONLY),
+     Parameter('plan', Parameter.POSITIONAL_ONLY),
+     Parameter('subs', Parameter.POSITIONAL_ONLY, default=None),
+     Parameter('metadata_kw', Parameter.VAR_KEYWORD)])
 
 
 class RunEngine:
@@ -551,19 +560,20 @@ class RunEngine:
             jsonschema.validate(doc, schemas[DocumentNames.event])
             self.dispatcher.process(DocumentNames.event, doc)
 
-    def __call__(self, plan, subs=None, *, raise_if_interrupted=False,
-                 **metadata_kw):
+    def __call__(self, *args, **metadata_kw):
         """Execute a plan.
 
-        Any keyword arguments other than those listed below will be interpreted
-        as metadata and recorded with the run.
+        Any keyword arguments will be interpreted as metadata and recorded with
+        any run(s) created by executing the plan. Notice that the plan
+        (required) and extra subscriptions (optional) must be given as
+        positional arguments.
 
         Parameters
         ----------
-        plan : generator
+        plan : generator (positional only)
             a generator or that yields ``Msg`` objects (or an iterable that
             returns such a generator)
-        subs: callable, list, or dict, optional
+        subs : callable, list, or dict, optional (positional only)
             Temporary subscriptions (a.k.a. callbacks) to be used on this run.
             For convenience, any of the following are accepted:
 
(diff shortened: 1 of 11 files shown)


hhslepicka:
I really like this approach! 👍 Mainly Part I.
Even not being the recommended approach it is still very common to make calls to RE inside functions and loops and this comes at hand to avoid the behavior mentioned.

danielballan:
This keeps failing due to the (unrelated) intermittent test failures of that #@$#@(*%@#$ ramp plan hahaha. Maybe finally fixed that in #858.

This should be good to merge if people are OK with the principle.

bluesky/run_engine.py
```diff
@@ -573,18 +575,27 @@ def __call__(self, plan, subs=None, *, raise_if_interrupted=False,
...
-            raise RunEngineInterrupted. False by default.
-
         Returns
         -------
         uids : list
             list of uids (i.e. RunStart Document uids) of run(s)
         """
+        # This scheme lets us make 'plan' and 'subs' POSITIONAL ONLY, reserving
+        # all keyword arguments for user metdata.
+        sig = Signature(
```
tacaswell: Can you define this outside of `__call__` so we can bind it back to call so introspection works as expected?

danielballan: How does that work? I've played around with `Signature` for a bit and I'm not sure that's possible. Got a link or a toy example?

danielballan: Signatures bind to arguments, not to functions. At least that's what `Signature.bind` does.

tacaswell: https://www.python.org/dev/peps/pep-0362/#visualizing-callable-objects-signature

` wrapper.__signature__ = sig`

danielballan: Right! Done and force-pushed.

danielballan:
The introspection now works:

```python
In [2]: RE = RunEngine({})

In [3]: RE.__call__?
Signature: RE.__call__(plan, subs=None, /, **metadata_kw)
Docstring:
Execute a plan.
<snip>
```

PR #858, opened 2017-10-18
TST: Yet another world-weary attempt to fix intermittent ramp plan failures.

danielballan:


tacaswell:
I am :-1: on this

danielballan:
It fails about 1/3 of the time these days. Can you think of a way that we can guarantee that the inner plan will be run X times?

tacaswell:
I am :+1: on my changes, but someone else should review them to be merged.

danielballan:
Glorious. Thanks!

Issue #847, opened 2017-10-15
issue with 'aborted' scans being listed as 'status = success' in stop metadata.


awalter-bnl:
Aborted scans at 21-ID are found to have 'status' equal to 'success' in the stop field of the metadata. This should be changed to 'failed' for aborted scans I believe.

danielballan:
Confirmed:

```python
In [11]: RE(count([], num=2, delay=5))
start {'uid': 'f9a2b621-f77d-4bbf-8c43-aa18717327d3', 'time': 1508189649.8308501, 'num_intervals': 1, 'scan_id': 1, 'hints': {'dimensions': [(('time',), 'primary')]}, 'plan_type': 'generator', 'plan_args': {'detectors': [], 'num': 2}, 'plan_name': 'count', 'detectors': [], 'num_points': 2}
^CA 'deferred pause' has been requested. The RunEngine will pause at the next checkpoint. To pause immediately, hit Ctrl+C again in the next 10 seconds.
Deferred pause acknowledged. Continuing to checkpoint.
^C
Your RunEngine is entering a paused state. These are your options for changing
the state of the RunEngine:

RE.resume()    Resume the plan.
RE.abort()     Perform cleanup, then kill plan. Mark exit_stats='aborted'.
RE.stop()      Perform cleanup, then kill plan. Mark exit_status='success'.
RE.halt()      Emergency Stop: Do not perform cleanup --- just stop.

Pausing...
Out[11]: ('f9a2b621-f77d-4bbf-8c43-aa18717327d3',)

In [12]: RE.abort()
Aborting: running cleanup and marking exit_status as 'abort'...
stop {'run_start': 'f9a2b621-f77d-4bbf-8c43-aa18717327d3', 'time': 1508189653.79558, 'uid': '542513f9-fd3b-4904-b3e0-5d94acbc7ecc', 'exit_status': 'success', 'reason': ''}
```

danielballan:
Thanks again for catching this @awalter-bnl. We now have it covered by a test to prevent accidental unnoticed breakages in the future.

PR #861, opened 2017-10-18
Fix mark success on abort

tacaswell:
## Description

`run_wrapper` and friends ensure that the plan always yields a run_stop which means the fall through in the RunEngine which sets the exit_status before hitting the run_close of lost resort.

## Motivation and Context

fixs #847 

## How Has This Been Tested?

added a test

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -20,7 +20,8 @@ from . import plan_patterns
 from .utils import (Struct, Subs, normalize_subs_input, root_ancestor,
                     separate_devices, apply_sub_factories, update_sub_lists,
                     all_safe_rewind, Msg, ensure_generator, single_gen,
-                    short_uid as _short_uid, RampFail, make_decorator)
+                    short_uid as _short_uid, RampFail, make_decorator,
+                    RunEngineControlExceptions)
 
 
 def planify(func):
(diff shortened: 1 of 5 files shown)


bluesky/tests/test_run_engine.py
```diff
@@ -1298,3 +1298,50 @@ def test_num_events(fresh_RE, db):
...
+
+    class Snowflake(Exception):
+        ...
+
+    @bp.run_decorator()
+    def bad_plan():
+        yield Msg('null')
+        raise Snowflake('boo')
+
+    with pytest.raises(Snowflake):
```
danielballan: I take it that this fails to happen?

tacaswell: it raises, but the run is marked as a success.

bluesky/plans.py
```diff
@@ -1153,6 +1166,75 @@ def finalize_wrapper(plan, final_plan, *, pause_for_debug=False):
     return ret
 
 
+def teef_wrapper(plan, *,
```
danielballan: This name is so bad it's good.

danielballan: Oh actually — I wasn’t going to suggest changing this but I have a good one: contingency_plan!

danielballan:
I like this a lot. Will take a closer look tomorrow!

stuartcampbell:
Can we put a comment in to spell out what `teef` means ?  I know what it is now, but future me might not be so bright! :)

prjemian:
Yes.  Name is evocative.  Must use restraint...

On Oct 18, 2017 7:21 PM, "Stuart Campbell" <notifications@github.com> wrote:

Can we put a comment in to spell out what teef means ? I know what it is
now, but future me might not be so bright! :)

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/NSLS-II/bluesky/pull/861#issuecomment-337764319>, or mute
the thread

danielballan:
I think the name has been changed to “contingency”. I have learned that abbreviations are particularly tough on nonnative speakers and am now trying to avoid them where practical.

danielballan:
@prjemian Ha. Was not aware just how evocative until I Googled it. Let’s avoid scandalizing potential Dutch collaborators.

tacaswell:
I have force-pushed potentially offensively named functions out of the history.

PR #863, opened 2017-10-19
Fix validate uniform eventstream

tacaswell:
## Description

We have always said that an event stream is homogeneous in terms of data keys.  We had been enforcing that this was true per descriptor, but not per stream name. 

## Motivation and Context

This was noted during review of #853

## How Has This Been Tested?

Had to update a few tests that were verifying the behavior on a per-descriptor level.

Fixed some tests that are broken by ophyd changes to Device.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1302,7 +1302,7 @@ class RunEngine:
             del self._monitor_params[obj]
         # Count the number of Events in each stream.
         num_events = {}
-        for (bundle_name, _), counter in self._sequence_counters.items():
+        for bundle_name, counter in self._sequence_counters.items():
             if bundle_name is None:
                 # rare but possible via Msg('create')
                 continue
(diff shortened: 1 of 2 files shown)


bluesky/run_engine.py
```diff
@@ -1696,9 +1702,9 @@ def _collect(self, msg):
         bulk_data = {}
         local_descriptors = {}  # hashed on obj_read, not (name, objs_read)
         for stream_name, data_keys in named_data_keys.items():
-            objs_read = frozenset(data_keys)
-            desc_key = (stream_name, objs_read)
+            desc_key = stream_name
             if desc_key not in self._descriptors:
+                objs_read = frozenset(data_keys)
```
danielballan: This didn't need to be moved inside the loop.

bluesky/run_engine.py
```diff
@@ -1713,11 +1719,17 @@ def _collect(self, msg):
...
-                self._descriptors[desc_key] = doc
+                self._descriptors[desc_key] = (objs_read, doc)
                 self._sequence_counters[desc_key] = count(1)
             else:
-                descriptor_uid = self._descriptors[desc_key]['uid']
-
+                objs_read, doc = self._descriptors[desc_key]
+                d_objs = frozenset(data_keys)
+                if d_objs != objs_read:
+                    raise RuntimeError("Miss matched objects read, "
```
danielballan: Mismatched the word, not Miss Matched the [children's clothing store](http://littlemissmatched.com/).

bluesky/run_engine.py
```diff
@@ -1522,8 +1522,12 @@ def _save(self, msg):
         objs_read = frozenset(self._objs_read)
 
         # Event Descriptor
-        desc_key = (self._bundle_name, objs_read)
-        if desc_key not in self._descriptors:
+        desc_key = self._bundle_name
+        d_objs, doc = self._descriptors.get(desc_key, (None, None))
+        if d_objs is not None and d_objs != objs_read:
+            raise RuntimeError("Miss matched objects read, expected {!s}, "
```
danielballan: Ditto re: mismatched

PR #871, opened 2017-10-23
DEV: negative adpative scan init

licode:


licode:
tests added.

bluesky/plans.py
```diff
@@ -2565,6 +2565,8 @@ def adaptive_scan(detectors, target_field, motor, start, stop,
     --------
     :func:`bluesky.plans.relative_adaptive_scan`
     """
+    assert 0 < min_step < max_step, "min_step and max_step must meet condition of max_step > min_step > 0"
```
danielballan: Would this better as a `ValueError` rather than an `AssertionError` since it is triggered by an illegal user-provided value?

bluesky/plans.py
```diff
@@ -2595,7 +2597,11 @@ def adaptive_core():
         past_I = None
         cur_I = None
         cur_det = {}
-        while next_pos < stop:
+        if stop >= start:
+            direction_sign = 1
+        else:
+            direction_sign = -1
+        while next_pos*direction_sign < stop*direction_sign:
```
danielballan: Minor point: this line needs whitespace around `*` to be PEP-8 compliant.

bluesky/plans.py
licode: sure, looks like assert is most often used in test.

tacaswell: CPython reserves the right to disregard assert statements at run time (for speed) so they are not good for runtime checks.

bluesky/plans.py
licode: I am kind of confused on how to control whitespace in if or while statement. good catch.

licode:
also added another test in relative_adaptive_scan.

bluesky/plans.py
```diff
@@ -2565,7 +2565,8 @@ def adaptive_scan(detectors, target_field, motor, start, stop,
     --------
     :func:`bluesky.plans.relative_adaptive_scan`
     """
-    assert 0 < min_step < max_step, "min_step and max_step must meet condition of max_step > min_step > 0"
+    if not 0 < min_step < max_step:
+        raise ValueError("min_step and max_step must meet condition of max_step > min_step > 0")
```
danielballan: This needs to be broken into two lines to fit inside the PEP8 79-character limit. This is one nice way to do this:

```python
        raise ValueError("min_step and max_step must meet condition of "
                         "max_step > min_step > 0")
```

licode: good catch.

bluesky/plans.py
```diff
@@ -2566,7 +2566,8 @@ def adaptive_scan(detectors, target_field, motor, start, stop,
     :func:`bluesky.plans.relative_adaptive_scan`
     """
     if not 0 < min_step < max_step:
-        raise ValueError("min_step and max_step must meet condition of max_step > min_step > 0")
+        raise ValueError("min_step and max_step must meet condition of"
```
danielballan: missed the trailing space here -- it will read "condition ofmaxstep"

danielballan:
OK, great. In order to keep commit history clean, I think it's worth squashing this down to one commit before we merge. Can you squash and force-push?

licode:
yep, that's also what I am thinking.

danielballan:
Good to go! Merging.

Issue #875, opened 2017-10-25
create a "tune" scan plan


prjemian:
tune a motor to the centroid of signal(motor)

This will implement a successful algorithm that has been used at the APS.

## Expected Behavior
Initially, traverse the range from start to stop with a specified number of points  Repeat with progressively smaller step size until the specified minimum step size is reached.  Rescans are centered on the cumulative centroid of signal(motor) with reduced range.

## Current Behavior
This is may be considered an alternative to the adaptive scan.
It is a useful addition to the suite of available plans.

## Possible Solution
Code will be coming on a branch.

## Steps to Reproduce (for bugs)
-n/a-

## Context
Part of a suite of tools that will automatically tune the beam line.

## Your Environment
bluesky

prjemian:
Completed with 5e41b2d

Issue #862, opened 2017-10-19
Better coupling of Documents into adaptive plans


tacaswell:
- `DocCollector` like call back to be subscribed from a plan to provide direct access to the documents in the plan
 - a consistent way for a plan to consult outside resources for direction, launder network communication through the run engine.  Maybe a way to ship a coroutine to the RE from a plan.

teddyrendahl:
👍 👍 

This is much needed for us. Leveraging things like `LiveFit` within a plan is powerful but slightly awkward at this point in time.

tacaswell:
We should also be very careful here, it is very easy to build 'callback soup' or over built frameworks.

@teddyrendahl What would you change about http://nsls-ii.github.io/cookbook/adaptive_fitting.html#sphx-glr-cookbook-adaptive-fitting-py to make it less awkard?

teddyrendahl:
A very good point. I need to think about it for a while.

I would say that most of the strangeness from our use case had to do with averaging/filtering. Things like `LiveFit` seem to occupy a weird space where they are kind of data processing, but if you are using it for control (the LiveFit informs the next step of the plan) it has to be integrated with the scanning logic as well.  For instance,  if we take a measurement as we experience a beam drop, this event can not be included in our `LiveFit`. Something that is very easy to do in offline analysis is awkward to put in the scanning logic largely because the `LiveFit` will add documents to the lmfit model anytime we read.

In my head, the issue you raised above would provide a tighter control over what events we send to our callbacks in general, but now that I read it again I might have been filling in some blanks with my own biases. I also agree that building a data analysis pipeline for my one use case of `LiveFit` in an adaptive plan is probably a bad idea.

Thoughts @ZLLentz?

tacaswell:
Can you reliably tell from looking at a single event if the beam has dropped or not?  If so it sounds like you just want a slightly more turned livefit the knows when to drop events on the floor.

A way to hook up an event filter to the RE (or to just subscribe a filter call back and subscribe the rest of your callbacks to it) would probably be a reasonable thing to want.

PR #876, opened 2017-10-25
#875 introduce tune_centroid() plan

prjemian:
tune a motor to the centroid of signal(motor) with progressively smaller range and step size

Move the motor to the final centroid at the end

## Description
adds `tune_centroid()` plan to bluesky.plans

## Motivation and Context
This algorithm has been used at APS (in SPEC macros) for some years.

## How Has This Been Tested?

Unit test contributed.  Also tested in ipython session using:

    motor = Mover('motor', {'motor': lambda x: x}, {'x': 0})
    det = SynGauss('det', motor, 'motor', center=-1.3, Imax=1e5, sigma=0.05)
    RE(tune_centroid([det], "det", motor, -1.5, -0.5, 0.01, 10))

Also tested (not shown here) using EpicsMotor and EpicsSignalRO with an `swait` record configured as a pseudo-detector.

```
In [1]: bluesky.__version__
Out[1]: '0.10.2.post17+g770df4e'

In [6]:     motor = bluesky.examples.Mover('motor', {'motor': lambda x: x}, {'x': 0})
   ...:     det = bluesky.examples.SynGauss('det', motor, 'motor', center=-1.3, Imax=1e5, sigma=0.021)
   ...:     RE(tune_centroid([det], "det", motor, -1.5, -0.5, 0.01, 10))
   ...: 
Transient Scan ID: 502     Time: 2017/10/25 00:47:13
Persistent Unique Scan ID: 'ed3e2b28-c3d7-4cac-97a1-6580141573d3'
New stream: 'primary'
|   seq_num |       time |      motor |        det |
|         1 | 00:47:13.4 |      -1.50 |       0.00 |
|         2 | 00:47:13.4 |      -1.39 |      12.87 |
|         3 | 00:47:13.4 |      -1.28 |   57126.92 |
|         4 | 00:47:13.4 |      -1.17 |       0.00 |
|         5 | 00:47:13.4 |      -1.06 |       0.00 |
|         6 | 00:47:13.4 |      -0.94 |       0.00 |
|         7 | 00:47:13.4 |      -0.83 |       0.00 |
|         8 | 00:47:13.5 |      -0.72 |       0.00 |
|         9 | 00:47:13.5 |      -0.61 |       0.00 |
|        10 | 00:47:13.5 |      -0.50 |       0.00 |
|        11 | 00:47:13.5 |      -1.50 |       0.00 |
|        12 | 00:47:13.5 |      -1.45 |       0.00 |
|        13 | 00:47:13.6 |      -1.40 |       0.89 |
|        14 | 00:47:13.6 |      -1.35 |    4729.94 |
|        15 | 00:47:13.6 |      -1.30 |   99297.17 |
|        16 | 00:47:13.6 |      -1.25 |    8268.87 |
|        17 | 00:47:13.6 |      -1.20 |       2.73 |
|        18 | 00:47:13.6 |      -1.15 |       0.00 |
|        19 | 00:47:13.6 |      -1.10 |       0.00 |
|        20 | 00:47:13.6 |      -1.06 |       0.00 |
|        21 | 00:47:13.6 |      -1.39 |       6.94 |
|        22 | 00:47:13.7 |      -1.37 |     389.51 |
|        23 | 00:47:13.7 |      -1.35 |    7333.19 |
|        24 | 00:47:13.7 |      -1.33 |   46310.78 |
|        25 | 00:47:13.7 |      -1.30 |   98103.95 |
|        26 | 00:47:13.7 |      -1.28 |   69711.79 |
|        27 | 00:47:13.7 |      -1.26 |   16616.57 |
|        28 | 00:47:13.7 |      -1.24 |    1328.59 |
|        29 | 00:47:13.8 |      -1.22 |      35.63 |
|        30 | 00:47:13.8 |      -1.19 |       0.32 |
|        31 | 00:47:13.8 |      -1.17 |       0.00 |
generator tune_centroid ['ed3e2b'] (scan num: 502)



all documents from last plan in variable: plan_documents
exit status: success
# descriptor(s): 1
# event(s): 31
Peak at  -1.29715791695
Out[6]: ('ed3e2b28-c3d7-4cac-97a1-6580141573d3',)

In [7]: motor.read()
Out[7]: {'motor': {'timestamp': 1508910433.830341, 'value': -1.2971579169521377}}
```

## Screenshot:
tune_centroid

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -2710,7 +2710,7 @@ def tune_centroid(
     
     Set `snake=True` if your positions are reproducible
     moving from either direction.  This will not necessarily
-    decrease the number of steps required to reach convergence.
+    decrease the number of traversals required to reach convergence.
     Snake motion reduces the total time spent on motion
     to reset the positioner.  For some positioners, such as 
     those with hysteresis, snake scanning may not be appropriate.  
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -2746,7 +2746,7 @@ def tune_centroid(
     Example
     -------
     motor = Mover('motor', {'motor': lambda x: x}, {'x': 0})
-    det = SynGauss('det', m1, 'm1', center=-1.3, Imax=1e5, sigma=0.05)
+    det = SynGauss('det', motor, 'motor', center=-1.3, Imax=1e5, sigma=0.05)
     RE(tune_centroid([det], "det", motor, -1.5, -0.5, 0.01, 10))
     """
     if step_factor <= 0:
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -2694,9 +2694,9 @@ def relative_adaptive_scan(detectors, target_field, motor, start, stop,
 def tune_centroid(
         detectors, signal, motor, 
         start, stop, min_step, 
-        num_points = 10, 
-        step_factor = 2,
-        snake = False,
+        num_points=10, 
+        step_factor=2,
+        snake=False,
         *, md=None):
     """
     plan: tune a motor to the centroid of signal(motor)
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -2694,7 +2694,7 @@ def relative_adaptive_scan(detectors, target_field, motor, start, stop,
 def tune_centroid(
         detectors, signal, motor, 
         start, stop, min_step, 
-        num_points=10, 
+        num=10, 
         step_factor=2,
         snake=False,
         *, md=None):
@@ -2735,8 +2735,8 @@ def tune_centroid(
         end of range, note: start < stop
     min_step : float
         smallest step size to use.
-    num_points : int, optional
-        number of points with each step size, default = 10
+    num : int, optional
+        number of steps with each step size, default = 10
     step_factor : float, optional
         used in calculating range when 
         maximum is found, note: step_factor > 0, default = 2
@@ -2755,9 +2755,9 @@ def tune_centroid(
         raise ValueError("min_step must be positive")
     if step_factor <= 0:
         raise ValueError("step_factor must be positive")
-    if (num_points - 1) <= 2*step_factor:
+    if (num - 1) <= 2*step_factor:
         raise ValueError(
-            "Increase num_points and/or decrease step_factor"
+            "Increase num and/or decrease step_factor"
             " or tune_centroid will never converge to a solution"
         )
     try:
@@ -2770,7 +2770,7 @@ def tune_centroid(
                          'motor': repr(motor),
                          'start': start,
                          'stop': stop,
-                         'num_points': num_points,
+                         'num': num,
                          'min_step': min_step,},
            'plan_name': 'tune_centroid',
            'hints': {},
@@ -2788,9 +2788,9 @@ def tune_centroid(
 
     @stage_decorator(list(detectors) + [motor])
     @run_decorator(md=_md)
-    def _tune_core(start, stop, num_points, signal):
+    def _tune_core(start, stop, num, signal):
         next_pos = start
-        step = (stop - start) / (num_points - 1)
+        step = (stop - start) / num
         peak_position = None
         cur_I = None
         cur_det = {}
@@ -2822,7 +2822,7 @@ def tune_centroid(
                 stop = np.clip(peak_position + step_factor*step, low_limit, high_limit)
                 if snake:
                     start, stop = stop, start
-                step = (stop - start) / (num_points - 1)
+                step = (stop - start) / num
                 next_pos = start
 
         # finally, move to peak position
@@ -2830,7 +2830,7 @@ def tune_centroid(
             # improvement: report final peak_position
             yield from mv(motor, peak_position)
 
-    return (yield from _tune_core(start, stop, num_points, signal))
+    return (yield from _tune_core(start, stop, num, signal))
 
 
 def one_nd_step(detectors, step, pos_cache):
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -2783,6 +2783,9 @@ def tune_centroid(
     else:
         _md['hints'].setdefault('dimensions', dimensions)
 
+    low_limit = min(start, stop)
+    high_limit = max(start, stop)
+
     @stage_decorator(list(detectors) + [motor])
     @run_decorator(md=_md)
     def _tune_core(start, stop, num_points, signal):
@@ -2817,6 +2820,8 @@ def tune_centroid(
                 # improvement: report current peak_position somehow
                 start = peak_position - step_factor*step
                 stop = peak_position + step_factor*step
+                start = max(min(start, high_limit), low_limit)
+                stop = max(min(stop, high_limit), low_limit)
                 if snake:
                     start, stop = stop, start
                 step = (stop - start) / (num_points - 1)
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -2731,8 +2731,6 @@ def tune_centroid(
         end of range, note: start < stop
     min_step : float
         smallest step size to use.
-        note: For EpicsMotors, ``min_step=min(min_step,motor.MRES)``
-        is used.
     num_points : int, optional
         number of points with each step size, default = 10
     step_factor : float, optional
@@ -2756,8 +2754,6 @@ def tune_centroid(
             "Increase num_points and/or decrease step_factor"
             " or tune_centroid will never converge to a solution"
         )
-    if isinstance(motor, EpicsMotor):
-        min_step = max(min_step, epics.caget(motor.prefix+".MRES"))
     _md = {'detectors': [det.name for det in detectors],
            'motors': [motor.name],
            'plan_args': {'detectors': list(map(repr, detectors)),
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -2747,6 +2747,8 @@ def tune_centroid(
     det = SynGauss('det', motor, 'motor', center=-1.3, Imax=1e5, sigma=0.05)
     RE(tune_centroid([det], "det", motor, -1.5, -0.5, 0.01, 10))
     """
+    if min_step <= 0:
+        raise ValueError("min_step must be positive")
     if step_factor <= 0:
         raise ValueError("step_factor must be positive")
     if (num_points - 1) <= 2*step_factor:
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -2736,7 +2736,7 @@ def tune_centroid(
     min_step : float
         smallest step size to use.
     num : int, optional
-        number of steps with each step size, default = 10
+        number of points with each traversal, default = 10
     step_factor : float, optional
         used in calculating range when 
         maximum is found, note: step_factor > 0, default = 2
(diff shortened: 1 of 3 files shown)


prjemian:
I believe this is ready to merge.

prjemian:
@danielballan, @tacaswell : please review

bluesky/plans.py
```diff
@@ -2691,6 +2691,148 @@ def inner_relative_adaptive_scan():
...
+            else:
+                in_range = start <= next_pos <= stop  # positive motion
+            
+            if in_range:
+                next_pos += step
+            else:
+                if sum_I == 0:
+                    return
+                peak_position = sum_xI / sum_I  # centroid
+                # TODO: report current peak_position in metadata
```
danielballan: Indeed, I don't think we have a very good story about how to get results like this out of a plan. We need one. Perhaps we can think about that in conjunction with #862.

bluesky/plans.py
```diff
@@ -2691,6 +2691,148 @@ def inner_relative_adaptive_scan():
     return (yield from inner_relative_adaptive_scan())
 
 
+def tune_centroid(
+        detectors, signal, motor, 
+        start, stop, min_step, 
+        num_points = 10,
```
danielballan: A trivial request: no spaces between keyword arguments and their default values for PEP8 compliance. I'd be happy to add that commit myself if you'd like.

prjemian: I'll get it.

bluesky/plans.py
prjemian: Right.  This case is easy since it sets the motor to the centroid, just read the motor position after RE finishes.  In general, we need a way to hook this up.  I'd want to be able to insert this content into a document.  Did not see an easy (or slightly complicated) example in the existing code.

teddyrendahl:
@prjemian Looks cool! We have some similar plans in use at SLAC. Have you thought about including any signal averaging? Our first pass looked something similar to above, but was susceptible when used on noisy signals

prjemian:
A simple smoother could be added.  For noisy signals, why not do that smoothing in an EPICS calc or swait record instead?  Seems more consistent with the other plans.

Something that does this with a smoothing factor, k, between 0 (no new values) to 1 (only new values):

```
memory = 0
with each new reading:
    value = memory * (1-k) + new_value * k
    memory = value
```

BTW, tried `tune_centroid()` on a signal consisting of a random number generator (`np.random.uniform()`) and it results in the motor being tuned to the vicinity of (start + stop)/2, as expected.  If signal is constant, then tune is exactly that value.

prjemian:
To apply that smoothing in an EPICS `swait` record, here is an example using [EPICS `swait` record support in ophyd](https://github.com/BCDA-APS/use_bluesky/blob/master/ipython_setup/startup/41-swait.py):

```
epics.caput("xxx:userCalcEnable", "Enable")
calc1 = swaitRecord("xxx:userCalc1")
calc1.reset()
calc1.channels.A.input_pv.put("xxx:userCalc1.VAL")
calc1.channels.B.value.put("0.5")   # smoothing factor
swait.calc.put("(1-B)*A+B*RNDM")
swait.scan.put("I/O Intr")
not_so_noisy = EpicsSignalRO('xxx:userCalc1', name='not_so_noisy')
```

You could connect up your noisy signal PV to, for example, channel C with these additions:

```
calc1.channels.C.input_pv.put("noisy:PV.VAL")
swait.calc.put("(1-B)*A+B*C")
```

teddyrendahl:
Doing it on the signal side is clean I agree. Unfortunately, it is harder if you want to correlate different signals together. I think this needs a solution outside the scope of this PR though. 

At SLAC, we do something very similar to this plan except we use `LiveFit` in the plan itself and query the model for the center. Seems to work well. Also allows for a little more flexibility i.e make target 90% of the max e.t.c. But no qualms with the implementation above, excited to try it out here.

danielballan:
I hope we can make it easier to incorporate LiveFit in plans in the future because I do like that approach, but I also like the directness of this approach and in particular that (unlike nonlinear least squares) it does not require an initial guess — beyond the range to scan of course.

danielballan:
Addressing the point of how to get results out of plan, what about introducing some plans that are instances of a callable class rather than a plain function?

```python
import bluesky.plans as bp
import inspect


class Foo:
    """
    Perform a scan and collect the readings (in leui of some interesting analysis.)
    """
    def __init__(self):a
        self.result = None
        
    def __call__(self, *args, **kwargs):
        result = []
        for msg in bp.scan(*args, **kwargs):
            if msg.command == 'read':
                reading = yield msg
                result.append(reading)
            else:
                yield msg
        self.result = result
    __call__.__signature__ = inspect.signature(bp.scan)

from bluesky import RunEngine
from ophyd.sim import motor, det
RE = RunEngine({})

foo = Foo()

RE(foo([det], motor, 1, 5 ,5))
foo.result  # has a bunch of collected data -- but this could an analysis result
```

prjemian:
We have begun discussion related to, but different than, this PR.  Time for new issue about this?

bluesky/plans.py
```diff
@@ -2691,6 +2691,148 @@ def inner_relative_adaptive_scan():
...
+        peak_position = None
+        cur_I = None
+        cur_det = {}
+        sum_I = 0       # for peak centroid calculation, I(x)
+        sum_xI = 0
+        
+        while abs(step) >= min_step:
+            yield Msg('checkpoint')
+            yield from mv(motor, next_pos)
+            yield Msg('create', None, name='primary')
```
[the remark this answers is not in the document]
tacaswell: Relying on `motor.name` here likely works in most cases, but is not bullet proof as the value that sets what goes into the reading for an EpicsMotor is `motor.user_readback.name`, (or generically the name of any child).  By convention we set that same value as the name of the parent, but 

```python
try:
    motor_name, = motor.hints['fields']
except (AttributeError, ValueError):
    motor_name = motor.name
```

and then using `motor_name` in the above code.

bluesky/plans.py
tacaswell: A complicated way to get this into the document stream now

```python
from ophyd.signal import Signal
sig = Signal(name='syn_signal', value=0)
...
yield from bp.mv(sig, your_value)
yield from trigger_and_read([sig], name='computed_results')
```

We do not currently have hooks into the `RunEngine` to pass a message with a document payload, but maybe that is something we should add.

When we started building this system out we had a number of long (and veering towards extensional) discussions about 'raw' data vs 'computed data' and what should be in the data coming out of the `RunEngine`.  We ended up settling on "data that you can read from hardware".   These were before we had the concept of streams inside of a run as pinned down as we do now and we have subsequently chipped away at this with `AttributeSignal` and `DerivedSignal` on ophyd objects.  

We probably should re-address this discussion in the near future, but I do not think it needs to be done for 1.0.

bluesky/plans.py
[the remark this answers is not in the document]
tacaswell: This code also needs to be defensive an check if the left hand side of returns from `yield` and `yield from` are None which is what happens if you `list(plan(...))` (or any of the other simulation methods).

danielballan:
@prjemian Fine by me! I couldn't tell from the code comments (the ones about `RE.md`) where you were hoping to solve the "get results out of plans" problem as part of this PR. I'm perfectly happy to address that separately. We'll continue this discussion in #877.

bluesky/plans.py
```diff
@@ -2691,6 +2691,148 @@ def inner_relative_adaptive_scan():
...
+                start = peak_position - step_factor*step
+                stop = peak_position + step_factor*step
+                if snake:
+                    start, stop = stop, start
+                step = (stop - start) / (num_points - 1)
+                next_pos = start
+
+        # finally, move to peak position
+        if peak_position is not None:
+            # RE.md["peak_position"] = peak_position
```
tacaswell: This is too late to make it into _this_ runs meta-data, but it will appear in the _next_ runs which would be extremely confusing.

bluesky/plans.py
```diff
@@ -2691,6 +2691,148 @@ def inner_relative_adaptive_scan():
...
+            yield from mv(motor, peak_position)
+
+    yield from _tune_core(start, stop, num_points, signal)
+    # print("Peak at ", RE.md["peak_position"])
+    return
```
tacaswell: This should be `return (yield from _tune_core(...)` as the `run_decorator` return the uid of the Start document created for the run and that will make sure it gets passed out (the RE will return the uids to the prompt, no matter what, that is a separate mechanism.

bluesky/plans.py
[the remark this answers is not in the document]
prjemian: @tacaswell : re: "... could be replaced by ...", how does the motor get moved in the suggested code?  Perhaps you meant replace starting from `yield Msg('create', None, name='primary')` through `yield Msg('save')`?  With that, I agree, shorter and cleaner.  Like this:

	ret = (yield from trigger_and_read(detectors + [motor]))
	cur_I = ret[signal]['value']
	sum_I += cur_I
	position = ret[motor_name]['value']
	sum_xI += position * cur_I

bluesky/plans.py
prjemian: Agreed.  Will mark this as a possible improvement.  And note in the docstring how to retrieve the result from the motor position (in keeping with the philosophy that plan results are retrieved from hardware).

bluesky/plans.py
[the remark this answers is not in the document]
tacaswell: Sorry for not being clear, I just meant the reading part

bluesky/plans.py
```diff
@@ -2756,8 +2760,12 @@ def tune_centroid(
             "Increase num_points and/or decrease step_factor"
             " or tune_centroid will never converge to a solution"
         )
+    try:
+        motor_name, = motor.hints['fields']
+    except (AttributeError, ValueError):
+        motor_name = motor.name
     _md = {'detectors': [det.name for det in detectors],
-           'motors': [motor.name],
+           'motors': [motor_name],
```
tacaswell: This one should still be `motor.name` so that it matches the device names in the descriptor!

bluesky/plans.py
```diff
@@ -2817,6 +2820,8 @@ def _tune_core(start, stop, num_points, signal):
                 # improvement: report current peak_position somehow
                 start = peak_position - step_factor*step
                 stop = peak_position + step_factor*step
+                start = max(min(start, high_limit), low_limit)
```
tacaswell: `np.clip` may be more readable here?

bluesky/plans.py
```diff
@@ -2691,6 +2691,150 @@ def inner_relative_adaptive_scan():
...
+        raise ValueError(
+            "Increase num_points and/or decrease step_factor"
+            " or tune_centroid will never converge to a solution"
+        )
+    try:
+        motor_name, = motor.hints['fields']
+    except (AttributeError, ValueError):
+        motor_name = motor.name
+    _md = {'detectors': [det.name for det in detectors],
+           'motors': [motor_name],
```
tacaswell: sorry for the double comment, the GH UI is being weird for me and hidding the other one as out-dated.

This should still be `motor.name` so that it can be used to look up the device in the descriptor (where the configuration is keyed on the name of the parent).

tacaswell:
Wait, is this 3-space indented?

tacaswell:
:sheep: NM, I can not count

danielballan:
One last question -- why make `num` mean 'number of strides' here when it means 'number of points' in all existing plans?

I won't take a position on which way is "right" -- it seems like a religious argument -- but the original choice was made to align with the `num` argument to `np.linspace`. At this point consistency seems desirable, but if you have a reason that strides makes more sense in this particular case I won't raise strong objections.

prjemian:
willing to adjust the implementation for uniformity, don't pull yet

prjemian:
OK, now ready for merge.  Wahoo!

danielballan:
Looks great, thanks! Waiting for Travis-CI to give us the green light....

tacaswell:
:tada:

PR #879, opened 2017-10-26
FIX: show live figure if made internally

CJ-Wright:
Call show when using an internally generated figure (like the image callbacks)


tacaswell:
Thanks.  This does have a weird edge case if interactive mode is not on in matplotlib (as the show will block holding up everything!) but if it does that it will be very obvious to the user!

PR #883, opened 2017-10-26
docstring fix

jrmlhermitte:
OCD fix
(maybe should also be `RE=RE`. also would suggest emphasizing that this is not necessary and `RE.subscribe` can be called after the fact as well)


tacaswell:
It actually _must_ be `RE=RE` as it is a keyword-only arg!

jrmlhermitte:
urgh yes you're right :-| oops!

tacaswell:
Just noticed this is your first bluesky PR! :tada:

jrmlhermitte:
:laughing:

PR #887, opened 2017-10-31
Making default plan names shorter

### LINKED ISSUE #740: Plan names are too long
We have received feedback that it would be nice to provide shorter names for plans like `relative_outer_product_scan`. I think we should consider changing the names we document and teach. There is no reason we can't support aliases for the existing names indefinitely, so this would not break user code.

Some ideas:

count (fine as is)
scan (fine)
relative_scan -> rel_scan
inner_product_scan -> joint_scan
outer_product_scan -> grid_scan

cc @stuwilkins

stuartcampbell:
The default plan names are being changed to something shorter.

fixes #740 

## Description
`outer_product_scan` --> `grid_scan`
`relative_outer_product_scan` --> `rel_grid_scan`
more to follow


stuartcampbell:
Just to check, I presume that the `plan_pattern` still remains as inner or outer ?

PR #888, opened 2017-10-31
Comparison between current bluesky algorithm and the one from chx on peak stats

licode:
comparison with chx work is added.

### DIFF
--- a/bluesky/callbacks/fitting.py
+++ b/bluesky/callbacks/fitting.py
@@ -199,7 +199,7 @@ class PeakStats(CollectThenCompute):
     cen : mid-point between half-max points on each side of the peak
     max : x location of y maximum
     min : x location of y minimum
-          crossings : crosses between y and middle line, which is
+    crossings : crosses between y and middle line, which is
           ((np.max(y) + np.min(y)) / 2). Users can estimate FWHM based
           on those info.
     fwhm : the computed full width half maximum (fwhm) of a peak.
--- a/bluesky/callbacks/fitting.py
+++ b/bluesky/callbacks/fitting.py
@@ -199,6 +199,10 @@ class PeakStats(CollectThenCompute):
     cen : mid-point between half-max points on each side of the peak
     max : x location of y maximum
     min : x location of y minimum
+    fwhm : the computed full width half maximum (fwhm) of a peak.
+    This is done by computing the number of times the data crosses the median.
+    More than two crosses may exist for multiple peaks or very noisy data.
+    The distance between the first and last crossing is taken to be the fwhm.
     """
 
     def __init__(self, x, y, edge_count=None):
@@ -286,4 +290,3 @@ class PeakStats(CollectThenCompute):
 
         # reset y data
         y = self.y_data
-        # insert lmfit
--- a/bluesky/callbacks/fitting.py
+++ b/bluesky/callbacks/fitting.py
@@ -199,10 +199,12 @@ class PeakStats(CollectThenCompute):
     cen : mid-point between half-max points on each side of the peak
     max : x location of y maximum
     min : x location of y minimum
+          crossings : crosses between y and middle line, which is
+          ((np.max(y) + np.min(y)) / 2). Users can estimate FWHM based
+          on those info.
     fwhm : the computed full width half maximum (fwhm) of a peak.
-    This is done by computing the number of times the data crosses the median.
-    More than two crosses may exist for multiple peaks or very noisy data.
-    The distance between the first and last crossing is taken to be the fwhm.
+           The distance between the first and last crossing is taken to
+           be the fwhm.
     """
 
     def __init__(self, x, y, edge_count=None):
@@ -213,6 +215,7 @@ class PeakStats(CollectThenCompute):
         self.max = None
         self.min = None
         self.nlls = None
+        self.crossings = None
         self.fwhm = None
         self.lin_bkg = None
         self._edge_count = edge_count
@@ -232,6 +235,7 @@ class PeakStats(CollectThenCompute):
         self.max = None
         self.min = None
         self.nlls = None
+        self.crossings = None
         self.fwhm = None
         self.lin_bkg = None
 
@@ -284,9 +288,9 @@ class PeakStats(CollectThenCompute):
 
         if _cen_list:
             self.cen = np.mean(_cen_list)
-
-        if len(_cen_list) >= 2:
-            self.fwhm = np.abs(_cen_list[-1] - _cen_list[0], dtype=float)
-
+            self.crossings = np.array(_cen_list)
+            if len(_cen_list) >= 2:
+                self.fwhm = np.abs(self.crossings[-1] - self.crossings[0],
+                                   dtype=float)
         # reset y data
         y = self.y_data
--- a/bluesky/callbacks/fitting.py
+++ b/bluesky/callbacks/fitting.py
@@ -282,7 +282,7 @@ class PeakStats(CollectThenCompute):
             self.cen = np.mean(_cen_list)
 
         if len(_cen_list) >= 2:
-            self.fwhm = float(_cen_list[-1] - _cen_list[0])
+            self.fwhm = np.abs(_cen_list[-1] - _cen_list[0], dtype=float)
 
         # reset y data
         y = self.y_data
--- a/bluesky/callbacks/fitting.py
+++ b/bluesky/callbacks/fitting.py
@@ -281,8 +281,8 @@ class PeakStats(CollectThenCompute):
         if _cen_list:
             self.cen = np.mean(_cen_list)
 
-        if len(_cen_list) == 2:
-            self.fwhm = float(_cen_list[1] - _cen_list[0])
+        if len(_cen_list) >= 2:
+            self.fwhm = float(_cen_list[-1] - _cen_list[0])
 
         # reset y data
         y = self.y_data
--- a/bluesky/callbacks/fitting.py
+++ b/bluesky/callbacks/fitting.py
@@ -273,8 +273,8 @@ class PeakStats(CollectThenCompute):
             _x = x[cr:cr+2]
             _y = y[cr:cr+2] - mid
 
-            dx = np.diff(_x)
-            dy = np.diff(_y)
+            dx = np.diff(_x)[0]
+            dy = np.diff(_y)[0]
             m = dy / dx
             _cen_list.append((-_y[0] / m) + _x[0])
 
--- a/bluesky/tests/test_scientific.py
+++ b/bluesky/tests/test_scientific.py
@@ -1,0 +1,19 @@
+import numpy as np
+from bluesky.plans import scan
+from bluesky.examples import motor, noisy_det, det
+from bluesky.callbacks.scientific import PeakStats
+
+def test_peak_statistics(fresh_RE):
+    """peak statistics calculation on simple gaussian function
+    """
+    RE = fresh_RE
+    x = 'motor'
+    y = 'det'
+    ps = PeakStats(x, y)
+    RE.subscribe(ps)
+    RE(scan([det], motor, -5, 5, 100))
+
+    assert ps.cen == 0
+    assert ps.com == 0
+    fwhm_gauss = 2*np.sqrt(2*np.log(2)) # theoretical value with std=1
+    np.allclose(ps.fwhm, fwhm_gauss, atol=1e-2)
--- a/bluesky/tests/test_scientific.py
+++ b/bluesky/tests/test_scientific.py
@@ -1,19 +1,89 @@
 import numpy as np
 from bluesky.plans import scan
-from bluesky.examples import motor, noisy_det, det
-from bluesky.callbacks.scientific import PeakStats
+from ophyd.sim import motor, noisy_det, det, SynGauss
+from bluesky.callbacks.fitting import PeakStats
 
-def test_peak_statistics(fresh_RE):
+
+def get_ps(x, y, shift=0.5):
+    """ peak status calculation from CHX algorithm.
+    """
+    ps = {}
+    x=np.array(x)
+    y=np.array(y)
+
+    PEAK=x[np.argmax(y)]
+    PEAK_y=np.max(y)
+    COM=np.sum(x * y) / np.sum(y)
+    ps['com'] = COM
+    ### from Maksim: assume this is a peak profile:
+    def is_positive(num):
+        return True if num > 0 else False
+
+    # Normalize values first:
+    ym = (y - np.min(y)) / (np.max(y) - np.min(y)) - shift  # roots are at Y=0
+
+    positive = is_positive(ym[0])
+    list_of_roots = []
+    for i in range(len(y)):
+        current_positive = is_positive(ym[i])
+        if current_positive != positive:
+            list_of_roots.append(x[i - 1] + (x[i] - x[i - 1]) / (abs(ym[i]) + abs(ym[i - 1])) * abs(ym[i - 1]))
+            positive = not positive
+    if len(list_of_roots) >= 2:
+        FWHM=abs(list_of_roots[-1] - list_of_roots[0])
+        CEN=list_of_roots[0]+0.5*(list_of_roots[1]-list_of_roots[0])
+
+        ps['fwhm'] = FWHM
+        ps['cen'] = CEN
+
+    else:    # ok, maybe it's a step function..
+        print('no peak...trying step function...')
+        ym = ym + shift
+        def err_func(x, x0, k=2, A=1,  base=0 ):     #### erf fit from Yugang
+            return base - A * erf(k*(x-x0))
+        mod = Model(  err_func )
+        ### estimate starting values:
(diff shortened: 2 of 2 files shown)


bluesky/callbacks/scientific.py
```diff
@@ -167,8 +167,8 @@ def compute(self):
         if _cen_list:
             self.cen = np.mean(_cen_list)
 
-        if len(_cen_list) == 2:
-            self.fwhm = float(_cen_list[1] - _cen_list[0])
+        if len(_cen_list) >= 2:
```
licode: For noisy data, we will get cen_list lengh is larger than 2. If we define it equal to 2, we get None as return on fwhm. 

This is consistent with chx algo now. https://github.com/NSLS-II-CHX/ipython_ophyd/blob/d2389af73506d6cf2f70e95a336b77a16267381f/profile_collection/startup/95-utilities.py#L275

bluesky/tests/test_scientific.py
```diff
@@ -0,0 +1,19 @@
...
+    """peak statistics calculation on simple gaussian function
+    """
+    RE = fresh_RE
+    x = 'motor'
+    y = 'det'
+    ps = PeakStats(x, y)
+    RE.subscribe(ps)
+    RE(scan([det], motor, -5, 5, 100))
+
+    assert ps.cen == 0
```
jrmlhermitte: maybe `np.isclose(ps.cen, 0, atol=1e-6)` or something similar? Just in case since they're floating point.

bluesky/callbacks/scientific.py
danielballan: Good catch! I notice that CHX's algo takes the absolute value. Would that be appropriate here as well?

bluesky/tests/test_scientific.py
licode: good catch

danielballan: Might be a good place to use https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.testing.assert_almost_equal.html although `isclose` is fine too.

bluesky/callbacks/scientific.py
jrmlhermitte: yeah good catch. maybe it's also worth doing the same for `self.cen`. if `_cen_list > 2`, it might not make sense anymore.
I'd be tempted (but this is probably overkill, also writing this naively.):
```python
_cen_list = _cen_list.sort()
self.cens = (_cen_list[1:] + _cen_list[:-1])*.5
self.fwhms = np.diff(_cen_list)
# choose the cen or peak, but allow user to look at the others? 
self.cen = self.cens[0]
self.fwhms = self.fwhm[0]
```

licode:
For noisy gaussian curve, the two algos look the same to me with accuracy high than 1e-6.

danielballan:
Great. This is exactly what I was hoping to get. Now if CHX encounters problems in the future, we can easily run both implementations against the same data.

licode:
After we merge this, I will add step function part. Or in the same PR?

danielballan:
I think it's better to do the step part separately. It will need documentation, more tests, and some API consideration.

bluesky/callbacks/scientific.py
licode: Good suggestion. I think this may work for late update when we want to deal with multiple peaks. 

But for single peak with lots of noise, the middle curve cuts multiple points on both the left and right of the peak. That's what we are focusing now. And this will fail I think.

jrmlhermitte: ok sounds good. i agree. this PR should be restricted to the `self.fwhm=None` issue. maybe a PR should be made following this one?
Great tests!!

bluesky/tests/test_scientific.py
```diff
@@ -83,5 +83,5 @@ def test_peak_statistics_compare_chx(fresh_RE):
     ps_chx = get_ps(ps.x_data, ps.y_data)
 
     np.allclose(ps.cen, ps_chx['cen'], atol=1e-6)
```
danielballan: How could this test fail?

danielballan:
This will need a big rebase on #892. Apologies, @licode.

licode:
Do you want to merge this one first, then 892?

licode:
I give loose test on cen, as for both chx and bluesky algorithm, I find this value varies a lot depending on the data. However, center of mass is very stable and a good estimation for noisy data.

danielballan:
No, #892 is critical for 1.0.0 (because it is an API change) whereas we really only _need_ this in time for the next cycle. I expect we will tag at least one more release before we roll out a wide update at NSLS-II, and this can easily be merged in time for that.

Have you considered using a random seed to make the tests deterministic?

licode:
squashed commits.

fuzz/README
```diff
@@ -1,2 +1,2 @@
 `fuzz.py` is attempting to break the RunEngine by throwing it a random
-sequence of messages. Execute the fuzz testing with `python fuzz.py`.
\ No newline at end of file
+sequence of messages. Execute the fuzz testing with `python fuzz.py`.
```
danielballan: Seems like your editor did something funny here.

bluesky/callbacks/fitting.py
```diff
@@ -281,8 +281,8 @@ def compute(self):
         if _cen_list:
             self.cen = np.mean(_cen_list)
 
-        if len(_cen_list) == 2:
-            self.fwhm = float(_cen_list[1] - _cen_list[0])
+        if len(_cen_list) >= 2:
```
danielballan: The commit message labels this a bug fix. Was `== 2` a bug? Itseems designed to silence the FWHM computation if the half-max crossings aren't clean and obvious.

licode: I think implementing this way not only covers the previous case when _cen_list == 2, but also extends the capability to deal with noisy data where crossings aren't clean and obvious, or more than 2. For noisy data, FWHM is difficult to get but this calculation can at least provide an initial guess.

danielballan: OK, that's reasonably compelling.

danielballan:
This seems close to ready, but the tests are currently failing. Can you revisit @licode?

licode:
Let me double check this to make sure test can pass.

licode:
Tests failed at test_disable? It runs ok for test_scientific.

danielballan:
We do a couple bec-related tests under the umbrella of `test_disable`, including this line which causes the failure:

```
assert ast.literal_eval(repr(bec.peaks)) == vars(bec.peaks)
```

It looks like this PR puts something in `bec.peaks` that breaks the eval-able repr. I can't think we this change would need to break that test. What does `bec.peaks` look like now if you run the tests locally?

licode:
I got the same error locally as shown in travis for test_bec.

danielballan:
That makes sense. The next step is to debug it. What does `bec.peaks` look like?

licode:
```
{
'com':
    {'det_a': 2.8761155398324294}
,
'cen':
    {'det_a': 3.5186421247177457}
,
'max':
    {'det_a': (4.0,
               0.77601445129398539)}
,
'min':
    {'det_a': (3.0,
               0.42649034303669409)}
,
'fwhm':
    {'det_a': array([ 2.36772926])}
,
'nlls':
    {'det_a': None}
,
}
```
ast.literal_eval doesn't know how to translate array([...]). I think it returns fwhm as list somewhere. Let me check.

bluesky/callbacks/fitting.py
danielballan: Could we second a second pair of eyes on this before merging? Seem OK @mrakitin @jrmlhermitte  @tacaswell? Too brittle?

danielballan:
That seems like the right fix to me. I'd like to get a second opinion on our discussion about FWHM before this goes in, just to be sure.

bluesky/callbacks/fitting.py
mrakitin: I like this implementation when we consider only external points better. As I mentioned a few months ago, eventually we may come up with two different definitions of FWHM:
- local FWHM, which will essentially be a list of FWHM values for each found peak (need a reliable peaks finder, e.g. https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.signal.find_peaks_cwt.html)
- global FWHM, which will calculate the width between two external peaks. Let's devote a separate PR for that.

bluesky/callbacks/fitting.py
```diff
@@ -273,16 +273,16 @@ def compute(self):
...
             m = dy / dx
             _cen_list.append((-_y[0] / m) + _x[0])
 
         if _cen_list:
             self.cen = np.mean(_cen_list)
 
-        if len(_cen_list) == 2:
-            self.fwhm = float(_cen_list[1] - _cen_list[0])
+        if len(_cen_list) >= 2:
+            self.fwhm = np.abs(_cen_list[-1] - _cen_list[0], dtype=float)
```
jrmlhermitte: Thanks @danielballan  I haven't used this. However, looking at this, I have a general set of comments:

1. one probably should document `fwhm` in line 201.

2. personally, I'd argue that it doesn't make sense (and might confuse the user) when more than two crossings are found. But it's probably better to output more data than nothing. I would add more to the comment so the user can know that this has been computed on data that contained more than two crossings. How about including like @mrakitin suggested a `fwhm_extended` and `fwhm`. Maybe `fwhm` and `fwhm2`?
Or at least something like `ncrossings`, which is the number of crossings found. But that may be overkill.
Another thing I noticed, if `cen_list < 2` we should probably set `self.fwhm = None` since this is a re-usable object (and same for `cen`).


3. In passing, I also found something that may be a little awkward. Why not use:
```py
def center_of_mass(y, x=None):
if x is None:
    x = np.arange(len(y))
return np.sum(x*y)/np.sum(y)
```
instead? There may be a reasoning I'm not seeing. (maybe set `y = np.abs(y)` as well)


I think for this PR, just documenting the fwhm, and setting `fwhm` and `cen` to `None` where appropriate is probably enough. Suggested documentation:
```
fwhm : the computed full width half maximum (fwhm) of a peak. 
This is done by computing the number of times the data crosses the median.
The distance between the first and last crossing is taken to be the fwhm.
```

You could maybe add something like crossings (just suggestion)
```
crossings : the points at which the data crosses the median
```
just for completeness. (i.e. what does my `fwhm` mean? for one peak, or for multiple crossings?

licode:
Thanks for all the suggestion! Doc on fwhm is added. I think fwhm is set as None as default. I agree that it will be better for another PR to deal with complicated cases like local and global FWHM together with crossings.

danielballan:
Given that there is no immediate rush to get this merged, I would rather see all three of those points above addressed in this PR, while we have the details fresh in mind. I like the idea of ncrossings and fwhm_extended. And while it is true that fwhm is None by default at init time, it looks to me like it retains a stale value from previous runs and needs to be reset to None again if the number of crossings is less than 2, as Julien pointed out.

As an aside, unrelated to the scope of this PR, this kind of bug (stale attributes from previous runs) could be avoided if we stopped subscribing CallbackBase classes directly and instead subscribe factory functions that regenerate a new instance per run. That is — assert that CallbackBase is “single use” and blows up if it is handed a second start document. This change has been verbally proposed by @tacaswell and I thought this moment was a good time to highlight that sometimes he is right about things. :-D

jrmlhermitte:
I agree with this just want to quickly add:
1. @danielballan suggests `ncrossings` while I suggested `crossings`. I prefer @danielballan idea (saving a number rather than a list)
2. There are cases where we may want state, we have one time use factories and accumulators (say a histogram plotter, or image stitcher). This might be something to think of as I think we may want to handle each separately (I think I can see the accumulators simply returning `self` whereas the factories return some new instance, but perhaps it may be a little more involved to implement)

licode:
OK. As more people vote for this, I will keep on adding work to this PR. I also find fwhm is set as None in compute() function, so i assume this is done to reset state value each time?

I will use fwhm and fwhm_extended, or fwhm_local and fwhm_global? any preference?

danielballan:
> I will use fwhm and fwhm_extended, or fwhm_local and fwhm_global? any preference?

We should keep the name `fwhm` in there for back-compatibility. I have no strong opinion on what to call the other one. Maybe `fwhm_permissive` or `fwhm_tolerant`?

mrakitin:
+1 for `fwhm` and `fwhm_extended`.

licode:
any idea on self.nlls ? No doc for this attribute.

mrakitin:
https://www.acronymfinder.com/Non_Linear-Least-Squares-(usually-seen-as-NLLS)-(NLLSQ).html?

jrmlhermitte:
no clue, but just noticed that the variables are already cleared so that comment doesn't need to be addressed. sorry, that was a "late at night after running 4 miles" comment :-P

danielballan:
My follow-up comment was a "I'm in bed reading this on my phone and it's worth a second look" comment. Glad you were right, @licode!

Yes, nlls is non-linear least squares and, as you can see, it is currently just a skeleton. It's always `None`. It should be documented as "not implemented" or implemented (but not necessarily in this PR).

licode:
if ncrossings is even number, the calculation is straight forward for local fwhm. however, we may come across the case when ncrossings is odd number? any idea? Or I just ignore the case when it is odd, with try.. except...

danielballan:
Hmm. Using try/except to give up based on parity seems fragile in a way that I would find surprising as a user. I had in mind that we would always use the outer-most zero-crossings, tacitly assuming that the inner ones are noise. What did others have in mind?

licode:
How about this?

if ncrossings is even, we just calculate fwhm as an array of individual each peak width.
if ncrossing is odd, fwhm = fwhm_extended = the difference of outmost crossing points. We assume within points are just noise, so fwhm equals to fwhm_extended in this case.

any comment?

licode:
any other issue is assert ast.literal_eval(repr(bec.peaks)) == vars(bec.peaks) from test_bec.py

if fwhm is an array of individual peak. ast.literal_eval(repr(bec.peaks)) fails as previous reported. array can't be recognized. 
```
'fwhm':
    {'det_a': array([ 2.36772926])}
```

we need to better implement ast.literal_eval(repr(bec.peaks)) ?

jrmlhermitte:
I vote to keep `fwhm` as one number, not an array (so if we do look at case > 2 then it should just be the largest distance). I worry this might break existing use (am I wrong that this is being used yet?)

I would also prefer to keep a variable called `self.crossings` as opposed to `self.fwhm_list`. But I will let you decide, these are just my two cents.

danielballan:
There is no way for us to change what literal_eval does; that is a proxy for how Python parsing itself works. Let’s just give one scalar value — isn’t that what users are likely to expect?

I was thinking that fwhm should retain the old, strict behavior, where it is None if there are not exactly two crossings. Anything else would break back-compatibility. This has been in fairly wide use for awhile now, so we should only change it if we consider the current behavior to be a bug.

The point of adding a separate one, fwhm_extended, is to suggest, “Here is a best guess that will exist for any number of crossings >= 2, but it might be confused by noise. Use at your own risk!”

Am I on the same page?

licode:
I agree it is better to keep fwhm as a scaler to follow standard definition. It may be a good idea to keep self.crossings, so users can at least estimate peak width by themselves for complicated cases. Will make decision next week.

mrakitin:
Just to give an idea from the Shadow3 ray tracing package: https://github.com/radiasoft/shadow3/blob/d419cb3e2e66b790566081ac47661bce8ba89e7b/ShadowToolsPrivate.py#L154-L156 - they seem to like the external crossings. I'm seconding on keeping `fwhm` as a number.

licode:
I set fwhm as a scaler. In order not to cause any confusion, fwhm_extended or others are not used. I added self.crossings to indicate all the cross points. So users can have better estimation by themselves.

mrakitin:
In #1002 I had to modify the FWHM to be [a dict](https://github.com/NSLS-II/bluesky/pull/1002/files#diff-177dabe93bbee02f265f7408541d9964R285), having coordinates which were used for calculation of FWHM to plot that correctly. Probably should be considered either here or in my PR.

bluesky/callbacks/fitting.py
```diff
@@ -199,6 +199,12 @@ class PeakStats(CollectThenCompute):
     cen : mid-point between half-max points on each side of the peak
     max : x location of y maximum
     min : x location of y minimum
+          crossings : crosses between y and middle line, which is
```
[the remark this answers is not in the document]
licode: good catch.

danielballan:
Thanks, @licode.

To summarize:

1. A new automated test ensures that bluesky's results agree with the results from @lwiegart's peak-fitting code. This allows to systematically check that bluesky's results are up to par. Presently, only a simple noisy Gaussian is tested; other cases will be easy to add in the future, as desired.
2. As discussed yesterday in a meeting with @cmazzoli, we will next extend PeakStats to handle N-dimensional data. This work will proceed in a separate PR, led by @licode. We will handle the straightforward cases first (min, max, com) and the more complex cases in a second round (fwhm, cen).

PR #893, opened 2017-10-31
API: re-organize / split plans.py for developer convenience 

### LINKED ISSUE #850: Split plans module into separate modules

tacaswell:
This will be force-pushed to

The nominal goal is to get all of the local imports into plans.py namespaced so tab-complete only gives you fully boxed plans.

@teddyrendahl @ZLLentz I suspect this will cause you a lot of breakage.

cntx.py is likely to be deleted, all of it's contents were already deprecated.

closes #850

### DIFF
--- a/bluesky/__init__.py
+++ b/bluesky/__init__.py
@@ -7,25 +7,7 @@ from .utils import IllegalMessageSequence
 from .utils import FailedStatus
 
 from .run_engine import RunEngine
-from .plans import SupplementalData
-
-# for back-compat
-from .plans import PlanBase
-from .plans import Count
-from .plans import AbsListScanPlan
-from .plans import DeltaListScanPlan
-from .plans import AbsScanPlan
-from .plans import LogAbsScanPlan
-from .plans import DeltaScanPlan
-from .plans import LogDeltaScanPlan
-from .plans import AdaptiveAbsScanPlan
-from .plans import AdaptiveDeltaScanPlan
-from .plans import PlanND
-from .plans import InnerProductAbsScanPlan
-from .plans import InnerProductDeltaScanPlan
-from .plans import OuterProductAbsScanPlan
-from .plans import OuterProductDeltaScanPlan
-from .plans import Tweak
+from .preprocessors import SupplementalData
 
 from ._version import get_versions
 __version__ = get_versions()['version']
--- a/bluesky/cntx.py
+++ b/bluesky/cntx.py
@@ -1,0 +1,254 @@
+from functools import wraps
+from contextlib import contextmanager
+import warnings
+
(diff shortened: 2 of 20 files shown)


tacaswell:
The selling point of these namespace is that you can do

```python
from bluesky import plns as bp
from bluesky import rplns as rbp
```

and then `bp.<tab>` or `rbp.<tab>` gives you reasonable (<15 entry) lists of options to make them discover-able.  

We keep back-api on everything else and if you want to import individually `from bluesky.plans import (count, grid_scan)` you can as well.

I think this is the easiest path to A) not breaking anything B) providing curated namespaces

danielballan:
Yes, but as you know, users who might reasonably see that code you wrote and then try:

```python
from bluesky.plns import scan
from bluesky.rplns import relative_scan
```

may well be surprised to find that they don't work!

What is scans.py? It's midnight before 1.0.0. Let's stick to the plan.

tacaswell:
The point of this whole exercise is to give a nice currated namespace,  if we are going to support the long plan names forever out of `bluesky.plans` the option are either this or making yet another module just to hold some imports.

Scaling those ambitions back to slightly better sorted modules for devs.

danielballan:
>  if we are going to support the long plan names forever out of `bluesky.plans`

As a collaboration, I think we can weight the benefits of a pristine namespace against the cost of slowly retiring the old names. Creating yet-another-namespace -- and one with import behavior that experienced Python users might find surprising -- is not something I want to do unilaterally the night before a release.

danielballan:
The controversial parts of this have been separated off into #896 and marked for consideration for a later release. Thanks, @tacaswell :- )

The scope of this PR is now in line with what folks signed off on in #850. Note that, in that issue, I originally proposed, "we can shim imports into plans.py with a warning about the new location." We have not actually done that here, judging that it would be better to fix the errors than generate a big slew of warnings.

If we have misjudged that, we can always add shims in a quick patch release and then walk the changes out more gradually. But I think we're correctly following the spirit on the consensus in #850.

PR #894, opened 2017-11-01
Making default plan names shorter

### LINKED ISSUE #740: Plan names are too long

stuartcampbell:
The default plan names are being changed to something shorter.
fixes #740

## Description
`relative_scan` --> `rel_scan`
`outer_product_scan` --> `grid_scan`
`relative_outer_product_scan` --> `rel_grid_scan`
`relative_spiral_fermat` -> `rel_spiral_fermat`
`relative_spiral` -> `rel_spiral`
`relative_list_scan` -> `rel_list_scan`
`relative_log_scan` -> `rel_log_scan`
`relative_adaptive_scan` -> `rel_adaptive_scan`


danielballan:
Thanks again for the hand, @stuartcampbell. I want to emphasize for anyone following along: **the old plan names will continue to work forever**, aliased in the source code. But we will make these new names the ones we document prominently and teach because we think they will make a better first impression on new users who are a bit overwhelmed by the verbosity in general.

stuartcampbell:
I was thinking, do you think we should add a simple test to make sure the legacy names continue to work ?

tacaswell:
I am having second thought about supporting them forever, they are going to clutter the namespace

danielballan:
Well, at worst we can move them through a _very slow_ deprecation cycle.

teddyrendahl:
@danielballan Just curious. Is this direct feedback you have that names are too long? I always thought longer more descriptive names benefited newer users, and shorter names were better for people who have to type this stuff hundreds of times per day?

danielballan:
@teddyrendahl Yes. The lengthy names virtually always come up when giving tutorials to non-computational people. "Users will never type all that." Eventually they come around when they realize that tab-complete really does help, but it's been a consistent gripe during the first couple months of adoption.

I hope that these are still fairly comprehensible. The only abbreviation in play is 'rel' for 'relative', which at least is less cryptic that 'dscan' in my opinion.

Does this seem like a mistake? Dissenting opinion would be useful to hear.

teddyrendahl:
Well. I like the lengthy descriptive names. You can be very explicit about what is happening. There are a ton of ways to get around long names (alias on import, tab-complete, e.t.c). I worry that scientists may complain about long names, but will suffer the consequences of readability when the code turns into a bunch of abbreviations and acronyms
```
bs.RE(bs.rpln.rel_log(...))
```
 However, something something about the customer always being right. If users have an issue then they have an issue and who am I to tell them what they like better or don't. I can always re-alias the other way.

teddyrendahl:
I also agree. Changing `relative` to `rel` is not a huge deal. Avoid `dscan` `ascan2` at all costs.

PR #896, opened 2017-11-01
ENH: put curated namespaces at the top level

tacaswell:
## Description

The selling point of these namespace is that you can do

```python
from bluesky import plns as bp
from bluesky import rplns as rbp
```

and then bp.<tab> or rbp.<tab> gives you reasonable (<15 entry) lists of options to make them discover-able.

The only downside is that 

```python
frob bluesky.plns import count
```

will not work is `plns` is not a module

### DIFF
--- a/bluesky/__init__.py
+++ b/bluesky/__init__.py
@@ -1,4 +1,5 @@
 import logging
+from types import SimpleNamespace
 logger = logging.getLogger(__name__)
 
 from .utils import Msg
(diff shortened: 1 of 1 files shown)


danielballan:
Any thoughts from Team SLAC and Team APS?

teddyrendahl:
Well. We don't have enough non-developers for this to really hurt us. 

That being said, it would be nice to be more explicit in the naming so that people realize they are not submodules, and are in fact namespaces. How do you feel about `plan_list` and `rplan_list` or something similar? I realize this is a little more verbose.

Thoughts @ZLLentz?

ZLLentz:
Can I have more context about the use cases for these namespaces? I can't think of a situation where I'm assembling scans in an IPython session, so I assume I'm missing something. I always work under the assumption that we carefully choose what to scan prior to the experiment, so having something that looks like a module but is not can only be confusing.

ZLLentz:
I found something interesting today (maybe a bit hacky). It's possible to add something like:
```
import sys

sys.modules['bluesky.plns'] = plns
sys.modules['bluesky.rplns'] = rplns
```
And then Python is tricked into thinking your two namespaces are real modules that have already been imported. A user can then do:
```
from bluesky.plns import count
```
and it will work as they expect. Python will say "Aha, I've already imported this! Let me grab my object from `sys.modules` and do a quick `getattr` for you."

I'm not sure if this an acceptable solution to the import from namespace problem, but I thought it was an interesting workaround. As I've been using `bluesky` more, I've started to understand the benefits of having namespaces like these that I didn't see back in November.

ZLLentz:
Fair enough Dan, I agree with your points. I'm planning to use this for something completely unrelated, but I noticed that this PR was still sitting open so I thought I'd share it here as well.

danielballan:
Closed as stale and deadlocked, please reopen if this is still a going concern.

stan-dot:
At the moment this is the behavior:
```
import bluesky.preprocessors as bpp
```

it seems to work well for DLS in regards to developer ergonomics and sounds roughly equivalent to the option suggested.

action suggestion: close the PR

PR #902, opened 2017-11-01
Added generated to toctree in state-machine

jrmlhermitte:
@danielballan


danielballan:
With this, now all of the .rst files generated by sphinx's 'autosummary' extension are stored under the directory `source/generated` rather than being dumped directly in `source/` alongside version-controlled files.

PR #906, opened 2017-11-01
TST: Add test to check for legacy plan names

stuartcampbell:
Add a test to check for that legacy plan names exist.

### DIFF
--- a/bluesky/tests/test_legacy_plans.py
+++ b/bluesky/tests/test_legacy_plans.py
@@ -8,5 +8,5 @@ def test_legacy_plan_names():
     assert bp.relative_spiral is bp.rel_spiral
     assert bp.relative_spiral_fermat is bp.rel_spiral_fermat
     assert bp.relative_list_scan is bp.rel_list_scan
-    assert relative_log_scan is rel_log_scan
-    assert relative_adaptive_scan is rel_adaptive_scan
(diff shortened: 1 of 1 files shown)


bluesky/tests/test_legacy_plans.py
```diff
@@ -0,0 +1,12 @@
...
+import bluesky.plans as bp
+
+def test_legacy_plan_names():
+    assert bp.outer_product_scan is bp.grid_scan
+    assert bp.relative_outer_product_scan is bp.rel_grid_scan
+    assert bp.relative_scan is bp.rel_scan
+    assert bp.relative_spiral is bp.rel_spiral
+    assert bp.relative_spiral_fermat is bp.rel_spiral_fermat
+    assert bp.relative_list_scan is bp.rel_list_scan
+    assert relative_log_scan is rel_log_scan
```
danielballan: Tests fail because `bp.` is missing here and on the line below.

stuartcampbell: Doh, sorry.  That's what you get for rushing on the way to a meeting with the facility director! :)

danielballan: :- D

danielballan:
Thanks!

danielballan:
Thanks!

PR #907, opened 2017-11-01
FIX: pseudo reset wrapper

### LINKED ISSUE #905: BUG: the 'reset_positions' step of relative_inner_product_scan doesn't merge multi-axis propertly
It seems to direct both the parent object and then the child object to move.

tacaswell:
If parent is in the set of things to be reset, do not reset the
children too.

closes #905


danielballan:
@tacaswell: fastest guns in the west.

PR #913, opened 2017-11-06
Abbreviate UID with 8 chars instead of 6.

danielballan:


danielballan:
Some context for this, belatedly: We have started to see collisions between the first six characters. Time to bump the default abbreviation to eight.

attn @gjwillms

jrmlhermitte:
looks good!

Issue #868, opened 2017-10-23
Plotting problems with scan_nd


cmazzoli:
Asked a scan in 2 motors, got it but figure prepared for getting a mesh in the same 2 motors instead
For details, see:
https://logbook.nsls2.bnl.gov/23-ID-1/#63611_1

````
In [79]: th2th_pos = cycler(theta, thetas) + cycler(delta, deltas)
In [81]: dets = [fccd]

In [82]: RE(mv(fccd.exposure, (1, 0, 1)))
Out[82]: ()
...

In [85]: RE(scan_nd(dets,th2th_pos))
Transient Scan ID: 86050     Time: 2017/10/22 20:17:40
Persistent Unique Scan ID: 'aab28779-4e4f-473c-8c06-2f9bd6d4fc32'
Transient Scan ID: 86050 @ 2017/10/22 20:17:40
Persistent Unique Scan ID: 'aab28779-4e4f-473c-8c06-2f9bd6d4fc32'
New stream: 'baseline'
New stream: 'primary'                                                                                                                                                         
+-----------+------------+------------+------------+-------------------+-------------------+                                                                                  
|   seq_num |       time |      delta |      theta | fccd_stats3_total | fccd_stats4_total |
+-----------+------------+------------+------------+-------------------+-------------------+
|         1 | 20:17:47.8 |    40.0000 |    19.8700 |         -16613114 |        -259419609 |
|         2 | 20:18:05.9 |    41.0000 |    20.3700 |            408778 |           5242892 |                                                                                  
|         3 | 20:18:10.8 |    42.0000 |    20.8700 |            334858 |           3953314 |                                                                                  
/home/xf23id1/conda_envs/dev/lib/python3.6/site-packages/ophyd/status.py:336: RuntimeWarning: divide by zero encountered in double_scalars                                    
  time_remaining = time_elapsed / fraction + self.settle_time
|         4 | 20:18:15.9 |    43.0000 |    21.3700 |            267732 |           3321314 |                                                                                  
|         5 | 20:18:21.7 |    44.0000 |    21.8700 |            201152 |           3414689 |                                                                                  
|         6 | 20:18:26.7 |    45.0000 |    22.3700 |            126287 |           3231229 |                                                                                  
|         7 | 20:18:35.3 |    45.9965 |    22.8700 |            105760 |           4082372 |                                                                                  
|         8 | 20:18:40.4 |    47.0000 |    23.3700 |             61969 |           3338174 |                                                                                  
[...]

Issue #867, opened 2017-10-23
inne_product_scan doesnt' plot correctly


bisogni:
By running an 'inner_product_scan', the automatic plot pops out however no data get really plotted. The template looks more like the one of a 'mesh/outer_product_scan'...

@danielballan @tacaswell can we have a quick fix?

![gc_diag_diode_gc_diag_grid_m3_diag_diode_m3_diag_grid_vs_pgmslt_d_bot_pgmslt_d_top](https://user-images.githubusercontent.com/24460486/30558084-4f8a7728-9c7e-11e7-9fa8-6ae58f26ffac.png)

ambarb:
@cmazzoli From our discussion about https://github.com/NSLS-II/bluesky/issues/868, maybe you have a point if inner_product_scan is still broken as @bisogni reported.  However, if DAMA makes scan_nd do as you ask for any 2D motor scan, then things like spiral scans will be broken in in the way that they plot.  @danielballan can you weigh in on this since i think  this has something to do with the _md exceptions related to hints?

cmazzoli:
Hi Andi,
Thanks, now I understand better.
Personally I think that DAMA has all the capability of taking the challenge
and recognizing of a cycler provides a linear scan (diff on motors have
always the same sign) or it is intended to be a mesh (periodic inversion of
sign on both).
However if they fear that they cannot do it, then we will adapt ;)
Please note that as far as I understand currently the 2D scan-nd assumed as
a mesh is the exception compared to the rest. Am I wrong?
Thanks again.
Claudio

ambarb:
Claudio,

Yes, that is my observation.  If you have more than 2 motors for `scan_nd` than bluesky will make separate line plots for each in the LivePlot window.

ambarb:
In essence, you could hack a th2th scan as you did earlier with scan_nd and just also add gamma scan from 0 to 0  (or wherever it is that you have it).  Or you may be able to get away with use tardis.l  in `scan()`

cmazzoli:
Yes, I got the same idea (extra motor, I would rather prefer a non
interfering one like eta for example but basically the idea is the same..).
However, I would rather prefer DAMA to correct the problem suitably, if
possible. Dan
@danallan, could you please let us know?
Thanks.
Claudio

danielballan:
Marked for completion before "1.0", meaning we should have a fix within the next week and half.

cmazzoli:
:thumbsup:

[...]

PR #917, opened 2017-11-16
Added hint to inner_prod scan

jrmlhermitte:
Some issues with best effort callback. Used a hint in the `relative_inner_product_scan` to help guide the best effort callback.

Sample code to run (toggle `False` to `True` to test between the outer and inner product scans):
(Note this is a copy and paste from documentation)
```python
from bluesky.plan_stubs import abs_set, wait

from bluesky.plans import (relative_outer_product_scan)
from bluesky.plans import relative_inner_product_scan

from bluesky.preprocessors import (run_decorator, stage_decorator,
                                    subs_decorator)
from bluesky.callbacks import LiveTable, LivePlot
from ophyd.sim import det4, motor1, motor2
from bluesky import RunEngine

from bluesky.utils import ts_msg_hook


bec = BestEffortCallback()

def grid_in_grid(samples):
    """
    Scan a grid around the neighborhood of each sample.

    Parameters
    ----------
    sample : dict
        mapping each sample's name to its (x, y) position
    """

    # In this example we hard-code the hardware and other parameters. For more
    # flexibility, they could instead be parameters to the function.
    detector = det4
    x = motor1
    y = motor2
    x_range = y_range = 0.2
    x_num = y_num = 5

    hints = dict()

    @subs_decorator([
                     LiveTable([detector, x, y]),
                     #LivePlot('motor2'),
                     bec,
                    ])
    def plan():
        for name, position in samples.items():
            # Prepare metadata.
            md = {'sample': name}  #, 'hints': hints}

            # Move to the cetner of the sample position.
            x_pos, y_pos = position
            yield from abs_set(x, x_pos)
            yield from abs_set(y, y_pos)
            yield from wait()

            # Scan a grid around that position.
            if False:
                yield from relative_outer_product_scan([detector],
                                                       x, -x_range, x_range,
                                                       x_num,
                                                       y, -y_range, y_range,
                                                       y_num,
                                                       True, md=md)

            else:
                yield from relative_inner_product_scan([detector],
                                                       x_num,
                                                       x, -x_range, x_range,
                                                       y, -y_range, y_range,
                                                       md=md)

    yield from plan()


# Example usage:

RE = RunEngine({})
RE.msg_hook = ts_msg_hook
RE.subscribe(print)



#bec = BestEffortCallback()
#RE.subscribe(bec)

samples = {'A': (1, 1),
           'B': (1, 2),
           'C': (1, 3),
           'D': (2, 1),
           'E': (2, 2),
           'F': (2, 3)}

RE(grid_in_grid(samples))
```

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -982,7 +982,7 @@ def relative_inner_product_scan(detectors, num, *args, per_step=None, md=None):
 
     _md['hints'] = default_hints
 
-    if 'hints' in md:
+    if md is not None and 'hints' in md:
         _md['hints'].update(md['hints'])
 
     @bpp.reset_positions_decorator(motors)
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -966,24 +966,27 @@ def relative_inner_product_scan(detectors, num, *args, per_step=None, md=None):
     """
     _md = {'plan_name': 'relative_inner_product_scan',
            }
-    _md.update(md or {})
+    md = md or {}
+    _md.update(md)
     motors = [motor for motor, start, stop in partition(3, args)]
 
-    # Default should assume the first motor is what to plot
-    # If another motor is desired, override with separate hints
-    if hasattr(motors[0], 'hints') and 'fields' in motors[0].hints:
-        fields = list()
-        for motor in motors:
-            fields.extend(motor.hints['fields'])
-        default_dimensions = [(fields, 'primary')]
-        default_hints = {'dimensions': default_dimensions}
-    else:
-        default_hints = {}
+    # Give a hint that the motors all lie along the same axis
+    # [(['motor1', 'motor2', ...], 'primary'), ] is 1D (this case)
+    # [ ('motor1', 'primary'), ('motor2', 'primary'), ... ] is 2D for example
+    fields = []
+    for motor in motors:
+        fields.extend(getattr(motor, 'hints', {}).get('fields', []))
 
+    default_dimensions = [(fields, 'primary')]
+
+    default_hints = {}
+    if len(fields) > 0:
+        default_hints.update(dimensions=default_dimensions)
+
+    # now add default_hints and override any hints from the original md (if
+    # exists)
     _md['hints'] = default_hints
-
-    if md is not None and 'hints' in md:
-        _md['hints'].update(md['hints'])
+    _md['hints'].update(md.get('hints', {}) or {})
 
     @bpp.reset_positions_decorator(motors)
     @bpp.relative_set_decorator(motors)
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -971,7 +971,7 @@ def relative_inner_product_scan(detectors, num, *args, per_step=None, md=None):
 
     # Default should assume the first motor is what to plot
     # If another motor is desired, override with separate hints
-    if 'fields' in motors[0].hints:
+    if hasatttr(motors[0], 'hints') and 'fields' in motors[0].hints:
         fields = list()
         for motor in motors:
             fields.extend(motor.hints['fields'])
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -972,7 +972,10 @@ def relative_inner_product_scan(detectors, num, *args, per_step=None, md=None):
     # Default should assume the first motor is what to plot
     # If another motor is desired, override with separate hints
     if 'fields' in motors[0].hints:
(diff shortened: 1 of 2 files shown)


jrmlhermitte:
This is a first pass for #868 and #867.

bluesky/plans.py
```diff
@@ -964,10 +964,18 @@ def relative_inner_product_scan(detectors, num, *args, per_step=None, md=None):
...
     """
-    _md = {'plan_name': 'relative_inner_product_scan'}
+    _md = {'plan_name': 'relative_inner_product_scan',
+           }
     _md.update(md or {})
     motors = [motor for motor, start, stop in partition(3, args)]
 
+    # update the hints separately for the best effort callback
+    # TODO : consider non-primary?
+    default_hints = {'dimensions': [((motors[0].name,), 'primary')]}
```
tacaswell: The 'primary' here is what data stream these things are in.  It is probably better to ask the motor for it's hints.  The value we care about here is probably `motors[0].user_readback.name` which is most cases is the same as `motors[0].name` but does not have to be.  

We also want to list _all_ of the motors in the dimensions, not just the first one.  This should probably be something like

```python
[(motor.hints.get('fields', ()), 'pirmary') for motor in motors]

```

but I am not sure that that nesting is correct.

jrmlhermitte: sounds good. it should not be a list of motor names but a choice of one for a 1D plot see the modification. not sure how to make it look nicer in terms of list comprehension. @danielballan should have some say as to the general API, if this is the right place to put it

danielballan:
To my comment from Slack:

> I think that the plan hints should give the complete story, and it should be up to BEC to decide whether it wants to ignore axes beyond the zeroth one. Someday we might get better high-dimensional viz, and it would be nice if it worked on old data.

As I load my memory of the design discussions around 'hints' back into mental RAM, I think we actually planned for this case (but neglected to implement it!). For a grid scan over two motors, the hint looks like:

```
[(('motor1',), 'primary), ('motor2',), 'primary')]
```

For an inner-product scan with two motors, the hint should look like:

```
[(('motor1', 'motor2'), 'primary')]
```

Each element in the list is an axis of the _scan_. That axis is often composed of just one motor, but as in the case of an inner-product scan, it can be composed of multiple motors moving together. This is why we specify `('motor1',)` is a single-element tuple -- we left room for the more general case.

jrmlhermitte:
Ok, here is a proposed change. There are two levels to this:
1. Inner product scan needs to provide a hint that suggests motors are along the same axis, i.e. for two motor names, 'motor1' and 'motor2', changed the assumed `dimension`s from :
```python
[(('motor1',), 'primary), ('motor2',), 'primary')]
```
to
```python
[(('motor1', 'motor2',), 'primary')]
```
That's the attempt I made in `plans.py`
2. Finally, there is an issue where this list is currently flattened and assumed each motor name is a dimension, meaning both these lists would change to:
```python
[(('motor1', 'motor2',), 'primary')]
```
I have modified the code so that we pick only one element of the name tuples in each list (`('motor1', 'motor2')` for example).
Thus we end for both cases with the same flattened list as before, but with the correct number of elements.

TODO (suggestions from Dan) 
- Right now, the first motor is assumed to be the x-axis. It could be interesting to allow the user to choose another motor (or require them to put the motor they want as x-axis first)
- It might be interesting to hold on to the extra motor information, for example, use `'motor2'` for the legend in the `inner_product_scan` case (would not make sense for `outer_product_scan`.

These could be a new PR with issue or handled here.

Thoughts? Suggestions?

bluesky/plans.py
```diff
@@ -964,10 +964,27 @@ def relative_inner_product_scan(detectors, num, *args, per_step=None, md=None):
...
     """
-    _md = {'plan_name': 'relative_inner_product_scan'}
+    _md = {'plan_name': 'relative_inner_product_scan',
+           }
     _md.update(md or {})
     motors = [motor for motor, start, stop in partition(3, args)]
 
+    # Default should assume the first motor is what to plot
+    # If another motor is desired, override with separate hints
+    if hasattr(motors[0], 'hints') and 'fields' in motors[0].hints:
```
tacaswell: Do we need this conditional?  Could we do this as

```python
for motor in motors:
    fields.extend(getattr(motor, 'hints', {}).get('fields', []))
```

?

jrmlhermitte: yes that's much more elegant thanks! made a few extra changes

danielballan:
Elegant! Your helpful comments should rescue us next time we have to remember how this works.

jrmlhermitte:
thanks!

PR #918, opened 2017-11-16
removed None

jrmlhermitte:
@danielballan small change

### DIFF
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -92,7 +92,7 @@ class BestEffortCallback(CallbackBase):
 
         # Prepare a guess about the dimensions (independent variables) in case
         # we need it.
-        motors = self._start_doc.get('motors') or None
+        motors = self._start_doc.get('motors')
         if motors is not None:
             GUESS = [([motor], 'primary') for motor in motors]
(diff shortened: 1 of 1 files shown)


bluesky/callbacks/best_effort.py
```diff
@@ -92,7 +92,7 @@ def start(self, doc):
 
         # Prepare a guess about the dimensions (independent variables) in case
         # we need it.
-        motors = self._start_doc.get('motors') or None
+        motors = self._start_doc.get('motors')
```
tacaswell: This changes behavior in the case when the `get` returns an empty list, not sure of the consequence of that though....

jrmlhermitte: hmm that's interesting. I'm more in favor of a consistent result that uncertain. Tests pass but I think this PR should be closed. What do yo think?

tacaswell:
I _think_ leaving the `or None` ensures that we don't get empty lists going through (and hence always have a 'decent' guess), but I leave to @danielballan.

danielballan:
I think if a plan hints `motors=[]`, that could mean something different than omitting `motors` from the hint. I think this PR should probably be merged.

PR #920, opened 2017-11-16
ENH #919, passes unit tests

### LINKED ISSUE #919: plans.tune_centroid() needs some updates
The new plan for `plans.tune_centroid()` needs to be improved.

## Expected Behavior
* `step_factor` should be easy to understand
* motion should not go beyond initial start & stop specified
* result should tune to the centroid

## Current Behavior
* `step_factor` now is used with the current step size to determine the range of the next pass
* motion sometimes goes beyond `stop` on first pass
* misses the peak and tunes to the shoulder often, especially if the background is high (low S/N)

## Possible Solution
* change `step_factor` (keep the name) to be *the factor to reduce the range of each successive iteration*
* enforce bounds on motion
* determine the centroid from each pass separately (cumulative centroid data can suffer from too much data with poor S/N)

scan a synthetic pseudo Voigt function over a broad range (say range >> FWHM), all three issues will likely present

## Context
Results are sub-optimal

prjemian:
Fixes #919: 

## Description

* change `step_factor` (keep the name) to be *the factor to reduce the range of each successive iteration*
* enforce bounds on motion
* determine the centroid from each pass separately

## Motivation and Context
#919

## How Has This Been Tested?
`coverage run run_tests.py`

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -697,14 +697,14 @@ def tune_centroid(
                     start, stop = stop, start
                 step = (stop - start) / (num - 1)
                 next_pos = start
-                #print("peak position = {}".format(peak_position))
-                #print("start = {}".format(start))
-                #print("stop = {}".format(stop))
+                # print("peak position = {}".format(peak_position))
+                # print("start = {}".format(start))
+                # print("stop = {}".format(stop))
 
         # finally, move to peak position
         if peak_position is not None:
             # improvement: report final peak_position
-            #print("final position = {}".format(peak_position))
+            # print("final position = {}".format(peak_position))
             yield from bps.mv(motor, peak_position)
 
     return (yield from _tune_core(start, stop, num, signal))
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -581,9 +581,9 @@ def tune_centroid(
     smaller step size until the minimum step size is reached.
     Rescans will be centered on the signal centroid
     (for $I(x)$, centroid$= \sum{I}/\sum{x*I}$)
-    with original scan range reduced by `step_factor`.
+    with original scan range reduced by ``step_factor``.
 
-    Set `snake=True` if your positions are reproducible
+    Set ``snake=True`` if your positions are reproducible
     moving from either direction.  This will not necessarily
     decrease the number of traversals required to reach convergence.
     Snake motion reduces the total time spent on motion
@@ -594,7 +594,7 @@ def tune_centroid(
 
     Note:  Ideally the signal has only one peak in the range to
     be scanned.  It is assumed the signal is not polymodal
-    between `start` and `stop`.
+    between ``start`` and ``stop``.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


bluesky/plans.py
```diff
@@ -570,7 +570,7 @@ def tune_centroid(
         detectors, signal, motor,
         start, stop, min_step,
         num=10,
-        step_factor=2,
+        step_factor=3.0,
```
jrmlhermitte: why is this 3 by default? mainly curious. seems like it would be 4 from doc?

prjemian: Choice of 3 is arbitrary.  Balances between rapid convergence and possibly missing a low S/N peak.

Seems as if the upper bound should be `num`.  Agree?

prjemian: Definitely tested in simulation.  Here's *the* [unit test for `tune_centroid()`](https://github.com/prjemian/bluesky/blob/master/bluesky/tests/test_scans.py#L280):

tacaswell:
As a complete aside, I came across 'golden section search' (https://en.wikipedia.org/wiki/Golden-section_search ) which is a way to find a unimodal peak with a minimum number of measurements.  I suspect that for most cases on beamlines the concern is getting between the data points (because motors are slow) but if there are cases where the count time is much longer than the move time that might be an interesting alternative optimization strategy.

prjemian:
Both of those algorithms sound like good alternatives and could be considered first or when `tune_centroid()` fails.

I'm motivated by instrumental data such as [these plots from our USAXS](http://usaxs.xray.aps.anl.gov/livedata/specplots/2017/10/10_15_Myers1/).  The linear-linear plots are tune-up scans.  Most of them scan the peak within the chosen range.  [This particular scan](http://usaxs.xray.aps.anl.gov/livedata/specplots/2017/10/10_15_Myers1/s74.png) would have benefited from `tune_centroid()` and a wider scanning range since it missed the peak ([found it the next scan of `mr`](http://usaxs.xray.aps.anl.gov/livedata/specplots/2017/10/10_15_Myers1/s81.png)).

bluesky/plans.py
```diff
@@ -581,7 +581,7 @@ def tune_centroid(
     smaller step size until the minimum step size is reached.
     Rescans will be centered on the signal centroid
     (for $I(x)$, centroid$= \sum{I}/\sum{x*I}$)
-    with a scan range of 2*step_factor*step of current scan.
+    with original scan range reduced by `step_factor`.
```
danielballan: An annoying sphinx detail: single backticks render the contents in italics, not as code. You need ``step_factor`` or :data:`step_factor` to make sphinx happy. (GitHub Markdown behaves different than sphinx; it render single- or double-backticks the same way.)

bluesky/plans.py
jrmlhermitte: great. I'm wondering if we should include a test for `step_factor` at least. Maybe at least something like:
```python
    with pytest.raises(ValueError):  
        # assert raises ValueError if step_factor < 1
        scan5 = bp.tune_centroid([det], 'det', motor, 5, 0, -0.1, 10, 
                                 step_factor=.5, snake=False)
        RE(scan5)
```
could not hurt to add in `test_tune_centroid`

PR #921, opened 2017-11-20
DOC: correct state-machine.html page

mrakitin:
Review/correction of the contents of the state-machine.html page.

## Description

Removed duplicating examples, corrected typos and added missing methods.

## Motivation and Context

Better documentation.

### DIFF
--- a/doc/source/state-machine.rst
+++ b/doc/source/state-machine.rst
@@ -174,6 +174,20 @@ and, if it detects a problem, it suspends execution. When it detects that
 conditions have returned to normal, it gives the RunEngine permission to resume
 after some interval. This can operate unattended.
 
+.. ipython::
+    :verbatim:
+
+    In [1]: RE(scan([det], motor, -10, 10, 15), LiveTable([motor, det]))
+    +------------+-------------------+----------------+----------------+
+    |   seq_num  |             time  |         motor  |           det  |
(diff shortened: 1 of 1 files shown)


doc/source/state-machine.rst
```diff
@@ -166,28 +166,14 @@ RE.state       Check if 'paused' or 'idle'.
...
-    +------------+-------------------+----------------+----------------+
-    |   seq_num  |             time  |         motor  |           det  |
-    +------------+-------------------+----------------+----------------+
-    |         1  |  16:46:08.953815  |          0.03  |        290.00  |
-    Suspending....To get prompt hit Ctrl-C to pause the scan
-    |         2  |  16:46:20.868445  |          0.09  |        279.00  |
-    |         3  |  16:46:29.077690  |          0.16  |        284.00  |
-    |         4  |  16:46:33.540643  |          0.23  |        278.00  |
-    +------------+-------------------+----------------+----------------+
-
```
mrakitin: @danielballan, I removed this example since exactly the same example appear later.

danielballan: I actually prefer this duplication to stay. My intention was to provide a quick, easy-to-skim example to help people understand what we mean by "suspension", even if they don't dig into the details of the full example just yet.

PR #923, opened 2017-11-27
DOC: Added hint on interactive plotting

jrmlhermitte:
I had a user ask me how to see the plots displayed. Since it's suggesting the use of ipython, we should give a hint on how to set up interactive mode for `LivePlot` (which is using matplotlib).

@danielballan thoughts?


hhslepicka:
I believe that it is already available in here:
http://nsls-ii.github.io/bluesky/callbacks.html#aside-making-plots-update-live

The `install_qt_kicker`, `install_nb_kicker` should solve this problem.

prjemian:
Ok with that

danielballan:
My first thought was to agree with @hhslepicka that our existing section on this is sufficient, but given that both you @jrmlhermitte and this user didn't find it right away, maybe that's not true.

In any case, `plt.ion()` is called as part of `%matplotlib ...` so it shouldn't need to be called explicitly. But, to be clear, `plt.ion()` and `install_XX_kicker()` do different things and are _both_ necessary.

hhslepicka:
With https://github.com/NSLS-II/bluesky/pull/807 what do you think of adding it to the LivePlot constructor? Call the kicker and install it in case it is not yet? This way you don't face this kind of issue...

danielballan:
I would bet that that is too magical for @tacaswell. It’s a feature not a bug that intermediate users must understand something about these libraries to use them; if we silently “do the right thing” for LivePlot, users might be surprised when other matplotlib functions aren’t automatically interactive because we didn’t teach them from the beginning. And there are conceivable cases where users don’t want to pay the performance cost of live drawing because they are just writing to a file. For novice users IPython profiles should be taking care of this.

Can we do more to surface this important detail in our docs?

doc/source/plans_intro.rst
```diff
@@ -73,9 +73,16 @@ Or, similarly, we can plot the data during the scan using ``LivePlot``.
...
 
+    import matplotlib.pyplot as plt; plt.ion()
     from bluesky.callbacks import LivePlot
     RE(scan([det], motor, 1, 5, 5), LivePlot('det', 'motor'))
 
+.. note::
+
+    You'll want to make sure your plotting system is in interactive mode to see
+    the plots update in real time. This is what the first line effectively
+    accomplishes.
```
stuartcampbell: Should we add a note that if you are using a jupyter notebook then you should use ```%matplotlib ...``` ?

danielballan: We have it [here](https://nsls-ii.github.io/bluesky/callbacks.html#aside-making-plots-update-live). I think I'd prefer to _link_ to that note so that we only have to update it in one place if our advice changes in the future (which I expect it will with Jupyter _Lab_).

jrmlhermitte:
Sorry for delay. Based on this conversation I believe the most appropriate solution is to include a link to the live plotting instructions. Ready for review. Thanks for input @danielballan @stuartcampbell @prjemian @hhslepicka   :-)

jrmlhermitte:
Note : related to #807 but I think we can push this first and resolve the next later.

jrmlhermitte:
thanks!

PR #930, opened 2017-12-13
Waterfall

CJ-Wright:
Add 1D waterfall plotting (this should also finally end xpdView). 

## Description
Add callback for making Waterfalls of 1D patterns.

## Motivation and Context
Lots of reduced data (even raw data) is expressed as 1D data, visualizing that would be nice.

## How Has This Been Tested?
That is a work in progress. I need an ophyd sim hardware that puts out a 1D pattern

### DIFF
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -492,10 +492,8 @@ class LiveWaterfall(CallbackBase):
         The units for the x and y axes
     """
 
-    def __init__(self, x_name, y_name, db=None, units=None,
+    def __init__(self, x_name, y_name, units=None,
                  window_title=None):
-        super().__init__((x_name, y_name,), db=db)
-        self.db = db
         self.x_name = x_name
         self.y_name = y_name
         self.units = units
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -47,6 +47,7 @@ class LivePlot(CallbackBase):
     >>> my_plotter = LivePlot('det', 'motor', legend_keys=['sample'])
     >>> RE(my_scan, my_plotter)
     """
+
     def __init__(self, y, x=None, *, legend_keys=None, xlim=None, ylim=None,
                  ax=None, fig=None, epoch='run', **kwargs):
         super().__init__()
@@ -124,9 +125,12 @@ class LivePlot(CallbackBase):
         if self.x == 'time' and self._epoch == 'run':
             new_x -= self._epoch_offset
 
+        self.update(new_x, new_y)
+        super().event(doc)
+
+    def update(self, new_x, new_y):
         self.update_caches(new_x, new_y)
         self.update_plot()
-        super().event(doc)
 
     def update_caches(self, x, y):
         self.y_data.append(y)
@@ -183,6 +187,7 @@ class LiveScatter(CallbackBase):
     --------
     :class:`bluesky.callbacks.LiveGrid`.
     """
+
     def __init__(self, x, y, I, *, xlim=None, ylim=None,
                  clim=None, cmap='viridis', ax=None, **kwargs):
         if ax is None:
@@ -295,6 +300,7 @@ class LiveGrid(CallbackBase):
     --------
     :class:`bluesky.callbacks.LiveScatter`.
     """
+
     def __init__(self, raster_shape, I, *,
                  clim=None, cmap='viridis',
                  xlabel='x', ylabel='y', extent=None, aspect='equal',
@@ -397,6 +403,7 @@ class LiveFitPlot(LivePlot):
         matplotib Axes; if none specified, new figure and axes are made.
     All additional keyword arguments are passed through to ``Axes.plot``.
     """
+
     def __init__(self, livefit, *, num_points=100, legend_keys=None, xlim=None,
                  ylim=None, ax=None, **kwargs):
         if len(livefit.independent_vars) != 1:
@@ -508,7 +515,7 @@ def plot_peak_stats(peak_stats, ax=None):
 
     if ps.lin_bkg:
         lb = ps.lin_bkg
-        ln, = ax.plot(ps.x_data, ps.x_data*lb['m'] + lb['b'],
+        ln, = ax.plot(ps.x_data, ps.x_data * lb['m'] + lb['b'],
                       ls='--', lw=2, color='k')
         arts['bkg'] = ln
 
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -6,6 +6,7 @@ import numpy as np
 import warnings
 
 from .core import CallbackBase, get_obj_fields
+from .waterfall import Waterfall
 
 
 class LivePlot(CallbackBase):
@@ -473,6 +474,52 @@ class LiveFitPlot(LivePlot):
     def stop(self, doc):
         self.livefit.stop(doc)
         # Intentionally override LivePlot.stop. Do not call super().
+
+
+class LiveWaterfall(CallbackBase):
+    """
+    Stream 1D images in a waterfall viewer.
+
+    Parameters
+    ----------
+    x_name : str
+        field name for x dimension
+    y_name: str
+        field name for y dimension
+    db: databroker.Broker instance
+        The Broker to fill the events if events are not filled
+    units: tuple of str
+        The units for the x and y axes
+    """
+
+    def __init__(self, x_name, y_name, db=None, units=None,
+                 window_title=None):
+        super().__init__((x_name, y_name,), db=db)
+        self.db = db
+        self.x_name = x_name
+        self.y_name = y_name
+        self.units = units
+
+        self.fig = plt.figure(window_title)
+
+        self.wf = Waterfall(fig=self.fig, unit=self.units)
+        self.i = 0
+        self.fig.show()
+
+    def start(self, doc):
+        self.i = 0
+        self.wf.key_list.clear()
+        self.wf.int_data_list.clear()
+
+    def event(self, doc):
+        y = doc['data'][self.y_name]
+        x = doc['data'][self.x_name]
+        self.update((x, y))
+        super().event(doc)
+
+    def update(self, data):
+        self.wf.update(key_list=[self.i], int_data_list=[data])
+        self.i += 1
 
 
 def plot_peak_stats(peak_stats, ax=None):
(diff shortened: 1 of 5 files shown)


CJ-Wright:
xref: https://gitter.im/NSLS-II/DAMA?at=5a319be0a2be4668289a5b03

CJ-Wright:
@tacaswell ready for review!

bluesky/callbacks/mpl_plotting.py
```diff
@@ -124,9 +126,12 @@ def event(self, doc):
         if self.x == 'time' and self._epoch == 'run':
             new_x -= self._epoch_offset
 
+        self.update(new_x, new_y)
+        super().event(doc)
+
+    def update(self, new_x, new_y):
```
CJ-Wright: Note that this change is so that we have a somewhat uniform interface, `update` will take in the raw data types and do the expected thing (add them to a list, plot them, etc). This allows us to use these callbacks without the event model, eg if you had already unpacked the data somewhere.

bluesky/callbacks/waterfall.py
```diff
@@ -0,0 +1,168 @@
...
+        line = event.artist
+        name = line.get_label()
+        line.axes.legend([name], handlelength=0,
+                         handletextpad=0, fancybox=True)
+        line.figure.canvas.draw_idle()
+
+    def _update_plot(self, x_offset_val=None, y_offset_val=None):
+        """core method to update x-, y-offset sliders"""
+        self.ax.set_facecolor('w')
+        self.ax.cla()
```
tacaswell: This is a very inefficient way to do this,  it is better to keep a mapping between the `Line2D` instances, the raw data they came from and the offset associated with that line and update as required.

bluesky/callbacks/waterfall.py
```diff
@@ -0,0 +1,168 @@
...
+        # callback for showing legend
+        self.canvas.mpl_connect('pick_event', self.on_plot_hover)
+        self.key_list = key_list
+        self.int_data_list = int_data_list
+        self.ax = self.fig.add_subplot(111)
+        self.unit = unit
+        self.label_size = label_size
+        self.tick_size = tick_size
+        # flag to prevent update
+        self.halt = False
```
tacaswell: Why do we need this state?

bluesky/callbacks/waterfall.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def on_plot_hover(self, event):
+        """callback to show legend when click on one of curves"""
+        line = event.artist
+        name = line.get_label()
+        line.axes.legend([name], handlelength=0,
+                         handletextpad=0, fancybox=True)
+        line.figure.canvas.draw_idle()
+
+    def _update_plot(self, x_offset_val=None, y_offset_val=None):
```
tacaswell: It is better for there to be one point of truth for the values of the offsets.  Have this method always consult the sliders (no args to this function) and register a dummy function that drops the val on to the slider callback or keep this function taking input, but drop it consulting the sliders, and use a partial to put the `val` from the callback into the correct kwarg.  You can then only update x or y data on the Artists.

bluesky/callbacks/mpl_plotting.py
```diff
@@ -468,6 +476,50 @@ def stop(self, doc):
         # Intentionally override LivePlot.stop. Do not call super().
 
 
+class LiveWaterfall(CallbackBase):
```
tacaswell: Is there any other name for this than 'waterfall'?  There is also a community that calls khymographs 'waterfall' plots so there is not universal agreement on the meaning.

How about `OffsetLines`, `ShiftedLines`, or `StackedLines` (`stackplot` is already a thing https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.stackplot.html) ?

bluesky/callbacks/mpl_plotting.py
tacaswell: The input to `update` below is of an entirely different type than the `update` here.  I do not fully understand this use case.  You still need `Start` and `Descriptor` documents to set up the `Callback` objects to be ready to take in data.  In what case would you have `Start` and `Descriptor` documents but not `Event` documents?  

In general callbacks may configure them selves based on the `Start` and `Descriptor` documents or do very different things based on which stream the `Event` is in there may not be a single `update` method with a fixed signature (see BestEffortCallback for example).

CJ-Wright: My hope was that a) we could use callbacks outside the event model by calling `update` with the actual data. It would be upon the user/dev to properly parse the above `start` and `descriptor` code and make certain that all the state was created as needed. This would mean that we could use this on data which didn't adhere to the event model. I'd really prefer not to have to rewrite everything just so I can pass raw data to it.

My understanding of `BestEffortCallback` is that it is a bit of an outlier since it is a gestalt of callbacks.  I'd imagine that most of these base mpl driven classes will not do vastly different things based off of the `Start` and `Descriptor` since they are designed to produce a specific plot from a specific type of data.

tacaswell: If that is the case then you should write functions independent of the callbacks that work on raw data and then use those functions in the callback.

The start/descriptor (plus maybe some values passed at init time) should be providing instructions to the callback about how to unpack the events and the `event` method is already the 'update' method.  I do not see the advantage of standardizing on a second method that requires understanding the internal state of the callback to use.  One of the biggest strengths of the current callback design is that users of the callbacks do not need to know _anything_ about what goes on inside (in principle).  We should be moving to making those abstractions less leaky.

BEC is probably prototypical of callbacks going forward with more brains / dispatch logic rolled in.

danielballan: I don't think `CallbackBase` should be used "outside of the event model." Its role its to communicate between the event model and generic Python functions. Anything outside of the event model could use some simpler callback mechanism without the four methods named for event model documents, right?

CJ-Wright: I guess, are there plans to break the callbacks so as to separate the update and event model logic?

danielballan: Not really. The `update` method was added in a refactor intended to make it easier for subclasses to do simple arithmetic on the data before plotting. I've recently thought it would be better to do that through a callback hook (`prepare_data` or something). Like `LivePlot` in general, this API is not particularly well thought out and not intended to be an extensible framework, more like a simple example. The only rock-solid API is `RunEngine.subscribe` and, equivalently, `CallbackBase`.

CJ-Wright: I'm thinking about this specifically since there are other applications of streaming which could benefit from the mpl wrapper chunks that this has done. (eg I really wouldn't want to try to re-implement the image viewer stuff or the scatter plot if I didn't have to and a solution to one seems like a solution to both)

danielballan: I see. What would it look like to make an image viewer (with some `update` method) that knows nothing of the event model and then an adapter between said image viewer and the event model which inherits from `CallbackBase` and calls `CJsAwesomeImageViewer.update`?

tacaswell: We really need to re-suscitate xray-vision...

CJ-Wright: That works too!
For live plot we could move most of the `start` document stuff to the init
```python
self.x_data, self.y_data = [], []
label = " :: ".join(
            [name for name in self.legend_keys])
kwargs = ChainMap(self.kwargs, {'label': label})
self.current_line, = self.ax.plot([], [], **kwargs)
self.lines.append(self.current_line)
self.legend = self.ax.legend(
            loc=0, title=self.legend_title).draggable()
```
Then have the `start` method overwrite the label/remake the legend. Have the event method unpack the data and hand it to update.

CJ-Wright: Note that this may also help with `LiveImage` as well, since we can create a version which assumes that the data is already loaded, and have a broker driven class which loads/feeds data to the `LiveImage` class.

tacaswell: > Note that this may also help with LiveImage as well, since we can create a version which assumes that the data is already loaded, and have a broker driven class which loads/feeds data to the LiveImage class.

I don't understand.  Broker _also_ emits documents so it should be the same code for live or post-fact.

> Then have the start method overwrite the label/remake the legend. Have the event method unpack the data and hand it to update.

To be clear 'update' is not part of the callback API

> For live plot we could move most of the start document stuff to the init

Why?  You do not have access to the information in the start document at `__init__` time

CJ-Wright: For `LiveImage` we have a broker filling events, if we took only data then we wouldn't need to worry about where the data was/which broker to hook into.

I understand it isn't a part of the callback API. My point is that there could be two APIs:
1. Classes which wrap MPL plots and take streaming data via an `update` method. Information not handled in the streaming data can be handled at init and changed later (via attributes). Note that this API doesn't really care where the data is coming from (eg some raw streaming firehose somewhere).
1. Classes which take event model and turn it into data (potentially by using a broker to fill events) which is plumbed into the first API's `update` and metadata which is used to change the first API's state to have the desired info (change the title, legends, etc.).

In the non event model case the `__init__` time info could be entered manually (eg, I want the title to be 'x' because that is the name of the stream to which I'm attached). In the case of the event model, this could be defaulted sanely, especially since it will be quickly overridden by the info from the documents.

tl;dr It might be interesting to consider a system where the "plotting live streaming data" parts of callbacks are separate from their "interpret documents and do something with them".

danielballan:
@CJ-Wright Are you interested in following up on the review comments?

CJ-Wright:
At some point yes, I'm sick today and I just haven't gotten around to it. It is on my to-do list though.

danielballan:
Feel better, @CJ-Wright! :-)

bluesky/callbacks/waterfall.py
chiahaoliu: So it seems better to manage ``Line2D`` instance for each curve and use ``set_xdata, ``set_ydata`` methods to update the array and then ``draw``?

(found example might be related: https://matplotlib.org/gallery/api/line_with_text.html#sphx-glr-gallery-api-line-with-text-py) 

Also, under this context do we need to worry about cleaning ``Axes`` before draw?

bluesky/callbacks/waterfall.py
chiahaoliu: This state is to compensate the weird ``ax.set_facecolor('w')`` call above. We can remove it for sure.

stan-dot:
@CJ-Wright is this still a desired feature?

tacaswell:
Still worth considering, but CJ has moved on.

stan-dot:
can we move the *idea* into an issue, close this PR and link to it as a possible implementation direction?

tacaswell:
@stan-dot Yes, that seems reasonable to me.

stan-dot:
as discussed, moved to the issue
https://github.com/bluesky/bluesky/issues/1774

PR #932, opened 2017-12-14
ENH: LiveStream

teddyrendahl:
## Description

The goal of this addition is to provide an outline for doing streaming analysis and feed this information to other out of the box `bluesky` callbacks. The callback uses the `streamz` module to send the event documents through an arbitrary graph network. Each document is then repackaged and sent to subscribers. 

## Motivation and Context
This solves a number of problems we have at SLAC where we want to process data as it is emitted and not necessarily wait for those things to propagate through a databroker. In order to not confuse event streams, the `LiveStream` mimics the document creation in the `RunEngine`, it creates its own sequence numbering, descriptors e.t.c. Here is an example below of running `bluesky.examples.stepscan`, and having two plots; one averaging every two shots, one showing the raw data.
stream_example

Obviously this is a wonky example, but I think it shows the power of this solution. I can do arbitrary analysis in my graph network and send all this through the built-in bluesky callbacks. I could imagine the use case where you have two databroker instances, one subscribed to the raw data and one subscribed to the processed information. This could provide an interesting first pass at analysis for user groups. At SLAC, we would want to use this to only show processed information to our `LiveFit`. Because any` bluesky` callback can subscribe to the processed data stream we can also plot the actual information that our fit is seeing without duplicating logic in multiple callbacks.

### Implementations Details
* The `streamz` module isn't strictly necessary. I think it makes sense for now to force the users to fit their processing pipeline into this form, but if this is not flexible enough, or we don't want the dependency, we can find a way around it.
* Recreating the document structure was kind of painful. It will also be annoying if you change how the RunEngine structures these dictionaries that the same change has to be made in `LiveStream`. It however, seemed liked a necessary evil.

## How Has This Been Tested?

Added tests to the bluesky and `dev-requirements` now includes streamz
<!--

### DIFF
--- a/bluesky/callbacks/stream.py
+++ b/bluesky/callbacks/stream.py
@@ -29,6 +29,7 @@ class LiveDispatcher(CallbackBase):
     RunEngine, effectively creating a small chain of callbacks
 
     .. code::
+
         # Create our dispatcher
         ld = LiveDispatcher()
         # Subscribe it to receive events from the RunEgine
@@ -69,8 +70,7 @@ class LiveDispatcher(CallbackBase):
         """
         Receive an event document from the original stream.
 
-        This should be reimplemented by a subclass, with ``super().event``
-        called as the last step in the method.
+        This should be reimplemented by a subclass.
 
         Parameters
         ----------
--- a/bluesky/callbacks/stream.py
+++ b/bluesky/callbacks/stream.py
@@ -2,7 +2,6 @@ import logging
 import time as ttime
 from collections import Iterable, ChainMap
 
-import streamz
 import jsonschema
 import numpy as np
 from event_model import DocumentNames, schemas
@@ -14,44 +13,39 @@ from ..utils import new_uid
 logger = logging.getLogger(__name__)
 
 
-class LiveStream(CallbackBase):
+class LiveDispatcher(CallbackBase):
     """Create a secondary event stream of processed data"""
-    def __init__(self, in_node, out_node=None):
-        self.in_node = in_node
-        # Use the same node as an output if None is given
-        self.out_node = out_node or self.in_node
-        # The output should propagate the modified Event document down the
-        # pipeline
-        self.event = self.in_node.emit
-        self.out_node.sink(self.send_event)
+    def __init__(self):
         # Public dispatcher for callbacks
         self.dispatcher = Dispatcher()
         # Local caches for internal use
         self.seq_count = 0  # Maintain our own sequence count for this stream
+        self.raw_descriptors = dict()  # Store raw descriptors for use later
         self._stream_start_uid = None  # Generated start doc uid
-        self._raw_descriptors = dict()  # List of past raw event descriptors
-        self._described = list()  # List of raw event desc. we have translated
-        self._last_descriptor = None  # List of sent descriptors
+        self._descriptors = dict()  # Dictionary of sent descriptors
 
-    def start(self, doc):
-        """Receive a stop document, re-emit it for this event stream"""
+    def start(self, doc, _md=None):
+        """Receive a start document, re-emit it for this event stream"""
         self._stream_start_uid = new_uid()
+        _md = _md or dict()
         # Create a new start document with a new uid, start time, and the uid
         # of the original start document. Preserve the rest of the metadata
         # that we retrieved from the start document
         md = ChainMap({'uid': self._stream_start_uid,
                        'original_run_uid': doc['uid'],
                        'time': ttime.time()},
-                      doc)
-        # Emit the start document for anyone subscribed to our Dispatcher
+                      _md, doc)
+        # Dispatch the start document for anyone subscribed to our Dispatcher
         self.emit(DocumentNames.start, dict(md))
+        super().start(doc)
 
     def descriptor(self, doc):
-        """Receive a descriptor and store for later reference"""
-        self._raw_descriptors[doc['uid']] = doc
+        """Store a descriptor"""
+        self.raw_descriptors[doc['uid']] = doc
         super().descriptor(doc)
 
-    def send_event(self, doc):
+    def process_event(self, doc, stream_name='primary',
+                      id_args=None, config=None):
         """
         Receive an event document
 
@@ -62,20 +56,26 @@ class LiveStream(CallbackBase):
         that it is not confused with the raw data stream being emitted from the
         RunEngine
         """
+        id_args = id_args or (doc['descriptor'],)
+        config = config or dict()
+        # Determine the descriptor id
+
+        desc_id = frozenset((tuple(doc['data'].keys()), stream_name, id_args))
         # If we haven't described this configuration
         # Send a new document to our subscribers
-        if doc['descriptor'] not in self._described:
-            prior_desc = self._raw_descriptors[doc['descriptor']]
+        if (stream_name not in self._descriptors or
+            desc_id not in self._descriptors[stream_name]):
             # Create a new description document for the output of the stream
             data_keys = dict()
             # Parse the event document creating a new description. If the key
             # existed in the original source description, just assume that it
             # is the same type, units and shape. Otherwise do some
             # investigation
+            raw_desc = self.raw_descriptors.get(doc['descriptor'], {})
             for key, val in doc['data'].items():
                 # Described priorly
-                if key in prior_desc['data_keys']:
-                    key_desc = prior_desc['data_keys'][key]
+                if key in raw_desc['data_keys']:
+                    key_desc = raw_desc['data_keys'][key]
                 # String key
                 elif isinstance(val, str):
                     key_desc = {'dtype': 'string',
@@ -89,29 +89,31 @@ class LiveStream(CallbackBase):
                     key_desc = {'dtype': 'number',
                                 'shape': []}
                 # Modify the source
-                key_desc['source'] = 'Stream: {}'.format(self.out_node.name
-                                                         or 'Unnamed Stream')
+                key_desc['source'] = 'Stream'
                 # Store in our new descriptor
                 data_keys[key] = key_desc
             # Create our complete description document
             desc = ChainMap({'uid': new_uid(), 'time': ttime.time(),
                              'run_start': self._stream_start_uid,
+                             'data_keys': data_keys, 'configuration': config,
                              'object_keys': {'stream':
                                              list(data_keys.keys())}},
-                            prior_desc)
+                            raw_desc)
             # Store information about our descriptors
-            self._last_descriptor = dict(desc)
-            self._described.append(doc['descriptor'])
+            desc = dict(desc)
+            if stream_name not in self._descriptors:
+                self._descriptors[stream_name] = dict()
+            self._descriptors[stream_name][desc_id] = desc
             # Emit the document to all subscribers
-            self.emit(DocumentNames.descriptor, self._last_descriptor)
+            self.emit(DocumentNames.descriptor, desc)
 
         # Clean the Event document produced by graph network. The data is left
         # untouched, but the relevant uids, timestamps, seq_num are modified so
         # that this event is not confused with the raw data stream
         self.seq_count += 1
+        desc_uid = self._descriptors[stream_name][desc_id]['uid']
         current_time = ttime.time()
-        evt = ChainMap({'uid': new_uid(),
-                        'descriptor': self._last_descriptor['uid'],
+        evt = ChainMap({'uid': new_uid(), 'descriptor': desc_uid,
                         'timestamps': dict((key, current_time)
                                            for key in doc['data'].keys()),
                         'seq_num': self.seq_count, 'time': current_time},
@@ -119,22 +121,25 @@ class LiveStream(CallbackBase):
         # Emit the event document
         self.emit(DocumentNames.event, dict(evt))
 
-    def stop(self, doc):
+    def stop(self, doc, _md=None):
         """Receive a stop document, re-emit it for this event stream"""
         # Create a new stop document with a new_uid, pointing to the correct
         # start document uid, and tally the number of events we have emitted.
         # The rest of the stop information is passed on to the next callback
+        _md = _md or dict()
+        num_events = dict((stream, len(self._descriptors[stream]))
+                          for stream in self._descriptors.keys())
         md = ChainMap(dict(run_start=self._stream_start_uid,
                            time=ttime.time(), uid=new_uid(),
-                           num_events={'primary': self.seq_count}),
+                           num_events=num_events),
                       doc)
         self.emit(DocumentNames.stop, dict(md))
         # Clear the local caches for the run
-        self._raw_descriptors.clear()
-        self._described.clear()
         self.seq_count = 0
-        self._last_descriptor = None
+        self.raw_descriptors.clear()
+        self._descriptors.clear()
         self._stream_start_uid = None
+        super().stop(doc)
 
     def emit(self, name, doc):
         """Emit a document, first checking the schema"""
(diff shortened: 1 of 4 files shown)


bluesky/callbacks/stream.py
```diff
@@ -0,0 +1,181 @@
...
+        self.dispatcher = Dispatcher()
+        # Local caches for internal use
+        self.seq_count = 0  # Maintain our own sequence count for this stream
+        self._stream_start_uid = None  # Generated start doc uid
+        self._raw_descriptors = dict()  # List of past raw event descriptors
+        self._described = list()  # List of raw event desc. we have translated
+        self._last_descriptor = None  # List of sent descriptors
+
+    def start(self, doc):
+        """Receive a stop document, re-emit it for this event stream"""
```
ZLLentz: docstring typo here

CJ-Wright:
You might take a look at previous attempts at this kind of thing. https://github.com/xpdAcq/SHED/tree/master/shed

bluesky/callbacks/stream.py
```diff
@@ -0,0 +1,181 @@
...
+                # Store in our new descriptor
+                data_keys[key] = key_desc
+            # Create our complete description document
+            desc = ChainMap({'uid': new_uid(), 'time': ttime.time(),
+                             'run_start': self._stream_start_uid,
+                             'object_keys': {'stream':
+                                             list(data_keys.keys())}},
+                            prior_desc)
+            # Store information about our descriptors
+            self._last_descriptor = dict(desc)
```
danielballan: I think instead of relying on recency to match a new event to its new descriptor, you should keep a mapping between original descriptor uids and their corresponding new descriptor documents.

bluesky/callbacks/stream.py
```diff
@@ -0,0 +1,181 @@
...
+            # Emit the document to all subscribers
+            self.emit(DocumentNames.descriptor, self._last_descriptor)
+
+        # Clean the Event document produced by graph network. The data is left
+        # untouched, but the relevant uids, timestamps, seq_num are modified so
+        # that this event is not confused with the raw data stream
+        self.seq_count += 1
+        current_time = ttime.time()
+        evt = ChainMap({'uid': new_uid(),
+                        'descriptor': self._last_descriptor['uid'],
```
danielballan: So this would be `self._desc_mapping[doc['descriptor']]`

bluesky/callbacks/stream.py
```diff
@@ -0,0 +1,181 @@
...
+
+class LiveStream(CallbackBase):
+    """Create a secondary event stream of processed data"""
+    def __init__(self, in_node, out_node=None):
+        self.in_node = in_node
+        # Use the same node as an output if None is given
+        self.out_node = out_node or self.in_node
+        # The output should propagate the modified Event document down the
+        # pipeline
+        self.event = self.in_node.emit
```
danielballan: :+1:

bluesky/callbacks/stream.py
teddyrendahl: Planning on another commit to fill out the docs. I'll make sure to fix this.

danielballan:
@CJ-Wright SLAC has some interest in Streams too and spun this up to kick off some conversation. Thanks for linking to pySHED. This PR is short enough that I can fully understand it. I'd like to see both this and pySHED applied to 5-10 use cases. I think that will help me understand this a lot better.

Summarizing impromptu lunch discussion about this PR:

We like this model. (To be clear, as of this writing, this is not exactly what's in the PR.)

```python
def graph_manager(start=None, descriptors=None, stop=None):
    # builds one Stream, no matter how many descriptors it gets,
    # closing over any metadata extracted from start and descriptor(s)
    return in_node, out_node

live_stream = LiveStream(graph_manager)

live_stream.subscribe(analysis_db.insert)
live_stream.subscribe(LivePlot(...))

RE.subscribe(live_stream)
```

Documents flow:
* from the RunEngine's `Dispatcher` to `LiveStream`,
* through a `streamz.Stream` graph constructed by the graph manager,
* out through LiveStream's `Dispatcher`
* into any user-registered callbacks that take in the processed data.

Every time `LiveStream` receives a new RunStart, EventDescriptor, or RunStop document it calls `graph_manager` anew. The `graph_manager` always returns `(in_node, out_node)` of some Stream. The `out_node` may be `None` (i.e. if the Stream writes to disk and nothing flows out).

```python
# on LiveStream.__init__
graph_manager()
# when we get our first start doc
graph_manager(start)
# when we get our first descriptor
graph_manager(start, [desc])
# when/if we get another
graph_manager(start, [desc, desc2])
# etc.
# when we get our stop
graph_manager(start, [desc, desc2], stop)
# when a new run begins...
graph_manager(start)
```

Each time, `graph_manager` may extract any information it needs (such as mask from the RunStart or an exposure time from the EventDescriptor), build an arbitrarily complex `Stream`, and return the Stream's in and out nodes back to `LiveStream`. When `LiveStream` receives a new Event document, it emits it into `in_node`. The `LiveStream` sinks the out node into its `Dispatcher`.

The `graph_manger` may be any callable. When it called multiple times by `LiveStream`, it may build a new `Stream` each time or return the same one repeatedly. `LiveStream` will check whether the in and out nodes returned by `graph_manager` are new and behave appropriately (sinking new out nodes to the Dispatcher).

In the case where `graph_manager` needs to stitch together information from two or more EventDescriptors (e.g. background subtraction) the `graph_manager` may need to first return a Stream that caches Events. When the `graph_maanger` has enough EventDescriptors to fully operate, it will build the proper Stream, drain the cache Stream into it, and then return that proper Stream to `LiveStream` for receiving all future Events.

Discussion still in progress....

CJ-Wright:
I have a couple of questions (I'm still thinking about this so it might not be exhaustive).
1. How will we pull data from other run starts?
1. How do different pipelines get plumbed together? (Eg pipeline1 is reduction to 1D, pipeline2 is fitting 1D with gaussians)
1. How do we handle accumulation over multiple run starts?
1. Can we get multiple outputs of the stream?

danielballan:
1. If a given application needs external data (included but not limited to data accessible in some databroker) that is the domain of the Stream created by the `graph_manager`. `LiveStream` doesn't need to know.
2. I expect steps will be strung together as part of one Stream. Part of the advantage of Streams is that this is pretty easy to do. But one could also string together multiple `LiveStream` objects, as:
    ```python
    ls1 = LiveStream(graph_manager1)
    ls2 = LiveStream(graph_manager2)
    RE.subscribe(ls1)
    ls1.subscribe(ls2)
    ```
3. The `graph_manager` may hold state across runs.
4. If the user needs access to multiple `Dispatchers` / out nodes, I would call that out of scope for this class. Of course, if all the user needs is a record of partial records, Stream nodes can export data to disk as a side effect.

CJ-Wright:
I ask about 2) because in our previous discussions we have talked about the concept of separating chunks of the computation from one another (via "metanodes"). I would argue that having one large `LiveStream` would be a bit of a problem, as data is only turned back into documents on the outgoing side of a dispatcher (I think?). Thus every time one wants to write data into a databroker or use a visualization callback we need to have an end of the `LiveStream`. If each `LiveStream` is treated like a single "metanode" then that allows a bit more flexibility.

ZLLentz:
Wall of thoughts that I was having trouble articulating earlier:

We should design the API of the core library to be simpler wherever possible, especially for users who are trying to accomplish simple tasks. The design goal of this `LiveStream` should be to handle the assembly and dispatching of a single synthetic event stream that is structurally identical to a raw event stream.

A user that simply wants to average n events before sending them off to a standard callback should only need to provide the logic for reading the `data` dictionary and choosing when to output a new `data` dictionary.

I do not like the proposed `graph_manager` API. If I want to retain some state, I'm encouraged to provide some object with a single monolithic function that assembles and returns `streamz` nodes, perhaps parsing up to three different document types. The problem here is that now I have an object with a single monolithic function, and I have no incentive to clean it up.

If I need an object anyway, we should expand the API to allow the user to specify a collection of simpler methods for handling `event`, `descriptor`, `start`, and `stop` documents separately. This is exactly the specification of a generic `bluesky` callback, which is okay. The specification of each of these parts can be shifted to be more appropriate for `streamz` integration, if built-in `streamz` support is a desired goal. It is not clear to me that this should be a goal, because only requiring a callable makes this class more reusable and allows the user manage their `streamz` objects in custom ways.

The basic user implements just a data pipeline and ignores the rest, while an advanced user can specify anything they need to change the state of the object (and therefore reconfigure the pipeline) and they still benefit from not having to fully assemble or dispatch the documents. Pipelines of arbitrary complexity can be built in the middle, and these can do whatever the user wants with respect to saving to disk, sending to a processing node, etc., but only one hook for outputting the assembled documents is required because we can create and send data into secondary `LiveStream` objects without much difficulty by passing them our nodes. I think it is okay to require an advanced user to have one unique `LiveStream` object per unique output stream.

TLDR:

- API that encourages the user to create an object should leverage the advantages of having an object to keep simple tasks simple.
- I would rather pass in a readable object of a known structure than a monolithic function from an object.
- Building `streamz` support into this class makes it harder to use, not easier.

danielballan:
@ZLLentz Good points. Some quick reflection:

We could think of all of our existing callbacks (LivePlot, LiveFit, etc.) as "sinks." Documents flow in but not out, and some side-effect is produced. Building on that idea, a "`CallbackBase` with a `Dispatcher`" sounds like a useful unit to put in our library. It's roughly similar to a `streamz.Stream` node, but it has four "channels" (implemented as methods) for the four document types.

I still think the `graph_manager` concept is a good adapter between our four-channel callbacks and the single channel provided by `streamz.Stream`. Our four channels have different semantics (Events vs 'metadata') and the `graph_manager` accounts for that in a nice way: the 'metadata' documents define/configure the pipeline, and then the Events flow through it. That separation of state is in some ways cleaner than existing`CallbackBase` subclasses, which share state between methods in a less structured way.

But you make a strong argument that bluesky should _not_ opine "Got a multi-step analysis process? You should probably install `streamz`!" Sticking a `Dispatcher` onto `CallbackBase` is a more straightforward path. The `graph_manager` idea may only be worth the cognitive overhead if other `streamz` features are desired. It's fundamentally solving the problem of adapting four channels into one, and this has advantages and disadvantages.

ZLLentz:
@danielballan I think we're on the same page now. I'm warming up to the idea of a `graph_manager` so long as it isn't the only way to re-emit a custom document stream.

I'm imagining now that we split `LiveStream` into two separate classes:

- A class that facilitiates a `Dispatcher` and provides methods like `emit_event(self, data)`, `configure_metadata(self, md)`, `subscribe(self, cb)`. This exists so that a user doesn't have to re-implement how to build documents.
- A class that has built-in `streamz` support hooked up in the way described by `graph_manager`, which creates or includes the emitter class and calls it's methods when appropriate.

Then we can have strong `streamz` support and we can still support very simple callbacks.

bluesky/callbacks/stream.py
jrmlhermitte: what if an event is not filled? should the stream do some checking? (will need a `db` reference passed in as arg/kwarg)

CJ-Wright:
@jrmlhermitte as requested, here is an example of an averaging pipeline with SHED.  https://github.com/xpdAcq/SHED/pull/124/files

teddyrendahl:
Alright. @CJ-Wright @jrmlhermitte @ZLLentz @danielballan 

Pushed a big update. I'm going to walk through my general process, feel free to disagree. I think the first draft of this was cool in concept, but there are two distinct ideas here that are worth decoupling. 
* A user should be able to repackage data and emit documents so that it can be read by anything that supports the `bluesky` callback methods. This should be done as part of the bluesky library
* Streaming analysis is a cool and easy way to represent complex logical operations. It also fits well in a bluesky callback because you could easily distribute CPU intensive operations. This however, does not belong in bluesky core. I don't think we are in any position to demand that bluesky users implement their analysis in a specific way.

In this light, I have changed the name of the core class to `LiveDispatcher`. This does not have anything to do with `streamz` out of the box. All it does is collect documents, sanitize them, and then re-emit them to subscribers. (Talking with @tacaswell I think this could be core of the `EventBundler` class someday). Now, if you want to modify these document in anyway, you subclass `LiveDispatcher` implement `def event(self, doc)` function to shoot documents through your pipeline. Whatever pipeline you create is then responsible for calling `LiveDispatch.process_events`. This function accepts a partial event document with `data` and a reference to the **raw** descriptor uid. This is so that our `LiveDispatch` knows when to emit a new descriptor if the low-level device configuration changes. There are also additional optional kwargs if you want to release descriptors for your pipeline configuration.

Now back to `LiveStream`. I'm not sure there is an elegant model that takes an arbitrary `streamz` network with no extra code and still supports a full feature set. Questions like these  are hard to answer for every possible pipeline:
* Should event descriptors **and** events flow through the network?
* Do we destroy the network at the conclusion of every run?
* What if you want multiple out nodes?

I don't think bluesky should be in the business of defining these things. @CJ-Wright is taking the correct approach and defining these things in a separate package. With the tools here you can make all the tools yourself by subclassing `LiveDispatch` and reimplenting the event hooks. Want to send all the descriptors through your network? subclass `descriptor`. Want to add a bunch of metadata about your network? subclass `start`. At the end of the day, all bluesky should care is that if return a document that it can be converted into the `EventModel`.

Now in these tests I have one very simple subclass of `LiveDispatch` that turns all the keys negative, and one that uses streamz to average shots. I agree with @jrmlhermitte that the `AverageStream` should just live in the tests suite and get copy pasta to an examples page somewhere.

Also, @jrmlhermitte. Thanks for the comments. Planning on cleaning the docs once things are a little more settled.

danielballan:
I agree that the scope you've settled on here is a good place to draw a line.

I'm not sure if it's the _only_ good place to draw a line, and someday I might advocate for separately including some interface like `graph_manager(start, [desc, desc2], stop)` in bluesky, but I agree these things are best hashed out outside the core library.

I like that your PR handles the bluesky-specific concerns (data_keys, uids, our four channels) by extending/combining existing ideas.

jrmlhermitte:
I like this and will try to read it later when I have more time. Just simply for the record after quickly reading this, the suggestion to move `AverageStream` to docs/tests/examples was @danielballan suggestion and I seconded it. :) Sorry that I must have forgotten to mention that....

When I passed through it, I just have two small details. Shoud `emit` be called `process` or something else perhaps? The fact that `process_event` must be called by the `event` should be documented by the class as it is a pretty important detail.

Looks good!

danielballan:
> The fact that `process_event` must be called by the `event` should be documented by the class as it is a pretty important detail.

Agreed. I might suggest that the base class should do this, just so that it is a working (if trivial) example.

bluesky/callbacks/stream.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def descriptor(self, doc):
+        """Store a descriptor"""
+        self.raw_descriptors[doc['uid']] = doc
+        super().descriptor(doc)
+
+    def process_event(self, doc, stream_name='primary',
+                      id_args=None, config=None):
+        """
+        Receive an event document
```
tacaswell: Please include what the event is revived from.   Keep the to / from receive/send verbiage clear seems to the hardest documentation challenge for any networky type project :wink:

teddyrendahl: Are you referring to keeping the `uid` from the original `Event` document or the entire thing? I thought about making this mandatory in the call, but I didn't want to enforce that a single processed event from the pipeline needed to be linked with a single raw event.

teddyrendahl:
Sorry, had a busy week upon returning to SLAC. Will try to put the polish this shortly before or after Jan 1. 

1) Seems reasonable to have the base class just pass through Event document. Just to clarify, we want.

```python
      def event(self, evt, **kwargs):
             self.process_event(evt, **kwargs)
```
and when when reimplementing you just call `super().event(evt)` to emit the document.

2) @jrmlhermitte I can change the name of the `emit` method. I was stealing the name from the `RunEngine` method of the same purpose, but if `process` is clearer than I will change it.

jrmlhermitte:
sounds good. @teddyrendahl sorry that's my mistake yes `emit` is good (I would vote to keep API similar to `RunEngine` unless someone disagrees)

teddyrendahl:
Documentation added to the `callbacks.rst`. The base `event` methods now calls `process_event`. After clarification from @tacaswell, this should be ready for final review

bluesky/callbacks/stream.py
danielballan: I think that "include what the event is revived from" is supposed to say "include what the event is _received_ from" in the docstring -- something like:

```
Receive an event document from the emit method.
```

danielballan:
@teddyrendahl Sorry for letting this language. I took a shot at clarifying @tacaswell on his behalf, above.

teddyrendahl:
Ah I understand. Should be more explicit now.

PR #945, opened 2018-01-10
Turned warning into DeprecationWarning

jrmlhermitte:
should be proper warning `DeprecationWarning`
@tacaswell

### DIFF
--- a/bluesky/suspenders.py
+++ b/bluesky/suspenders.py
@@ -503,7 +503,8 @@ class SuspendOutBand(_SuspendBandBase):
             a generator, list, or similar containing `Msg` objects
     """
     def __init__(self, *args, **kwargs):
-        warn("bluesky.suspenders.SuspendOutBand is deprecated.")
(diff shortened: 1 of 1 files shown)


tacaswell:
can you add `stackleve=2` as well?

tacaswell:
nvm

danielballan:
Actually, I would like to revert this change. I use UserWarnings for these because they are the only type of warnings that display anything by default. DeprecationWarnings (and others) are silent by default. Example:

```python
# foo.py
import warnings

def deprecation():
    warnings.warn('this does not display by default', DeprecationWarning)

def user():
    warnings.warn('this displays by default', UserWarning)
```

```python
In [1]: import foo

In [2]: foo.deprecation()  # nothing happens!

In [3]: foo.user()
/Users/dallan/Documents/Repos/playbooks/foo.py:5: UserWarning: this displays by default
  warnings.warn('this displays by default', UserWarning)
```

PR #947, opened 2018-01-12
ENH: add a x2x_scan

tacaswell:
## Description

This is a generic version of a theta-2theta scan.

## Motivation and Context

Watching @ambarb do scans while testing the tardis at CSX she typed this signature into `relative_inner_product_scan` as a bit of wishful API.

@ambarb Did I understand this correctly and is this useful?  Do you have a better name?

## How Has This Been Tested?

Still needs test

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -614,7 +614,7 @@ def tune_centroid(
         number of points with each traversal, default = 10
     step_factor : float, optional
         used in calculating new range after each pass
-        
+
         note: step_factor > 1.0, default = 3
     snake : bool, optional
         if False (default), always scan from start to stop
@@ -1411,6 +1411,52 @@ def fly(flyers, *, md=None):
     for flyer in flyers:
         yield from bps.collect(flyer)
     yield from bps.close_run()
+
+
+def x2x_scan(detectors, motor1, motor2, start, stop, num, *,
+             per_step=None, md=None):
+    """
+    Relatively scan over two motors in a 2:1 ratio
+
+    This is a generalized version of a theta2theta scan
+
+    Parameters
+    ----------
+    detectors : list
+        list of 'readable' objects
+
+    motor1, motor2 : Positioner
+        The second motor will move half as much as the first
+
+    start, stop : float
+        The relative limits of the first motor.  The second motor
+        will move between ``start / 2`` and ``stop / 2``
+
+    per_step : callable, optional
+        hook for cutomizing action of inner loop (messages per step)
+        See docstring of bluesky.plan_stubs.one_nd_step (the default) for
+        details.
+
+    md : dict, optional
+        metadata
+
+
+    """
+    _md = {'plan_name': 'x2x_scan',
+           'plan_args': {'detectors': list(map(repr, detectors)),
+                         'motor1': motor1.name,
+                         'motor2': motor2.name,
+                         'start': start, 'stop': stop, 'num': num,
+                         'per_step': repr(per_step)}
+           }
+
+    _md.update(md or {})
+    return (yield from relative_inner_product_scan(
+        detectors, num,
+        motor1, start, stop,
+        motor2, start / 2, stop / 2,
+        per_step=per_step,
+        md=_md))
 
 
 relative_list_scan = rel_list_scan  # back-compat
--- a/bluesky/tests/test_scans.py
+++ b/bluesky/tests/test_scans.py
@@ -479,3 +479,22 @@ def test_relative_fermat_spiral(RE, hw):
     approx_multi_traj_checker(RE, scan,
                               _get_fermat_data(start_x, start_y),
                               decimal=2)
+
+
+def test_x2x_scan(RE,hw):
+    y_start = -1
+    y_stop = 4
+    y_num = 11
+    expected_traj = []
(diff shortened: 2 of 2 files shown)


ambarb:
yes, that is correct.  the generic option in bluesky is more flexible because it lets you decide if the range of two theta is 2x theta.  spec's th2th assumes this is what you would like so you only need to input the range for 2theta.

ambarb:
@tacaswell yes, after looking at the doc string your `x2x_scan` does what we discussed.  And I think your name for the scan makes good sense.  

However, the 2x motor is usually the detector and the frame of reference for the scientist creating the scan.  You may consider switching the order of the motors and the name.

For reference to discussion:
spec always did this a little backward as your scan arguments would be for the fast motor (and it would then move the second motor half as much) so the name for the scan (`th2th`) was the inverse of what was the underlying d2scan.
https://certif.com/spec_manual/mac_3_9_2.html

bluesky/plans.py
```diff
@@ -1413,6 +1413,50 @@ def fly(flyers, *, md=None):
     yield from bps.close_run()
 
 
+def x2x_scan(detectors, motor1, motor2, start, stop, num, *,
+             per_step=None, md=None):
+    """
+    Relatively scan over two motors in a 1:2 ratio
+
+    This is a pasteurized version of a theta2theta scan
```
danielballan: As much as I love metaphorical language, "generalized" might be friendlier here.

tacaswell:
Should this be flipped to move the second motor half of the first?

bluesky/plans.py
tacaswell: I suspect that is me losing to the spell-checker....

tacaswell:
flipped it so that the second motor moves half as much (as `2xx_scan` is not a valid function name in python).

danielballan:
I'm comfortable with the name `x2x_scan` even though its slightly misleading. I think most users of this plan will be reaching for it because they know about `theta2theta` and are accustomed to SPEC's (arguably backward) convention.

danielballan:
I think this is good to merge but I'd like @ambarb to take a last look and sign off before we merge it.

tacaswell:
and it needs a test!

ambarb:
Do you guys want someone from the beamline side to test it?

tacaswell:
@awalter-bnl Can I had this off to you?   Can you

 - make sure the name and signature make sense and match your expectations
 - add a test (something like listify it and assert that the sets have the right relationships)

You should be able to push directly to my branch.

awalter-bnl:
Hi @tacaswell , I can take a look at this. Can you provide some direction as to where the tests are usually kept? This way I can follow the existing layout for this new test.

danielballan:
The tests of plans are a little scattered about for historical reasons, but [`test_plans.py`](https://github.com/NSLS-II/bluesky/blob/master/bluesky/tests/test_plans.py#L184) is a good place to put new ones.

Have you encountered pytest _fixtures_ before? Notice that the tests take `RE` and `hw` ("hardware") as arguments. When pytest runs the tests, it inspects the tests' function signatures and automatically produces a RunEngine instance `RE` and a `SimpleNamespace` (a generic, dottable object) `hw` for use in the test so that we don't have to type that boilerplate every time or update it in hundreds of different places when something changes. Something like `hw.motor` and `hw.det` are appropriate in this case.

Also notice the usage `MsgCollector`, which is useful for comparing that _actual_ messages to the expected ones. That is worth imitating. But no need to imitate @tacaswell's spelling of 'expected' as 'expecte'. #thedangersofautocomplete ;- )

awalter-bnl:
Thanks @danielballan I had a 'brief' introduction when doing the 'spiral_square' plan, but Maksim did most of the test function work. I will ask questions if I have any !-)

mrakitin:
For reference: https://github.com/NSLS-II/bluesky/pull/1001/files#diff-7d8393473493d7106f01ce5900958096R554

awalter-bnl:
@tacaswell I have checked as requested, and added the test to tests/test_scans.py (including running it to ensure there are no issues). I don't see any issues with the name or signature.

Take a look at the test and let me know if you see any issues.

bluesky/tests/test_scans.py
```diff
@@ -479,3 +479,22 @@ def test_relative_fermat_spiral(RE, hw):
     approx_multi_traj_checker(RE, scan,
                               _get_fermat_data(start_x, start_y),
                               decimal=2)
+
+
+def test_x2x_scan(RE,hw):
```
danielballan: Style thing: needs a space after this comma. (Soon we'll probably apply a linter, as many large project do, so these kinds of things we'll actually be _tested_.)

danielballan:
Thanks, @awalter-bnl and @mrakitin!

awalter-bnl:
Thanks @danielballan, I missed that space but it has now been fixed.

bluesky/tests/test_scans.py
danielballan: Yes, there is a lot of cleaning up to do on all of our projects. I have a branch (somewhere...) that attempted to do this for bluesky but it's probably hopelessly out of date now. Might as well start over.

PR #951, opened 2018-01-16
Fix liveraster

tacaswell:
```python
from bluesky import RunEngine
from ophyd.sim import motor1, motor2, det4
import bluesky.plans as bp
import ophyd
install_qt_kicker()


ophyd.setup_ophyd()
bec = BestEffortCallback()
RE = RunEngine()
RE.subscribe(bec)
det4.exposure_time = 1

# try the permutations of this changing the low -> high or high -> low sense of scans
RE(bp.grid_scan([det4], motor1, 7, 0, 7, motor2, 0, 5, 5, True))
```


danielballan:
Can you explain more about what was wrong with our original thinking? I'm confused by the relationship between 'origin' and 'extent'. Do we really need to hard-code 'origin'?

danielballan:
(P.S. Thank you for tackling this tedious issue!!)

bluesky/callbacks/best_effort.py
```diff
@@ -263,7 +263,7 @@ def descriptor(self, doc):
             gridding = self._start_doc.get('hints', {}).get('gridding')
             if gridding == 'rectilinear':
                 self._live_grids[doc['uid']] = {}
-                x_key, y_key = dim_fields
+                slow, fast = dim_fields
```
danielballan: :+1: Local variable names matter. :-D

tacaswell:
Thanks @jrmlhermitte !

PR #952, opened 2018-01-17
DOC: Fixed some syntax in doc on callbacks

jrmlhermitte:
I'm sure there could be more typos, but I just noticed this one in passing.


jrmlhermitte:
thanks!

PR #953, opened 2018-01-17
Remove the deprecation warning

jrmlhermitte:
Undoing #945 (sorry @danielballan ).

As an alternative, we can get together and vote on a force push to remove the commit but I personally don't feel comfortable with this.


tacaswell:
Major :-1: on force pushing (except to your own dev branches).

danielballan:
Haha, this is certainly _not_ the most thrash-y pair of commits in bluesky's git history -- no big deal. Definitely no need to force-push.

@stuartcampbell I want to check that you are OK with this reversion before merging since you :+1:-ed the original. Do you agree that user visibility bests semantics in this case?

PR #901, opened 2017-11-01
Merge doc- and test-requirements.txt

mrakitin:
Closes #900.

PR #954, opened 2018-01-17
add doct to requirements

CJ-Wright:
Imported in `blueksy.callbacks.broker`

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,3 @@
-attrs
 coverage
 databroker
 doctr
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ numpy
 super_state_machine
 toolz
 tqdm
+doct
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,6 @@
+attrs
 cycler
+doct
 event-model
 historydict
 jsonschema
@@ -6,4 +8,3 @@ numpy
 super_state_machine
 toolz
 tqdm
-doct


CJ-Wright:
Ready for Review!

CJ-Wright:
done, although I don't really understand the dev vs not dev reqs.

tacaswell:
can you move `attrs` too?

mrakitin:
It's used by travis: https://github.com/NSLS-II/bluesky/blob/7e49f70fe128ac9b8198435b90781f70ee1d7277/.travis.yml#L29. I think the idea was to install bluesky with less dependencies, but if users want full functional, then optional dev dependencies should be installed. @danielballan, @tacaswell, any comments?

tacaswell:
Ohh, I understand this now, never mind.

`requirement` is things you need too run, `dev-requirements` are things you need to run the tests.

We should probably just remove the use of `doct`....

danielballan:
Thanks for addressing this. I agree that we should just remove doct. In light of stated plans for databroker, this should just use plain dicts. (Search for the phrase "custom dictionary" in the [databroker release notes](https://nsls-ii.github.io/databroker/whats_new.html).)

CJ-Wright:
Just for context I ran into this because the conda-forge build of bluesky caused breakages from this missing dep.
@danielballan Is that going to be in the release that goes out to the beamlines?
@mrakitin Would there be any interest in breaking this up a bit more (build, run, test, doc)?

mrakitin:
@CJ-Wright, not sure, maybe not. We tended to merge the requirements. See, e.g. https://github.com/NSLS-II/bluesky/pull/901/files.

danielballan:
@CJ-Wright Yes, in fact that was released last cycle (including at XPD).

CJ-Wright:
I meant if the total removal of doct was going to go into a release this cycle.

danielballan:
Oh, sorry. No, not this cycle. This cycle, we _just_ got some of the more complex beamlines updated to the version of databroker where we deprecated it, so we should let it sit for a cycle or two before full removing.

CJ-Wright:
Can this go in?

tacaswell:
I am :-1: on merging this, will have a PR to just hide the import shortly.

danielballan:
Agreed -- we should just remove doct or hide the import.

PR #957, opened 2018-01-19
FIX/DOC: Fix a hilarious bug in ts_msg_hook, and document ts_msg_hook.

danielballan:


bluesky/utils.py
```diff
@@ -926,7 +926,7 @@ def ensure_uid(doc_or_uid):
 
 
 def ts_msg_hook(msg):
-    t = '{:%H:%m:%S}'.format(datetime.datetime.now())
+    t = '{:%H:%M:%S.%f}'.format(datetime.datetime.now())
     msg_fmt = '{: <17s} -> {!s: <15s} args: {}, kwargs: {}'.format(
```
mrakitin: For future reference: https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior

Issue #961, opened 2018-01-24
inner_product_scan plots extra axis


jrmlhermitte:
```py
RE(inner_product_scan([hutchb_diag], 11, jj_slits.inboard, -2, 2, jj_slits.outboard, -1.8, 
    ...: -1.8+4), hints={'fields': [hutchb_diag.stats1.total.name], 'dimensions' : [([jj_slits.outbo
    ...: ard.name, jj_slits.inboard.name], 'primary')]})

```
plots an extra axis:

<img src="https://user-images.githubusercontent.com/705366/35359611-c8bedeb8-0128-11e8-970a-c74194c8f463.png" width=200 alt="hutchb_diag_stats1_total_jj_slits_inboard_vs_jj_slits_outboard-1" />

Only the bottom plot should be present.

jrmlhermitte:
it seems the issue was resolved for relative scans but not absolute. i think the BEC hints logic should move from relative to absolute

jrmlhermitte:
Here is a script to reproduce results, and the plots one by one:
```python
from bluesky.plans import (relative_outer_product_scan,
                           relative_inner_product_scan,
                           outer_product_scan,
                           inner_product_scan)
from bluesky.plan_stubs import mov

from ophyd.sim import det4, motor1, motor2
from bluesky import RunEngine
from bluesky.callbacks.best_effort import BestEffortCallback

RE = RunEngine({})
# subscribe BEC
bec = BestEffortCallback()
RE.subscribe(bec)


# move motor to a reproducible location
RE(mov(motor1, 0))
RE(mov(motor2, 0))
RE(relative_outer_product_scan([det4], motor1, -1, 1, 20, motor2, -2, 2, 20, True))
RE(outer_product_scan([det4], motor1, -1, 1, 20, motor2, -2, 2, 20, True))

# move motor to a reproducible location
RE(mov(motor1, 0))
RE(mov(motor2, 0))
RE(relative_inner_product_scan([det4], 20, motor1, -1, 1, motor2, -2, 2))
RE(inner_product_scan([det4], 20, motor1, -1, 1, motor2, -2, 2))
```

![bec_outer_1](https://user-images.githubusercontent.com/705366/35361874-fe251ae2-0130-11e8-95cf-758c7906c04b.png)
![bec_outer_2](https://user-images.githubusercontent.com/705366/35361875-01ab4268-0131-11e8-95af-ac0a93e762ab.png)
[...]

PR #963, opened 2018-01-24
BUG : Fix inner product

jrmlhermitte:
See issue https://github.com/NSLS-II/bluesky/issues/961

So far, this first commit fixes the inner product. It now looks like:
inner_product

need to fix so that first plot is not plotted

### DIFF
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -112,7 +112,9 @@ class BestEffortCallback(CallbackBase):
             warn("We are ignoring the dimensions hinted because we cannot "
                  "combine streams.")
 
-        # TODO : Add option to select multiple dim_fields
+        # for each dimension, choose one field only
+        # the plan can supply a list of fields. It's assumed the first
+        # of the list is always the one plotted against
         self.dim_fields = [fields[0]
                            for fields, stream_name in dimensions]
 
@@ -120,10 +122,10 @@ class BestEffortCallback(CallbackBase):
         # motivation for this is that when plotting, we find dependent variable
         # by finding elements that are not independent variables
         # TODO : add list comprehension
-        self.independent_vars = list()
+        self.all_dim_fields = list()
         for fields, stream_name in dimensions:
             for field in fields:
-                self.independent_vars.append(field)
+                self.all_dim_fields.append(field)
 
 
         _, self.dim_stream = dimensions[0]
@@ -173,7 +175,7 @@ class BestEffortCallback(CallbackBase):
         if stream_name == self.dim_stream:
             # Ensure that no independent variables ('dimensions') are
             # duplicated here.
-            columns = [c for c in columns if c not in self.independent_vars]
+            columns = [c for c in columns if c not in self.all_dim_fields]
 
             if self._table_enabled:
                 self._table = LiveTable(list(self.dim_fields) + columns)
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -121,12 +121,9 @@ class BestEffortCallback(CallbackBase):
         # make distinction between flattened fields and plotted fields
         # motivation for this is that when plotting, we find dependent variable
         # by finding elements that are not independent variables
-        # TODO : add list comprehension
-        self.all_dim_fields = list()
-        for fields, stream_name in dimensions:
-            for field in fields:
-                self.all_dim_fields.append(field)
-
+        self.all_dim_fields = [field
+                               for fields, stream_name in dimensions
+                               for field in fields]
 
         _, self.dim_stream = dimensions[0]
 
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -175,7 +175,8 @@ class BestEffortCallback(CallbackBase):
             columns = [c for c in columns if c not in self.all_dim_fields]
 
             if self._table_enabled:
-                self._table = LiveTable(list(self.dim_fields) + columns)
+                # plot everything, independent or dependent variables
+                self._table = LiveTable(list(self.all_dim_fields) + columns)
                 self._table('start', self._start_doc)
                 self._table('descriptor', doc)
 
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -111,9 +111,21 @@ class BestEffortCallback(CallbackBase):
             dimensions = GUESS  # Fall back on our GUESS.
             warn("We are ignoring the dimensions hinted because we cannot "
                  "combine streams.")
+
         # TODO : Add option to select multiple dim_fields
         self.dim_fields = [fields[0]
                            for fields, stream_name in dimensions]
+
+        # make distinction between flattened fields and plotted fields
+        # motivation for this is that when plotting, we find dependent variable
+        # by finding elements that are not independent variables
+        # TODO : add list comprehension
+        self.independent_vars = list()
+        for fields, stream_name in dimensions:
+            for field in fields:
+                self.independent_vars.append(field)
+
+
         _, self.dim_stream = dimensions[0]
 
         # Print heading.
@@ -161,7 +173,7 @@ class BestEffortCallback(CallbackBase):
         if stream_name == self.dim_stream:
             # Ensure that no independent variables ('dimensions') are
             # duplicated here.
-            columns = [c for c in columns if c not in self.dim_fields]
+            columns = [c for c in columns if c not in self.independent_vars]
 
             if self._table_enabled:
                 self._table = LiveTable(list(self.dim_fields) + columns)
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -1,3 +1,8 @@
+'''
+    Best Effort Callback.
+    For instructions on how to test in a simulated environment please see:
+        tests/interactive/best_effort_cb.py
+'''
 from cycler import cycler
 from datetime import datetime
 from io import StringIO
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -200,7 +200,10 @@ class LiveScatter(CallbackBase):
         ax.margins(.1)
         self._xdata, self._ydata, self._Idata = [], [], []
         self._norm = mcolors.Normalize()
+        self._minx, self._maxx, self._miny, self._maxy = (None,)*4
 
+        self.xlim = xlim
+        self.ylim = ylim
         if xlim is not None:
             ax.set_xlim(xlim)
         if ylim is not None:
@@ -233,12 +236,28 @@ class LiveScatter(CallbackBase):
         super().event(doc)
 
     def update(self, x, y, I):
+        # if one is None all are
+        if self._minx is None:
+            self._minx = x
+            self._maxx = x
+            self._miny = y
+            self._maxy = y
+
         self._xdata.append(x)
         self._ydata.append(y)
         self._Idata.append(I)
         offsets = np.vstack([self._xdata, self._ydata]).T
         self.sc.set_offsets(offsets)
         self.sc.set_array(np.asarray(self._Idata))
+
+        if self.xlim is None:
+            minx, maxx = np.minimum(x, self._minx), np.maximum(x, self._maxx)
+            self.ax.set_xlim(minx, maxx)
+
+        if self.ylim is None:
+            miny, maxy = np.minimum(y, self._miny), np.maximum(y, self._maxy)
+            self.ax.set_ylim(miny, maxy)
+
         if self.clim is None:
             clim = np.nanmin(self._Idata), np.nanmax(self._Idata)
             self.sc.set_clim(*clim)
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -697,8 +697,8 @@ def scan_nd(detectors, cycler, *, per_step=None, md=None):
            }
     _md.update(md or {})
     try:
-        dimensions = [([motor.hints['fields'] for motor in cycler.keys], 'primary')
-                      ]
+        dimensions = [(motor.hints['fields'], 'primary')
+                      for motor in cycler.keys]
     except (AttributeError, KeyError):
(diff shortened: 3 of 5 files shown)


jrmlhermitte:
inner_fixe

last commit fixed inner product as well

jrmlhermitte:
It all works and it's good to go. Here is the script I used to test (needed a `db` instance as well to check the descriptor, just in case...)

```py
from bluesky.plans import (relative_outer_product_scan,
                           relative_inner_product_scan,
                           outer_product_scan,
                           inner_product_scan)
from bluesky.plan_stubs import mov

from ophyd.sim import det4, motor1, motor2
from bluesky import RunEngine

from databroker import Broker

# db setup
config = temp_config()
tempdir = config['metadatastore']['config']['directory']

def cleanup():
    shutil.rmtree(tempdir)
db = Broker.from_config(config)


RE = RunEngine({})
# subscribe BEC
bec = BestEffortCallback()
RE.subscribe(bec)
RE.subscribe(db.insert)



# move motor to a reproducible location
RE(mov(motor1, 0))
RE(mov(motor2, 0))
RE(relative_outer_product_scan([det4], motor1, -1, 1, 20, motor2, -2, 2, 20, True))
RE(outer_product_scan([det4], motor1, -1, 1, 20, motor2, -2, 2, 20, True))

# move motor to a reproducible location
RE(mov(motor1, 0))
RE(mov(motor2, 0))
RE(relative_inner_product_scan([det4], 20, motor1, -1, 1, motor2, -2, 2))
RE(inner_product_scan([det4], 20, motor1, -1, 1, motor2, -2, 2))
```

jrmlhermitte:
Summary of what happened:
Previous PR fixed the outer product scan. However, it changed the interpretation of a key variable used to guess independent and depdendent vars. Here is a summary the of the vars, in the BestEffortCallBack object:
- `self.dim_fields` : the dependent variables that we should be plotting against
- `self.independent_vars` : the independent variables of the system
- `self.fields` : the dependent variables of the system

note that in the case of inner product, there are two independent variables, but one is chosen for plotting so that `dim_fields` is a smaller subset of `independent_vars` (and they're not equal).

I tried my best to document so that if we encounter more issues we can troubleshoot them. I hope `independent_vars` is a reasonable variable name.


Also, the first commit moves the hints from the `relative_inner_product_scan` to the `inner_product_scan` for the obvious reason that the former calls the latter, but both should contain the same hints,

bluesky/callbacks/best_effort.py
```diff
@@ -111,9 +111,21 @@ def start(self, doc):
...
+
         # TODO : Add option to select multiple dim_fields
         self.dim_fields = [fields[0]
                            for fields, stream_name in dimensions]
+
+        # make distinction between flattened fields and plotted fields
+        # motivation for this is that when plotting, we find dependent variable
+        # by finding elements that are not independent variables
+        # TODO : add list comprehension
+        self.independent_vars = list()
```
tacaswell: We (@danielballan @stuartcampbell and I) spent a fair amount of time talking about the names on this and settled on 'dimensions' over 'independent' and 'exogenous',  I would rather not re-introduce that ambiguity.   Could this be 

```python
self._no_plot = {f for fields, stream in dimensions for f in fields}
```

?

bluesky/callbacks/best_effort.py
```diff
@@ -161,7 +173,7 @@ def descriptor(self, doc):
         if stream_name == self.dim_stream:
             # Ensure that no independent variables ('dimensions') are
             # duplicated here.
-            columns = [c for c in columns if c not in self.dim_fields]
+            columns = [c for c in columns if c not in self.independent_vars]
```
tacaswell: This will also remove the extra dimensions from the table which I suspect they do not want.

bluesky/callbacks/best_effort.py
jrmlhermitte: there should be two types of fields: the independent vars we want to plot and the ones we don't want to plot. how about dim_fields and dim_fields_selected?

bluesky/callbacks/best_effort.py
jrmlhermitte: It only will if `dim_fields` contains *all* the motors. In some cases it does not. Currently, `dim_fields` is interpreted as the number of independent axes (so one means 1d plot, two means 2d plot)
in the case of inner_product_scan  something like dimensions = (['motor1', 'motor2'], 'primary') was transformed into dimensions= (['motor1'], 'primary') (where it's just assumed we plot versus first motor)
 (done here: https://github.com/NSLS-II/bluesky/pull/963/files#diff-3b88b6e90432890fa4a5951ebfc3b6d0L115)

bluesky/plans.py
```diff
@@ -697,8 +697,8 @@ def scan_nd(detectors, cycler, *, per_step=None, md=None):
            }
     _md.update(md or {})
     try:
-        dimensions = [(motor.hints['fields'], 'primary')
-                      for motor in cycler.keys]
+        dimensions = [([motor.hints['fields'] for motor in cycler.keys], 'primary')
```
tacaswell: I don't think this is right as it says for a general scan_nd there is only one dimension and it is all of the keys.

In general this should be n-dimensions (so `len(dimensions) > 1`) each of which has 1 fiield.  It should be the job of `inner_product_scan` which knows that there is really only 1 dimension to replace the default / general case with the specialized case.

jrmlhermitte: yes this is erroneous and should not have been touched thanks! removed

jrmlhermitte:
Fixed some portions of code. For setting dimentions:
* inner_product on motor1, motor2 and we can to plot versus motor2:
`{'dimensions' : [(['motor2', 'motor1'], 'primary')]}`
* outer_product on motor1, motor2 (and we obviously plot versus both):
`{'dimensions' : [(['motor2'], 'primary'),  (['motor1'], 'primary')]}`

bluesky/callbacks/best_effort.py
```diff
@@ -161,7 +172,7 @@ def descriptor(self, doc):
         if stream_name == self.dim_stream:
             # Ensure that no independent variables ('dimensions') are
             # duplicated here.
-            columns = [c for c in columns if c not in self.dim_fields]
+            columns = [c for c in columns if c not in self.all_dim_fields]
 
             if self._table_enabled:
                 self._table = LiveTable(list(self.dim_fields) + columns)
```
tacaswell: This should use `all_dim_fields` instead.

tacaswell:
Can you add a test for `outer_product_scan`, `inner_product_scan`, and `scan_nd` with 3 motors to verify that dimensions has the right length?

danielballan:
Also see `bluesky/tests/interactive/example.ipy` which is an interactive test meant to be validated by a human. That might be a good place to add some of these checks also.

bluesky/callbacks/best_effort.py
jrmlhermitte: good catch! livetable should plot everything

jrmlhermitte:
thanks will add tests later. i like examples.ipy. wish i had seen it before before i scrambled across documentation to get a sim working :-D

danielballan:
I should add a really noticeable comment about it at the top of `best_effort.py`....

jrmlhermitte:
A few things added:
- a test ( see https://github.com/NSLS-II/bluesky/pull/963/files#diff-c792e0c1490691a38226918d14b94356R592V)
- logic to autoscale x and y limits when doing the scatter plot (https://github.com/NSLS-II/bluesky/pull/963/files#diff-3b7aa3fc7a1649edaa3cca7f6b707052R253)
- another interactive test of which the last three produce figures below.
(https://github.com/NSLS-II/bluesky/pull/963/files#diff-1f51be1ac985c3aaac8fe652189d6dacR62)

hints_line
hints_scatter
hints_gridding_rect

jrmlhermitte:
fixed a small plotting in example, they now all overlay ( this was a choice of plotting limits, not issue in bec)
det4_vs_motor1

jrmlhermitte:
There was a small issue with the test. Also, I was running way too long of a scan: 20*20*30=12000 steps! I decreased this to 2*2*3=12....
If tests pass, this is good to go for me. Let me know if there are any other comments

bluesky/callbacks/mpl_plotting.py
```diff
@@ -233,12 +236,28 @@ def event(self, doc):
...
         self._xdata.append(x)
         self._ydata.append(y)
         self._Idata.append(I)
         offsets = np.vstack([self._xdata, self._ydata]).T
         self.sc.set_offsets(offsets)
         self.sc.set_array(np.asarray(self._Idata))
+
+        if self.xlim is None:
+            minx, maxx = np.minimum(x, self._minx), np.maximum(x, self._maxx)
+            self.ax.set_xlim(minx, maxx)
```
jrmlhermitte: @tacaswell I tried different functions that appeared to automatically reset limits, i could not get them to work correctly, so i have left a manual limit setting per event for now.

jrmlhermitte:
the tests pass, but then Travis fails on py3.5 for some reason. not sure why?

jrmlhermitte:
```
bluesky/tests/test_examples.py::test_sample_md_dict_requirement /home/travis/.travis/job_stages: line 57:  3524 Aborted                 (core dumped) coverage run run_tests.py
```

bluesky/callbacks/mpl_plotting.py
danielballan: I believe the auto-limiters include nice margins, which we are forgoing here. Can you open an issue to revisit this and hopefully use matplotlib's relimiter in the future? Doesn't have to block this PR.

tacaswell: Did you use https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.relim.html?highlight=relim#matplotlib.axes.Axes.relim ?

bluesky/plans.py
```diff
@@ -764,7 +765,29 @@ def inner_product_scan(detectors, num, *args, per_step=None, md=None):
...
+    default_dimensions = [(x_fields, 'primary')]
+
+    default_hints = {}
+    if len(x_fields) > 0:
+        default_hints.update(dimensions=default_dimensions)
+
+    # now add default_hints and override any hints from the original md (if
+    # exists)
+    _md['hints'] = default_hints
+    _md['hints'].update(md.get('hints', {}) or {})
```
tacaswell: Do not think you need the `or {}`?

bluesky/tests/test_callbacks.py
```diff
@@ -586,3 +587,47 @@ def event(self, doc):
...
     RE(count([hw.direct_img]))
+
+
+def test_plotting_hints(RE, hw, db):
+    ''' This tests the run and checks that the correct hints are created.
+        Hints are mainly created to help the BestEffortCallback in plotting the
+        data.
+        Use a callback to do the checking.
+    '''
+    class HintChecker(CallbackBase):
```
tacaswell: It is not documented, but https://github.com/NSLS-II/bluesky/blob/master/bluesky/tests/utils.py#L30 is helpful for this sort of test.

bluesky/plans.py
jrmlhermitte: can `None`  be passed to `md`?

bluesky/tests/test_callbacks.py
jrmlhermitte: great, let's use this. ill replace `HintChecker`.

bluesky/callbacks/mpl_plotting.py
jrmlhermitte: @danielballan yes i thought of writing my own, but then realized, sometimes we may want an offset, sometimes a percentage. ill try `relim` again just in case, but yes i did. ill get back to you. thanks!

jrmlhermitte: It still doesn't work. I'm in interacitve mode. Here's code that can repdroduce this:
```py
from pylab import *
ion()
clf();plot([1,2,3,], [4,5,6])
xlim(0,1)
gca().relim()
```

jrmlhermitte: (does not change limits back to normal). Is this potentially a bug? (or my poor usage?)

jrmlhermitte:
ok tests pass. The `relim` function doesn't seem to work on my setup. I won't put it in. Let me know if this is good to go. (And I'll make an issue for the `relim` problem in `bluesky` , which may potentially go to `matplotlib`)

danielballan:
I’m tardy in reviewing this, but I’m halfway through. We’ll get this merged tomorrow I promise but I’d like to finish my close read first.

On Thu, Feb 1, 2018 at 7:18 PM -0500, "Julien Lhermitte" <notifications@github.com<mailto:notifications@github.com>> wrote:


ok tests pass. The relim function doesn't seem to work on my setup. I won't put it in. Let me know if this is good to go. (And I'll make an issue for the relim problem in bluesky , which may potentially go to matplotlib)

bluesky/callbacks/mpl_plotting.py
```diff
@@ -200,7 +200,10 @@ def __init__(self, x, y, I, *, xlim=None, ylim=None,
         ax.margins(.1)
         self._xdata, self._ydata, self._Idata = [], [], []
         self._norm = mcolors.Normalize()
+        self._minx, self._maxx, self._miny, self._maxy = (None,)*4
```
danielballan: `self._minx = self._maxx = self._miny = self._maxy = None` is a slightly more boring way to do this.

bluesky/tests/test_callbacks.py
danielballan: Neat!

bluesky/callbacks/mpl_plotting.py
danielballan: I can reproduce that. I checked `gca()._stale == False`; it does not seem to be a drawing issue.

tacaswell: That is the expected behavior `relim` computes the new data limits, but does not change them.  You need to call `autoscale_view`. 

However, reading the docstring it does not take into account `Collection` instances.  Do not rember the details of why off the top of my head though.

bluesky/callbacks/mpl_plotting.py
jrmlhermitte: thanks for the suggestion!

PR #968, opened 2018-02-01
MNT: move `doct` import from top-level

tacaswell:
alternative to #954


danielballan:
I'm not sure I understand why this needed to be a doct in the first place. All the other callbacks that stash the RunStart document just stash it as a dict.

tacaswell:
because it get run through a format string which may be using '.' access because it makes the string easier to read/write.

danielballan:
Ah -- good enough!

PR #969, opened 2018-02-01
first pass on documentation for Msg

jrmlhermitte:
Still needs to be elaborated on. All with push permissions on this repo, feel free to push to this branch.

### DIFF
--- a/doc/source/msg.rst
+++ b/doc/source/msg.rst
@@ -8,13 +8,43 @@ Message Protocol
 Overview
 --------
 
-The `RunEngine` oversees all commands of a run, intelligently rewinding and
-re-running commands during a run in order to produce a clean result. How does
-it work? On the basic level, it works by storing a list of atomistic operations
-to be run, and running them in sequence. Each of these atomistic operations are
-represented by a `bluesky.Msg` object. Thus, every plan designed for use by the
-`RunEngine` is a list of `Msg` objects. To fully capture the provenance of a
-command, four items are needed:
+A *plan* is a sequence of atomic operations describing a data acquisition
+procedure. Each operation is represented by a ``bluesky.Msg`` ("message")
+object. A plan may be implemented as a simple list of messages:
+
+.. code-block:: python
+
+    # (Behold, the most boring data acquisition ever conducted!)
+    plan = [Msg('open_run'), Msg('close_run')]
+
+or as a generator the yields messages one at time:
+
+.. code-block:: python
+
+    def plan():
+        yield Msg('open_run')
+        yield Msg('close_run')
+
+The above examples are equivalent. For more sophisticated uses, the second one
+is more powerful, as it can incorporate loops, conditionals, adaptive logic ---
+generally any Python code.
+
+But, crucially, the plan code itself must not communicate with hardware.
+(You should never put ``epics.caput(...)`` in a plan!) Rather, each operation
+is represented by a ``Msg`` object that *describes* what should be done. This
+makes it safe to introspect the plan for error-checking, simulation, and
+visualization purposes --- without touching real hardware. For example, we
+could print each message in the plan like so:
+
+.. code-block:: python
+
+    plan = [Msg('open_run'), Msg('close_run')]
+
+    # a very, very simple 'plan simulator'
+    for msg in plan:
+        print(msg)
+
+A ``Msg`` has four members, accessible as attributes:
 
 - command
 - obj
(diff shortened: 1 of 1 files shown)


doc/source/msg.rst
```diff
@@ -8,21 +8,33 @@ Message Protocol
 Overview
 --------
 
-The `bluesky.Msg` object is a `namedtuple` subclass which has the fields
+The `RunEngine` oversees all commands of a run, intelligently rewinding and
+re-running commands during a run in order to produce a clean result. How does
+it work? On the basic level, it works by storing a list of atomistic operations
+to be run, and running them in sequence. Each of these atomistic operations are
+represented by a `bluesky.Msg` object. Thus, every plan designed for use by the
+`RunEngine` is a list of `Msg` objects. To fully capture the provenance of a
```
tacaswell: This is a very picky comment, but 'list' is confusing because it can be conflated with `list` (the built-in python type).  'Sequence' or 'series' are better words.

tacaswell: I don't really like 'provenance' here, it feels off.  I have started to think of the `Msg` objects as analogs of machine instructions, python byte code, or an RPC call.

danielballan:
I couldn't make pushing to this branch work, so I opened a PR: https://github.com/jrmlhermitte/bluesky/pull/1/files

It's a pretty big change, but it retains most of the ideas in your first pass. I particularly like the way you describe messages as "atomic".

danielballan:
Oh, nevermind, I was trying to push to @jrmlhermitte's remote, but this is branch is on the upstream fork. Pushed.

jrmlhermitte:
great thanks! I'll close the PR on my fork then :-)

doc/source/msg.rst
```diff
@@ -8,25 +8,108 @@ Message Protocol
...
+    def plan():
+        yield Msg('open_run')
+        yield Msg('close_run')
+
+The above examples are equivalent. For more sophisticated uses, the second one
+is more powerful, as it can incorporate loops, conditionals, adaptive logic ---
+generally any Python code.
+
+But, crucially, the plan code itself must not communicate with hardware.
+(You should never put ``epics.caput(...)`` in a plan!) Rather, each operation
```
jrmlhermitte: :+1:

doc/source/msg.rst
```diff
@@ -8,25 +8,108 @@ Message Protocol
...
 
-``command`` must be one of a controlled list of commands, ``obj`` is the
-object to apply the command to and ``args`` and ``kwargs`` are arguments to
-the command.  Any ``args`` or ``kwargs`` not consumed by the run engine are
-passed through to the calls on the objects.
-
-The `RunEngine` has a registry which is used to dispatch the `Msg` objects
-based on the value of the `Msg.cmd`.  By default a basic set of commands are
-registered, but users can register their own functions to add custom commands.
+where ``command`` must be one of a controlled list of commands, ``obj`` is the
```
jrmlhermitte: good idea to remove discussion on `NamedTuple` overall

doc/source/msg.rst
```diff
@@ -8,25 +8,108 @@ Message Protocol
...
+object (i.e. Device) to apply the command to, if applicable, and ``args`` and
+``kwargs`` are arguments to the command.
+
+The execute the plan, the RunEngine consumes it, one message at a time.
+
+.. code-block:: python
+
+    def very_simple_run_engine(plan):
+        for msg in plan:
+            # Process the msg.
```
tacaswell: should we include a link to https://nsls-ii.github.io/bluesky/run_engine.html

prjemian: 👍

CJ-Wright:
seems nice

doc/source/msg.rst
```diff
@@ -104,27 +187,44 @@ of immediate motion
 trigger
 +++++++
 
+.. automethod:: bluesky.run_engine.RunEngine._trigger
```
jrmlhermitte: I added these but the docs don't look so nice with them. It would be nice to take the docstring just from these functions and paste them here.

danielballan: Tough call. Reuse is good, especially because it help us avoid accidental inconsistencies, but readability is also good.

danielballan:
Per discussion in private channels with @jrmlhermitte and @tacaswell, we plan to remove the `automethod` block and instead write new documentation, per command, with this structure:

1. Affect on hardware
2. Affect on RE (affected caches, etc)
3. Msg args/kwargs expected
4. Related plans

jrmlhermitte:
Yes, and in **affect on hardware**, I think we decided that a distinction should be made between Msg's that call ophyd methods and Msg's that don't.

doc/source/msg.rst
mrakitin: @danielballan, I don't see it here. Did you push?

jrmlhermitte: done

jrmlhermitte:
I removed the autodocs and wrote in some rough description of the `Msg` objects.
I also added in the other missing messages from the command registry.

I think this needs two things:
1. I think this still needs some work. Personally, I like the doc in `configure` the best. I would like to work all other docstrings to have this form.
2. We should separate the messages that result in calls to methods on `obj.method()` and expect `Status` objects from the rest. I think we could break this up into:
  1. Messages that call objects and
      1.  expect status objects
      2. other data (like read() collect() etc)
  2. Messages that do not call objects

Also, we have one method `wait_for`, which waits on `Future`s. I'm wondering if all ophyd methods should be converted to return `asyncio` Futures rather than status obejcts in the future? I believe they do the same thing, contain callbacks. I'm not familiar with the `CAThreads` much and if the conversion would be simple but to first order naively it does. This is a digression.

doc/source/msg.rst
```diff
@@ -186,16 +186,20 @@ of immediate motion.
...
 +++++++
 
-This will call the ``obj.trigger`` method and cache the returned status object.
+This will call the ``obj.trigger`` method and cache the returned status object
+and caches the returned status object.
 
 
 unstage
 +++++++
+Instruct the RunEngine to unstage the object. This calls ``obj.unstage()`` and
```
danielballan: Stage and unstage return a list of the Device(s) that were (un)staged, not status objects.

Why not status objects? We debated this one for awhile. Operations performed during staging are supposed to involve twiddling configuration, and should happen fast. Staging should not involve lengthy `set` calls.

Why a list of the objects staged? Staging a Device causes that Device's component Devices (if any) to also be staged. All of these children are added to a list, along with `[self]`, and returned by `Device.stage()`, so that the _plan_ can keep track of what has been staged, like so:

```python
devices_staged = yield Msg('stage', device)
```

Why would the plan want to know that? It needs to avoid accidentally trying to stage something twice, such as a staging a parent and then trying to also stage its child. It's important to avoid that because staging something redundantly raises an error.

jrmlhermitte: thanks @danielballan ! i ll add most of your wording to the PR.

danielballan:
> Personally, I like the doc in configure the best. I would like to work all other docstrings to have this form

:+1: Looks good. We should keep expanding `configure` to include the five points in my last comment as well.

> I think we could break this up into [...]

Sensible.

> I'm wondering if all ophyd methods should be converted to return asyncio Futures rather than status obejcts in the future

I think it would have been technically possible to do this in the beginning, and I argued for it for the first year or so, but Ken made a strong case that we needed a place to stash additional state, such as `success` and `device` (the latter mostly for debugging purposes). From there, we grew a bunch more optional attributes related to supporting progress bar. Futures have no way of expressing, "This task is 53% done."

There was also something about how Status objects are a bridge between the Land of Threads in ophyd and the Land of Coroutines in bluesky, but to be honest I'm still not technically clear why a plain old Future wouldn't have been able to serve that purpose. Maybe @tacaswell has that straight.

awalter-bnl:
I just found this while going through un-merged PR's. Is there a reason this is yet to be merged? I think it is an improvement which we should move forward on.

danielballan:
Is it comprehensive yet?

danielballan:
I haven't checked whether this is comprehensive, but in the spirit of "some docs are better than no docs" let's do as @awalter-bnl suggested above and merge. Re-running the CI just to make sure this builds cleanly....

PR #971, opened 2018-02-05
FIX: Test and fix bug in repeat plan.

danielballan:


tacaswell:
I do not understand why this fixes things :disappointed:

danielballan:
@tacaswell This was new to me also! It turns out that Python cannot `send` into a `tee` object, so `ensure_generator` must be applied _after_ teeing.

PR #974, opened 2018-02-14
REF: 'scan' is just 'inner_product_scan' passed one motor.

danielballan:
This change was agreed on in the lead up to the 1.0.0 release
but de-prioritized because it is just a refactor, not an API break.

* The old `scan` and `rel_scan` code has been retired but retained
  in the codebase for future reference, renamed `_scan_1d` and
  `_rel_scan_1d`.
* New implementations of `scan` and `rel_scan` -- fully backward-compatible
  of course! -- build around the old `inner_product_scan` code.
  They now accept multiple motors:

  ```
  scan(dets, motor, -1, 1, 3)  # scan -1 to 1 in 3 steps
  scan(dets, xmotor, -1, 1, ymotor, -1, 1, 5)  # scan x and y along a diagonal in 5 steps
  ```
* The old ``inner_product_scan`` and ``relative_inner_product_scan`` will still
  be supported forever. They pass through to ``scan`` and ``rel_scan``, rearranging
  the arguments because they expect the number of steps to come _first_. This
  was *not* a popular API choice, so we expect people to be happier just using
  ``scan`` and ``rel_scan``. ``inner_product_scan`` will probably be forgotten.
* No tests have been touched as a part of this change.

The only downside is that the implementation of ``scan`` is now more complex,
which makes it less approachable. To ameliorate this downside, we retain the original
as ``_scan_1d`` for reference and (in a forthcoming PR) we provide better
documentation for writing custom plans.


bluesky/plans.py
```diff
@@ -913,17 +929,31 @@ def inner_rel_grid_scan():
 
 
 def relative_inner_product_scan(detectors, num, *args, per_step=None, md=None):
+    # For rel_scan, num is the _last_ positional arg instead of the first one.
+    # Notice the swapped order here.
+    yield from rel_scan(detectors, *args, num, per_step=per_step, md=md)
+
+
+def rel_scan(detectors, *args, per_step=None, md=None):
```
tacaswell: is it worth doing enough processing to accept `num` as a kwarg or the last positional?

bluesky/plans.py
```diff
@@ -751,15 +765,17 @@ def inner_product_scan(detectors, num, *args, per_step=None, md=None):
     :func:`bluesky.plans.grid_scan`
     :func:`bluesky.plans.scan_nd`
     """
+    num = args[-1]
```
tacaswell: Does this need a `if not len(args) - 1 % 3 == 0: raise` logic?

bluesky/plans.py
```diff
@@ -726,17 +726,31 @@ def inner_scan_nd():
 
 
 def inner_product_scan(detectors, num, *args, per_step=None, md=None):
+    # For scan, num is the _last_ positional arg instead of the first one.
+    # Notice the swapped order here.
+    yield from scan(detectors, *args, num, per_step=None, md=None)
```
tacaswell: should we pass through `plan_name='inner_product_scan'`

bluesky/plans.py
danielballan: That check happens in `plan_patters.py`. (I know because I tripped it.) Doesn't seem necessary to do it again here.

bluesky/plans.py
danielballan: Hmm. Probably. Will add that.

bluesky/plans.py
danielballan: What do you think? I considered it but decided against it.

tacaswell: I am +0.5, we should consider these headers part of the API, but `scan` is nicer.

danielballan: OK, done. 👍

danielballan:
Travis-CI's status report indicates that they are having issues today. But the test phase of the builds has completed: tests passed.

teddyrendahl:
Am I understanding the motivation correctly that `inner_product_scan` allows `n` motors and that adding that functionality to the base `scan` function just makes it `inner_product_scan`? 

I'm certainly not going to block this because it doesn't break API but I'm not entirely in love with the concept. Importing `scan` when you want a `1d` scan and `inner_product_scan` when you want `2d` doesn't seem too huge a burden. In fact if you wanted one import you could just import `inner_product_scan` and pass it one motor. I guess my main concern would be that we overload the `scan` syntax and make it far more confusing than it should be. If I'm trying to put myself in a users head and you show me, this my next question would be "Can I pass two different step counts and make an outer product scan?" Just seems like a slippery slope.

In the end, this doesn't break anything, and I haven't dug through the scanning library quite enough to understand how much this helps the maintenance of the library. Just giving my thoughts.

danielballan:
> Am I understanding the motivation correctly that `inner_product_scan` allows `n` motors and that adding that functionality to the base scan function just makes it `inner_product_scan`?

Yes.

> [...] Just seems like a slippery slope.

I'm sympathetic to this argument. I definitely prefer "many small functions with well-defined functionality" to "cram a bunch of kwargs and magic into a a couple sprawling functions." The latter is often tempting to new users/developers but harmful in the long run.

While I was initially against this change -- not sure who first pitched it -- I have come to around to thinking that this case is clear-cut enough that we can give our users one less name to remember without falling too far down this slope. Maybe it comes down to whether you think scan is a special case of inner-product or a special case of outer-product. I'll grant that the argument-munging going on here has a bad code smell.

In summary, I'm not strongly for or against this. Other thoughts?

bluesky/plans.py
```diff
@@ -938,19 +985,19 @@ def relative_inner_product_scan(detectors, num, *args, per_step=None, md=None):
...
-    _md = {'plan_name': 'relative_inner_product_scan',
+    _md = {'plan_name': 'rel_scan',
            }
     md = md or {}
     _md.update(md)
     motors = [motor for motor, start, stop in partition(3, args)]
 
     @bpp.reset_positions_decorator(motors)
     @bpp.relative_set_decorator(motors)
-    def inner_relative_inner_product_scan():
```
mrakitin: Good job on getting rid of this confusing name!

bluesky/plans.py
```diff
@@ -938,19 +985,19 @@ def relative_inner_product_scan(detectors, num, *args, per_step=None, md=None):
     :func:`bluesky.plans.inner_product_scan`
     :func:`bluesky.plans.scan_nd`
     """
-    _md = {'plan_name': 'relative_inner_product_scan',
+    _md = {'plan_name': 'rel_scan',
            }
```
mrakitin: Why not to place `}` to the previous line? Just a cosmetic suggestion...

jrmlhermitte:
I think this looks good. I am :+1: on num being the last arg. This is a frequent complaint I've heard across beamlines for those who were used to specs dscan.

teddyrendahl:
Just thought of something that may be a slightly sneaky result of this PR.

The `one_1d_step` and `one_nd_step` have slightly different signatures. I didn't see anywhere where you check that there is a single dimension and use the `one_1d_step` signature instead? This will break existing code that assumes the `one_1d_step` signature and calls `scan`.

tacaswell:
:1st_place_medal: for @teddyrendahl for the paranoid sharp eye!

danielballan:
Yep, that would definitely break a thing or two on our floor that I know of! I think there is a graceful fix, will post one tomorrow.

danielballan:
Addressed in https://github.com/NSLS-II/bluesky/pull/988

PR #981, opened 2018-02-20
TEMP: Delete 0MQ tests to check if they are hanging Travis.

danielballan:


danielballan:
Indeed, the 0MQ tests were the issue. See #982.

PR #985, opened 2018-02-20
Print reg

jrmlhermitte:
After discussion with @danielballan this is a good idea I think!

```
In [7]: RE.print_command_registry(verbose=False)
List of available commands
create : Trigger the run engine to start bundling future obj.read() calls for
save : Save the event that is currently being bundled
drop : Drop the event that is currently being bundled
read : 
monitor : 
unmonitor : 
null : 
stop : 
set : 
trigger : 
sleep : Sleep the event loop
wait : Block progress until every object that was triggered or set
checkpoint : Instruct the RunEngine to create a checkpoint so that we can rewind
clear_checkpoint : Clear a set checkpoint
rewindable : Set rewindable state of RunEngine
pause : Request the run engine to pause
collect : 
kickoff : Start a flyscan object
complete : 
configure : Configure an object
stage : Instruct the RunEngine to stage the object
unstage : Instruct the RunEngine to unstage the object
subscribe : 
unsubscribe : 
open_run : Instruct the RunEngine to start a new "run"
close_run : Instruct the RunEngine to write the RunStop document
wait_for : Instruct the RunEngine to wait until msg.obj has completed. Better
input : 

```

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1412,11 +1412,6 @@ class RunEngine:
         obj = msg.obj
         # actually _read_ the object
         ret = obj.read(*msg.args, **msg.kwargs)
-        # Ask the object for any datum documents is has cached.
-        if hasattr(obj, 'collect_asset_docs'):
-            datum_docs = list(obj.collect_asset_docs(*msg.args, **msg.kwargs))
-        else:
-            datum_docs = []
 
         if self._bundling:
             # if the object is not in the _describe_cache, cache it
@@ -1441,9 +1436,15 @@ class RunEngine:
             # add this object to the cache of things we have read
             self._objs_read.append(obj)
 
-            # stash the results
+            # Stash the results, which will be emitted the next time _save is
+            # called --- or never emitted if _drop is called instead.
             self._read_cache.append(ret)
-            self._asset_docs_cache.extend(datum_docs)
+            # Ask the object for any resource or datum documents is has cached
+            # and cache them as well. Likewise, these will be emitted if and
+            # when _save is called.
+            if hasattr(obj, 'collect_asset_docs'):
+                self._asset_docs_cache.extend(
+                    obj.collect_asset_docs(*msg.args, **msg.kwargs))
 
         return ret
 


danielballan:
It might also be good to have a public API where you can programmatic check the command registry. For example:

```
'set' in RE.list_command_registry()
```

Should we make `RunEngine._command_registry` public? Should we add yet another method (`list_command_registry`)? I think I prefer the latter because I consider the _keys_ of `RE._command_registry` public but the _values_ -- the RunEngine coroutine methods -- private.

jrmlhermitte:
I like that idea. Perhaps it should still be a method of an instance of `RE` and not `staticmethod`? The main reason I mention that is I could see perhaps one day a feature request for allowing the registering of custom commands? Or is that too dangerous?
I've added:
```py
In [3]: RE.commands()
Out[3]: 
['create',
 'save',
 'drop',
....
```

(it could also be a property maybe)

bluesky/run_engine.py
```diff
@@ -305,6 +305,32 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
 
         self.loop.call_soon(self._check_for_signals)
 
+    def commands(self):
```
jrmlhermitte: should it be a `@property`?

danielballan:
> The main reason I mention that is I could see perhaps one day a feature request for allowing the registering of custom commands?

We are on the same page: https://nsls-ii.github.io/bluesky/msg.html#registering-custom-commands  :-D

There is only one use of this on the floor. I intend to add it to the msg documentation as an example.

bluesky/run_engine.py
danielballan: I think that make sense. Can you also add these to the list of Attributes in the RunEngine's top-level docstring?

jrmlhermitte:
errors are from this (a bad toctree ref and also fstrings not compat with py3.5). ill change a little later....

jrmlhermitte:
the bugs were sphinx related. I updated the docs.
(Found out that I can't put `commands` in the `Attributes` part of docstring because it's a `Property`)

jrmlhermitte:
I added tests for these functions. They're pretty much the code copied and pasted, but it will make sure we're aware of any API changes.

bluesky/run_engine.py
```diff
@@ -305,6 +311,61 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
...
+        commands = "List of available commands\n"
+
+        for command, func in self._command_registry.items():
+            docstring = func.__doc__
+            if verbose is False:
+                docstring = docstring.split("\n")[0]
+            commands = commands + "{} : {}\n".format(command, docstring)
+        print(commands)
+
+        return commands
```
tacaswell: If you do `RE.print_command_registry()` won't this result in it printing twice?  Once from the print and once from the returned value?

jrmlhermitte: yes you're right thanks! probably a return value would not be expected for this one

tacaswell:
@jrmlhermitte This needs a rebase.

I am :+1: on merging this.

bluesky/run_engine.py
```diff
@@ -305,6 +311,60 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
...
+        >>> from bluesky import RunEngine
+        >>> RE = RunEngine()
+        >>> # Print a very verbose list of currently registered commands
+        >>> RE.print_command_registry(verbose=True)
+        '''
+        commands = "List of available commands\n"
+
+        for command, func in self._command_registry.items():
+            docstring = func.__doc__
+            if verbose is False:
```
danielballan: `if not verbose` is a bit more Pythonic.

bluesky/run_engine.py
```diff
@@ -181,6 +181,12 @@ class RunEngine:
         It is set to ``bluesky.run_engine.PAUSE_MSG`` by default and
         can be modified based on needs.
 
+    Properties
```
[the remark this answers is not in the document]
jrmlhermitte: thanks! made the change

bluesky/run_engine.py
jrmlhermitte: thanks! done

danielballan:
I tried to rebase this over some unrelated fixes to make sure everything was still good, and ended up pushing the wrong branch back up (oops). Since said branch was already merged, GitHub marked this PR as merged.

Fortunately, no commits were lost. I pushed the original content of this PR to https://github.com/NSLS-II/bluesky/pull/1029. Sorry for the runaround.

PR #987, opened 2018-02-20
API: Deprecate caching_repeater; limit 'repeat' plan to callables.

danielballan:
The notion of repeating a literal sequence of Msgs is problematic,
because some Msgs' contents depends on replies from the RunEngine.
For example, `Msg('unsubscribe', ...)` contains a ``token`` that
comes from the RunEngine's reply to `Msg('subscribe')`. Therefore,
it does not really make sense to support the usage, "Repeat this
literal sequence of Msgs." Repeater utilities should require
callables, so that adaptiveness can be fully supported.

attn @ZLLentz

### DIFF
--- a/doc/source/plans.rst
+++ b/doc/source/plans.rst
@@ -649,6 +649,4 @@ These are useful utilities for defining custom plans and plan preprocessors.
    :nosignatures:
 
     broadcast_msg
-    repeater
-    caching_repeater
     repeat
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -6,6 +6,7 @@ import operator
 from functools import reduce
 from collections import Iterable
 import time
+import warnings
 
 try:
     # cytools is a drop-in replacement for toolz, implemented in Cython
@@ -837,6 +838,8 @@ def caching_repeater(n, plan):
     --------
     :func:`bluesky.plans.repeater`
     """
+    warnings.warn("The caching_repeater will be removed in a future version "
+                  "of bluesky.")
     it = range
     if n is None:
         n = 0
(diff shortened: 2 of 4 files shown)


danielballan:
Note that since we have not released bluesky since `repeat` was added, we can make this change without breaking the API per se.

doc/source/plans.rst
```diff
@@ -649,6 +649,4 @@ These are useful utilities for defining custom plans and plan preprocessors.
    :nosignatures:
 
     broadcast_msg
-    repeater
```
danielballan: I only recently listed these here in the first place (since the most recent tag). Now, since `caching_repeater` is being deprecated and `repeater` is inferior to `repeat`, I'm rolling back my decision to list them here.

ZLLentz:
I'm assuming this PR is because someone has been trying to repeat a complicated series of messages that really doesn't work for the reasons you outlined above. Of course, we are removing the capability to repeat a very simple string of messages, which works, but I think this is a good change. In general, I think we should show some restraint with Python's ability with multiple input types because it can make the code confusing and unmanageable. With a library like `bluesky` that acts as a useful tool for other projects, we should especially avoid laying traps for the user.

I'd rather take this out before the release and consider adding it in later if there is a need. There should be a way to inspect list plans and warn the user about these issues, but perhaps that's overkill for a feature-rich for-loop.

danielballan:
Great. We're on the same page.

Indeed, this came up because someone was trying to repeat a plan that used `subs_decorator`. That's a pretty common use case, and it alarmed me to realize that this failed on that case.

I would definitely support investigating ways to re-instate this feature in that, to borrow your phrase, avoids laying traps for the user.

PR #988, opened 2018-02-20
FIX/TST: Accept one_1d_step or one_nd_step-like signatures in per_step.

danielballan:
This fixes a regression in the backward compatibility of `scan`. If
a user passes in a ``per_step`` parameter with has a 1D signature
(a signature like ``one_1d_step``) this PR uses a local "adapter"
function to translate this into ``one_nd_step``.

It also adds validation and user-friendly error messages in the event
that the ``per_step`` parameter provided by the user does not have one
of the expected signatures.

There is still a potential for breaking user code: we now insist that
the parameter *names* in ``per_step`` functions match the ones we expect
-- either ``one_1d_step`` and ``one_nd_step``. Before, users could technically
get away with a ``per_step`` function that had different parameter names
because we called the ``per_step`` function with positional arguments.
However, I think it is fair to assert that we always had the right to pass in
the arguments by keyword, i.e. to require that when we specify an "expected
signature" that this includes the names. So I would not consider this an
API break.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1,4 +1,5 @@
 import sys
+import inspect
 from itertools import chain
 from functools import partial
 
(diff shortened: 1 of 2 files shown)


teddyrendahl:
Looks good. Thanks for taking the time to make succinct error messages. In my opinion, one of the hardest parts of learning to use`bluesky` is that tracebacks can be very obscure.

danielballan:
Agreed. We've heard that feedback from a couple directions, and I think it's probably the most important general area of development in bluesky and ophyd at this point.

PR #990, opened 2018-02-21
First test of a streaming trigger and read

jrmlhermitte:
(Not meant to be merged, meant for discussion)
We have a case, where a beamline wants access to their datum (filename of data) before data has been written. This would help save time as their data can come from a long (30min) acquisition of a time series.

A solution to this, as @danielballan mentioned is to have read() come before wait().
Could it be a good idea to have this within the `trigger_and_read` plan stub?
Another option is to add a read before and after the wait explicitly (no checking for a `streaming_data` bool etc).


jrmlhermitte:
oops this was wrong. i had the wrong branch... closing this and opening the correct one...

PR #993, opened 2018-02-23
Added read_streaming method

jrmlhermitte:
## Description
A discussion on extending `read` to work more efficiently with 'streaming' devices. A 'streaming' device is defined to be a device that can return information on where to retrieve its data before it has completed filling it. This then would allow a user to read their data before their device has finished filling their data.

A simple example is getting the filename of an image time series that a detector has not completed in filling (for example: during long acquisitions).

## Motivation
We have a case, where a beamline wants access to their datum (filename of data) before data has been written. This would help save time as their data can come from a long (30min) acquisition of a time series.

This is a result from discussion with @danielballan .
The idea is that two event streams should be emitted: a 'stream_primary' and 'primary' event stream.
In the event that the stream_primary contains no events, nothing (including a descriptor) should be emitted.

## How to test this
```py3
from databroker import Broker
from bluesky import RunEngine
from ophyd.sim import img, det1, det2, NumpySeqHandler


db = Broker.from_config(temp_config())
img.reg = db.reg

RE = RunEngine({})

RE.subscribe(db.insert)
# convienently print docs in some way
def print_doc(name, doc):
    print("Got document {}\n".format(name))
    print("Contents : {}\n".format(doc))
RE.subscribe(print_doc)


uid, = RE(count([img, det1, det2]))
```

NOTE : this description has been strongly edited. Here are some old comments which I believe are not so relevant anymore:

The idea for this will be to have an object such as
```py
class MyOphydObj(Device):
    ...
    def read(self):
        return ...# a filename
    def read_streaming(self):
        return ... # filename
obj = MyOphydObj()
```
The `read` message will eventually call `obj.read()` and the `read_streaming` message will eventually call `read_streaming`, generating events for both in separate streams.


bluesky/plan_stubs.py
```diff
@@ -743,6 +748,12 @@ def inner_trigger_and_read():
             if hasattr(obj, 'trigger'):
                 no_wait = False
                 yield from trigger(obj, group=grp)
+        yield from create(streaming_name)
+        for obj in devices:
+            reading = (yield from read_streaming(obj))
+            if reading is not None:
+                ret.update(reading)
+        yield from save()
```
danielballan: In case anyone else wonders...

I had to remind myself what happens if we 'create' and 'save' with no 'read' messages between. The RunEngine just treats that as a no-op. (So this code block behaves fine if no devices implement `read_streaming`.)

bluesky/plan_stubs.py
```diff
@@ -743,6 +748,12 @@ def inner_trigger_and_read():
             if hasattr(obj, 'trigger'):
                 no_wait = False
                 yield from trigger(obj, group=grp)
+        yield from create(streaming_name)
+        for obj in devices:
+            reading = (yield from read_streaming(obj))
```
danielballan: I guess a `read_streaming` plan stub needs to be defined also, before tests are added.

bluesky/plan_stubs.py
```diff
@@ -743,6 +748,12 @@ def inner_trigger_and_read():
...
+            if reading is not None:
+                ret.update(reading)
```
danielballan: I guess there is no point in capturing these results because we capture redundant results below. If these reading are different than the later readings something is broken. Perhaps, if anything, we should collect these results separately and asset that the subsequent readings are equal.

bluesky/plan_stubs.py
jrmlhermitte: yep, thanks for writing this, found it by reading `_save` before writing this :)

bluesky/plan_stubs.py
jrmlhermitte: oh yes, you're right... forgot to yield a `Msg`...thanks!

bluesky/plan_stubs.py
jrmlhermitte: yes you're right, these are returned results. i copied and pasted without looking in depth... :-|

bluesky/plan_stubs.py
```diff
@@ -748,15 +768,19 @@ def inner_trigger_and_read():
             if hasattr(obj, 'trigger'):
                 no_wait = False
                 yield from trigger(obj, group=grp)
+        # read from the streaming devices
         yield from create(streaming_name)
         for obj in devices:
             reading = (yield from read_streaming(obj))
```
danielballan: No need to assign `yield from read_streaming(obj)` to a variable now that it isn't used.

danielballan:
Reflecting on this, it's been bothering me a little that we seem to accumulating a large number of methods named `read_*`. Two was OK; three is starting to make me worry that we may not be done yet and we could end up with an unwieldy number of these. While I was mulling this over, @tacaswell hit me up on Slack with this same concern, independently.

When I think about the difference between `read_streaming()` and `read()`, it doesn't seem like the same _kind_ of distinction as the one between `read_configuration()` and `read()`. When we `read_streaming()` we aren't asking for a _different_ set of fields; we're asking for a (possibly empty) _subset_ of fields from `read()` that are safe to read eagerly. The plan is asking the Device for a new kind of information to help it generalize over the Devices that can stream, Devices that cannot stream, and mixtures thereof. Because this is different in kind from the distinction between `read()` and `read_configuration()`, that suggests we should be employing a different language feature.

Instead of:

```
if hasattr(obj, 'read_streaming'):
    ret = obj.read_streaming()
```

what about:

```
try:
    ret = obj.read(streaming=True)
except TypeError:
    # This method did not expect the keyword argument 'streaming'.
    pass
```

tacaswell:
It might also be worth considering adding the expectation that if you pre-maturely call `read` on a device with `streaming=False` that is not ready to be read (that is you have not waited for the trigger to report that it is done) it will raise an exception.   Moves a bit of state down into the device (if it is currently acquiring) and surfaces parts of that back to the plan (via an exception) if the plan does something wrong.

jrmlhermitte:
how about `streaming_attrs`? I think this makes more sense.

danielballan:
I agree we should have `streaming_attrs`. That doesn't solve the whole problem, because `.*_attrs` only specifies one level in the hierarchy. For example, `eiger4m.streaming_attrs` will be `['hdf5']` but we also need to know the contents of `eiger4m.streaming_attrs.hdf5` and so on. The `read_*` functions walk that tree. We could conceivable ask the _plans_ to walk that tree, but since we already have a pattern for that it seems natural to reuse it. (Currently, the `.*_attrs` are an ophyd implementation detail, not actually touched by bluesky at all.) Does that make sense?

jrmlhermitte:
oops i deleted my comment too fast. i agree for the same reason. i need to think about it when i can sit down at a computer :-D
I think it's worth thinking about this, and it may touch flyers as well. For ex: `read(type=flyer)` `read(type='streamer')` etc. will write later

My old comment was something like:
> How about `streaming_attrs`?

klauer:
I don't like this section of code as catching `TypeError` can mask other issues:
```python
try:
    ret = obj.read(streaming=True)
except TypeError:
    # This method did not expect the keyword argument 'streaming'.
    pass
```

To keep the same functionality, I would suggest using `inspect.signature` to see if the kwarg exists instead.

danielballan:
What we are trying at CHX:

```py
class EigerBase(...):
    def read(streaming=False):
        if streaming:
            key = self._image_name  # this comes from the SingleTrigger mixin
            return {key: super().read()[key]}
        else:
            return super().read()
```

jrmlhermitte:
@klauer, @danielballan helped me address your comment. very good idea, thanks for the keen eye!

jrmlhermitte:
Some changes were necessary (this is a work in progress):
- describe now needs a `streaming=True` kwarg so it knows what to describe
- the `_describe_cache` needs to be cleared per `create` not per `open_run`/`close_run` since now describe can return different results (depending on the `stream=True` flag)

bluesky/run_engine.py
```diff
@@ -1377,18 +1379,21 @@ def _read(self, msg):
         if self._bundling:
             # if the object is not in the _describe_cache, cache it
             if obj not in self._describe_cache:
+                self.describe_cache[obj] = dict()
```
danielballan: There is a nice trick we use in a couple places would make these two lines unneeded (Lines 1381 and 1382, as of this writing). In `RunEngine.__init__`, make `self._describe_cache` a `defaultdict(dict)` instead of a plain `dict`. Then you can check `if 'regular' not in self._describe_cache[obj]:` directly, knowing that `self._describe_cache[obj]` will be magically initialized to `dict()` if it needs to be.

bluesky/run_engine.py
```diff
@@ -1377,18 +1379,21 @@ def _read(self, msg):
         if self._bundling:
```
danielballan: Just to underline the intent here, if a plan tries to 'read' without having first asked to 'create', bundling will be `False`. We still go ahead and read the device and send the result back to the plan. The reading can be used by the plan to do adaptive things; it won't be captured by the RunEngine or emitted in a document.

Below this `if` statement, now having established that we _are_ bundling, we should separately check that the `self._read_type` is either `None` or `'regular'`. (I'm assuming that a given Event can only include readings that we invoked using a consistent read type.)

bluesky/run_engine.py
```diff
@@ -1504,6 +1576,12 @@ def _save(self, msg):
 
             Msg('save')
         """
+        read_type = self._read_type
+        # no read type means no data saved
+        if read_type is None:
+            return
+        self._read_type = None
+        # TODO: maybe bundling should be the read_type instead of bool?
```
danielballan: I think that works, and anything that reducing the number of stateful flags is a good thing! :- )

danielballan:
Another thought we don't want to lose track of -- do we want to name the stream `'partial_primary`? Do we want to add a separate key like `partial=True` or `read_type='partial'`?

jrmlhermitte:
@danielballan I have thought about this more and feel it might be too complex. I worry about giving too many hidden features to ophyd objects. We discussed this before, but let me bring the discussion again. Do we want:
1. The motivation of this PR is to eventually get to:
```py3
dev._streaming_attrs = [...]
RE(count([dev]))
```
2. An alternative way of doing this is not to touch the ophyd objects and modify the usage of count:
```py3
RE(count([dev.roi1, dev.roi2, dev.img], [dev.img]))
```
PR #1006 experiments with that. It was quite easy for me to get a working prototype for that one.

danielballan:
Thanks for implementing (2) so that we can make a concrete comparison. To lift some comments from a Slack conversation a couple days ago:

> The overall question is, "Is adding more complexity to the ophyd/bluesky interface worse than forcing users to write special plans like `streaming_count([eiger4m.hdf5], [eiger4m.stats1, eiger4m.stats5])`?" I would accept _some_ new complexity in order to avoid the having to give up on `count` for AD, but not an unlimited amount of complexity.

Building on that, I notice that `dev.streaming_attrs = [...]` is something that Device developers do, one time, where as the modified `count` asks the users to change their behavior every time. I don't expect the list of `streaming_attrs` to be tweaked interactively by users; a component is either safe to read before it is done triggering, or it is not. So, I would expect the users to strongly prefer (1).

I'm mostly concerned about the general trend of making the ophyd/bluesky interface more complex, increasing the number of methods and behaviors that one needs to learn in order to fully understand it. I'm focused on considering:

* Is our proposed interface for Approach (1) as simple as it can be?
* Does it definitely fully capture of the complexity we need it to, or are we going to run into places that we need to fall back to Approach (2) anyway?

bluesky/run_engine.py
jrmlhermitte: interesting. that's a good feature i wasn't aware about, getting a reading without dispatching documents. thanks for the clarification!

jrmlhermitte:
Thanks for the feedback. I have simplified the logic so that a new message is not necessary (rather, `streaming` should be a kwarg).

I tested with and without a streaming device. See the code in the edited first post above here to test.
To test  a streaming device, make this modification in ophyd.sim (or create your own with this format):

https://github.com/NSLS-II/ophyd/compare/master...jrmlhermitte:ophyd-streaming?expand=1

bluesky/tests/test_plans.py
```diff
@@ -171,13 +171,22 @@ def test_reset_wrapper(hw, RE):
     RE(bp.relative_inner_product_scan([], 1,
                                       p.pseudo1, 0, 1,
                                       p.pseudo2, 0, 1))
-    expecte_objs = [p, None, None,
```
danielballan: Heh heh heh. Dollars to donuts this (and the amazingly consistent misspellings that follow) was brought to us by emacs tab-complete.

jrmlhermitte:
Tests pass, had to make a few modifications (basically number of messages generated are larger when calling read).

Still some points to consider:
- We may want to add more tests. I think we'll have a better handle on that when we try at CHX (ex: if something no longer works at CHX, then we probably should have tested for that)
- we may want to add a new tag to descriptors. Right now, we distinguish *streaming* events from *non-streaming* by a user customizable prefix added to the stream name. However, we may likely want to filter out streaming documents (i.e. for `db.insert` and thus may need a more robust way of distinguishing them. @danielballan 's comment about this as well (I agree with you):
> At some point we should circle back on whether we want a separate flag in the descriptor, and a new parameter in the `create()` plan to match.
>I’m like 70/30 leaning toward a separate flag at this point.
- we may want the same logic for flyers (`collect(streaming=True)`), but I think we should defer it for now. Im actually surprised that this use case didn't come before time series!

jrmlhermitte:
job keeps crashing on
`bluesky/tests/test_examples.py::test_sleepy /home/travis/.travis/job_stages: line 57:  3526 Aborted                 (core dumped) coverage run run_tests.py`

any idea why?

jrmlhermitte:
backport for v1.2.0 (pre-asset refactor) branched here: https://github.com/jrmlhermitte/bluesky/tree/read-streaming-backport
Will test this on CHX sometime soon

jrmlhermitte:
yes, to summarize (more later):
- we will add a `read_mode` property to `ophyd` objects
- this `read_mode` property will consist of a `dict` of type (i.e. 'streaming') and state (ex: 'never', 'always', 'ondemand')
- this will require:
-- modify `trigger_and_read` to grab devices which can be streamed through this
-- modify the `describe_cache` to index on the hash of the `stream_name` , `Msg.args` and `Msg.kwargs`

tacaswell:
xref https://github.com/NSLS-II/ophyd/pull/528

PR #676, opened 2017-03-30
Square spiral

danielballan:
Supercedes https://github.com/NSLS-II/bluesky/pull/653 (from @awalter-bnl's master branch) and adds cleanup. Needs:

- [x] Tests
- [x] Entries added the plans documentation page

PR #1001, opened 2018-03-05
Square spiral 2

### LINKED ISSUE #646: add 'spiral_square_pattern' to common plans file
I have written a spiral_square plan (like spiral but follows a square path not a circular path) to 40-ESM_plans.py. It may be useful to move this over to the standard plans. It requires both spiral_square and spiral_square_pattern.

awalter-bnl:
square spiral update, supercedes #676 

## Description
This is an update to square_spiral_pattern (and the documentation section of square_spiral and relative_square_spiral) to allow for any arbitrary n*m matrix to be used. previously only even*even and odd*odd matrices where possible. It involved generating the cycler path in a more general way that allows for any arbitrary n*m matrix to be used.

## Motivation and Context
This resolves the issue with pull request  #676 raised by  @tacaswell whereby the limitation to even*even and odd*odd matrices may not have been needed

Closes #646


bluesky/plan_patterns.py
```diff
@@ -67,6 +67,99 @@ def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
...
+    y_num : float (must be even if x_num is even and must be odd if x_num is odd, if not it is increased by 1 to ensure this)
+        number of y axis points
+
+    Returns
+    -------
+    cyc : cycler
+    '''
+    x_points, y_points = [], []
+
+    if x_num%2==0:
```
danielballan: There should be spaces around operators like `==` and `=` below.

bluesky/plan_patterns.py
```diff
@@ -67,6 +67,99 @@ def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
...
+
+    delta_x = x_range/(x_num-1)
+    delta_y = y_range/(y_num-1)
+    num_ring = max(x_num,y_num)
+    x_max=x_center + delta_x * (x_num-1)/2
+    x_min=x_center - delta_x * (x_num-1)/2
+    y_max=y_center + delta_y * (y_num-1)/2
+    y_min=y_center - delta_y * (y_num-1)/2
+
+    for n,i_ring in enumerate(range(num_st, num_ring+1,2)):
```
danielballan: Need a spaces after both commas in this line.

bluesky/plan_patterns.py
```diff
@@ -67,6 +67,99 @@ def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
...
+        x_ring_max=x_center + delta_x * (n+offset)
+        y_ring_max=y_center + delta_y * (n+offset)
+        x_ring_min=x_center - delta_x * (n+offset)
+        y_ring_min=y_center - delta_y * (n+offset)
+
+        for n in range(1, i_ring):
+            x = x_ring_min+delta_x*(n-1)
+            y = y_ring_min
+
+            if ( (x <= x_max) and (x>= x_min) and (y<=y_max) and (y>=y_min)     ):
```
danielballan: The outer parens here aren't technically necessary. If you want them for readability's sake, I'd say that's a judgment call, but PEP8 would have you remove the extra  spaces inside the outer parens.

danielballan:
I also wanted to resurface the consensus we reached around how to handle asymmetric spirals https://github.com/NSLS-II/bluesky/pull/675#issuecomment-337685135. I'm fine with tackling that in a separate PR, just didn't want to lose track of it.

awalter-bnl:
I forgot about this one, while I am not completely against the use of ‘asym_spiral’ after witnessing the use of bluesky on the floor I am starting to think that have a very small number of plans with options may be easier for the beamline staff and users to work with than a whole series of slightly different plans.

My thoughts for this specific case would be have one plan called ‘spiral’ and have an optional argument ‘pattern’ that takes the values ’standard’, ‘square’ or ‘fermat’. In this idea by including an optional ‘delta_y’ parameter the scan would become asymmetric but if it is not included it is not.

I have found that people have trouble remembering the names for all the built-in plans (and in some cases are re-writing similar ones), so having a smaller number of ‘plans'  would be easier for them to remember. to get details about options available in the plan they can then just use  ‘plan_name?’ to get the help documentation. As another suggestion I would also consider adding ‘rel’ as an optional boolean argument to all the scans instead of having different plans for relative motion scans.

Anyway this is just a suggestion, I am sure you guys have discussed this a lot and I am just providing feedback from my experience’s on the floor.

cheers
Andrew




On Mar 5, 2018, at 9:44 AM, Dan Allan <notifications@github.com<mailto:notifications@github.com>> wrote:


I also wanted to resurface the consensus we reached around how to handle asymmetric spirals #675 (comment)<https://github.com/NSLS-II/bluesky/pull/675#issuecomment-337685135>. I'm fine with tackling that in a separate PR, just didn't want to lose track of it.

bluesky/plan_patterns.py
mrakitin: I'd recommend using [flake8](https://pypi.python.org/pypi/flake8), which hits all inconsistencies with PEP8. We will meet with @awalter-bnl later today to discuss/fix it.

danielballan:
> Anyway this is just a suggestion, I am sure you guys have discussed this a lot and I am just providing feedback from my experience’s on the floor.

This is useful feedback.

From the developer's perspective, trying to avoid painting ourselves into a corner, the "safe" course it to make many small, separate functions. We should push back against that tendency where can do so with confidence. We recently merged `inner_product_scan` and `scan`. I'm glad we waited to do this until we were sure we understood the consequences. The thing we are trying to avoid at all costs is growing _coupled keyword arguments_ -- cases were changing one parameter alters the meaning of another parameter or causes it to be silently ignored. While these kinds of interactions might make perfect sense to any given _author_ of such a function, they are well known to cause confusion in general for users.

In the specific case of spiral patterns, would all these functions take the same args other than pattern?

awalter-bnl:
I understand your point, reduction should always be done so as to to not cause more confusion. But many separate plans also adds to confusion so these things need to be done with thought ;-).

I am a strong supporter of the combination of scan and inner_product. In my mind the ideal scenario is to have as a basic set the following list of plans.

Count
Scan ( including inner product)
Grid
Spiral ( with the different patterns)

Scan and grid would then have the optional Boolean ‘rel’ argument that determines if it is a relative scan or not. These 4 plans would then do 90% of what is required and would be easy to remember, with the plan? Command used to give details easily.

In the spiral case I would have all other arguments the same, however I haven’t looked at them to see if that is possible.






On 5 Mar 2018, at 12:38 pm, Dan Allan <notifications@github.com<mailto:notifications@github.com>> wrote:


Anyway this is just a suggestion, I am sure you guys have discussed this a lot and I am just providing feedback from my experience’s on the floor.

This is useful feedback.

From the developer's perspective, trying to avoid painting ourselves into a corner, the "safe" course it to make many small, separate functions. We should push back against that tendency where can do so with confidence. We recently merged inner_product_scan and scan. I'm glad we waited to do this until we were sure we understood the consequences. The thing we are trying to avoid at all costs is growing coupled keyword arguments -- cases were changing one parameter alters the meaning of another parameter or causes it to be silently ignored. While these kinds of interactions might make perfect sense to any given author of such a function, they are well known to cause confusion in general for users.

In the specific case of spiral patterns, would all these functions take the same args other than pattern?

mrakitin:
@danielballan, @tacaswell, the PR is ready for your review. Please let @awalter-bnl and me know if this can be improved somehow. Thanks!

mrakitin:
This works perfect - the tests pass, and the docs are generated correctly, but we weren't able to get a plot with BEC. Here is a sample code we tried:
```py
from bluesky.callcacks.best_effort import BestEffortCallback
from bluesky.plans import spiral_square
from bluesky.simulators import summarize_plan
from ophyd.sim import hw
from matplotlib import pyplot as plt


RE = RunEngine({})
bec = BestEffortCallback()
RE.subscribe(bec)

hw = hw()
det = hw.det
motor1 = hw.motor1
motor2 = hw.motor2

plt.ion()

# summarize_plan(spiral_square([det], motor1, motor2, 0, 0, 1, 1, 11, 11))
RE(spiral_square([det], motor1, motor2, 0, 0, 1, 1, 11, 11))
```
How can it be solved?

awalter-bnl:
So just to clarify a points here, we are trying to plot this in the same style as grid_scan, not as the other spirals. This makes more sense because after the scan is complete there is no real difference in the data between a square spiral and a grid scan (the main reason it is useful as it combines the benefits of starting in the centre of the range and a result that is an n*m matrix). This may be the issue.



On Mar 6, 2018, at 4:45 PM, Maksim Rakitin <notifications@github.com<mailto:notifications@github.com>> wrote:


This works perfect - the tests pass, and the docs are generated correctly, but we weren't able to get a plot with BEC. Here is a sample code we tried:

from bluesky.callcacks.best_effort import BestEffortCallback
from bluesky.plans import spiral_square
from bluesky.simulators import summarize_plan
from ophyd.sim import hw
from matplotlib import pyplot as plt


RE = RunEngine({})
bec = BestEffortCallback()
RE.subscribe(bec)

hw = hw()
det = hw.det
motor1 = hw.motor1
motor2 = hw.motor2

plt.ion()

# summarize_plan(spiral_square([det], motor1, motor2, 0, 0, 1, 1, 11, 11))
RE(spiral_square([det], motor1, motor2, 0, 0, 1, 1, 11, 11))

How can it be solved?

danielballan:
> we weren't able to get a plot with BEC. 

Your code example doesn't include `install_kicker()` to make the plots live-update, so any plots would be drawn after the scan completes. Is that the issue?

> we are trying to plot this in the same style as grid_scan

For the documentation, you mean? Do you want to use https://nsls-ii.github.io/bluesky/generated/bluesky.simulators.plot_raster_path.html#bluesky.simulators.plot_raster_path?

Apologies if I'm missing the point.

mrakitin:
I've tried that:
```py
%matplotlib qt
install_qt_kicker()
```
and then ran the code above, but still getting an empty plot. Did we misconfigure the hints?

awalter-bnl:
we are trying to plot this in the same style as grid_scan

For the documentation, you mean? Do you want to use https://nsls-ii.github.io/bluesky/generated/bluesky.simulators.plot_raster_path.html#bluesky.simulators.plot_raster_path?

Apologies if I'm missing the point.

Sorry I think you are missing the point so let me try a better description. The aim is to have this plot using LiveGrid, as used for grid_scan plotting, instead of LiveScatter, which is used for the other 2 spiral plans.

danielballan:
When I run @mrakitin's example, I get a blank figure, so it sounds like I can reproduce what @mrakitin is seeing. I also notice this warning in the console:

```
UserWarning: Need both 'shape' and 'extents' in plan metadata to create LiveGrid.
```

The bec gets part way through trying to create a figure and then discovers it doesn't have all the information it needs and bails. It would have been easier for you to catch this problem if we had made this an _error_ instead of a warning -- a nice case study in why warnings are generally useless -- but in this case we want to allow a scan to proceed even if bec doesn't have all the metadata it wants to make the best visualization. Our reasoning is, "The data is still being saved, and you can always visualize it later."

So, to finish this add something like:

```python

   _md = {'shape':(y_num, x_num),
           'extents': ((y_center - y_range, y_center + y_range),
                       (...)),
            ...

```

I'm not 100% sure I have the order of y and x right there, FYI.

awalter-bnl:
Thanks Dan,
I have added this and got the plotting to now work including having the right values set to the right axes, except that the labels for the axes are swapped. I was not able to work out where the axes labels where set from, and so was not able to fix this. I assume it is one of the metadata inputs, which one is it taken from?

cheers
Andrew


On Mar 8, 2018, at 9:02 AM, Dan Allan <notifications@github.com<mailto:notifications@github.com>> wrote:


When I run @mrakitin<https://github.com/mrakitin>'s example, I get a blank figure, so it sounds like I can reproduce what @mrakitin<https://github.com/mrakitin> is seeing. I also notice this warning in the console:

UserWarning: Need both 'shape' and 'extents' in plan metadata to create LiveGrid.


The bec gets part way through trying to create a figure and then discovers it doesn't have all the information it needs and bails. It would have been easier for you to catch this problem if we had made this an error instead of a warning -- a nice case study in why warnings are generally useless -- but in this case we want to allow a scan to proceed even if bec doesn't have all the metadata it wants to make the best visualization. Our reasoning is, "The data is still being saved, and you can always visualize it later."

So, to finish this add something like:

   _md = {'shape':(y_num, x_num),
           'extents': ((y_center - y_range, y_center + y_range),
                       (...)),
            ...


I'm not 100% sure I have the order of y and x right there, FYI.

danielballan:
@awalter-bnl I pushed a fix to your branch.

bluesky/plans.py
```diff
@@ -1397,12 +1397,13 @@ def spiral_square(detectors, x_motor, y_motor, x_center, y_center, x_range,
                          'x_range': x_range, 'y_range': y_range,
                          'x_num': x_num, 'y_num': y_num,
                          'per_step': repr(per_step)},
-                         'plan_name': 'spiral_square',
-                         'plan_pattern': 'spiral_square',
-                         'shape': (y_num,x_num),
-                         'extents': ((y_center - y_range, y_center + y_range),
-                                     (x_center - x_range, x_center + x_range)),
-                         'hints': {},
+                        'plan_name': 'spiral_square',
```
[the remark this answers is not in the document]
mrakitin: The problem is that this block of code
```py
{'detectors': list(map(repr, detectors)),
                         'x_motor': repr(x_motor), 'y_motor': repr(y_motor),
                         'x_center': x_center, 'y_center': y_center,
                         'x_range': x_range, 'y_range': y_range,
                         'x_num': x_num, 'y_num': y_num,
                         'per_step': repr(per_step)}
```
should be indented at the same level as
```py
                        'plan_name': 'spiral_square',
                        'plan_pattern': 'spiral_square',
                        'shape': (y_num, x_num),
                        'extents': (
                            (y_center - y_range / 2, y_center + y_range / 2),
                            (x_center - x_range / 2, x_center + x_range / 2)),
                        'hints': {},
```
so `'per_step':` and above should have 1 more level of indention. Anyway, flake8 is not happy at line 1400, not sure how to align it correctly. Trying...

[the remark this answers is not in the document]
mrakitin: Fixed that in 91c4ad68766304ad0c9de0ed037e7d71c7703fd9.

mrakitin:
We performed some more tests with @awalter-bnl, and the live plot does not seem to produce correct results.

Here is the test code:
```py
%matplotlib qt
install_qt_kicker()

from bluesky.plans import spiral_square, grid_scan
from bluesky.simulators import summarize_plan, plot_raster_path
from ophyd.sim import hw
from matplotlib import pyplot as plt


RE = RunEngine({})
bec = BestEffortCallback()
RE.subscribe(bec)

hw = hw()
det = hw.det4
motor1 = hw.motor1
motor2 = hw.motor2

motor1.delay = 0.1
motor2.delay = 0.1

plt.ion()

# summarize_plan(spiral_square([det], motor1, motor2, 0, 0, 1, 1, 11, 11))
RE(spiral_square([det], motor1, motor2, 0, 0, 1, 2, 11, 11))

RE(grid_scan([det], motor2, -1, 1, 11, motor1, -0.5, 0.5, 11, True))

plot_raster_path(spiral_square([det], motor1, motor2, 0, 0, 1, 2, 11, 11),
                 'motor1', 'motor2')
```

Here are the results:

`spiral_square` | `grid_scan`
-|-
det4_vs_motor1_motor2_4_spiral | det4_vs_motor1_motor2_3_grid

The raster path for `spiral_square` is:
det4_vs_motor1_motor2_3_raster

danielballan:
`LiveRaster` sorts by sequence number, not (x, y) position. You'll need to fall back to using `LiveScatter` (by changing the `'gridding` hint) or teach `LiveRaster` how to deal with rasters that have non-trivial orderings.

mrakitin:
That's what I got when changed `gridding` to `raster` in hints:
det4_vs_motor1_motor2_raster_spiral
I don't like the axes are swapped, but it seems the behavior is consistent with `grid_scan`.

danielballan:
Hmm. The inconsistency in axes between LiveScatter and LiveGrid seems like a bug that could have been included in #951. Are you up for taking on those dragons again, @tacaswell?

mrakitin:
@danielballan, I think they are still consistent, we changed the order of motors in the scan:
```py
RE(grid_scan([det], motor2, -1, 1, 11, motor1, -0.5, 0.5, 11, True))
```
If I change the order of motors in `spiral_square`, it produces same axes ordering:
det4_vs_motor1_motor2_spiral_reverse_order

danielballan:
Oh, OK. Is this good to merge then?

danielballan:
I gather than @awalter-bnl would have preferred a LiveRaster-like plot. Making a LiveRaster more sophisticated seems out of scope for this PR. If that gets added later, we can always update the hints to 'rectilinear' again at that time.

danielballan:
Actually, do we want a separate terms -- something like `{'gridding': 'rectilinear_sequential'}` (like a grid) and `{'gridding': 'rectilinear_special'}` (like a spiral)? For now, `{'gridding' : 'rectilinear_special'}` will be the same add no gridding hint at all, so it will have no immediate effect, but if `LiveRaster` gets smarter in the future, it will be abel to interpret old data.

mrakitin:
I'm not sure. I just pushed my update, but the live plot does not seem as nice as `grid_scan`'s one. I'd still try to
> teach `LiveRaster` how to deal with rasters that have non-trivial orderings.

@awalter-bnl and I discussed it and it looks like a good feature to accept arbitrary trajectories. Maybe it's worth a separate issue and PR. @danielballan, what do you think?

mrakitin:
OK, I'm slower in typing 😄. You answered my question in https://github.com/NSLS-II/bluesky/pull/1001#issuecomment-371614792. Let's then merge this PR if everybody's happy with it, and I'll open a separate issue for making `LiveRaster` smarter.

danielballan:
Agreed.

One small nit-pick: I'm not sure I like introducing the new key `'raster'` into hints. Nothing else uses that key yet, and BEC doesn't know to look for it yet. Since we already have `'gridding'`, should we just come up with a new _value_ as opposed to a new key? Maybe `{'gridding': 'rectilinear_nonsequential'}` since it is rectilinear but we can't rely on `event['seq_num']` for ordering?

mrakitin:
OK, created #1003 to follow up later.

@danielballan, I agree, will update it with `rectilinear_nonsequential` as suggested.

prjemian:
Imagine a `raster` scan that is not rectilinear, such as a [lissjous](https://github.com/BCDA-APS/APS_BlueSky_tools/issues/25).

danielballan:
In Thebes lissjous case I think there should be no “gridding” hint at all, in which case the Best-Effort Callback will fall back on the most generic way of representing 2D scans, LiveScatter.

Do we forsee any other categories for 2D scans beyond:
- sequentially-ordered grid
- grid with some arbitrary ordering (as in square spiral)
- non-gridded

danielballan:
Confusing typo there — “Thebes” should be “the”. Thanks, iPhone!

mrakitin:
Just to avoid confusion for the future us: here is what the following code produces:
```py
%matplotlib qt
install_qt_kicker()

from bluesky.plans import spiral_square, grid_scan
from bluesky.simulators import summarize_plan, plot_raster_path
from ophyd.sim import hw


RE = RunEngine({})
bec = BestEffortCallback()
RE.subscribe(bec)

hw = hw()
det = hw.det4
motor1 = hw.motor1
motor2 = hw.motor2

motor1.delay = 0.1
motor2.delay = 0.1

plot_raster_path(spiral_square([det], motor1, motor2, 0, 0, 1, 2, 11, 11),
                 'motor1', 'motor2')

RE(spiral_square([det], motor1, motor2, 0, 0, 1, 2, 11, 11))

RE(grid_scan([det], motor1, -0.5, 0.5, 11, motor2, -1, 1, 11, True))
```

pr1001_final

I'm not very happy with the order of axes in the right two plots (i.e. `motor1` is vertical, and `motor2` is horizontal axis), but it's probably how live plotting works (the behavior differs from what `plot_raster_path` suggests). If I change the motors order in arguments of the two plans, this still does not match the trajectory (should be transposed). @danielballan, any reasoning why `motor1` takes vertical axis?

awalter-bnl:
@mrakitin this is perhaps something we may want to change in the plan. For grid_scan the outer (or slowest)axis is first and it works progressively inward. So for a 2D grid_scan the order is y_axis then x_axis. So in your routines you have motor 1 and motor 2 around the wrong way to have them plot correctly. It appears that scatter plot is flipping the axis from what it should be as For the square_spiral I have followed the spiral approach which is to order them as x_axis then y_axis. When plotting using raster plot it matches the plot raster path alignment.

Perhaps the order of specifying axes should be standardized in one direction or the other. Given that for N-dimensional scans the grid_scan idea works better perhaps this should also be used for the spiral scans?



On Mar 8, 2018, at 5:23 PM, Maksim Rakitin <notifications@github.com<mailto:notifications@github.com>> wrote:


Just to avoid confusion for the future us: here is what the following code produces:

%matplotlib qt
install_qt_kicker()

from bluesky.plans import spiral_square, grid_scan
from bluesky.simulators import summarize_plan, plot_raster_path
from ophyd.sim import hw


RE = RunEngine({})
bec = BestEffortCallback()
RE.subscribe(bec)

hw = hw()
det = hw.det4
motor1 = hw.motor1
motor2 = hw.motor2

motor1.delay = 0.1
motor2.delay = 0.1

plot_raster_path(spiral_square([det], motor1, motor2, 0, 0, 1, 2, 11, 11),
                 'motor1', 'motor2')

RE(spiral_square([det], motor1, motor2, 0, 0, 1, 2, 11, 11))

RE(grid_scan([det], motor1, -0.5, 0.5, 11, motor2, -1, 1, 11, True))

[pr1001_final]<https://user-images.githubusercontent.com/13209176/37179767-c0916d08-22f4-11e8-84f7-d8b277c36f2c.png>

I'm not very happy with the order of axes in the right two plots (i.e. motor1 is vertical, and motor2 is horizontal axis), but it's probably how live plotting works (the behavior differs from what plot_raster_path suggests). If I change the motors order in arguments of the two plans, this still does not match the trajectory (should be transposed). @danielballan<https://github.com/danielballan>, any reasoning why motor1 takes vertical axis?

mrakitin:
Thanks for the explanation, @awalter-bnl! I'm not sure if there is something else to do for this PR. Do we want to change ordering of the axes within this PR?

awalter-bnl:
@mrakitin I think the order of the axes is a larger issues that should be looked at in detail across all plans I think this is a larger issue than this PR. As I followed the spiral and fermat_spiral plan arrangement for the spiral_square plan I think we should pull as is.

mrakitin:
Thanks @awalter-bnl! I agree that the proposed refactoring has a larger scope.
Thanks @danielballan for your review!

mrakitin:
@tacaswell, I think it's a good candidate for NSLS-II deployment 2018-2. Some beamlines are really interested in it.

awalter-bnl:
I agree with @mrakitin there are at least 2 that I know are waiting for this to be deployed !-)

mrakitin:
Travis consistently fails for Python 3.5 with:
```
bluesky/tests/test_simple_api.py::test_plans_motors_no_hints[grid_scan-args3-kwargs3] /home/travis/.travis/job_stages: line 57:  3533 Aborted                 (core dumped) coverage run run_tests.py
The command "coverage run run_tests.py" exited with 134.
0.16s$ coverage report -m
Name    Stmts   Miss  Cover   Missing
-------------------------------------
No data to report.
```
@danielballan, any ideas why?

mrakitin:
3 or 4...

danielballan:
Wild. I have no idea. Things I would try:

* Install a Python 3.5 conda environment and try to reproduce that failure locally.
* Either locally (if possible to reproduce) or on Travis, run the tests on commits search for the specific commit between the master branch and the tip of this branch that introduces the failure.

tacaswell:
'power-cycled' the PR incase we were caching a bad version of some dependency.

Core dumps are real bad.

This is also happening on master:

https://travis-ci.org/NSLS-II/bluesky/jobs/373640546

tacaswell:
but the last build from the PR worked https://travis-ci.org/NSLS-II/bluesky/builds/373521021and the master build from 2 weeks ago https://travis-ci.org/NSLS-II/bluesky/builds/368926608

PR #1002, opened 2018-03-08
ENH: improve peak stats output on bec-plots

### LINKED ISSUE #934: ENH: display more information for peak stats in legends
## Expected Behavior

It would be very useful to have the peak stat values in the legend, e.g. some or all values can be displayed:
screenshot-2017-12-15 what s new - 2017 holiday edition

## Current Behavior

<kbd>Shift</kbd>+<kbd>P</kbd> currently displays:
screenshot-2017-12-15 what s new - 2017 holiday edition 1

## Possible Solution

Implement the values printed next to the stat names.

## Context

Requested by @zhernenkov, @yugangzhang, @lwiegart (the list is growing).

## Your Environment

http://try.nsls2.bnl.gov/ and  Ubuntu dev machine.

mrakitin:
Add more peak stats information on bec-plots.

## Description

The PR covers more verbose output of the peak stats on the best effort callback plots.

## Motivation and Context

Resolves #934.

## How Has This Been Tested?

All existing tests have passed.

## Reproduce it with:
```py
%matplotlib qt
install_qt_kicker()

from bluesky.plans import count, scan, rel_scan
from ophyd.sim import motor, det

RE = RunEngine({})
bec = BestEffortCallback()
RE.subscribe(bec)

RE(rel_scan([det], motor, -1.5, 2, 15))
```

det_vs_motor

### DIFF
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -504,12 +504,13 @@ class LivePlotPlusPeaks(LivePlot):
                 for artist in self.__arts:
                     artist.set_visible(True)
         elif self.__arts is not None:
-                for artist in self.__arts:
-                    artist.set_visible(False)
+            for artist in self.__arts:
+                artist.set_visible(False)
         self.ax.legend(loc='best')
         self.ax.figure.canvas.draw_idle()
 
     def plot_annotations(self):
+        # Vertical lines:
         styles = iter(cycler('color', 'kr'))
         vlines = []
         for style, attr in zip(styles, ['cen', 'com']):
@@ -518,10 +519,42 @@ class LivePlotPlusPeaks(LivePlot):
             if self.ax in self.__labeled:
                 label = '_no_legend_'
             else:
-                label = attr
+                label = f'{attr}={val:.2f}'
             vlines.append(self.ax.axvline(val, label=label, **style))
+
+        # Horizontal lines:
+        styles = iter(cycler('color', 'bm'))
+        hlines = []
+        for style, attr in zip(styles, ['min', 'max']):
+            val = self.peak_results[attr][self.y]
+            # Only put labels in this legend once per axis.
+            if self.ax in self.__labeled:
+                label = '_no_legend_'
+            else:
+                label = f'{attr}={val[1]:.2f}'
+            hlines.append(self.ax.axhline(val[1], label=label, **style))
+
+        # Horizontal arrows (e.g., FWHM):
+        attr = 'fwhm'
+        val = self.peak_results[attr][self.y]
+        harrows = []
+        if val:
+            max_crds = self.peak_results['max'][self.y]
+            min_crds = self.peak_results['min'][self.y]
+            style = {'color': 'g'}
+            fwhm = val['fwhm']
+            fwhm_x = val['cen_list']
+            fwhm_y = (max_crds[1] + min_crds[1]) / 2
+            label = f'{attr}={fwhm:.2f}'
+            # Fake line for the legend:
+            hlines = self.ax.plot([], [], label=label, **style)
+            harrows.append(self.ax.annotate(
+                '', xytext=(fwhm_x[0], fwhm_y),
+                xy=(fwhm_x[1], fwhm_y),
+                arrowprops=dict(arrowstyle="<->", **style)))
+
         self.__labeled[self.ax] = None
-        self.__arts = vlines
+        self.__arts = vlines + hlines + harrows
 
     def stop(self, doc):
         self.check_visibility()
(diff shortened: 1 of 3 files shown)


danielballan:
> any ideas where and how they should be hidden?

I remember trying to fix this in the original PR that introduced the Best-Effort Callback. If I remember correctly, matplotlib made this hard to do -- technically _possible_ but arguably not worth the code complexity it would introduce -- but there was a fix coming in matplotlib. It's possible that fix has since been released. Not sure if updating matplotlib is enough or if further action is now required on our part. @tacaswell can probably fill in the rest.

danielballan:
> Any ideas how to combine them into one?

Maybe change

```python
for style, attr in zip(styles, ['cen', 'com']):
```

to

```python
for style, attr, meth in zip(styles, ['cen', 'com', 'min', 'max'], ['vline', 'vline', 'hline', 'hline']):
    ...
    getattr(self.ax, meth)(...)
```

It's not immediately obvious to me that would be an improvement overall. It's a judgement call between repetition and simplicity vs "DRY" and some complexity to support it. I can see this one either way.

bluesky/callbacks/best_effort.py
```diff
@@ -518,10 +519,42 @@ def plot_annotations(self):
             if self.ax in self.__labeled:
                 label = '_no_legend_'
             else:
-                label = attr
+                label = '{}={:.2f}'.format(attr, val)
```
danielballan: :+1:

danielballan: It might be nice to make this format string configurable via `__init__` and some attribute, like `label_format='{attr}={val:.2f}'`. The main use case would be to let people increase precision. Might as well give them control over the whole string.

prjemian:
And a default setting for the user-supplied format string if it raises an exception.

bluesky/callbacks/best_effort.py
mrakitin: I like the idea, will move it to the attribute. I also agree about the precision, we should not stick to 2 decimal points.

mrakitin:
It's WIP, don't merge yet, please. I'll implement @prjemian's suggestion on raising an exception. Also, we with @tacaswell played a bit with draggable text boxes, it looks like we can improve the behavior for multiple plots with them rather than placing all the items into the legend (imagine you have 3+ curves, the plot will be overcrowded).

bluesky/callbacks/fitting.py
```diff
@@ -282,7 +282,9 @@ def compute(self):
             self.cen = np.mean(_cen_list)
 
         if len(_cen_list) == 2:
-            self.fwhm = float(_cen_list[1] - _cen_list[0])
+            self.fwhm = {'fwhm': float(_cen_list[1] - _cen_list[0]),
```
danielballan: @mrakitin If you refactor this PR onto #888, can you use `self.fwhm` and `self.crossings` instead?

mrakitin: Oh, cool! I didn't realized that looking at the current master. Should I wait until #888 is merged?

danielballan: That's probably easiest.

prjemian: Is this going to break existing code that expects self.fwhm to be a number, not a dict?

danielballan: @prjemian It would, but I don't believe it will be necessary to make this change in the end.

bluesky/callbacks/best_effort.py
```diff
@@ -432,7 +432,7 @@ def clear(self):
 
 
 class PeakResults:
-    ATTRS = ('com', 'cen', 'max', 'min', 'fwhm', 'nlls')
+    ATTRS = ('com', 'cen', 'max', 'min', 'fwhm', 'nlls', 'crossings')
```
mrakitin: @danielballan, I think this part was missed in #888, so I added the `crossings` attribute here as well. This caused a problem with `test_bec.py` in Travis (also reproduced in my dev env). I solved it by changing the type of the `crossings` from `ndarray` to `list` (see [below](https://github.com/NSLS-II/bluesky/pull/1002/files#diff-177dabe93bbee02f265f7408541d9964R291)).

danielballan:
Status update: This is an improvement on the status quo in my view. @mrakitin would still like to add more interactivity: not simply displaying the _latest_ parameters in the legend, but making it possible to go back to parameters for older curves by clicking on those curves.

bluesky/callbacks/best_effort.py
```diff
@@ -85,6 +86,10 @@ def disable_plots(self):
         "Do not plot anything."
         self._plots_enabled = False
 
+    def format_labels(self, fmt):
```
tacaswell: Should this be a property?

tacaswell:
bigger fish to fry this week and should be re-factored for new call back scheme anyway.

stan-dot:
@tacaswell  I made an updated version of this PR here https://github.com/bluesky/bluesky/pull/1782


should this PR be closed?

stan-dot:
(misclick)

PR #1006, opened 2018-03-16
Try streaming read version 2

jrmlhermitte:
## Description
This is an alternative example on how to implement reading from streaming devices, meant for discussion with #993.

This has been tried on v1.2.0 (not including asset refactor yet. this is in case we decide we want to backport)

## Motivation
The motivation for this is that in some cases involving time series from area detectors, acquisitions can take a while (30 min or so). During this acquisition, it's actually not possible from bluesky to know what the filename of the detector is before the fact! It would be nice to allow this feature, and in a general fashion that may benefit other experiments.

## How has this been tested?
See the script below. Should add tests later if this implementation is chosen.

```py3

from databroker import Broker
from bluesky import RunEngine
from ophyd.sim import img, det1, det2, NumpySeqHandler


db = Broker.from_config(temp_config())
img.reg = db.reg

RE = RunEngine({})

#db.reg.register_handler('NPY_SEQ', NumpySeqHandler)

#RE.subscribe(lambda name, doc: print(name))
RE.subscribe(db.insert)

# to see the documents as they go
RE.subscribe(print)

uid, = RE(count([img, det1, det2], [img]))
```

and:
```
In [5]: db[-1].table()
Out[5]: 
                              time  det1  \
seq_num                                    
1       2018-03-16 08:12:00.158367   5.0   

                                          img      det2  
seq_num                                                  
1        2622cdcb-b471-4db0-8ad6-2afbedeb2cd8  1.764994  

In [6]: db[-1].table("streaming_primary")
Out[6]: 
                              time                                   img
seq_num                                                                 
1       2018-03-16 08:12:00.137234  2622cdcb-b471-4db0-8ad6-2afbedeb2cd8
```


jrmlhermitte:
After discussion with @danielballan I am convinced that #993 is the way to go. Closing.

PR #1011, opened 2018-03-26
DOC: add `%matplotlib notebook` to tutorials docs

mrakitin:
Tutorials page is missing `%matplotlib notebook` directive. Fixing this issue here.

### DIFF
--- a/doc/source/tutorial.rst
+++ b/doc/source/tutorial.rst
@@ -137,6 +137,8 @@ To start, let's use the all-purpose
 
 .. code-block:: python
 
+    %matplotlib notebook
+
     from bluesky.callbacks.best_effort import BestEffortCallback
     bec = BestEffortCallback()
 


tacaswell:
I am mostly :-1: on this, the text above talks about setting up the terminal with qt5.

mrakitin:
OK, I agree it's a bit out of scope. The reason I suggested this addition is that I tried to copy-paste the code snippets into a notebook at http://try.nsls2.bnl.gov/, and at some point got a nasty error message:
```py
bec = BestEffortCallback()

# Send all metadata/data captured to the BestEffortCallback.
RE.subscribe(bec)

# Make plots update live while scans run.
from bluesky.utils import install_kicker
install_kicker()
```

```python
NotImplementedError                       Traceback (most recent call last)
<ipython-input-2-7e8b6286ffe4> in <module>()
      7 # Make plots update live while scans run.
      8 from bluesky.utils import install_kicker
----> 9 install_kicker()

/opt/conda/envs/tutorial/lib/python3.6/site-packages/bluesky/utils.py in install_kicker(loop, update_rate)
    719     else:
    720         raise NotImplementedError("The matplotlib backend {} is not yet "
--> 721                                   "supported.".format(backend))
    722 
    723 

NotImplementedError: The matplotlib backend module://ipykernel.pylab.backend_inline is not yet supported.
```

As an alternative, we may have a `Note` box, where we warn users if they use Jupyter notebooks, then they will need to use the magic command. Thoughts?

tacaswell:
we should probably sort out how to change the default backend to notebook from inline on try.nsls2.bnl.gov

mrakitin:
Agree. I haven't found a place where to configure it to have a notebook backend by default, but there must be something about it. Closing this PR and will submit an issue to the docker tutorial project.

PR #1013, opened 2018-03-30
BUG: Suspenders now call .resume() upon resuming

stuartcampbell:
The suspenders where calling the obj.pause() upon being triggered, but were not calling the obj.resume() when they resumed.

## Description
Added a coroutine `_resume()` and then registered (And called) a `Msg` called `resume` to run it.

## Motivation and Context
The FastCCD detector on CSX turns off the HDF5 file capture when paused, and as resume is not being called on the suspender resume, the detector never starts writing data to a file again.  

For the Ctrl-C type pause/resume everything works as expected.

## How Has This Been Tested?
We (myself and @jrmlhermitte) ran numerous tests on the beamline to make sure this fixed the observed issue on CSX.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -256,7 +256,7 @@ class RunEngine:
         self._msg_cache = deque()  # history of processed msgs for rewinding
         self._rewindable_flag = True  # if the RE is allowed to replay msgs
         self._plan_stack = deque()  # stack of generators to work off of
-        self._response_stack = deque([None])  # resps to send into the plans
+        self._response_stack = deque()  # resps to send into the plans
         self._exit_status = 'success'  # optimistic default
         self._reason = ''  # reason for abort
         self._task = None  # asyncio.Task associated with call to self._run
@@ -420,7 +420,7 @@ class RunEngine:
         self._deferred_pause_requested = False
         self._plan_stack = deque()
         self._msg_cache = deque()
-        self._response_stack = deque([None])
+        self._response_stack = deque()
         self._exception = None
         self._run_start_uids.clear()
         self._exit_status = 'success'
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1039,12 +1039,12 @@ class RunEngine:
                     # side of the yield in the plan will be moved past
                     resp = self._response_stack.pop()
                     # The case where we have a stashed exception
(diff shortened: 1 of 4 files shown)


stuartcampbell:
I know we need to fix the tests, but I wanted to just get the PR in so people can have a look to make sure it's not nonsense! 😄

tacaswell:
@cmazzoli Congratulations on helping to identify a long standing set of bugs in the core of the RunEngine!

I think that the failure of the loop invariant had little impact in practice as the test that were failing (other than the obvious off-by-one) were all very unhappy failure paths.  If we had examples of plans on the floor that were aggressively catching exceptions and were interrupted and  expecting to get the right values back we might have seen it, but would be very surprised if this ever reared it's head on the floor.

tacaswell:
The nightly failure is a cython / cpython clash.  Those issues have all been fixed upstream, but apparently not in the nightly travis enviroment yet.

stuartcampbell:
@danielballan I think that whether it's a Msg or not in the future is up for debate.  I think we just need a consistent behaviour so that the 'resume' (however it is injected) is done.

tacaswell:
I wonder if it makes sense to say the _only_ way to change the internal state of the RE is via a message / co routine?

bluesky/run_engine.py
```diff
@@ -999,10 +1008,19 @@ def _run(self):
         """
         pending_cancel_exception = None
         self._reason = ''
+        # sentinel to decide if need to add to the response stack or not
+        sentinel = object()
         try:
             self.state = 'running'
             while True:
+                assert len(self._response_stack) == len(self._plan_stack)
```
jrmlhermitte: :+1:

bluesky/run_engine.py
```diff
@@ -1016,17 +1034,25 @@ def _run(self):
...
                     # correct place
                     yield from asyncio.sleep(0, loop=self.loop)
+                    # always pop off a result, we are either sending it back in
+                    # or throwing an exception in, in either case the left hand
+                    # side of the yield in the plan will be moved past
+                    resp = self._response_stack.pop()
                     # The case where we have a stashed exception
-                    if self._exception is not None:
+                    if (self._exception is not None or
+                            isinstance(resp, Exception)):
```
jrmlhermitte: good catch as well

PR #1018, opened 2018-04-10
FIX: clean up the internal state of the RE earlier in _save

tacaswell:
Move the logic to re-set the internal state of the RE to not bundling
to above where we do anything with descriptors.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1560,6 +1560,17 @@ class RunEngine:
 
         # Event Descriptor documents
         desc_key = self._bundle_name
+
+        # This is a separate check because it can be reset on resume.
+        seq_num_key = desc_key
+        if seq_num_key not in self._sequence_counters:
+            counter = count(1)
+            counter_copy1, counter_copy2 = tee(counter)
+            self._sequence_counters[seq_num_key] = counter_copy1
+            self._teed_sequence_counters[seq_num_key] = counter_copy2
+        self._bundling = False
+        self._bundle_name = None
+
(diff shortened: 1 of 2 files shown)


danielballan:
It looks like this breaks a couple of tests in a real way.

bluesky/run_engine.py
```diff
@@ -1560,6 +1560,17 @@ def _save(self, msg):
...
+
+        # This is a separate check because it can be reset on resume.
+        seq_num_key = desc_key
+        if seq_num_key not in self._sequence_counters:
+            counter = count(1)
+            counter_copy1, counter_copy2 = tee(counter)
+            self._sequence_counters[seq_num_key] = counter_copy1
+            self._teed_sequence_counters[seq_num_key] = counter_copy2
+        self._bundling = False
+        self._bundle_name = None
```
jrmlhermitte: line 1601 needs `self._bundle_name`
I recommend:
`bundle_name = self._bundle_name`
...

tacaswell: indeed, we already had grabbed this a few lines above as `desc_key`.

tacaswell:
fixed and force-pushed.

bluesky/run_engine.py
jrmlhermitte: yes! looks good to me

tacaswell:
@jrmlhermitte why do you want a rebase?

jrmlhermitte:
this was for problems with callbacks like such:
```py3
def callback(name, doc):
    if name == 'descriptor':
        raise ValueError


```

PR #1019, opened 2018-04-12
MNT: tweak document validation 

tacaswell:
## Description

This is coupled with the PR into event-model (https://github.com/NSLS-II/event-model/pull/15).

## Motivation and Context

 - allow both list and tuple to be accepted as json "array"
 - put a list in the descriptor for the interruption event stream to allow for stricter schema (this is required for tests to pass with https://github.com/NSLS-II/event-model/pull/15).

## How Has This Been Tested?

Running the bluesky tests against a local version of event-model

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ before_install:
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
   - pip install -r requirements.txt
+  - pip install --upgrade cython
   - pip install -r dev-requirements.txt
   - pip install codecov
   - python setup.py install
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -26,6 +26,7 @@ before_install:
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
   - pip install -r requirements.txt
+  - pip install --upgrade cython
```
jrmlhermitte: why upgrade cython?

tacaswell: to try and make it work with the nightly environment.  It is currently failing on a pandas bug (which is fixed on master and will be released soon)

jrmlhermitte: sounds good thanks!

PR #1020, opened 2018-04-12
Added magics

jrmlhermitte:
WIP : Adding some ipython magics for bluesky for better diagnostics.

## Description
This is a WIP adding a series of bluesky magics that will allow the user to find what Devices they have available.

## Motivation and Context
Users often need to see a list of defined ophyd devices. This will help with that.

## How Has This Been Tested?

Tested at PDF beamline so far:
```py3
In [1]: %detectors
Python name              Ophyd Name          
========================================
Test_Cam1                Test_Cam1           
Cam2                     Cam2                
camera                   Cam2                
stats_plugin             Cam2_stats5         
pe1                      pe1                 
```

## Discussion
This is currently just searching the user name space. It might be potentially useful to search all `Signals` (children of `Devices`).

### DIFF
--- a/bluesky/magics.py
+++ b/bluesky/magics.py
@@ -6,6 +6,7 @@
 # ip.register_magics(BlueskyMagics)
 
 import asyncio
+import warnings
 from bluesky.utils import ProgressBarManager
 from bluesky import RunEngine, RunEngineInterrupted
 from IPython.core.magic import Magics, magics_class, line_magic
@@ -117,7 +118,7 @@ class BlueskyMagics(Magics):
         ''' List all available detectors.'''
         # also make sure it has a name for printing
         label = 'detector'
-        devices = _labeled_devices(user_ns=self.shell.user_ns)
+        devices = get_labeled_devices(user_ns=self.shell.user_ns)
         cols = ["Python name", "Ophyd Name"]
         print("{:20s} \t {:20s}".format(*cols))
         print("="*40)
@@ -129,7 +130,7 @@ class BlueskyMagics(Magics):
         ''' List all available motors.'''
         # also make sure it has a name for printing
         label = 'motor'
-        devices = _labeled_devices(user_ns=self.shell.user_ns)
+        devices = get_labeled_devices(user_ns=self.shell.user_ns)
         # ignore the first key
         positioners = [positioner[1] for positioner in devices[label]]
         _print_positioners(positioners, precision=self.FMT_PREC)
@@ -203,13 +204,14 @@ def _print_positioners(positioners, sort=True, precision=6):
     print('\n'.join(lines))
 
 
-def _labeled_devices(user_ns=None, maxdepth=6):
-    ''' Returns list of all devices that are labeled.
+def get_labeled_devices(user_ns=None, maxdepth=6):
+    ''' Returns dict of labels mapped to devices with that label
 
         Parameters
         ----------
         user_ns : dict, optional
             The namespace to search on
+            Default is to grab the namespace of the ipython shell.
 
         maxdepth: int, optional
             max recursion depth
@@ -221,14 +223,14 @@ def _labeled_devices(user_ns=None, maxdepth=6):
         Examples
         --------
         Read devices labeled as motors:
-            objs = _labeled_devices()
+            objs = get_labeled_devices()
             my_motors = objs['motors']
     '''
     # could be set but lists are more common for users
     obj_list = collections.defaultdict(list)
 
     if maxdepth <= 0:
-        print("Recursion limit exceeded")
+        warnings.warn("Recursion limit exceeded")
         return obj_list
 
     if user_ns is None:
@@ -242,11 +244,11 @@ def _labeled_devices(user_ns=None, maxdepth=6):
         if not key.startswith("_"):
             if is_parent(obj):
                 labels = getattr(obj, '_ophyd_labels_', set())
-                obj_list.update(_labeled_devices(user_ns=obj.__dict__,
-                                                 maxdepth=maxdepth-1,))
+                obj_list.update(get_labeled_devices(user_ns=obj.__dict__,
+                                                    maxdepth=maxdepth-1,))
             else:
                 if hasattr(obj, '_ophyd_labels_'):
-                    # inherit parent labels
+                    # don't inherit parent labels
                     labels = obj._ophyd_labels_
                     for label in labels:
                         obj_list[label].append((key, obj))
@@ -258,7 +260,7 @@ def is_parent(dev):
     # return whether a node is a parent
     # should not have component_names, or if yes, should be empty
     # read_attrs needed to check it's an instance and not class itself
-    return (hasattr(dev, 'component_names') and len(dev.component_names) > 0
+    return (isinstance(dev, type) and len(dev.component_names) > 0
             and hasattr(dev, 'read_attrs'))
 
 
--- a/bluesky/magics.py
+++ b/bluesky/magics.py
@@ -115,7 +115,8 @@ class BlueskyMagics(Magics):
     def detectors(self, line):
         ''' List all available detectors.'''
         # also make sure it has a name for printing
-        logic = lambda x : is_detector(x) and hasattr(x, 'name')
+        def logic(x):
+            return is_detector(x) and hasattr(x, 'name')
         devices = _which_devices(cls_whitelist=logic, cls_blacklist=None,
                                  user_ns=self.shell.user_ns)
         cols = ["Python name", "Ophyd Name"]
@@ -128,15 +129,14 @@ class BlueskyMagics(Magics):
     def motors(self, line):
         ''' List all available detectors.'''
         # also make sure it has a name for printing
-        logic = lambda x : is_epics_motor(x) and hasattr(x, 'name')
-        devices = _which_devices(cls_whitelist=logic, cls_blacklist=None,
-                                 user_ns=self.shell.user_ns)
-        cols = ["Python name", "Ophyd Name"]
-        print("{:20s} \t {:20s}".format(*cols))
-        print("="*40)
-        for name, obj in devices:
-            print("{:20s} \t {:20s}".format(name, str(obj.name)))
-
+        def logic(x):
+            return is_epics_motor(x) and hasattr(x, 'name')
+        positioner_list = _which_devices(cls_whitelist=logic,
+                                         cls_blacklist=None,
+                                         user_ns=self.shell.user_ns)
+        # ignore the first key
+        positioners = [positioner[1] for positioner in positioner_list]
+        _print_positioners(positioners, precision=self.FMT_PREC)
 
     @line_magic
     def wa(self, line):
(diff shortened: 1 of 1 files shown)


bluesky/magics.py
```diff
@@ -13,6 +13,10 @@
 from operator import attrgetter
 from . import plans as bp
 from . import plan_stubs as bps
+
+from ophyd.areadetector.base import ADBase
```
tacaswell: Can we do this without and ophyd dependency?

bluesky/magics.py
```diff
@@ -110,6 +114,27 @@ def ct(self, line):
...
+        ''' List all available detectors.'''
+        devices = _which_devices(cls_whitelist=[ADBase], cls_blacklist=None)
+        cols = ["Python name", "Ophyd Name"]
+        print("{:20s} \t {:20s}".format(*cols))
+        print("="*40)
+        for name, obj in devices:
+            print("{:20s} \t {:20s}".format(name, obj.name))
+
+    @line_magic
+    def motors(self, line):
```
tacaswell: This duplicates the functionality of `wa` below.

jrmlhermitte:
changed whitelist/blacklist to functions

```py3
In [1]: %motors
Python name              Ophyd Name          
========================================
Det_1_X                  Det_1_X             
Det_1_Y                  Det_1_Y             
Det_1_Z                  Det_1_Z             
Det_2_X                  Det_2_X             
Det_2_Y                  Det_2_Y             
Det_2_Z                  Det_2_Z             
Grid_X                   Grid_X              
Grid_Y                   Grid_Y              
Grid_Z                   Grid_Z              
ECS_tel_guide            ECS_tel_guide       

In [2]: %detectors
Python name              Ophyd Name          
========================================
Test_Cam1                Test_Cam1           
Cam2                     Cam2                
camera                   Cam2                
stats_plugin             Cam2_stats5         
pe1                      pe1                 
```

bluesky/magics.py
jrmlhermitte: thanks, we'll need to combine them

mrakitin:
I like the idea, it's really good!

teddyrendahl:
We do something [similar](https://github.com/pcdshub/hutch-python/blob/master/hutch_python/namespace.py) at SLAC. We decided to make `types.Namespace` objects which is a little more programmatically useful than a print out. We also made sure that you search through `Component` devices as well. For instance if you want all your `EpicsMotor` objects, you might want to check that you grab the `X`, `Y` and `Z` components of your `XYZStage`.

Like the idea! Interested to see the solution you come up with.

tacaswell:
It would also be interesting to look at some sort of decorator / registration that stashed weak-refs at device creation time.

We know quite a bit when we create these devices, we should find a way to exploit that rather than re-finding things after the fact.

jrmlhermitte:
@teddyrendahl I like your implementation!

So I think we're doing to check for attributes rather than check for classes to avoid importing in ophyd devices (@tacaswell suggested it here and I agree with it).
But the recursion is what I was thinking about implementing, and I had the same idea, you need two types : the device type (which can eventually contain a component leaf) and the desired type (the leaf).
I've implemented it here as well with `leaf_logic` (logic to determine if we have a leaf) or `node_logic` (logic to determine if we have a node). I assume the node is a subset of a leaf so I check for the leaf first.

Need to improve on this later. So far, it does not look asethetic:
```py3
In [3]: %signals
Python name              Ophyd Name          
========================================
det                      det                 
motor1readback           motor1              
motor1setpoint           motor1_setpoint     
motor1.motor2readback    motor2              
motor1.motor2setpoint    motor2_setpoint     
motor1.motor2det2        det2                
motor1.motor2.pseudo3x3.pseudo3x3_pseudo1readback        pseudo3x3_pseudo1   
motor1.motor2.pseudo3x3.pseudo3x3_pseudo1setpoint        pseudo3x3_pseudo1_setpoint
motor1.motor2.pseudo3x3.pseudo3x3_pseudo1.pseudo3x3_pseudo2readback      pseudo3x3_pseudo2   
motor1.motor2.pseudo3x3.pseudo3x3_pseudo1.pseudo3x3_pseudo2setpoint      pseudo3x3_pseudo2_setpoint
motor1.motor2.pseudo3x3.pseudo3x3_pseudo1.pseudo3x3_pseudo2.pseudo3x3_pseudo3readback    pseudo3x3_pseudo3   
motor1.motor2.pseudo3x3.pseudo3x3_pseudo1.pseudo3x3_pseudo2.pseudo3x3_pseudo3setpoint    pseudo3x3_pseudo3_setpoint
motor1.motor2.pseudo3x3.pseudo3x3_pseudo1.pseudo3x3_pseudo2.pseudo3x3_pseudo3sig         pseudo3x3_sig       

``
comments appreciated. thanks!

**Comment by jrmlhermitte** [comment]
> It would also be interesting to look at some sort of decorator / registration that stashed weak-refs at device creation time.
This requires a bit of state. I don't know if we want this. I vote more for passive, but less accurate methods over active but more accurate. In the end, if a motor is missing from the magic, I don't think it will cause too big of a fuss with BL scientists. An active method requires more support and trouble shooting.
 What do you think @tacaswell ?

**Comment by tacaswell** [comment]
The trade off is an explicit 'opt-in' which does require some state someplace and some maintenance in the profile, but if a device is missing or (in the view of BL staff) erroneously listed, it is fixed by adjusting the opt-in in the profile.  If it is all implicit / auto-discovery we don't have any state but if we the heuristic don't work it is much harder to fix on the floor (or we have to deal with conflicting heuristics from different beamlines).

**Comment by danielballan** [comment]
Randomly had a long discussion about this with @stuartcampbell and @tacaswell which I will now try to summarize:

## Proposed solution

Add a new optional keyword argument to `OphydObj` that takes a list of string labels which will be assigned to a new (private?) attribute. As in:

```py
class OphydObj:
   def __init__(..., labels=None):
        if labels is None:
            labels = []
        self._ophyd_labels_ = labels
```

(We are not necessarily suggesting the _name_ `labels`; this is just an illustration.) Then, magics or other utilities can scrape the namespace like this:

```py
import collections

for obj in globals():
    objects = collections.defaultdict(list)
    if hasattr(obj, '_ophyd_labels_'):
        for label in obj._ophyd_labels_:
            objects[label] = obj
```

Then, a general-purpose magic can display all labeled objects, visually grouped by label, or show a specific labels. Magics like `%ct` and `%wa` can be bound to specific label names like `'detectors'` and `'positioners'` respectively.

There is some precedence for this technique. Squatting an attribute name as a way of duck-typing is  also used by IPython's display hook (`_ipython_html`, etc).

## Advantages over current situation

* Users can name their own categories, not limited to 'detectors' and 'positioners'.
* A device can be sorted into multiple categories, like 'favorites' and 'detectors'.
* Instead of coordinating configuration in two places, like this:
    ```py
    motor = EpicsMotor('...', name='motor')
    # then, far away, in some other file...
    BlueskyMagics.positioners.append(motor)
    ```
    the configuration is all in one place:
    ```py
    motor = EpicsMotor('...', name='motor', labels=['positioners'])
    ```
* If an object is later deleted from the namespace, nothing bad happens. In the _current_ system, the names in `BlueskyMagics.positioners` can get out of sync with the variables names.
* If a profile script is re-executed with `%run`. In the current system, it is easy to end up with a list that includes duplicates.
* Overall, this avoids globally-managed state and keeps configuration local to each object. We have anecdotally found that global state is tempting but tends to grow complex and confusing. (It's why we removed `global_state.py` and underhook a lot of hassle to do so!)

## Considered Alternatives

* A special function that closes over a global list and registers an instance as a side-effect, like
   ```py
   motor = register(EpicsMotor('...', name='motor'))
   ```
* Scraping the namespace looking for attributes we already have, like `.position`.

jrmlhermitte:
I agree. I've added it [here](https://github.com/NSLS-II/ophyd/pull/538/files). We should move discussion there. We should merge and tag first before using bluesky magics? I can open a PR that tests by pulling from master on bluesky.

jrmlhermitte:
Okay, I updated to take in @danielballan 's comments. Needs this pr [here](https://github.com/NSLS-II/ophyd/pull/538).

Here is a gist example of some possible cases to run on [here](https://gist.github.com/jrmlhermitte/1e62deb334d677554e34512272d66762).


Tests are not passing, will investigate.

jrmlhermitte:
py3.6 passes but py3.5 fails. I'm not sure why?

bluesky/magics.py
```diff
@@ -110,53 +113,161 @@ def ct(self, line):
     positioners = []
     FMT_PREC = 6
 
+    @line_magic
+    def detectors(self, line):
+        ''' List all available detectors.'''
+        # also make sure it has a name for printing
+        label='detector'
```
danielballan: whitespace here (and similar where it occurs in later methods)

bluesky/magics.py
```diff
@@ -110,53 +113,161 @@ def ct(self, line):
...
+            value = v.__class__.__name__  # e.g. 'DisconnectedError'
+            low_limit = high_limit = offset = ''
+
+        lines.append(LINE_FMT.format(p.name, value, low_limit, high_limit,
+                                     offset))
+    print('\n'.join(lines))
+
+
+def _labeled_devices(user_ns=None, maxdepth=6):
+    ''' Returns list of all devices that are labeled.
```
danielballan: I found "Return list" to be misleading. It's always hard to be specific enough with 79 characters...maybe "Returns dict of labels mapped to devices with that label"?

bluesky/magics.py
```diff
@@ -110,53 +113,161 @@ def ct(self, line):
...
+
+        Parameters
+        ----------
+        user_ns : dict, optional
+            The namespace to search on
```
danielballan: Please mention what default behavior is

bluesky/magics.py
```diff
@@ -110,53 +113,161 @@ def ct(self, line):
...
 
-            lines.append(LINE_FMT.format(p.name, value, low_limit, high_limit,
-                                         offset))
-        print('\n'.join(lines))
+def is_parent(dev):
+    # return whether a node is a parent
+    # should not have component_names, or if yes, should be empty
+    # read_attrs needed to check it's an instance and not class itself
+    return (hasattr(dev, 'component_names') and len(dev.component_names) > 0
+            and hasattr(dev, 'read_attrs'))
```
danielballan: I think `not isinstance(dev, type)` is safer than `hasattr(dev, 'read_attrs')`, even though this will probably be fine.

bluesky/magics.py
```diff
@@ -110,53 +113,161 @@ def ct(self, line):
...
+        --------
+        Read devices labeled as motors:
+            objs = _labeled_devices()
+            my_motors = objs['motors']
+    '''
+    # could be set but lists are more common for users
+    obj_list = collections.defaultdict(list)
+
+    if maxdepth <= 0:
+        print("Recursion limit exceeded")
```
danielballan: Cool, seems reasonable to give up at a certain point rather than raise. I would prefer a warning here so that i can be tested and, if desired, suppressed.

bluesky/magics.py
```diff
@@ -110,53 +113,161 @@ def ct(self, line):
...
+        # return of commands)
+        # also check its a subclass of desired classes
+        if not key.startswith("_"):
+            if is_parent(obj):
+                labels = getattr(obj, '_ophyd_labels_', set())
+                obj_list.update(_labeled_devices(user_ns=obj.__dict__,
+                                              maxdepth=maxdepth-1,))
+            else:
+                if hasattr(obj, '_ophyd_labels_'):
+                    # inherit parent labels
```
danielballan: The current code does _not_ inherit parent labels, right? I _think_ that's probably the best behavior. (I suppose we could have a keyword arg to turn propagation on or off, but I don't think we necessarily need to build that until we know for sure that we need it for something.) Anyway, is this comment vestigial?

bluesky/magics.py
```diff
@@ -110,53 +113,161 @@ def ct(self, line):
...
+        else:
+            value = v.__class__.__name__  # e.g. 'DisconnectedError'
+            low_limit = high_limit = offset = ''
+
+        lines.append(LINE_FMT.format(p.name, value, low_limit, high_limit,
+                                     offset))
+    print('\n'.join(lines))
+
+
+def _labeled_devices(user_ns=None, maxdepth=6):
```
danielballan: Can this be a public function in case it's useful interactively?

bluesky/magics.py
jrmlhermitte: great idea! done

bluesky/magics.py
jrmlhermitte: sounds good. i used warnings.warn

bluesky/magics.py
jrmlhermitte: oops! fixed

bluesky/magics.py
jrmlhermitte: good idea. changed to `get_labeled_devices()`

jrmlhermitte:
new way 
```

In [1]: %run -i ../pyscripts/0092-test-pvmagics.py

In [2]: BlueskyMagics.positioners = []

In [3]: %wa
motor
        Python name              Ophyd Name          
        ========================================
        dev                      newdevice           
        dev2                     newdevice2          
other
        Python name              Ophyd Name          
        ========================================
        dev2                     newdevice2          
        cmp_dev                  compound            
detector
        Python name              Ophyd Name          
        ========================================
        det_dev                  newdet              
        cmp_dev                  compound            

```

will double check it tomorrow. what do you think?

bluesky/magics.py
```diff
@@ -107,55 +109,203 @@ def ct(self, line):
...
+
+    @positioners.setter
+    def positioners(self, val):
+        warnings.warn("positioners is deprecated." +
+                      "Please use the newer labels feature")
+        self._positioners = val
+
+    @property
+    def detectors(self):
+        warnings.warn("positioners is deprecated." +
```
danielballan: positioners -> detectors and likewise below

bluesky/magics.py
```diff
@@ -107,55 +109,203 @@ def ct(self, line):
...
+    cols = ["Python name", "Ophyd Name"]
+    print(prefix + "{:20s} \t {:20s}".format(*cols))
+    print(prefix + "="*40)
+    for name, obj in devices:
+        print(prefix + "{:20s} \t {:20s}".format(name, str(obj.name)))
+
+def are_positioners(devs):
+    # only true if all are positioners
+    # takes ((name, dev),...) tuple
+    logic = True
```
danielballan: This `logic` loop is clever, but `all(is_positioner(obj) for _, obj in devs)` might be simpler.

bluesky/magics.py
```diff
@@ -107,55 +109,203 @@ def ct(self, line):
         self._ensure_idle()
         return None
 
-    positioners = []
+    _positioners = []
+    _detectors = []
+
+    @property
```
danielballan: Now that I look at this written out I realize it isn't going to work. Properties only work their magic on _instances_ of classes. We need `BlueskyMagics.positioners` to return `[]`, but in fact it would do this:

```py
In [16]: class BlueskyMagics():
    ...:     _positioners = []
    ...:     @property
    ...:     def positioners(self):
    ...:         print('hey')
    ...:         return self._positioners
    ...:     @positioners.setter
    ...:     def positioners(self, val):
    ...:         self._positioners = val
    ...:

In [17]: BlueskyMagics.positioners
Out[17]: <property at 0x10d524278>
```

The good news is (heh heh) we have to use metaclasses to make this work the way we want. But it works:

```
In [21]: class MetaclassForClassProperties(type):
    ...:     @property
    ...:     def positioners(self):
    ...:         print('hey')
    ...:         return self._positioners
    ...:     @positioners.setter
    ...:     def positioners(self, val):
    ...:         print('hey')
    ...:         self._positioners = val
    ...:

In [22]: class BlueskyMagics(metaclass=MetaclassForClassProperties):
    ...:     _positioners = []
    ...:
    ...:

In [23]: BlueskyMagics.positioners  # test access
hey
Out[23]: []

In [24]: BlueskyMagics.positioners.append('a')  # test mutatation
hey

In [25]: BlueskyMagics.positioners  # test access again
hey
Out[25]: ['a']

In [26]: BlueskyMagics.positioners = ['a', 'b']  # test setting
hey

In [27]: BlueskyMagics.positioners  # test access again
hey
Out[27]: ['a', 'b']
```

bluesky/magics.py
danielballan: This needs a trailing space after `.` You can leave out the `+`. See example:

```py
In [29]: ('one line '
    ...:  'another line')
    ...:
Out[29]: 'one line another line'
```

jrmlhermitte:
thanks @danielballan . hmm, there still seems to be a problem:
```
In [29]: get_ipython().register_magics(BlueskyMagics)
    ...: 
/home/lhermitte/research/projects/bluesky/bluesky/magics.py:44: UserWarning: detectors is deprecated. Please use the newer labels feature
  "Please use the newer labels feature")
/home/lhermitte/research/projects/bluesky/bluesky/magics.py:32: UserWarning: positioners is deprecated. Please use the newer labels feature
  "Please use the newer labels feature")
```

I'm not sure how to fix that. maybe add a sentinel we unset after class instantiation? it should be temporary anyway

jrmlhermitte:
continued on #1031 This merge was accidental (merged from the wrong tab)
--

PR #1022, opened 2018-05-04
FIX: LiveDispatcher should work even if we do not see a start document

teddyrendahl:
## Description

Finally got around to playing with the `LiveDispatcher` and noticed this slightly annoying gotcha. If you subscribe a `LiveDispatcher` **after** the original start document, the secondary stream outputs an event document **before** ever outputting a start document. This is certainly confusing to subscribers and we also never perform any of the necessary start-up done in `LiveDispatcher.start`

This quick fix just checks right before we emit our first event that we emitted a start document, if not we force an `self.start()` with an empty dictionary.

### DIFF
--- a/bluesky/callbacks/stream.py
+++ b/bluesky/callbacks/stream.py
@@ -54,7 +54,7 @@ class LiveDispatcher(CallbackBase):
         # of the original start document. Preserve the rest of the metadata
         # that we retrieved from the start document
         md = ChainMap({'uid': self._stream_start_uid,
-                       'original_run_uid': doc['uid'],
+                       'original_run_uid': doc.get('uid', None),
                        'time': ttime.time()},
                       _md, doc)
         # Dispatch the start document for anyone subscribed to our Dispatcher
@@ -115,6 +115,10 @@ class LiveDispatcher(CallbackBase):
         """
         id_args = id_args or (doc['descriptor'],)
         config = config or dict()
+        # If we haven't seen a start document make sure we at least give an
+        # metadata-less start document to our subscribers
+        if not self._stream_start_uid:
+            self.start({})
         # Determine the descriptor id
         desc_id = frozenset((tuple(doc['data'].keys()), stream_name, id_args))
         # If we haven't described this configuration


tacaswell:
I would prefer we drop documents on the floor _until_ we see a start document and then forward the rest on (and the same with events / descriptors).

ZLLentz:
That approach means we can't subscribe any `LiveDispatcher` subclasses during any open run. It seems like an arbitrary restriction.

tacaswell:
Correct, we don't support 'catch up' any where else (and went through this discussion with the 0mq proxy previously).

For callbacks that just need to _see_ a start document, emitting an empty dict is fine, but for callbacks that make use of the information is the start document this approach will confusingly fail down stream.  When I think of the 'stream integrity' of a document stream the content of the documents is at least as important as the order of them!

You can go the other route and cache documents as they go by, but then you have to decide when to drop them (which can get messy) and exposes you to resource issues etc.   At that point we are re-implementing a message bus in python!

danielballan:
Being able to "catch up" is an obviously useful feature, and one that we should establish a recommended path for. I agree with @tacaswell  that providing this robustly requires more infrastructure than we can build here -- a real message bus. Sending an empty dict to downstream consumers breaks too many assumptions that they are otherwise allowed to make (e.g. all start documents have an accurate 'time' and a 'uid'). This change would force all callbacks to account for a new special case.

> You can go the other route and cache documents as they go by, but then you have to decide when to drop them (which can get messy) and exposes you to resource issues etc. At that point we are re-implementing a message bus in python!

@ZLLentz and @teddyrendahl -- If you feel strongly that you need a way to subscribe mid-run, I think a simple caching implementation might be worth having, as long as it were presented with a _huge asterisk_ flagging resource issues and recommending that users graduate to a proper message bus when they do anything at scale.

danielballan:
P.S. Speaking of "establishing a recommended path" for robust message buses integrated with bluesky, see https://github.com/NSLS-II/bluesky/pull/1023.

danielballan:
OK, yes I was misunderstanding. Thanks for clarifying.

I'm still concerned that this could create confusing bugs for callbacks. A callback might reasonably expect to see a `'plan_name'` (an item guaranteed by the RunEngine to be present, but not required by the spec -- an intentional distinction). Recently, it might even be safe to assume `'hints'` if the user is only using standard plans and hints-compliant plans. I would personally rather wait for the next full run that have to worry about special-casing a stub `start` document in my consumers.

Moreover, if we missed the 'start' document, aren't we likely to have missed one or more 'descriptor' documents as well? What happens then?

It's still possible we're talking past each other a little here. Should we set up a quick call to hash this out a bit more?

teddyrendahl:
No, I think we are now hitting the same notes. I wasn't sure how concerned I should be about it so I am glad you guys are inquisitive! In my head there were two standard use cases:

**Basic** -> Created in global context, guaranteed to see all documents.
```python
ld = LiveDispatcher(...)
ld.subscribe(cb)
RE.subscribe(ld)
RE(count(...))
```
**Advanced** -> Created in a plan where all the oddities are self-contained. This LiveDispatcher only lives within the context of this plan so you do not have to worry about shenanigans with arbitrary callbacks getting attached. This plan does not work as written unless ld.start({}) is called manually

```python
def stub_move from_filtered_fit():
       ld = LiveDispatcher(...)  # Use LiveDispatcher to filter data  
       lf = LiveFit(...)   # Fit of cleaned data
       ld.subscribe(lf)
       RE.subscribe(ld)
       for _ in range(...):
            yield from trigger_and_read(...)  # Collect data
       yield from set(motor, lf.model.intercept)   # Interactive decision based on fit    
```


The once case I was really concerned about is lets say you setup one of these LiveDispatchers with a DataBroker, and also store the raw data with a separate DataBroker. In this case if you don't receive that start document you run the risk of storing a bunch of processed data with no way to link it back to raw data runs. This seems very bad. I think we need to think of a better way around this. I can also make the "Advanced" method just call `LiveDispatcher.start()` manually and circumvent this if I am know I am an expert and not breaking anything. I am going to close this PR for now and maybe we can think about a better implementation later. I do have two thoughts I want to pontificate about for a second:

1) When you guys have a chance I think it would be worth checking out the original inspiration for this PR https://github.com/pcdshub/nabs/blob/master/nabs/plan_stubs.py. The idea is basically a `plan_stub` that will take detectors and return the average of them. Super powerful for adaptive plans and only eight lines of real Python code!

2) @danielballan I'm a little concerned about having to have callbacks that depend on `md` dictionaries that are formatted with non EventModel keys. I would argue that any callback that depends on fields like `"hints"` with no contingency for if that is missing is probably in the wrong. I understand that these are included in most default plans but I hope we design at least the built-in `bluesky` callbacks to only use information guaranteed in the `EventModel`.

tacaswell:
> @danielballan I'm a little concerned about having to have callbacks that depend on md dictionaries that are formatted with non EventModel keys. I would argue that any callback that depends on fields like "hints" with no contingency for if that is missing is probably in the wrong. I understand that these are included in most default plans but I hope we design at least the built-in bluesky callbacks to only use information guaranteed in the EventModel.

Part of the idea is to start paving the cow-paths.  We are going to have to (soonish?) start having extensions to the EventModel to specify this sort of thing.  Likely based on what Nexus has already done.  We also need a way to allow variation between BL / domain / facility.

PR #1023, opened 2018-05-08
Kafka

jrmlhermitte:
## Description

Pushing this before I forget. We should add a kafka callback.

## Motivation and Context
Allow for an easy way to send kafka messages.

## How Has This Been Tested?
Tests need to be added. I'm not sure how to go about this. Install kafka on Travis?

For now, this was tested:
```py3
from bluesky.callbacks.kafka import Producer, ConsumerDispatcher
from threading import Thread

from databroker import Broker

db = Broker.named("chx")

docs = db[-2].documents()

prod = Producer(bootstrap_servers=['cmb01:9092'], topic="test")
consumer = ConsumerDispatcher(bootstrap_servers=['cmb01:9092'], topic="test")

def printdoc(name, doc):
    print("Got a document with name {}".format(name))
    print("Contents : {}".format(doc))

consumer.subscribe(printdoc)

print("starting consumer thread")
th = Thread(target=consumer.start)
# so we can quit ipython without waiting on thread
th.daemon=True
th.start()
import time
time.sleep(1)
print('done')


for nds in docs:
    print("sending {} document".format(nds[0]))
    prod(*nds)
```

## More discussion
Should we include the run engine ID and PID in the message passed through the KafkaDispatcher?


bluesky/callbacks/kafka.py
```diff
@@ -0,0 +1,106 @@
...
+import pickle
+
+from bluesky.callbacks.core import CallbackBase
+
+from ..run_engine import Dispatcher, DocumentNames
+
+import time
+
+
+class Producer(CallbackBase):
```
tacaswell: If you are not going to use the simple dispatch method from `CallbackBase` I would not bother inheriting from it.

bluesky/callbacks/kafka.py
```diff
@@ -0,0 +1,106 @@
...
+    Example
+    -------
+    >>> producer = Producer(bootstrap_servers=['localhost:9092'], topic='analysis')
+    """
+    def __init__(self, bootstrap_servers, topic):
+        self._publisher = KafkaProducer(bootstrap_servers=bootstrap_servers)
+        self._topic = topic
+
+    def __call__(self, name, doc):
+        message = b' '.join([name.encode(), pickle.dumps(doc)])
```
tacaswell: How tied are we to pickle here?  Short term this is probably ok, but mid to long term it would be better to use an encoding that can be read from non-python client.

bluesky/callbacks/kafka.py
```diff
@@ -0,0 +1,106 @@
...
+        future = self._publisher.send(self._topic, message)
+        # call this to make sure that this blocks until sent
+        # (or else message may not send)
+        # TODO : handle case when this failed
+        result = future.get(timeout=60)
```
tacaswell: This can block the RE event loop, this should probably be order 1-2 seconds.

bluesky/callbacks/kafka.py
```diff
@@ -0,0 +1,106 @@
...
+    async def _poll(self):
+        while True:
+            #print("getting a consumer message")
+            message = self._consumer.poll()  # 1 sec timeout
+            if len(message):
+                for key, vals in message.items():
+                    if key.topic == self._topic:
+                        for val in vals:
+                            message = val.value
+                            name, doc = message.split(b' ', 2)
```
tacaswell: `partition` may be better here.

`maxsplit` is the maximum number of splits (not number of parts) so this should be 1?

jrmlhermitte:
The only way to test for this code is to run with Kafka as far as I can tell. Should we install kafka on Travis?
Here is the .yml for kafka [here](https://github.com/wvanbergen/kafka/blob/master/.travis.yml).

danielballan:
Hmm. I'm torn on whether to install Kafka. I don't want to grow a ton of heavy dependencies in the test suite. Is this a sign that it's time to start spinning off certain callbacks in special packages (`bluesky-kafka`)?

On one hand, we do already have 0MQ test dependency. On the other hand, 0MQ is a light proof-of-concept for small-scale use and testing, whereas Kafka is a scalable service.

I think I'm talking myself into `bluesky-kafka`, a separate new Python package -- the first of what will probably be several as bluesky grows new integrations. How do others feel about it?

prjemian:
I agree.  Sounds useful, sounds like an option, sounds like a separate
package.

mrakitin:
@danielballan, I like the idea.

CJ-Wright:
I like the idea, I think that many things could be moved to a callbacks specific package.

tacaswell:
Agreed in group meeting, @jrmlhermitte will make a new project for this.

jrmlhermitte:
moved here:
https://github.com/nsls-ii/bluesky-kafka

will add a travis test as  WIP PR

Issue #1027, opened 2018-05-09
python 3.5 segfaulting on travis


tacaswell:
Check the version of dependancies

https://travis-ci.org/NSLS-II/bluesky/builds/368602338

The 3.5 build used to pass, failed when restarted.

## Steps to Reproduce (for bugs)

1.
2.
3.

mrakitin:
Also observed in #1001.

tacaswell:
This seems to have fixed its self.

Issue #538, opened 2016-08-08
LivePlot with nbAgg does not live-update


danielballan:
Leaving this here to lure @tacaswell. The people are [clamoring](https://github.com/NSLS-II/Bug-Reports/issues/131) for it.

danielballan:
I see.

@tacaswell For tutorials it would be useful to have this, irrespective of our position that people should not move motors from notebooks. In advance of the SciPy talk you said you thought you could crack this. Would you reconsider?

PR #1031, opened 2018-05-11
FIX: Follow up to #1020

danielballan:


jrmlhermitte:
This either needs to go in or else we remove the commits added by #538. I wanted to state for the record that I accidentally merged that into bluesky (merged from the wrong tab).

jrmlhermitte:
continued on #1031 This merge was accidental (merged from the wrong tab)

jrmlhermitte:
What needs to be done (a subset):
- add labels to sim devices (@danielballan is working on this)
- remove the ref to `__opyd_labels__` in docs and replace with initialization of a mock device (I can do this)

danielballan:
Labels on sim devices is done and merged via https://github.com/NSLS-II/ophyd/pull/543

danielballan:
One build failed due to #1027, the other passed.

jrmlhermitte:
thanks @danielballan !

danielballan:
Thanks for having the vision to make this better. In the end we didn't take the exact path that you first proposed, but both bluesky and ophyd ended up better because you pushed on the issue.

jrmlhermitte:
This was begun with @mrakitin and I at PDF so thanks @mrakitin as well! :)

Issue #1033, opened 2018-05-14
DOC: update this documentation


prjemian:
https://github.com/NSLS-II/bluesky/blob/680260126bd97d15dcef3340d9ad73d42975bf6a/bluesky/plan_stubs.py#L711

should point to `bluesky.plan_stubs.close_run`

    :func:`bluesky.plan_stubs.close_run`

PR #1034, opened 2018-05-14
Close run docs

awalter-bnl:
updating the doc links for clsoe_run and open_run

## Description
modified the documentation for close_run and open_run.

## Motivation and Context
This fixes issue #1033

### DIFF
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -708,7 +708,7 @@ def open_run(md=None):
 
     See Also
     --------
-    :func:`bluesky.plans.close_run`
+    func:`bluesky.plans_stubs.close_run`
     """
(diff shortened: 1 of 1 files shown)


bluesky/plan_stubs.py
```diff
@@ -708,7 +708,7 @@ def open_run(md=None):
 
     See Also
     --------
-    :func:`bluesky.plans.close_run`
+    func:`bluesky.plans_stubs.close_run`
```
danielballan: A leading colon was lost here. The `:func:` is sphinx code for, "Try to find a function with this name and turn this into a link to its API documentation."

danielballan:
Thanks for the prompt fix @awalter-bnl!

PR #1036, opened 2018-05-19
FIX: Monitor must emit descriptor before event.

danielballan:
Hey, look a RunEngine bug!

In current master, it possible for a monitored signal
to emit an Event document from its subscription thread before
the corresponding EventDescriptor document is emitted by the
RunEngine's thread, thus confusing consumers who see a descriptor ID
referring to an EventDescriptor that they have not yet received.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,9 @@ before_install:
 
 install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
-  - pip install -r requirements.txt
   - pip install --upgrade cython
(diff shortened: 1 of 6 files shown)


danielballan:
To save the next person from trying to reconstruct the intent here:

1. This PR started as a one-line bug-fix.
2. There is also an unrelated fix to make `plan_mutator` keep working on Python 3.7.
3. And some unrelated updates to some tests that became broken by a change in ophyd behavior.

PR #1042, opened 2018-05-29
BLD: Include more files in distribution.

danielballan:
Most importantly, include requirements.txt, which is opened by
setup.py.

In current distribution, installing from a downloaded source tarball
does not work!

h/t @mrakitin for noticing the missing requirements.txt file.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,9 @@
-include versioneer.py
 include bluesky/_version.py
+recursive-include doc/source *
+include doc/Makefile doc/make.bat
+include LICENSE
+include MANIFEST.in
+include README.rst
+include requirements.txt
+include setup.py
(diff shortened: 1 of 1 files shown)


mrakitin:
Shouldn't we also update the `package_data` list? https://github.com/NSLS-II/bluesky/blob/e4793fec81d994ed4ec47574f85519ecdee97cf0/setup.py#L19

Here is [another example](https://github.com/CFN-softbio/pycxdgui/blob/29b3c0e74c09c4f44538c1c33dda5243376285bd/setup.py#L16-L22) we did with @jrmlhermitte.

tacaswell:
`package_data` controls what is installed by `setup.py install` the manifest controls what `setup.py sdist` bundles up.

PR #1044, opened 2018-05-31
FIX: Show-stopping typo in flyer asset collection.

danielballan:
This typo is only hit if using a Flyer than implements `collect_asset_docs()`.

Are there any on the ring other than the one that @mrakitin are setting up with @dhidas
for HXN?


tacaswell:
maybe CSX, but they are pinned to last version for re-buliding-c-things reasons.  HXN has very custom flyers that I did not try to re-write, and I think everyone else is doing event-per-row flyscanning.

PR #1046, opened 2018-06-01
switch to e notation

jrmlhermitte:
Some scientists have mentioned having difficulty reading from devices with small values.
Is it possible to switch to E notation?
This is meant for discussion. What does precision returned in EPICS mean? Is it numbers after leading decimal, or number of significant digits?

tested on:
```py
import bluesky.plans as bp
from bluesky import RunEngine

from ophyd.sim import motor1, motor2

RE = RunEngine()

bec = BestEffortCallback()

RE.subscribe(bec)

RE(bp.count([motor1, motor2]))

```

before:
```
In [1]: %run -i ../pyscripts/0098-test-livetable-array.py
Transient Scan ID: 1     Time: 2018/06/01 11:51:55
Persistent Unique Scan ID: 'b156de18-414c-4a5e-abd6-fe80a0bfc835'
New stream: 'primary'
|   seq_num |       time |     motor1 |     motor2 |
|         1 | 11:51:55.4 |      0.000 |      0.000 |
generator count ['b156de18'] (scan num: 1)


```
after 
```
In [1]: %run -i ../pyscripts/0098-test-livetable-array.py
Transient Scan ID: 1     Time: 2018/06/01 11:49:41
Persistent Unique Scan ID: '87b2c0c2-806c-4b29-a902-92f30e11f1e7'
New stream: 'primary'
|   seq_num |       time |     motor2 |     motor1 |
|         1 | 11:49:41.2 |  0.000e+00 |  0.000e+00 |
generator count ['87b2c0c2'] (scan num: 1)

```

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -191,10 +191,10 @@ class LiveTable(CallbackBase):
         Function to call to 'print' a line.  Defaults to `print`
     '''
     _FMTLOOKUP = {'s': '{pad}{{{k}: >{width}.{prec}{dtype}}}{pad}',
-                  'f': '{pad}{{{k}: >{width}.{prec}{dtype}}}{pad}',
+                  'e': '{pad}{{{k}: >{width}.{prec}{dtype}}}{pad}',
                   'g': '{pad}{{{k}: >{width}.{prec}{dtype}}}{pad}',
                   'd': '{pad}{{{k}: >{width}{dtype}}}{pad}'}
-    _FMT_MAP = {'number': 'f',
+    _FMT_MAP = {'number': 'e',
                 'integer': 'd',
                 'string': 's',
                 }


prjemian:
Q: *"What does precision returned in EPICS mean? Is it numbers after leading decimal, or number of significant digits?"*
A: number of digits to the right of the decimal point (whether %f or %e; good question what happens with %g) seems to be the common use in OPIs

danielballan:
As an aside, it sure is nice to have @prjemian watching bluesky issues these days. There is no substitute for experience.

danielballan:
This is unequivocally an _improvement_ but I'm questioning whether I should include it in today's bug-fix release. You could certainly argue that the current behavior, which often just shows `0` where this PR shows actually useful information, is a bug. On the other hand, I wouldn't be _surprised_ if we get at least one email from someone who finds the new behavior surprising or disruptive to their current usage. I could be wrong... but I'm inclined to save this for 1.4.0 and _not_ push it out in a bug fix 1.3.x release.

Am I being too paranoid? If people disagree with this and argue for releasing it ASAP we'll just put it in the _next_ 1.3.x release.

prjemian:
If you think it unnecessary, delay and review further.  OTOH, display of `0` *v.* useful number sounds more like this is necessary.  One compromise might be `%g` notation which uses fixed notation for mid-range numbers (abs(log10(abs(value))) < 6 or so) and scientific notation for numbers outside this range

```
In [1]: "%g" % 3.14157
Out[1]: '3.14157'

In [2]: "%g" % 3.14157e-17
Out[2]: '3.14157e-17'

In [3]: "%g" % 3.14157e-1
Out[3]: '0.314157'

In [4]: "%g" % 3.14157e17
Out[4]: '3.14157e+17'

In [5]: "%g" % 3.14157e7
Out[5]: '3.14157e+07'

In [6]: "%g" % 3.14157e5
Out[6]: '314157'
```

The table you show, for two motor positions *should* show fixed notation.  No one wants to see `1.72e-9` when the real intent is that the motor is at zero.

Can the display format be specified by the user?  With the default value as would normally be chosen?

awalter-bnl:
I think this is a definite improvement, but also agree that some will want the original format to continue. I like @prjemian suggestion that the display preference be specified by the user.

The question is do we want to do this for individual channels (ideal but time consuming to set up) or a global setting?

Anyway holding this off till we work out how to implement is the best bet in my eyes.

danielballan:
Big :+1: for `%g`. Not exactly sure how best to parameterize the customization because the formatting code is not simple, and injecting user preferences is tricky. Adding _this one_ options isn't bad (`float_format=...`) but I don't like where that trend leads. I'm tempted to just give people `%g` in bluesky v1.3.0 and wait to see if anyone really asks for additional customization. In the meantime, they can always write their own variant of `LiveTable`.

danielballan:
We test LiveTable against expected output, hard-coded in the test suite as a string. That string will need to be updated before this can be merged.

tacaswell:
punting out another release.

I am not 100% convinced that switching to `g` is better.   Many of the cases where I have seen issues on the floor it is because the IOC is reporting a precision of 0 so you would get

```python
In [48]: '{:.0g}'.format(1.23456789/1000000)
Out[48]: '1e-06'

In [49]: '{:.0f}'.format(1.23456789/1000000)
Out[49]: '0'

In [50]: '{:.0e}'.format(1.23456789/1000000)
Out[50]: '1e-06'

In [51]: 
```

prjemian:
I agree.  This is exactly the original point.  The `.PREC` field of the PV should be set properly.  For small numbers (say smaller than 0.001), automatically default to E notation and with some realistic precision if `.PREC` = 0?

danielballan:
I guess minimally we could document this is as a common problem so we have something to link people to?

prjemian:
agreed (both https://github.com/NSLS-II/bluesky/pull/1046#issuecomment-417111181 and https://github.com/NSLS-II/bluesky/pull/1046#issuecomment-420325989)

danielballan:
It looks like the consensus between @prjemian, @tacaswell and myself is that we should document this as a common problem and recommend the solution, "Fix the misconfigured IOCs!"

danielballan:
Opened https://github.com/NSLS-II/bluesky/issues/1115 to remind us to document this. Closing the PR.

Please reopen if I have misinterpreted our consensus.

PR #1047, opened 2018-06-04
FIX: catch tight_layout errors

tacaswell:
If over-constrained tight_layout may raise `ValueError`.  Catch that
and move on assuming that an ugly plot is better than blowing up.

https://gitter.im/NSLS-II/DAMA?at=5b12a1ba4eaffb692d76db00


danielballan:
The 3.6 build passed. Merging.

Issue #713, opened 2017-08-02
Add an est_time plan tool


danielballan:
`eta_time(scan([det], motor, 1, 10, 10))` would be cool. This would be a simulator that uses whatever it can find out about the hardware to estimate how long a plan would take to run and, say, return a number of seconds. It would not touch any hardware.

awalter-bnl:
This is an old one, but want to add my encouragement. This would be really useful for us, it would of course probably assume a fixed time for motor moves which wion't be so accurate but a great start.

danielballan:
Can you help us brainstorm what we would need to know from the hardware, to first order? In my mind, a "zeroth-order solution" would make blunt assumptions like "All motors take 1 second to move, period." I imagine we would need:

* movable hardware to report a an ETA to a given position like `motor.eta(5)` or `temp.eta(273)`
* detectors to consistently exposure their exposure time through a consistently-named attribute, and (better yet) give us some estimate of the overhead time (triggering time - acquiring time)

Maybe we should develop software loops for empirically measuring these things and storing the relevant data in a cached file somewhere.

awalter-bnl:
I agree getting an accurate measure of the time taken for the motors is difficult. For the detectors, almost all have a 'aquisition period' variable and this could be mapped to a consistently named attribute in the setup file, or if the variable is not present then a estimate could be hard-wired.

For the motors, a standard motor record from a delta tau has a 'motor velocity' attribute (in units/s) already, which can be used to give a time estimate for a move of x units. Non-motor record motors may need to introduce a hard-wired estimate like the detectors without aquisition time in the setup.

awalter-bnl:
Just thought I’d add, I am happy to discuss in person if it helps, just throw something on my calender

danielballan:
We can't prioritize this until after our "1.0" deadline (a week from Wednesday) but I think we can draft something usable not long after that.

awalter-bnl:
So I will wait to progress much on this until Dan is back. But this morning I did some testing at SIX in order to see if we could get an accurate eta for a 'move' and for a 'count'. I think it is possible but a little more work needs to be done. 

for 'move':
[...]

PR #1048, opened 2018-06-04
time estimates for plans

awalter-bnl:
This is part of a system that allows for time estimation for plans. It implements a telemetry and time estimation process at the ophyd object level ([NSLS-II/Ophyd PR # 567](https://github.com/NSLS-II/ophyd/pull/567)) to estimate the time to make a 'cmd' on an object. 'cmd' is the commands past in the plan messages.

An est_time('plan') function, similar to summarise_plans('plan'), then steps through each message in order to estimate the time to perform each step and hence the entire plan. 

This fixes issue #713 when combined with [NSLS-II / Ophyd PR # 567](https://github.com/NSLS-II/ophyd/pull/567)

### DIFF
--- a/bluesky/simulators.py
+++ b/bluesky/simulators.py
@@ -1,6 +1,7 @@
 from warnings import warn
 from bluesky.preprocessors import print_summary_wrapper
 from collections import namedtuple
+import inspect
 
 _TimeStats=namedtuple('TimeStats','est_time std_dev')
 _MsgStats=namedtuple('MsgStats', 'msg est_time std_dev')
@@ -362,8 +363,24 @@ class EstTimeSimulator():
                 elif msg.command == 'set':
                     self._plan_history['set'][msg.obj.name] = msg.args[0]
 
-                object_est = obj.est_time(cmd = msg.command, 
-                                                plan_history = self._plan_history, vals = msg.args)
+                #determine what values are required for time estimation and find them
+                obj_est_time = get_attr(obj,'est_time.'+cmd)
+                arg_names = inspect.signature(obj_est_time)
+                args = []
+                for arg_name in arg_names:
+                    try: 
+                        args.append(self._plan_history['set'][arg_name])
+                    except KeyError:
+                        try:
+                            args.append(getattr(obj, arg_name).position)
+                        except AttributeError:
+                            print('{}.est_time.{} requires a {} attribute but none can be \
+                                    found'.format(obj.name, cmd, arg_name) )
+                            raise
+
+                #ask the object for a time estimation. 
+                object_est = obj_est_time(*args)
+
                 return object_est
 
             elif msg.command == self._flyer_start_cmds:
@@ -372,8 +389,24 @@ class EstTimeSimulator():
                 obj = msg.obj
                 out_time = (0,0)
                 for pos in msg.obj._steps:
-                    object_est = obj.est_time(cmd = 'set', plan_history = self._plan_history, 
-                                                    vals = [pos])
+                     #determine what values are required for time estimation and find them
+                    obj_est_time = get_attr(obj,'est_time.'+cmd)
+                    arg_names = inspect.signature(obj_est_time)
+                    args = []
+                    for arg_name in arg_names:
+                        try: 
+                            args.append(self._plan_history['set'][arg_name])
+                        except KeyError:
+                            try:
+                                args.append(getattr(obj, arg_name).position)
+                            except AttributeError:
+                                print('{}.est_time.{} requires a {} attribute but none can be \
+                                    found'.format(obj.name, cmd, arg_name) )
+                            raise
+
+                    #ask the object for a time estimation. 
+                    object_est = obj_est_time(*args)
+
                     out_time = self.combine_est(out_time, object_est)
                     self._plan_history['set'][msg.obj._mot.name] = pos
 
--- a/bluesky/simulators.py
+++ b/bluesky/simulators.py
@@ -4,8 +4,10 @@ from collections import namedtuple, defaultdict
 import inspect
 
 _TimeStats = namedtuple('TimeStats', 'est_time std_dev')
-_MsgStats = namedtuple('MsgStats', 'message stop_time start_time')
-_GroupStats = namedtuple('GroupStats', 'stop_time start_time')
+_CumulativeStats = namedtuple('CumulativeStats',
+                              'message stop_time start_time')
+_MsgStats = namedtuple('MsgStats', 'message process_time action_time')
+_GroupStats = namedtuple('GroupStats', 'delta_time elapsed_time')
 
 
 def plot_raster_path(plan, x_motor, y_motor, ax=None, probe_size=None, lw=2):
@@ -144,26 +146,30 @@ _group_end_cmds = ['wait']
 # The commands that aren't  groupable but still have statistics.
 _timed_cmds = ['stage', 'unstage']
 # The commands that only require network communication.
-_communication_cmds = ['read', 'describe', 'open_run', 'close_run']
-# Define some times associated with network communications and internal
-# state tasks. I am not yet sure how to accurately estimate these times as
-# they are unlikely to be 'static' and measuring them for statistics will
-# take as much time as performing them, they are including in-case a good
-# solution for this is found.
-_communication_time = _TimeStats(0, 0)  # An est_time/std_dev named tuple
-_internal_state_time = _TimeStats(0, 0)  # An est_time/std_dev named tuple
+_communication_cmds = ['read', 'describe']
+# The commands that emit documents to databroker.
+_emit_cmds = ['open_run', 'close_run']
+# Define some times associated with network communications ('communication'),
+# internal processing ('internal') and databroker document emit time ('emit').
+# These are defined in a global dictionary to allow for the status_objects to
+# access them.
+Process_time = {'communication': _TimeStats(0, 0),
+                'internal': _TimeStats(0, 0),
+                'emit': _TimeStats(0, 0)}
 
 
-def est_time(plan):
-    '''The generator function estimates the time to complete a plan with
-    a yield at every message.
+def est_delta_time(plan):
+    '''A generator function that yields an estimate of the time to complete
+    each message in 'plan'.
 
     This method takes in a plan and yields out a named tuple with the structure
-    (msg, estimated stop time, estimated start time). Where msg is the message
-    from the plan and stop/start times are (est_time, std_dev) named tuples
-    which provides the end/start time for each message taking the scan start as
-    zero. Note that for 'groupable' actions (like 'set' and 'trigger') the
-    action msg will start and stop within the group time.
+    (msg, process_time, action_time). Where msg is the message
+    from the plan and process_time/ action_time are (est_time, std_dev) named
+    tuples which provide the time delta to process/perform the message
+    respectively. Note that for actions that do not return a status object the
+    action_time will be 'None' while for 'wait' and 'sleep' messages the
+    action_time will also be 'None' and the 'process_time' will be the wait/
+    sleep time.
 
     Parameters
     ----------
@@ -174,93 +180,104 @@ def est_time(plan):
     -----------------
     out_time : named tuple.
         A named tuple containing 3 items, the message specifed in the plan,
-        and the est_time/ std_dev named tuple for the estimated stop and start
-        times for the message, 'wait' messages return the start and stop times
-        for the entire group.
+        and the est_time/ std_dev named tuple for the process and action
+        times for the message.
     '''
 
     # create a reference to a dictionary called plan_history to capture set
     # changes during the plan for later modification.
     plan_history = {'set': {}, 'trigger': {}}
     # create a dictionary for storing the info regarding groups in progress.
-    groups = defaultdict(_TimeStats())
-
-    # used to hold the cumulative time for the plan
-    out_time = _TimeStats(0, 0)
+    groups = defaultdict(_GroupStats)
 
     for msg in plan:
         # case 1: commands that are 'groupable'
         if msg.command in _group_start_cmds:
             # add set and trigger commands to plan_history
             if msg.command == 'set':
-                plan_history['set'][msg.obj.name] = msg.args[0]
+                plan_history['set'][msg.obj.name] = (msg.args, msg.kwargs)
             elif msg.command == 'trigger':
                 if msg.obj.name in list(plan_history['trigger'].keys()):
                     plan_history['trigger'][msg.obj.name] += 1
                 else:
                     plan_history['trigger'][msg.obj.name] = 1
 
-            message_time = combine_est(
-                            obj_est(msg, plan_history=plan_history),
-                            _communication_time)
-            start_time = out_time
-            stop_time = combine_est(start_time, message_time)
+            process_time = Process_time['communication']
+            action_time = obj_est(msg, plan_history=plan_history)
 
             try:
                 group_time = groups[msg.kwargs.get('group')]
-            except KeyError:
+            except (KeyError, TypeError):
                 group_time = _GroupStats(_TimeStats(0, 0), _TimeStats(0, 0))
 
             # update the times for the group.
+            new_delta_time = combine_est(group_time.elapsed_time, action_time)
             groups[msg.kwargs.get('group')] = _GroupStats(combine_est(
-                                                 group_time.stop_time,
-                                                 stop_time,
+                                                 group_time.delta_time,
+                                                 new_delta_time,
                                                  method='max'),
-                                                 combine_est(
-                                                 group_time.start_time,
-                                                 start_time,
-                                                 method='min'))
+                                                 group_time.elapsed_time)
+
             # yield out the times for the message
-            yield _MsgStats(msg, stop_time, start_time)
+            yield _MsgStats(msg, process_time, action_time)
 
         # case 2: commands that end 'groups'
         elif msg.command in _group_end_cmds:
             group_time = groups.pop(msg.kwargs.get('group'))
 
-            start_time = group_time.start_time
-            out_time = combine_est(out_time, group_time.stop_time)
+            process_time = group_time.delta_time
+            action_time = None
 
-            yield _MsgStats(msg, out_time, start_time)
+            yield _MsgStats(msg, process_time, action_time)
 
         # case 3: commands that have stats but are not 'groupable'
         elif msg.command in _timed_cmds:
-            start_time = out_time
-            out_time = combine_est(out_time,
-                                   obj_est(msg, plan_history=plan_history))
+            process_time = obj_est(msg, plan_history=plan_history)
+            action_time = None
 
-            yield _MsgStats(msg, out_time, start_time)
+            yield _MsgStats(msg, process_time, action_time)
 
         # case 4: commands that have an associated communication time
         elif msg.command in _communication_cmds:
-            start_time = out_time
-            out_time = combine_est(out_time, _communication_time)
+            process_time = Process_time['communication']
+            action_time = None
 
-            yield _MsgStats(msg, out_time, start_time)
+            yield _MsgStats(msg, process_time, action_time)
 
-        # case 5: sleep command
+        # case 5: commands that have an associated databroker emit time
+        elif msg.command in _communication_cmds:
+            process_time = Process_time['emit']
+            action_time = None
+
+            yield _MsgStats(msg, process_time, action_time)
+
+        # case 6: sleep command
         elif msg.command is 'sleep':
-            start_time = out_time
-            out_time = combine_est(out_time, _TimeStats(msg.args[0], 0))
-            out_time = combine_est(out_time, _communication_time)
+            process_time = _TimeStats(msg.args[0], 0)
+            action_time = None
 
-            yield _MsgStats(msg, out_time, start_time)
+            yield _MsgStats(msg, process_time, action_time)
 
-        # case 6: commands that only require internal processing
+        # case 7: commands that only require internal processing
         else:
-            start_time = out_time
-            out_time = combine_est(out_time, _internal_state_time)
+            process_time = Process_time['internal']
+            action_time = None
 
-            yield _MsgStats(msg, out_time, _communication_time)
+            yield _MsgStats(msg, process_time, action_time)
+
+        # before moving to a new msg increase all group elapsed times.
+        for key in groups:
+            if msg.command in _group_end_cmds:
+                extra_time = combine_est(group_time.delta_time,
+                                         group_time.elapsed_time,
+                                         method='subtract')
+            else:
+                extra_time = process_time
+
+            prev_group = groups[key]
+            groups[key] = _GroupStats(prev_group.delta_time,
+                                      combine_est(prev_group.elapsed_time,
+                                                  extra_time))
 
 
 def combine_est(est_time_1, est_time_2, method='sum'):
@@ -382,8 +399,71 @@ def obj_est(msg, plan_history={}):
         object_est = obj_est_time(*args)
 
         return object_est
+
     else:
         return _TimeStats(0, 0)
+
+
+def est_time(plan):
+    '''A generator function that yields a cumulative estimated time for each
+    message in 'plan'.
+
+    This method yields out a named tuple with the structure (message, stop_time
+    , start_time). Where message is the message from the plan and stop/start
+    times are (est_time, std_dev) named tuples which provides the end/start
+    time for each message taking the scan start as zero. 'Wait' messages yield
+    the times for the group as a whole.
+
+    Parameters
+    ----------
+    plan : generator.
+        The bluesky plan that the estimated_time is to be estimated for.
+
+    Returns
+    -----------------
+    out_time : named tuple.
+        A named tuple containing 3 items, the message specifed in the plan and
+        the est_time/ std_dev named tuples for the stop and start times for the
+        message, 'wait' messages yield times for the group they conclude.
+    '''
+
+    # Define the cumulative timeand a group time dictionary for tracking group
+    # start times.
+    cumulative_time = _TimeStats(0, 0)
+    groups = defaultdict(_TimeStats)
+
+    for delta_time in est_delta_time(plan):
+        # if it is a groupable command
+        if delta_time.message.command in _group_start_cmds:
+            # if it is the first command for this group store the start time.
+            if delta_time.message.kwargs.get("group") not in groups.keys():
+                groups[delta_time.message.kwargs.get("group")] =\
+                    cumulative_time
+
+            start_time = cumulative_time
+            cumulative_time = combine_est(cumulative_time,
+                                          delta_time.process_time)
+            stop_time = combine_est(cumulative_time, delta_time.action_time)
+
+            yield _CumulativeStats(delta_time.message, stop_time, start_time)
+
+        # if this is a group ending command.
+        elif delta_time.message.command in _group_end_cmds:
+            start_time = groups.pop(delta_time.message.kwargs.get('group'))
+            stop_time = combine_est(start_time, delta_time.process_time)
+            cumulative_time = combine_est(cumulative_time, stop_time,
+                                          method='max')
+
+            yield _CumulativeStats(delta_time.message, stop_time, start_time)
+
+        # For all other cases.
+        else:
+            start_time = cumulative_time
+            cumulative_time = combine_est(cumulative_time,
+                                          delta_time.process_time)
+
+            yield _CumulativeStats(delta_time.message, cumulative_time,
+                                   start_time)
 
 
 def est_time_per_group(plan):
@@ -444,7 +524,8 @@ def est_time_run(plan):
             yield estimated_time
         elif estimated_time.message.command is 'close_run':
             stop_time = estimated_time.stop_time
-            yield _MsgStats(estimated_time.message, stop_time, start_time)
+            yield _CumulativeStats(estimated_time.message, stop_time,
+                                   start_time)
         else:
             yield estimated_time
 
@@ -475,14 +556,15 @@ def est_time_per_run(plan):
         concludes the run.
     '''
 
+    in_run = False
     for estimated_time in est_time_run(plan):
         if estimated_time.message.command == 'open_run':
-            inner_estimated_time = estimated_time
-            while inner_estimated_time.message.command is not 'close_run':
-                inner_estimated_time = next(est_time_run(plan))
-
-            yield inner_estimated_time
-        else:
+            in_run = True
+            yield estimated_time
+        elif estimated_time.message.command == 'close_run':
+            in_run = False
+            yield estimated_time
+        elif not in_run:
             yield estimated_time
 
 
@@ -508,27 +590,34 @@ def print_est_time(plan, est_time_func=est_time_per_run):
     run_id = -1  # This will not be used once runs can occur in parallel
 
     for estimated_time in est_time_func(plan):
-        time_string = f'   ->  start: \
-                      {estimated_time.start_time.est_time}+/-\
-                      {estimated_time.start_time.std_dev},  stop: \
-                      {estimated_time.stop_time.est_time}+/-\
-                      {estimated_time.stop_time.std_dev}'
+        time_string = f' -> start:'
+        time_string += f'{estimated_time.start_time.est_time}+/-'
+        time_string += f'{ estimated_time.start_time.std_dev}, stop:'
+        time_string += f'{estimated_time.stop_time.est_time}+/-'
+        time_string += f'{ estimated_time.stop_time.std_dev}'
 
-        if estimated_time.command == 'set':
-            print(f'move {estimated_time.message.obj.name} to \
-                  {estimated_time.message.args[0]}' + time_string)
+        if estimated_time.message.command == 'set':
+            print(f'move {estimated_time.message.obj.name:} to ' +
+                  f'{estimated_time.message.args[0]}' + time_string)
 
-        elif estimated_time.command == 'sleep':
+        elif estimated_time.message.command == 'sleep':
             print(f'wait for {estimated_time.message.args[0]}' + time_string)
 
-        elif estimated_time.command == 'close_run':
+        elif estimated_time.message.command == 'open_run':
+            run_id += 1
+            print(f'open_run {run_id}' + time_string)
+
+        elif estimated_time.message.command == 'close_run':
             print(f'close_run {run_id}' + time_string)
 
-        elif estimated_time.command == 'wait':
-            print(f'wait on group \
-                  {estimated_time.message.kwargs.get("group")}' +
+        elif estimated_time.message.command == 'wait':
+            print(f'wait on group ' +
+                  f'{estimated_time.message.kwargs.get("group")}' +
                   time_string)
 
+        elif hasattr(estimated_time.message.obj, 'name'):
+            print(f'{estimated_time.message.command} ' +
+                  f'{estimated_time.message.obj.name}' + time_string)
+
         else:
-            print(f'{estimated_time.message.command} \
-                    {estimated_time.message.obj.name}' + time_string)
+            print(f'{estimated_time.message.command}' + time_string)
--- a/bluesky/simulators.py
+++ b/bluesky/simulators.py
@@ -2,6 +2,7 @@ from warnings import warn
 from bluesky.preprocessors import print_summary_wrapper
 from collections import namedtuple
 import inspect
+import sys
 
 _TimeStats=namedtuple('TimeStats','est_time std_dev')
 _MsgStats=namedtuple('MsgStats', 'msg est_time std_dev')
@@ -344,6 +345,8 @@ class EstTimeSimulator():
             dictionary
         """
 
+        set_dict={}
+
         if msg.obj is not None:
 
             if msg.command is 'sleep':
@@ -361,22 +364,32 @@ class EstTimeSimulator():
                     else:
                         self._plan_history['trigger'][msg.obj.name] = 1
                 elif msg.command == 'set':
+                    try:
+                        set_dict['start_pos'] = self._plan_history['set'][msg.obj.name]
+                    except KeyError:
+                        set_dict['start_pos'] = msg.obj.position
+
+                    set_dict['target'] = msg.args[0]
                     self._plan_history['set'][msg.obj.name] = msg.args[0]
 
                 #determine what values are required for time estimation and find them
-                obj_est_time = get_attr(obj,'est_time.'+cmd)
-                arg_names = inspect.signature(obj_est_time)
+                obj_est_time = getattr(msg.obj.est_time, msg.command)
+                arg_names = inspect.signature(obj_est_time).parameters
                 args = []
                 for arg_name in arg_names:
-                    try: 
-                        args.append(self._plan_history['set'][arg_name])
+                    try:
+                        args.append(set_dict[arg_name])
                     except KeyError:
-                        try:
-                            args.append(getattr(obj, arg_name).position)
-                        except AttributeError:
-                            print('{}.est_time.{} requires a {} attribute but none can be \
-                                    found'.format(obj.name, cmd, arg_name) )
-                            raise
+                        try: 
+                            args.append(self._plan_history['set'][arg_name])
+                        except KeyError:
+                            print (self._plan_history)
+                            try:
+                                args.append(getattr(obj, arg_name).position)
+                            except AttributeError:
+                                print('{}.est_time.{} requires a {} attribute on {} but none can be \
+                                    found'.format(obj.name, msg.command, arg_name, obj) )
+                                raise
 
                 #ask the object for a time estimation. 
                 object_est = obj_est_time(*args)
@@ -390,7 +403,7 @@ class EstTimeSimulator():
                 out_time = (0,0)
                 for pos in msg.obj._steps:
                      #determine what values are required for time estimation and find them
-                    obj_est_time = get_attr(obj,'est_time.'+cmd)
+                    obj_est_time = getattr(obj.est_time,msg.cmd)
                     arg_names = inspect.signature(obj_est_time)
                     args = []
                     for arg_name in arg_names:
@@ -401,7 +414,7 @@ class EstTimeSimulator():
                                 args.append(getattr(obj, arg_name).position)
                             except AttributeError:
                                 print('{}.est_time.{} requires a {} attribute but none can be \
-                                    found'.format(obj.name, cmd, arg_name) )
+                                    found'.format(obj.name, msg.command, arg_name) )
                             raise
 
                     #ask the object for a time estimation. 
--- a/bluesky/simulators.py
+++ b/bluesky/simulators.py
@@ -116,7 +116,7 @@ def check_limits(plan):
                                      "".format(msg.obj.name, pos, (low, high)))
 
 
-def est_time(plan, print_output = True):
+class EstTime():
     """
     Estimates a time for a plan to be completed.
     This function estimates the time it will take (est_time) to complete the plan defined by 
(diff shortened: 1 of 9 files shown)


bluesky/simulators.py
```diff
@@ -114,3 +114,276 @@ def check_limits(plan):
...
+
+        Return Parameters
+        -----------------
+        out_est_time : list.
+            The combined est_time/std_dev tuple.
+        """
+        out_est_time = [0,0]
+        if method == 'sum':
+            out_est_time[0] = est_time_1[0] + est_time_2[0]
+            out_est_time[1] = est_time_1[1] + est_time_2[1]
```
danielballan: Shouldn't the uncertainties be combined in a way that accounts for [propagation of errors](https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulas)?

bluesky/simulators.py
```diff
@@ -114,3 +114,276 @@ def check_limits(plan):
...
+        method : string, optional.
+            The method to use for the combination of est_time_1 and est_time_2, default is 
+            'sum'.
+
+        Return Parameters
+        -----------------
+        out_est_time : list.
+            The combined est_time/std_dev tuple.
+        """
+        out_est_time = [0,0]
```
danielballan: This might be a good place to use a `namedtuple`, which would support both the normal tuple usage (item access by position) _and_ access by name, such as `out_est_time.value` and `out_est_time.uncertainty`.

bluesky/simulators.py
```diff
@@ -114,3 +114,276 @@ def check_limits(plan):
...
+
+            else:
+                out_est_time = est_time_1
+
+        return out_est_time
+
+
+
+    def obj_est_time(msg, val_dict):
+        """
```
danielballan: Functions that mutate their inputs should document prominent that they do so. It can be useful, and seems useful in the case, but it can also be _very surprising_ so it needs to be clearly flagged.

bluesky/simulators.py
```diff
@@ -114,3 +114,276 @@ def check_limits(plan):
...
+                return [msg.args[0], 0], val_dict
+
+            elif msg.command in (_run_start_cmds + _timed_cmds):
+                obj =  msg.obj
+                #The if-elif section below is used to track how many 'triggers' have occured since
+                #the last 'unstage'.
+                if msg.command == 'unstage':
+                    val_dict['trigger'][msg.obj.name] = 0
+                elif msg.command == 'trigger':
+                    if msg.obj.name in list(val_dict['trigger'].keys()):
```
danielballan: Removing `list` here would save from making a needless copy.

bluesky/simulators.py
```diff
@@ -114,3 +114,276 @@ def check_limits(plan):
...
+            The combined est_time/std_dev tuple.
+        """
+        out_est_time = [0,0]
+        if method == 'sum':
+            out_est_time[0] = est_time_1[0] + est_time_2[0]
+            out_est_time[1] = est_time_1[1] + est_time_2[1]
+
+        elif method == 'max':
+            if est_time_1[0] < est_time_2[0]:
+                out_est_time = est_time_2
```
danielballan: Combining uncertainties here is tricky. Maybe a soft maximum function is the right approximation?

bluesky/simulators.py
```diff
@@ -114,3 +114,276 @@ def check_limits(plan):
...
+                out_est_time = combine_est_time(out_est_time, grp_est_time)
+
+            else:
+                object_est_time, val_dict = obj_est_time(msg, val_dict)
+                out_est_time = combine_est_time(out_est_time, object_est_time)
+        return out_est_time, val_dict
+
+
+    #Define some variables used in the following.
+    val_dict = {'set':{}, 'trigger':{} } #this holds information on the updated values.
```
danielballan: Now for a signature naming-related nitpick! :-D

I don't find `val_dict` particularly descriptive. I usually see variables with the kind role with `state` in the name, which also hints that this is is something we are going to mutating a lot. Maybe just name it `state`?

prjemian: `state` means *so* many different things and is likely to already have implementation in cases where time estimates would be a valuable add-on.  At a loss how to make this name more specific (and less likely to conflict)-- ideas?

bluesky/simulators.py
prjemian: Isn't this identical:

```
if msg.obj.name in val_dict['trigger']:```

File: bluesky/simulators.py (hunk 1 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: These are not 'errors' but standard deviations. There are many ways to combine errors, this is just the simplest approach which works on a very general level. I am happy to discuss other approaches.

File: bluesky/simulators.py (hunk 2 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: I will look into this, I wasn't aware of 'namedtuples'.

File: bluesky/simulators.py (hunk 3 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: I thought I had clearly flagged this, I will make it clearer in the doc string.

File: bluesky/simulators.py (hunk 4 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: OK will do, for some reason I thought that this wouldn't work.

File: bluesky/simulators.py (hunk 5 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: Here it is not so tricky, because we are assuming that the time taken is that of the larger of the two. Therefore the error should also be the error associated with the larger of the two.

File: bluesky/simulators.py (hunk 6 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: I too am not so happy with val_dict, I will think about what else it can be but don't like state for the same reasons as @prjemian

Comment (prjemian) [reply bluesky/simulators.py]: `msg_info` ?  Seems concise, descriptive for its context, and not likely in use

**Comment by awalter-bnl** [comment]
@danielballan, relating to your big picture thoughts. 

1. I am happy to move the smaller function from the larger one, however I wanted to avoid polluting the namespace with a bunch of smaller functions. What do you think of _function to prevent this instead?

2. I agree we should look into this, but leave it for a future PR.

3. Each ophyd obj now already has an attribute function that provides this tuple for a given command, so I don't see the benefit of this approach. The 'granularity' you are requesting will then be provided by all the sub-functions that I will pull out.

File: bluesky/simulators.py (hunk 1 above)
Comment (danielballan) [reply bluesky/simulators.py]: Yes, I mean “error” as in “standard error” — standard deviation. Possible clash of continental terms?

I think that adding them in quadrature is understood to be the best general method when the sources of deviation are many and independently distributed (central limit theorem).

File: bluesky/simulators.py (hunk 6 above)
Comment (tacaswell) [reply bluesky/simulators.py]: This is pretty well scoped (it is not actually publicly exposed at all right now!) so I am not worried about collisions with other uses of `state`.  I like `state` quite a bit here, but I think that this is a candidate for a class (like the RE) to handle this via the visitor pattern (ex https://en.wikipedia.org/wiki/Visitor_pattern ).

Comment (prjemian) [reply bluesky/simulators.py]: ok with naming.  Agree there is more complexity here; a class is warranted.

File: bluesky/simulators.py (hunk 5 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: OK so I hadn't considered that in some cases the standard deviation could increase the time of one above the maximum time. This seems easy enough to work around, I will make this change too.

File: bluesky/simulators.py (hunk 6 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: I quite like msg_info, but maybe plan_info is better as it is keeping track of what items have been 'modified' in the plan up to this point. As I mentioned above I am not exactly sure how this would work as a class, but will think about it.

File: bluesky/simulators.py (hunk 3 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: Not sure exactly how to do this as a class, but I will think about it. Is the idea that an instance of the 'class' will have the dictionary as an attribute?

File: bluesky/simulators.py (hunk 6 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: To try to clarify the use of val_dict, essentially to know how long it takes to 'move' a device it is necessary to know the starting point. As this can be modified by an earlier message in the plan when stepping through messages without performing the action I need to keep track of any previous changes to this that would have been caused by the plan itself. There are also other changes (settle_time etc) that may be 'changed' during a plan, and can change the time estimate, that need to be tracked.

Comment (prjemian) [reply bluesky/simulators.py]: Complex numbers are probably the best example I can think of.  Special math operations for them.  This `msg_info` (or is it now `plan_history`) class could combine the special operations with the historical info such as a database.  Not sure I well understand what is required.  Assume the doc strings will describe in some detail.

File: bluesky/simulators.py (hunk 1 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: OK will look into this.

**Comment by awalter-bnl** [comment]
No Problem, on any of these points. I will make the changes. For number 3 I will think about the best way to approach this. Due to the complications of 'groups' which are performed in parallel I am not sure it is a simple as you predict. If you send in a series of msgs with a 'group' in the middle do you want it to return the time for each msg or 1 for the group and 1 for each msg inside the group? I am guessing the later but am not totally sure how you plan to arrange the progress bar for groups.

File: bluesky/simulators.py (hunk 6 above)
Comment (awalter-bnl) [reply bluesky/simulators.py]: I thin the docstrings already describe pretty well, if not let me know and I will update them. Essentially, as an example, think of an 'area detector' detector. To a first approximation the est_time for a triggering can be written as est_time = settle_time + num_images * acquire_time.

In a plan we could easily change the acquire_time part of the way through the plan. If this happens then we would need this 'updated' value in order to estimate the time for any following triggers of this detector in the plan. As the generator function does not actually perform this change, it needs to be 'tracked' which is what I use val_dict(name change pending) for.

Similar things occur for other actions (set for example by definition changes values). Hopefully this is a little clearer. I think I have a better way to do this using classes, I will write it up and see how it goes.

**Comment by awalter-bnl** [comment]
I have pushed a number of commits related to the issues described above. I think I have got all of the requests, with the exception of the move to put the telemetry recording on status objects. This is because I am having some issues understanding how status objects work from the code, I do intend to make this change once that is clearer to me however. Notes on the changes are below, each numbered item should have a corresponding commit for ease of following.

1. I turned the simulator est_time into a class 'EstTimeSimulator' and an instantiation 'est_time' because:
     - This allows easy access to the sub functions as requested by @danielballan 
     - This allows for a class definition of val_dict, which no longer has to be passed in and out of sub  
     classes, as requested by @danielballan  and @tacaswell  
2. Modifed EstTimeSimulator (bluesky), sim.py (ophyd), device.py (ophyd) and  EstTime (ophyd) to change how val_dict is used and its name.
    - in EstTimeSimulator (bluesky) it is a class attribtue self._plan_history.
    -  This means it is not longer passed into and returned from functions as requested by  
    @danielballan  and @tacaswell  and has a new name as requested by @danielballan and @prjemian.
    - in EstTime (ophyd), device.py (ophyd), sim.py (ophyd) it is an input only variable called 
    plan_history.
3. Changed the stats and est_time to namedtuples
    - Also updated some of the names for these to make it clearer.
4. remove telemetry and est_time on read.
    - At the request of @danielballan   
5. changed the method for combining std_dev's.
    - At  the request of @danielballan   
6. Added a new function obj.est_time.msg_est(plan)
    - This was done at the request of @danielballan  to help with a plan progress bar.
    - Contrary to @danielballan's request I have not made this a 'base' function of the main 
    obj.est_time(plan) simulator function as I think this would add a second iteration through 
    the list of messages which will add an undue processing time burden on obj.est_time(plan).

**Comment by danielballan** [comment]
Description looks good. Will dig into code as soon as I can. Happy to help walk through the Status code (and document it better in the process!)

**Comment by awalter-bnl** [comment]
Thanks @Daniel Allan , I will take you up on the offer of a discussion of status objects. Perhaps next week when you aren’t in Chicago.

**Comment by danielballan** [comment]
> For number 3 I will think about the best way to approach this. Due to the complications of 'groups' which are performed in parallel I am not sure it is a simple as you predict. If you send in a series of msgs with a 'group' in the middle do you want it to return the time for each msg or 1 for the group and 1 for each msg inside the group? I am guessing the later but am not totally sure how you plan to arrange the progress bar for groups.

I think that it should work like this:

```
# msg ->  est_time
Msg('set', motor1, ..., group='A')  -> 0
Msg('set', motor2, ..., group='A')  -> 0
Msg('set', motor3, ..., group='B')  -> 0
Msg('wait', 'A')  -> max(motor1.est_time.set(...), motor2.est_time.set(...))
Msg('wait', 'B')  -> motor3.est_time.set(...)
```

This is an accurate representation of how long each message will take to process: 'set' messages are instantaneous for RE; only the 'wait' messages take time.

**Comment by awalter-bnl** [comment]
Thanks @d
at present I do the former not the later, very similar to what you suggest, without separating by 'groups'. I will work on changing this.

**Comment by awalter-bnl** [comment]
I have updated this so that the est_time.group_eta function can handle multiple parallel 'groups'.

**Comment by danielballan** [comment]
This class has ["two methods one of which is `__init__`"](https://www.youtube.com/watch?v=o9pEzgHorH0) -- plus some utility methods that are tucked into the class for scoping / namespace-decluttering reasons. Let's refactor it as a _function_ plus utility functions. The scoping / namespace-decluttering is already provided by the module itself:

```py
from bluesky import est_time
est_time.some_function(...)
```

awalter-bnl:
Have made the change requested by @danielballan (removing EstTimeSimulator class) and tested.

Sorry for forgetting we had discussed this earlier, for posterity I am still not sure why this is better than the class structure. However as it doesn't prevent any of the aims of this PR being realized I ahve made the change.

bluesky/simulators.py
```diff
@@ -114,3 +121,401 @@ def check_limits(plan):
...
+_plan_history: dict.
+    A dictionary containing information on values updated during the plan. It has key:arg 
+    pairs with keys relating to message components where each arg is a dictionary 
+    containing devices (detectors, axes) that have been updated, as keywords and a value 
+    indicating the update status. For 'set' the update status is the latest 'position' that
+    it is to be set and for 'trigger' the update status is the number of times since the 
+    beginning of the scan or since an 'unstage' event that the device has been triggered.
+
+"""
+#The commands that start/end a 'run'.
```
danielballan: Most -- maybe all -- of these lists strike me as unneeded obfuscation. There is only one command that starts are run, `'open_run'`, and it is not likely to change in the future. Why not just type `'open_run'` directly where it's needed?

It occurs to me you might intend this as path to registering custom user commands (message types) with this system. User commands are somewhat limited: a user can't add a comment that interacts with private RunEngine state, including opening/closing runs or doing anything with groups. It is safe to hard-code these command names, for the foreseeable future.

bluesky/simulators.py
```diff
@@ -114,3 +121,401 @@ def check_limits(plan):
...
+#The commands that start/end a 'group', but not flyer commands.
+_group_start_cmds = ['set','trigger']   
+_group_end_cmds = ['wait']
+#The commands that don't start/end a 'group/plan' but are still timed.
+_timed_cmds = ['stage', 'unstage', 'sleep']
+#Commands that are 
+_flyer_start_cmds = ['kickoff']
+#create a reference to a dictionary called plan_history to capture set changes during the 
+#plan for later modification.
+_plan_history = {'set':{}, 'trigger':{}}
```
danielballan: This shouldn't be a module global; it should be defined internally to `est_time` and passed explicitly into `obj_est`, `group_est`, etc. A mutable global creates the possibility for unintended interactions. Can be useful, but no reason to open Pandora's box in this case.

bluesky/simulators.py
```diff
@@ -114,3 +121,401 @@ def check_limits(plan):
...
+    plan : generator.
+        The bluesky plan that the est_time is to be estimated for.
+
+    RETURN PARAMETERS
+    -----------------
+    out_time : list.
+        A list containing a msg/est_time/std_dev named tuple for each message in the plan.
+    '''
+    
+    out_time = []
```
danielballan: Rather than accumulating the messages in a list and then returning the list, this can be implemented as generator, yielding out one estimate at a time. This makes it more flexible. If the caller _wants_ a list, then can of course call `list(msg_est(plan))`.

bluesky/simulators.py
```diff
@@ -114,3 +121,401 @@ def check_limits(plan):
...
+        The first message of the group that was detected in plan_ETA.
+    
+    Return Parameters
+    -----------------
+    out_time : list.
+        The combined est_time/std_dev pair.
+
+    Updated Parameters
+    ------------------
+    _plan_history : dict.
```
danielballan: This should be promoted to a parameter and passed in by all callers.

bluesky/simulators.py
```diff
@@ -114,3 +121,401 @@ def check_limits(plan):
...
+#The commands that don't start/end a 'group/plan' but are still timed.
+_timed_cmds = ['stage', 'unstage', 'sleep']
+#Commands that are 
+_flyer_start_cmds = ['kickoff']
+#create a reference to a dictionary called plan_history to capture set changes during the 
+#plan for later modification.
+_plan_history = {'set':{}, 'trigger':{}}
+
+def est_time(plan, print_output = True):
+    '''The call method.
```
danielballan: Remove 'the call method' now that this is a function.

bluesky/simulators.py
```diff
@@ -114,3 +121,401 @@ def check_limits(plan):
...
+
+    This method takes in a plan, and an optional print_output kwarg, and returns an estimate for 
+    the time to complete for the plan, in addition it returns an time estimate for any subplans. 
+
+    PARAMETERS
```
danielballan: Sphinx expects the capitalization `Parameters`.

bluesky/simulators.py
```diff
@@ -114,3 +121,401 @@ def check_limits(plan):
...
+    ----------
+    plan : generator.
+        The bluesky plan that the est_time is to be estimated for.
+    print_output : boolean, optional.
+        Indicates if the return values should also be printed to the command line in a human 
+        readable way, default value is 'True'.
+
+    RETURN PARAMETERS
```
danielballan: Sphinx expects this to be spelled `Returns`

bluesky/simulators.py
```diff
@@ -114,3 +121,401 @@ def check_limits(plan):
...
+
+        if msg.command is 'sleep':
+            return [msg.args[0], 0]
+
+        elif msg.command in (_run_start_cmds + _group_start_cmds + _timed_cmds):
+            obj =  msg.obj
+            #The if-elif section below is used to track how many 'triggers'/'sets' have 
+            #occured since the last 'unstage'/start of plan.
+            if msg.command == 'unstage':
+                _plan_history['trigger'][msg.obj.name] = 0
```
danielballan: `del _plan_history['trigger'][msg.obj.name]` might be a clearer way of indicating "not applicable" than `0`.

bluesky/simulators.py
```diff
@@ -114,3 +121,401 @@ def check_limits(plan):
...
+            out_time = combine_est(out_time, rn_time)
+        if msg.command in (_group_start_cmds + _flyer_start_cmds):
+            grp_time = group_est(msg, plan)
+            out_time = combine_est(out_time, grp_time)            
+
+        else:
+            object_est = obj_est(msg)
+            out_time = combine_est(out_time, object_est)
+
+        if print_output == True:
```
danielballan: `if print_output` is sufficient.

bluesky/simulators.py
awalter-bnl: I am not convinced that hard-coding these is a good idea. My strong preference is to leave it as is in case things change in the future to make it easier to change.

bluesky/simulators.py
awalter-bnl: This was done explicitly because @tacaswell was very against me passing a variable in to a function that was modified in the function and the returned. So a decision needs to be made between @tacaswell and @danielballan about which approach to use as the requests are mutually exclusive.

bluesky/simulators.py
awalter-bnl: As mentioned above, this was done to avoid having the functions take this in as an argument and return a possibly modified version.

bluesky/simulators.py
awalter-bnl: This is set as zero as the 'value' is the number of triggers since the last 'unstage' not just an indication if it is staged. I did it this way in preparation for the custom devices like at CSX where some staging is done in the first trigger.

bluesky/simulators.py
awalter-bnl: fixed

bluesky/simulators.py
awalter-bnl: fixed

bluesky/simulators.py
awalter-bnl: fixed

bluesky/simulators.py
awalter-bnl: fixed

awalter-bnl:
I disagree @danielballan , they way it is written it absolutely takes into account sets/groups that occur outside of runs. I have chosen to also provide the time for the 'runs', as a separate list to the total time, to give the users a better understanding of which part of the plan might be taking to long and to help them organise the plans easier as the 'macro' type plans that most users will write will contain several 'runs' as their basic building block. I have treated 'runs' in the same way as 'groups', in that the times are separated out and added to the total time for the plan once the end of the 'run'/'group' has been found.

To clarify the total 'plan' time is not just the sum of the times for each 'run' but the sum of everything that occurs inside 'runs' AND everything that occurs between 'runs' and so can be longer than the sum of all 'runs'. I have tested the code to make sure this works by adding 'yield from mv....' and 'yield from wait..' lines between 'yield from scan......' lines to a test plan I was using. 

I have specifically NOT written the code so that it first runs 'msg_est' and then go back through this and sort into 'runs', 'groups' etc. because this means counting through the messages more than once, and hence doubles the time overhead of calling est_time.

danielballan:
> I have specifically NOT written the code so that it first runs 'msg_est' and then go back through this and sort into 'runs', 'groups' etc. because this means counting through the messages more than once, and hence doubles the time overhead of calling est_time.

Agreed that needing to iterate through the plan twice would be a bad design. I think this can be done in a single pass. Let me sketch out a little more specifically the design I had in mind with [this comment](https://github.com/NSLS-II/bluesky/pull/1048#issuecomment-396098919).

```py
# These constants should be revised using the test stand.
COMMUNICATION_TIME = (0.1, 0.1)  # It takes 0.1 +/- 0.1 to send a command to the IOC.
INTERNAL_STATE_TIME = (3e-5, 3e-5)  # It takes ~3e-5 seconds to process msgs like 'checkpoint'.

def est_time(plan):
    "Takes in a plan and yields out (time, std_err) per message."
    history = {}
    groups = defaultdict(list)
    for msg in plan:
        command = msg.command
        # Case 1: commands that generate a status object
        if command in ('set', 'trigger', 'kickoff', 'complete'):
            # TODO something involving history here...
            t = getattr(msg.obj.est_time, command)(...)
            groups[msg.kwargs.get('group')].append(t)
            yield COMMUNICATION TIME
        # Case 2: waiting on a status object cached by Case 1
        elif msg.command == 'wait':
            times = groups.pop(msg.kwargs.get('group'))
            t, std_err = combine(times, 'max')
            # TODO Update other groups (subtract t).
            yield t, std_err
        # Case 3: commands that communicate with the IOC but are quick
        elif msg.command in ('read', 'describe'):
            yield COMMUNICATION_TIME
        # Case 4: TODO 'collect' -- should scale with number of data points
        # Case 5: temporary special-case: these _should_ return a status object but don't yet
        elif command in ('stage', 'unstage'):
            yield (0, 0)  # temporary, unrealistic
        # Case 6: all other commands that do not involve network communication, are fast
        else:
            yield INTERNAL_STATE_TIME

def est_time_per_run(plan):
    "Takes a plan and yields out total (time, std_err) per run."
    times = []  # Accumulate times in current run.
    in_run = False  # Tracks whether we are currently inside a run.
    a, b = itertools.tee(plan)
    for msg, est in zip(a, est_time(b)):
        if msg == 'open_run':
            in_run = True
        if msg == 'close_run':
            in_run = False
            yield combine(times, 'max')
        if in_run:
            times.append(est)
```

Things I like about this general design:

* We obviate the discussion cited above about a global dict passing around mutable state. The state lives as a local var inside `est_time`. It is not shared between functions.
* We can imagine several reductions/aggregations following the pattern of `est_time_per_run`. Someone could write `est_time_total`, `est_time_per_sample` or `est_time_per_point` using the same pattern, and all the "tricky" logic is safely contained in `est_time`.

danielballan:
One clarification that occurred to me I should state: Note that because `est_time` is a _generator_, we aren't looping through the messages twice. We're capturing output _as it comes out_ and tallying. It's layered, but it's not doubling the loop.

awalter-bnl:
OK so I am slowly becoming convinced, I see two issues to be resolved first however:

ISSUE 1.
It would need another function on top of this to return the times in a useful way for users (e.g. a 'print_summary' style wrapper for printing or some other method for exporting the data to a GUI). 
- This may not be a major issue but want to get feedback on what people think regarding this as the API. 
- What do people think of a single API with boolean keywords used to determine what is printed?
    - eg. est_time_plan(plan, total = true, runs = true, .......)
    **NOTE: I am using the 'summarize_plan' naming convention here, what do people think about this?

ISSUE 2
I think that 95% of the time the users will want both the total estimated time AND the time per run. In your setup this would still require stepping through the message list twice, once when running est_time_per_run(plan) and once when running est_time_total(plan). 
- I see two options to get around this:
    - both would require the 'yields' to give 3 values 'name', est_time and std_dev with 'name' being
    something like 'msg x', 'group y' or 'run z' which can be used in the 'print' wrapper described above.

    1. Make est_time_per_run also yield the time estimate for msg's between the runs, including the 
    'name' parameter described above. Then we can simply pass est_time_per_run(est_time(plan)) as 
    the argument of est_time_total to output both. This relies on the fact that both est_time and 
    est_time_per_run are generators with the same structure for the output and so either can be the
    input of est_time_total. The 'name' can then be used to distinguish what type of 'thing' the time 
    estimate relates too. 

     2. we could make est_time_per_run(plan) also return the total time for the plan as the last 'yield'. If 
    we gave it the name 'total' we could distinguish it from those with the name 'run'. 

FURTHER THOUGHTS
In principal I like number 2 as it is explicit, but not strongly enough to insist. Number 1 has the benefit that it is easy to nest any future reductions in which could be useful.

I think that adding the 'name' as part of each yield would make a generic 'print' wrapper easy (the first issue I mentioned). It would just need to print the following on each yield:
`print (f.'{name}: est time = {est_time}, std_dev = {std_dev}')`

danielballan:
> What do people think of a single API with boolean keywords used to determine what is printed?
eg. `est_time_plan(plan, total = true, runs = true, .......)`
**NOTE: I am using the 'summarize_plan' naming convention here, what do people think about this?

The job of `summarize_plan` is to remove information. That is, when`list(plan)` is too much information, `summarize_plan` is there to construct an incomplete but often sufficient and nicer-to-read representation. It does not add any new information, so there is not much need to access its output programmatically -- might as well just dump to the screen.

The job of `est_time`, in contrast, is to provide additional information. That information should be programmatically accessible so that it can be reused by other components (GUIs, user-defined functions, etc).

This usage seems user-friendly to me, and unlike prints it is also programmatically useful:

```py
>>> est_time_per_run(plan)
[EstTime(time=10, std_err=1), EstTime(time=5, std_err=0.3), EstTime(time=10, std_err=1)]
```

If I want the total I can do this:

```py
>>> est_time_total(plan)
EstTime(time=20, std_err=2.3)
```

or this:

```py
>>> sum([item.item for item in est_time_per_run(plan)])
20
```

or this:

```py
>>> combine(est_time_per_run(plan))
EstTime(time=20, std_err=2.3)
```

As an additional convenience, I would not object to some high-level function that prints something like:

```
Total Estimated time: 20 +/ 2.3
Per run breakdown:
(etc.)
```

> I think that 95% of the time the users will want both the total estimated time AND the time per run. In your setup this would still require stepping through the message list twice, once when running est_time_per_run(plan) and once when running est_time_total(plan).

I think you are right that that is the common case and that we need a robust solution for it that ideally does not require iterating through the messages twice.

It just occurred to me that, when we extend the RunEngine to make it possible to have multiple runs open concurrently as we plan to do someday, the relationship between total time and time during and between runs will not be straightforward. Runs can overlap. We should try to avoid making the 'one run at a time' assumption in here to avoid future pain. Nonetheless the time 'per run' will remain a useful concept for the common case of non-overlapping runs.

Here's a proposal that solves your problem ("Users commonly want time per run and total time") and the problem I foresee ("Time per run is ill-defined.") I think `est_time_per_run` -- or whatever we call it -- should report the _cumulative_ `EstTime` each time a run is closed. This is well-defined even if runs overlap, it incorporates any time spent between runs, and tells the user what they probably want to know: "How long until I can see all of dataset X?" If we want to include the time spent after the last run closes, we can issue one last message at the end.

awalter-bnl:
So I want to push back a little here, and it is something that I was also considering for the group_est time as well as the potentially overlapping 'runs'.

I think the idea of a 'cumulative est_time' is a really good idea, but I think it would be beneficial if we extended the output tuple to include it AND the time for the run/group defined by 'name'. This would result in a named tuple with the values ('name', 'cumulative estimate', 'section?? estimate'). In this case I imagine that 'cumulative estimate' and 'section?? estimate' are themselves named tuples with the values ('est_time', 'std_dev'). (I am not convinced of the name 'section' to relate to the 'group', 'run' or 'non-groupable action' that is referenced by name but don't have a better option). 

The reason I like this is, as you say, cumulative est_time outputs when a particular part of the scan is done and by default provides the total est_time at the end. However there is still some information that is lost. By also including the total time for a 'group'/'run' or 'non-groupable action' then we are not losing any time information.

As a use case I could imagine the extra info of the total 'plan'/ 'group' or 'non-groupable action' times would be useful for a user trying to determine which part of a 'macro-like plan' is taking the longest and where the best point to change things to reduce the time is. I could also imagine using this extra info to make a 'visual representation', similar to a schedule bar-graph, showing when different 'sections' start and stop throughout the plan which would be a neat and useful GUI addition that can help with scheduling 'macro-like plans' with many runs inside.

What are everyones thoughts?

danielballan:
> I think the idea of a 'cumulative est_time' is a really good idea, but I think it would be beneficial if we extended the output tuple to include it AND the time for the run/group defined by 'name'. 

Can you give an example of what this would look like? I'm not clear on where the 'name' come from.

awalter-bnl:
The 'name' is really just to distinguish between like actions/groups/runs. For example a run might be
for a run:
`('run_1', (cumulative_est_time, cumulative_std_dev), (section_est_time, section_std_dev))`
for a group:
`('group_'+msg.group, (cumulative_est_time, cumulative_std_dev), (section_est_time, section_std_dev))`
for a non-groupable action:
`(msg.cmd+'_1', (cumulative_est_time, cumulative_std_dev), (section_est_time, section_std_dev))`

The aim of this is to produce an output something like:
`$ list[est_time_per_run(plan)]
[('group_A', ........), ('run_1', .......), ('sleep_1', ........), ('group_B', .........), ('run_2',........), .................`

Does this make it a little clearer?

danielballan:
I think we agree on:

* We need the low-level generator (along the lines of my `est_time` sketch) to provide maximally-granular information -- time per `Msg` -- such as would be suitable to support a progress bar.
* We need a high-level generator (cumulative, perhaps) to provide a reduced/simplified view for human consumption.

I think you are proposal an intermediate variant, which includes a level of detail somewhere between these two. I think at this point we should consider moving to a visual representation rather than a text-based one -- as you proposed above. A intermediate text-based representation looks like it will be quite verbose and hard to parse for realistic plans. But a visualization, whether a simple static plot or an interactive GUI, can be derived from the low-level generator, and this lets us avoid trying to mind-read which things the user doesn't want to see ("details") and which things are important for them to see that day. It also avoids the need to try to infer "sections" from what is technically a flat list.

danielballan:
I do think it's useful to consider other reduced representations. I mentioned `est_time_per_point` above -- having in mind the time per Event.

danielballan:
Another thought, lifted from early conversations on this topic -- Should we add `Msg('section', ...)` which plans can use to declare sections useful for progress bars and other consumers? One path might be to say, "This is a thing we know can do and probably want to do someday," but leave it out of scope for this PR.

Apologies for the stream-of-consciousness posts.

awalter-bnl:
> But a visualization, whether a simple static plot or an interactive GUI, can be derived from the low-level generator

I am not sure this is true, as I can see that it would also want to include 'runs', or other possible 'arrangments' and so to properly produce the visualization I mentioned above we would need to run through the messages twice, once in `est_time` and once in `est_time_per_run`. As this information already exists from the `est_time` call inside `est_time_per_run` then I am thinking we should output it as well. I think this approach makes the visualization much easier, as one call gives all of the required information in a list where the positions corresponds to the completion order.

> A intermediate text-based representation looks like it will be quite verbose and hard to parse for realistic plans

I don't think I agree with this, I see the parsing to be quite simple. For example to parse out the cumulative estimate from the yielded value, `val`, from `est_time_per_run` I would use:
`val.cumulative.est_time or val.cumulative.std_dev`
`val.name` would then be an identifying string that can be used in visualizations/ prints and `val.section.est_time/val.section.std_dev` could be used to get the total time for the visualizations/ prints or for further processing.

To be clear, I don't think we should be providing 'every' group and action time in `est_time_per_run`, just those that occur between runs. Everything inside the runs will be contained within the 'run' time, as is proposed to be done for groups in `est_time`.  I think that this resolves the issue of the verbosity of this 'intermediate' approach. We could also limit it to 'actions' that we know take significant time and leave out those that are essentially instantaneous.

> I do think it's useful to consider other reduced representations. I mentioned `est_time_per_point ` above -- having in mind the time per Event.

As an example I would think that this kind of plan would provide the above mentioned tuples for each Event and for each 'action' that occurs between events, following the same pattern as for my description of `est_time_per_run`.

danielballan:
> I don't think I agree with this, I see the parsing to be quite simple. For example to parse out the cumulative estimate from the yielded value, val, from est_time_per_run I would use:
val.cumulative.est_time or val.cumulative.std_dev [...]

Oh, I didn't mean hard to parse programmatically, I meant hard to visually read on the screen.

> To be clear, I don't think we should be providing 'every' group and action time in est_time_per_run, just those that occur between runs. Everything inside the runs will be contained within the 'run' time, as is proposed to be done for groups in est_time. I think that this resolves the issue of the verbosity of this 'intermediate' approach. 

I strongly dislike the shift in granularity depending on whether we are in a run.

...But I actually think we don't need to settle those discussions in order to proceed.

In order to un-block progress here, how about this: can you get to an implementation of something like [`est_time` in this comment](https://github.com/NSLS-II/bluesky/pull/1048#issuecomment-400036029) but with two revisions derived from our subsequent conversation:

1. The output `EstTime` should always be in cumulative terms.
2. Yield out `msg, est_time_tuple` instead of just `est_time_tuple`.

That would give us all the information, unfiltered, and enable us to write many variations with just one pass through he plan. Observe how `est_time_per_run` becomes simpler to do:

```py
def est_time_per_run(plan):
    "Takes a plan and yields out total (time, std_err) per run."
    for msg, est in est_time(plan):
        if msg.command == 'close_run':
            yield est
```

Your preferred variation is also easy to implement:

```py
def est_time_per_run(plan):
    "Takes a plan and yields out (label, (time, std_err)) per run."
    runs_open = 0
    group_counter = 0
    run_counter = 0
    for msg, est in est_time(b):
        if msg.command == 'open_run':
            runs_open +=1
        elif msg.command == 'close_run':
            runs_open -= 1
            run_counter += 1
            yield f'run_{run_counter}', est
        elif not runs_open and msg.command == 'wait':
            group_counter += 1
            yield f'group_{group_counter}', est
```


Are you satisfied that that implementation for `est_time` is sufficient to support the high-level APIs we need? If so, should we table the discussion of which high-level APIs are useful until we have the low-level `est_time` fully sorted out?

bluesky/tests/test_est_time.py
```diff
@@ -0,0 +1,105 @@
...
+    assert len(TelemetryUI.telemetry) == length + 1 #check an entry has been added to the dictionary
+
+    length = len(TelemetryUI.telemetry) #find the current length of the telemetry dictionary
+    RE(count([det1])) #stage, trigger and unstage det 1
+    assert len(TelemetryUI.telemetry) == length + 1 #check an entry has been added to the dictionary 
+                                                   #(3 once stage and unstage are added)
+
+
+#set up the telemetry dictionary with some known values. 
+def setup_telemetry():
```
danielballan: This is a good place to use a pytest fixture.

bluesky/tests/test_est_time.py
```diff
@@ -0,0 +1,105 @@
+from bluesky.simulators import est_time
+from ophyd.sim import hw
+from bluesky.run_engine import RunEngine
+from ophyd.telemetry import TelemetryUI
+from bluesky.plans import scan, count
+from bluesky.plan_stubs import mv, sleep
+import pytest
+
+#setup some testing objects
+RE=RunEngine()
```
danielballan: Instead of sharing instances across the module, make a fresh one per test by using pytest fixtures. This avoids unexpected and hard-to-debug interactions between separate tests.

bluesky/tests/test_est_time.py
```diff
@@ -0,0 +1,105 @@
...
+                                 {'action': 'unstage', 
+                                  'estimation':{'std_dev': float('nan'), 'time': float('nan')},
+                                  'object_name': 'det1', 'time': {'start': 22, 'stop': 22.95}}])
+                                                                #add 2 'trigger' values for det1
+        
+
+def test_EpicsMotorEstTime():
+    #test the time estimation for epics motorlike devices.
+    setup_telemetry()
+    assert round(motor1.est_time.set(1,2,1,4).est_time,4) == 4.9975
```
danielballan: The standard way to do float comparisons in tests as `assert_almost_equal` or `assert_allclose`. You can find examples elsewhere in the bluesky tests.

bluesky/tests/test_est_time.py
awalter-bnl: I will look into pytest fixtures.

bluesky/tests/test_est_time.py
awalter-bnl: see above

bluesky/tests/test_est_time.py
awalter-bnl: I will look at both of these options that I didn't find in my initial investigation of pytest.

awalter-bnl:
I am now seeing your point of view regarding having only msg's and groups included in the base `est_time` function. I think our differences now occur in the output structure.

> In order to un-block progress here, how about this: can you get to an implementation of something like est_time in this comment but with two revisions derived from our subsequent conversation:
>1. The output EstTime should always be in cumulative terms.
>2. Yield out msg, est_time_tuple instead of just est_time_tuple

>That would give us all the information, unfiltered, and enable us to write many variations with just one pass through he plan. 

So this does not solve the main problem that we are disagreeing on (at least from what I understand), which is that we _are_ losing information by not also returning the estimated time (or start time) for each part as well as the cumulative stop time. For example: 

- without the est_time for each part we could not make the 'schedule bar-graph' visual representation I mentioned earlier as we would need both _start_ and _stop_ times for each element, not just the stop readout given by cumulative time. This information is, in particular, lost for 'groups' in your proposed structure, but is also not 100% extractable from the proposed output for 'runs' (particularly if we have multiple overlapping 'runs').
 
-  I see having both as also being useful for de-bugging, in that it will be very easy to determine if an action is being missed (by the large difference between stop time of one action and the start time of the next). This indicates that we have an action that is clearly missed in the output of either the base `est_time` function or a higher level function.

So the question that is holding up the progress at present is what output structure do we need/want? As the proposed base `est_time` function also includes 'groups' then we will need to resolve this before working on that. In order to ensure that no information is lost I see two options (I am happy to consider others):

1. We output both the cumulative stop time and the time to complete each element (msg, group, run).
2. We output a cumulative _start_ as well as stop time for each element (msg, group, run).

In terms of making it easier to read and understand the output perhaps 2. is better but I don't have a huge preference either way.

danielballan:
I'm not completely following. Let's focus on (1).

> We output both the cumulative stop time and the time to complete each element (msg, group, run).

Help me understand what this means.

* Time per msg: I think you can get this by comparing (subtracting) two subsequent items.
* Time per group: I guess you could say that "time per group" is the time elapsed from the first 'set'/'trigger'/'kickoff'/'complete' that uses a given group label to the time that that group is waited on. This is easy to compute by, again, subtracting the cumulative time of that first labeled message from the time of the wait message. Any consumer of `est_time` can cache that info and compute on the fly, in one pass.
* Time per run: Similarly, I can watch 'open_run' go back, note the cumulative time, and subtract that from the matching 'close_run'.

What information are we missing?

awalter-bnl:
> Time per msg: I think you can get this by comparing (subtracting) two subsequent items.

Agreed we do not have issues with time per msg.

> Time per group: I guess you could say that "time per group" is the time elapsed from the first 'set'/'trigger'/'kickoff'/'complete' that uses a given group label to the time that that group is waited on. This is easy to compute by, again, subtracting the cumulative time of that first labeled message from the time of the wait message. Any consumer of est_time can cache that info and compute on the fly, in one pass.

Your version of `est_time` only outputs the cumulative stop time on the 'wait' messages not on every 'set'/'trigger'/'kickoff'/'complete' message. It also does not output this on the first such message, and so this information is lost based on the current structure. We could have it 'yield' on the first message, but then we are confusing the output by saying that 'if it is  a 'set'....... type message it is a 'start time' otherwise it is a 'stop time'. I prefer to include a cumulative start and stop time on each yield to avoid this confusion.

> Time per run: Similarly, I can watch 'open_run' go back by, note the cumulative time, and subtract that from the matching 'close_run'.

In principle this can be done for runs, however I could imagine a version of `est_time_run` where we only output info on each run element. While the msgs and groups that occur between runs would be included in the cumulative stop time of `est_time`, your output of only cumulative time for such a higher level function would loose the information regarding times between runs.

Is this clearer?

danielballan:
> Your version of est_time only outputs the cumulative stop time on the 'wait' messages not on every 'set'/'trigger'/'kickoff'/'complete' message.

I don't understand this. The proposed `est_time` would output this:

```py
...  # assume we are showing the middle of some long est_time output
(Msg('close_run'), EstTime(100, 0.001))
(Msg('set', motor1, 3, group='A'), EstTime(100.1, 100.1))  # first usage of group 'A'
(Msg('set', motor2, 3, group='A'), EstTime(100.2, 100.141))
(Msg('wait', None, group='A'), EstTime(113, 3))  # for some slow motor
...
```

So the "group time" for 'A' would be `113 - 100`.

awalter-bnl:
So for those playing at home, there was a long discussion between @danielballan and myself this morning trying to resolve the differences of opinions we have on this. We have come up with the following path forward:

- `est_time` will follow the structure that @danielballan suggested above.

- The output will be a named tuple following the structure `(msg, cumulative_time, time_for_msg)` with `cumulative_time` and `time_for_msg` also being named tuples with the structure `(est_time, std_dev)`. Due to the structure of `est_time` a `wait` message returns the time for the group it waits on in `time_for_msg`.

- All higher level `est_time_....` functions will have the same output structure.

I will make these changes in due course.

danielballan:
Per our chat, @awalter-bnl means `time_for_msg` to be `time_for_this_unit` (run, wait group, etc.). I have said I still wonder if that aggregation would be better done one layer up, as different consumers of this information might be interested in different aggregates (by sample instead of by run, for example).

I am also not sure that "higher level est_time_.... functions will have the same output structure," as this seems like a good application for low-level and high-level APIs that may have different return types.

But we have better alignment than we did before, and I reiterate these concerns here only to flag points of future discussion. I think @awalter-bnl should implement exactly what he specified above without regard to my tentative concerns, and then we can resume the discussion from there.

awalter-bnl:
The latest commit represents a major change in the structure/ API, this has not been tested or been checked for PEP8, this will be done once we have agreement that this is the structure/ API we want to use.

Notes on the new structure/API

-	Basic output structure is now a named tuple of the form `(message, stop_time, start_time)` where.
o	`message` is the full message yielded by `plan`
o	 `stop_time` and `start_time` are named tuples of the form `(est_time, std_dev)`
-	A base level function, with maximum granularity, is now provided via `est_time(plan)`
•	This yields for each `message`, with the yield for `wait` messages being the times for the associated group as a whole
-	Higher level functions reduce the verbosity of `est_time(plan)`
o	Examples: 
•	`est_time_run(plan)` which modifies the `start_time` in the `close_run` yield of `est_time(plan)` to be the start of the associated run. 
•	`est_time_per_run(plan)` which takes `est_time_run(plan)` and removes the yields between `open_run` and `close_run` messages.
•	`est_time_per_group(plan)` which takes `est_time(plan)` and removes the yields for ‘groupable’ messages.
•	Many others that have not been thought of yet !-)
-	Output functions take these ‘times’ and do something with them.
o	Examples:
•	`print_est_time(plan, est_time_func)` which takes the yields from `est_time_func(plan)` and prints them in a human readable format.
•	I would like to make `plot_est_time` which is like `print_est_time` but makes a ‘schedule bar graph’ plot out of them.
o	Getting the exact ouput that Dan wants could also be done using an output function, this is not true in the other direction.
•	Example. `>set(motor1, x , ‘groupA’), set(motor2, y, ‘groupB’), wait( ‘groupB’), set(motor3, z, ‘groupA’)` is a valid construct but in it would not be possible to ‘locate’ the `motor3` move inside `groupA` without having the start and stop time for the third `set` yielded, particularly if it finishes before the `motor1` move.

A few points to discuss:
-	Looking at the structure here I am wondering if it is worth removing the start and stop time for the `group` (yielded on the `wait` msg) from `est_time`. 
o	It could then be done in  a higher level function `est_time_group`, in the same way it is done for runs in `est_time_run`, this would be more consistent. 
-	I like this structure (with both low level and high level functions having the same API) because it makes the ‘Output functions’ generic, giving the user the ability to ‘select’ the verbosity level for a given function.
-	I am still not convinced that ‘communication time’ and ‘internal_state_time’ can be accurately provided. 
o	I don’t think they are ‘static’ and they will change from beamline to beamline and day to day (perhaps even hour to hour).
o	They have been included in case someone comes up with a good solution for estimating them.
-	If we want to include ‘communication_time’ independently into the times defined for messages that are estimated from telemetry (which they are in this draft) then we will need to make significant changes to the status_objects.
o	This is because at present this time is not accounted for in the estimation of the time for the move that is stored in the telemetry. As a result including it again here will mean it is counted twice.

bluesky/simulators.py
```diff
@@ -114,3 +120,424 @@ def check_limits(plan):
...
+
+    # used to hold the cumulative time for the plan
+    out_time = _TimeStats(0, 0)
+
+    for msg in plan:
+        # case 1: commands that are 'groupable'
+        if msg.command in _group_start_cmds:
+            # add set and trigger commands to plan_history
+            if msg.command == 'set':
+                plan_history['set'][msg.obj.name] = msg.args[0]
```
danielballan: Why pick off just `msg.args[0]` here? What about any others `args` and `msg.kwargs`?

bluesky/simulators.py
```diff
@@ -114,3 +120,424 @@ def check_limits(plan):
...
+            start_time = out_time
+            out_time = combine_est(out_time, _communication_time)
+
+            yield _MsgStats(msg, out_time, start_time)
+
+        # case 5: sleep command
+        elif msg.command is 'sleep':
+            start_time = out_time
+            out_time = combine_est(out_time, _TimeStats(msg.args[0], 0))
+            out_time = combine_est(out_time, _communication_time)
```
danielballan: Is this a copy/paste mistake? There are no network calls involved in sleep.

bluesky/simulators.py
```diff
@@ -114,3 +120,424 @@ def check_limits(plan):
...
+    the scan or since an 'unstage' event that the device has been triggered.
+"""
+# GROUP THE DIFFERENT COMMAND TYPES
+# The commands that start/end a 'group', but not flyer commands.
+_group_start_cmds = ['set', 'trigger', 'kickoff']
+_group_end_cmds = ['wait']
+# The commands that aren't  groupable but still have statistics.
+_timed_cmds = ['stage', 'unstage']
+# The commands that only require network communication.
+_communication_cmds = ['read', 'describe', 'open_run', 'close_run']
```
danielballan: 'read' and 'describe' are very different from 'open_run' and 'close_run'. The last two are internal state-related.

awalter-bnl:
@danielballan all 3 of your comments are valid, however before we get into these 'details' I want to agree that the overall structure and API are what we want. This will avoid me spending a lot of time resolving these small issues only to have to rewrite large portions of the code later anyway when it is decided that the structure and API are not what we want (as happened last time). 

Can I please get feedback on the structure and API first, once this is done we can go over the the details.

awalter-bnl:
I have cleaned up the code with flake8

danielballan:
Yes. @tacaswell and I had a thought about how to resolve this at SciPy but need some time to write it up. Maybe @tacaswell can get to it before I get back from vacation?

danielballan:
Notes from call with @awalter-bnl:

We will go with this API: `est_time(plan) -> yields (processing_time, action_time)` where `processing_time` is always an `EstTime` representing that time it takes to process a given msg and `action_time` is `None` for synchronous messages (open_run, checkpoint, etc.) or another `EstTime` for any messages that return status objects, representing their time to complete

I am bothered by the return type instability of this design --- some message types return `(EstTime, None)` while others return `(EstTime, EstTime)` --- but this is a reflection of the return type instability of what RE `send`s back to the plan for a given Msg (status object or not status object).

Higher level consumers have all the information they need to produce aggregates like "time per run", "time per group," or "which was the slow motor in this group.

It will eventually be important for the various constant times (internal state update time, communication time, document emitting time) to be configurable somehow in case the caller knows things about local configuration, such as how time-costly the callbacks are, that it wants to incorporate. For now we can use guesses --- `0` is a good zeroth-order approximation. Down the road, maybe we can get some data on communication time by doing spot checks with `EchoRequest`.

awalter-bnl:
Latest commits, updated the API as described above. Also added tests for each of the est_time functions. In progress new commits to the ophyd (https://github.com/NSLS-II/ophyd/pull/567/files) in order to remove the 'communication time' from the current status object telemetry estimates.
A few important points:
- I have 'renamed' est_time to est_delta_time (if this is not liked then I can change it again suggestions).
- I have written a new ''est_time' which uses 'est_delta_time' and converts to a cumulative start/stop time (as was the previous outputting.
- This allows the rest of the est_time functions and the print_est_time function to remain unchanged.

awalter-bnl:
FYI looking into why the tests are failing on Travis when they work fine locally !!

awalter-bnl:
OK so the error for all Travis builds relates to the lack of the related ophyd PR (https://github.com/NSLS-II/ophyd/pull/567/files). This should work once that PR is merged.

danielballan:
> This should work once that PR is merged.

To test this against the corresponding ophyd changes, you can add a line to `.travis.yml` to install ophyd specifically from that git branch. We will then revert the change before finally merging. We don't do this often, but it is appropriate for large coordinated changes like this one. For example, the last time I recall doing it was for the "asset refactor" work.

awalter-bnl:
@danielballan I am reluctant to do this because:
1. The ophyd branch work will need to be merged first anyway (this does the telemetry side of things so I think we should aim for that to be merged, deployed and generating data prior to merging this side of things so we can get some more test data).
2. Given the timeline for how often this is revisited I will definitely forget to undo this change prior to merging and who knows when it will then be noticed.

danielballan:
Fine by me.

awalter-bnl:
Based on a discussion with @danielballan, and the issues found trying to rebase the ophyd counterpart to the current master, I am closing this off. Potentially a new PR may be opened in the future.

PR #1054, opened 2018-06-14
Live grid image mirroring

awalter-bnl:
## Description
Add the kwarg 'axes_positive' which is a list in the form ['x_direction', y_direction']. x_direction can take the values 'right' or 'left' and y_direction can take the values 'up' or 'down'. These values define the positive direction of the corresponding x or y axis.

## Motivation and Context
At SRX they scan the sample in order to produce 'images'. This scanning process leads to the y axis being inverted, as the moving the sample 'up' moves the beam 'down' on the sample. This will allow them to flip the direction of y_axis in order to have LiveGrid display the image correctly.

### DIFF
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -310,6 +310,14 @@ class LiveGrid(CallbackBase):
     ax : Axes, optional
         matplotib Axes; if none specified, new figure and axes are made.
 
+    x_positive: string, optional
+        Defines the positive direction of the x axis, takes the values 'right'
+        (default) or 'left'.
+
+    y_positive: string, optional
+        Defines the positive direction of the y axis, takes the values 'up'
+        (default) or 'down'.
+
     See Also
     --------
     :class:`bluesky.callbacks.LiveScatter`.
@@ -317,7 +325,7 @@ class LiveGrid(CallbackBase):
     def __init__(self, raster_shape, I, *,
                  clim=None, cmap='viridis',
                  xlabel='x', ylabel='y', extent=None, aspect='equal',
-                 ax=None):
+                 ax=None, x_positive='right', y_positive='up'):
         if ax is None:
             fig, ax = plt.subplots()
         ax.cla()
@@ -336,6 +344,8 @@ class LiveGrid(CallbackBase):
         self.im = None
         self.extent = extent
         self.aspect = aspect
+        self.x_positive = x_positive
+        self.y_positive = y_positive
 
     def start(self, doc):
         if self.im is not None:
@@ -351,10 +361,27 @@ class LiveGrid(CallbackBase):
                             extent=extent, aspect=self.aspect,
                             origin='lower')
 
-        # make sure the 'sense' of the axes is 'y values increasing up'
+        # make sure the 'positive direction' of the axes matches what is defined in
+        #axes_positive
+        xmin, xmax = self.ax.get_ylim()
+        if ((xmin > xmax and self.x_positive == 'right') or
+                (xmax > xmin and self.x_positive == 'left')):
+            self.ax.set_xlim(xmax, xmin)
+        elif ((xmax >= xmin and self.x_positive == 'right') or
+                (xmin >= xmax and self.x_positive == 'left')):
+            self.ax.set_xlim(xmin, xmax)
+        else:
+            raise ValueError('x_positive must be either "right" or "left"')
+
         ymin, ymax = self.ax.get_ylim()
-        if ymin > ymax:
+        if ((ymin > ymax and self.y_positive == 'up') or
+                (ymax > ymin and self.y_positive == 'down')):
             self.ax.set_ylim(ymax, ymin)
+        elif ((ymax >= ymin and self.y_positive == 'up') or
+                (ymin >= ymax and self.y_positive == 'down')):
+            self.ax.set_ylim(ymin, ymax)
+        else:
+            raise ValueError('y_positive must be either "up" or "down"')
 
         self.im = im
         self.ax.set_title('scan {uid} [{sid}]'.format(sid=doc['scan_id'],
--- a/bluesky/magics.py
+++ b/bluesky/magics.py
@@ -346,7 +346,7 @@ def get_labeled_devices(user_ns=None, maxdepth=6):
 
 
     # Convert from defaultdict to normal dict before returning.
-    return dict(obj_list)
+    return {k: sorted(v) for k, v in obj_list.items()}
 
 
 def is_parent(dev):


bluesky/callbacks/mpl_plotting.py
```diff
@@ -351,10 +358,28 @@ def start(self, doc):
...
-        if ymin > ymax:
+        if (ymin > ymax and self.axes_positive[1] == 'up') or \
+                        (ymax > ymin and self.axes_positive[1] == 'down'):
             self.ax.set_ylim(ymax, ymin)
+        elif (ymax >= ymin and self.axes_positive[1] == 'up') or \
+                        (ymin >= ymax and self.axes_positive[1] == 'down'):
+            self.ax.set_ylim(ymin, ymax)
+        else:
+            raise ValueError('axes_positive[1] must be either "up" or "down"')
+
```
mrakitin: Too many blank lines?

bluesky/plan_patterns.py
```diff
@@ -66,6 +66,99 @@ def spiral(x_motor, y_motor, x_start, y_start, x_range, y_range, dr, nth, *,
     cyc += cycler(y_motor, y_points)
     return cyc
 
+        
+def spiral_square_pattern(x_motor, y_motor, x_centre, y_centre, x_range, y_range, x_num,y_num):
```
mrakitin: I thought we implemented it within #1001. Should it be removed from *this* PR?

bluesky/plans.py
```diff
@@ -1331,7 +1331,66 @@ def rel_spiral(detectors, x_motor, y_motor, x_range, y_range, dr, nth,
                               x_range, y_range, dr, nth, tilt=tilt,
                               per_step=per_step, md=_md))
 
+def spiral_square(detectors, x_motor, y_motor, x_centre, y_centre, x_range,
```
mrakitin: See my comment above. Covered by #1001, should be removed.

awalter-bnl: This is a github error that  I will try to rectify now. for some reason it added the commits from the square_spiral branch along with my new ones.

bluesky/plans.py
```diff
@@ -1329,9 +1329,9 @@ def rel_spiral(detectors, x_motor, y_motor, x_range, y_range, dr, nth,
     return (yield from spiral(detectors, x_motor, y_motor,
                               x_motor.position, y_motor.position,
                               x_range, y_range, dr, nth, tilt=tilt,
-                              per_step=per_step, md=_md))
-
-
+                              per_step=per_step, md=_md))  
+  
+
```
mrakitin: These are odd blank spaces... Should not be there since the file is not touched.

awalter-bnl: So if you look at the difference there is none between the old and new file, not sure why they are still included as 'changed lines' or how to fix this.

awalter-bnl:
Pushed some new commits to fix the style issues raised by @mrakitin

bluesky/plans.py
```diff
@@ -1471,7 +1471,6 @@ def inner_relative_spiral():
 
     return (yield from inner_relative_spiral())
 
-
 def ramp_plan(go_plan,
```
mrakitin: I think it misses one line, that's why it shows the difference.

awalter-bnl:
Thanks @mrakitin I think I solved the issue now.

mrakitin:
The changes are good now, thanks @awalter-bnl! We still need to do a "git surgery". Let's take care of it next Monday.

awalter-bnl:
thanks @mrakitin, I attempted a few git fixes today without much luck. I will look into it more tomorrow, but if I don't have any luck then some assistance next week will be appreciated.

bluesky/callbacks/mpl_plotting.py
```diff
@@ -309,6 +309,12 @@ class LiveGrid(CallbackBase):
 
     ax : Axes, optional
         matplotib Axes; if none specified, new figure and axes are made.
+        
+    axes_positive : list, optional
```
danielballan: Matplotlib also has a parameter to control this called [origin] -- ([example](https://matplotlib.org/2.0.2/examples/pylab_examples/image_origin.html)). We actually use it in this function. This makes me wonder:

1. Should we just expose origin? Do people really need to flip left-to-right or is controlling the vertical sign good enough for most uses?
2. If we do add this separate setting, does it pass the right parameter to origin to ensure the two features don't fight each other in unpredictable ways?

bluesky/callbacks/mpl_plotting.py
```diff
@@ -317,7 +323,7 @@ class LiveGrid(CallbackBase):
     def __init__(self, raster_shape, I, *,
                  clim=None, cmap='viridis',
                  xlabel='x', ylabel='y', extent=None, aspect='equal',
-                 ax=None):
+                 ax=None, axes_positive = ['right', 'up']):
```
danielballan: We avoid mutable defaults except in special circumstances. This should be implemented as a `tuple` -- or maybe as two separate parameters?

bluesky/callbacks/mpl_plotting.py
awalter-bnl: @tacaswell is COMPLETELY against using origin, see discussion [here](https://3.basecamp.com/3733838/buckets/4066447/messages/1128589265). It appears that introducing origin means we open ourselves up to some unexpected and weird behaviour. From that discussion it appears that this is the best compromise.

bluesky/callbacks/mpl_plotting.py
awalter-bnl: I am happy to make this a tuple. Perhaps a named tuple is a better option? what are everyones thoughts on named tuples?

danielballan: A namedtuple won't have much impact here, the user won't be able to access it (except via `inspect.signature`, which would be a very special case). A tuple is fine.

bluesky/callbacks/mpl_plotting.py
danielballan: OK, I just caught up. Was not aware of this latest opinion on 'origin'. Sounds good.

bluesky/callbacks/mpl_plotting.py
mrakitin: @awalter-bnl, once the tuple is implemented, we will be ready to merge it after a small "git surgery" :)

tacaswell: Would it be better to do this as two parameters so it is easier to set just one of them and let the other be the default?  This would also get is out of the mutable default issue.

bluesky/callbacks/mpl_plotting.py
```diff
@@ -336,6 +342,7 @@ def __init__(self, raster_shape, I, *,
         self.im = None
         self.extent = extent
         self.aspect = aspect
+        self.axes_positive = axes_positive
```
tacaswell: It might be more readable to unpack this into two private attributes here (and do the validation here).

bluesky/callbacks/mpl_plotting.py
awalter-bnl: I am getting to it @mrakitin hopefully by the end of the week !-)

bluesky/callbacks/mpl_plotting.py
```diff
@@ -351,11 +361,28 @@ def start(self, doc):
...
-        # make sure the 'sense' of the axes is 'y values increasing up'
+        # make sure the 'positive direction' of the axes matches what is defined in 
+        #axes_positive
+        xmin, xmax = self.ax.get_ylim()
+        if (xmin > xmax and self.x_positive == 'right') or \
+                        (xmax > xmin and self.x_positive == 'left'):
+            self.ax.set_xlim(xmax, xmin)
+        elif (xmax >= xmin and self.x_positive == 'right') or \
+                        (xmin >= xmax and self.x_positive == 'left'):
+            self.ax.set_ylim(xmin, xmax)
```
tacaswell: This should be xlim

tacaswell:
@awalter-bnl can you please squash this to one commit?

bluesky/callbacks/mpl_plotting.py
```diff
@@ -309,6 +309,14 @@ class LiveGrid(CallbackBase):
...
     ax : Axes, optional
         matplotib Axes; if none specified, new figure and axes are made.
+        
+    x_positive: string, optional
+        Defines the positive direction of the x axis, takes the values 'right' 
+        (default) or 'left'.
+        
+    y_positive: string, optional
+        Defines the positive direction of the y axis, takes the values 'up'
+        (default) or 'left'.
```
jrmlhermitte: do you mean 'down'?

awalter-bnl: yes I do I will fix this

awalter-bnl:
@tacaswell I _think_ I was able to squash the commits, but here in the conversation it still shows 23 commits (my squash commit is shown). Can you confirm if it has squashed correctly? otherwise I will make another attempt !-)

awalter-bnl:
I think the failure is unrelated (it is a bluesky magics failure).

PR #1056, opened 2018-06-19
update ramp_plan in plans.py to enable customized choose whether add …

ke-zhang-rd:
…pre data or not

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1475,7 +1475,7 @@ def rel_spiral_square(detectors, x_motor, y_motor, x_range, y_range,
 def ramp_plan(go_plan,
               monitor_sig,
               inner_plan_func,
-              pre_data_enable=False,
+              take_pre_data=False,
               timeout=None,
               period=None, md=None):
     '''Take data while ramping one or more positioners.
@@ -1535,7 +1535,7 @@ def ramp_plan(go_plan,
             fail_time = time.time() + timeout
 
         # take a 'pre' data point
-        if pre_data_enable == True:
+        if take_pre_data == True:
             yield from inner_plan_func()
         # start the ramp
         status = (yield from go_plan)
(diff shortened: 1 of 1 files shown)


ke-zhang-rd:
function ramp_plan will automatically add a pre data point. Try fix by give user optional  to choose whether add that or not

danielballan:
Thanks, @ke-zhang-rd. This reinforces my sense that `ramp_plan` has to be _so_ parameterized that it might be better to just rewrite it as needed on a case-by-case basis, but I guess that ship has sailed.

Can we brainstorm possible names? I see what you mean by `pre_data_enable` but I wonder if we can think of something more descriptive, something that might be more obvious to someone who is encountering this function for the first time. Ideas:

* take_initial_point
* initial_datapoint
* take_pre_point

Others?

ke-zhang-rd:
@danielballan 
You mean rewrite ramp_plan?  Make several case-by-case function instead of one parameterized function? 

How about "**take_pre_data**"?

danielballan:
> You mean rewrite ramp_plan? Make several case-by-case function instead of one parameterized function?

Yeah.

> How about "take_pre_data"?

That works for me.

ke-zhang-rd:
@danielballan,
Should I set default as "take_pre_data = True" OR False? 
True might be more compatible to current working code.

danielballan:
Yes, I think it's much better to avoid an API break and default to `True`.

ke-zhang-rd:
thanks @tacaswell 

now ramp_plan is compatible with existing code. If bealine need get rid of pre data, just set it as false? Is that good logic?

Also, what is the workflow of merger pull request? Is that during group meeting?

danielballan:
> If beamline need get rid of pre data, just set it as false? Is that good logic?

Yes

> Also, what is the workflow of merger pull request? Is that during group meeting?

It depends on the PR. Things with a large impact get a lot of review and sign-off from our collaborators at other facilities. Simple changes can be merged once one or two people have had a look.

I am curious -- how did this first come up? From your explorations of `ramp_plan`?

ke-zhang-rd:
No, from one deployment before at IOS. Tom found it.

Should I request someone review it?  since this is a small change

danielballan:
I feel comfortable unilaterally merging this one. Thanks for the improvement!

ke-zhang-rd:
Gotcha.

PR #1064, opened 2018-07-05
ENH: prototype allow passing message prefixes to zmq

jrmlhermitte:
## Description
Give "topics" to zmq messages managed by the `RemoteDispatcher` and `Publisher`.

## Motivation and Context

From discussion with @CJ-Wright 
He would like to multiplex to multiple zeroMQ instances from the same RunEngine.
A first step to that would be to allow passing of arbitrary prefixes to ZMQ messages that can be filtered upon receipt. This is similar to the idea how we use topics with Kafka (this prefix is basically like a topic, although of course the implementation is much different)

I propose that we allow adding a prefix to the zmq messages.

I would have opened an issue, but I found writing it out was an issue starting point for discussion.

This needs to be tested if there is general consensus that this is a good idea.

thanks!

## Alternatives
The user would have to do the filtering themselves downstream.
It is assumed that all the data is going through to the same zmq proxy.
Note that **all** messages still go through to the subscriber.

There is no notion of recipient based filtering.

## How Has This Been Tested?
Not tested yet

### DIFF
--- a/bluesky/callbacks/zmq.py
+++ b/bluesky/callbacks/zmq.py
@@ -35,8 +35,6 @@ class Publisher:
     >>> publisher = Publisher('localhost:5567', RE=RE)
     """
     def __init__(self, address, *, RE=None, zmq=None, prefix=''):
-        if prefix is None:
-            prefix = ''
         if zmq is None:
             import zmq
         if isinstance(address, str):
--- a/bluesky/callbacks/zmq.py
+++ b/bluesky/callbacks/zmq.py
@@ -24,6 +24,7 @@ class Publisher:
     zmq : object, optional
         By default, the 'zmq' module is imported and used. Anything else
         mocking its interface is accepted.
+    prefix : message prefix
 
     Example
     -------
@@ -33,7 +34,9 @@ class Publisher:
     >>> RE = RunEngine({})
     >>> publisher = Publisher('localhost:5567', RE=RE)
     """
-    def __init__(self, address, *, RE=None, zmq=None):
+    def __init__(self, address, *, RE=None, zmq=None, prefix=None):
+        if prefix is None:
+            prefix = ''
         if zmq is None:
             import zmq
         if isinstance(address, str):
@@ -43,8 +46,13 @@ class Publisher:
         self.hostname = socket.gethostname()
         self.pid = os.getpid()
         url = "tcp://%s:%d" % self.address
-        self._prefix = b'%s %d %d ' % (self.hostname.encode(),
-                                       self.pid, id(RE))
+        if prefix is not None:
+            self._prefix = b'%s %s %d %d ' % (prefix, self.hostname.encode(),
+                                              self.pid, id(RE))
+        else:
+            self._prefix = b'%s %s %d %d ' % (self.hostname.encode(), self.pid,
+                                              id(RE))
+
         self._context = zmq.Context()
         self._socket = self._context.socket(zmq.PUB)
         self._socket.connect(url)
@@ -204,7 +212,8 @@ class RemoteDispatcher(Dispatcher):
     >>> d.start()  # runs until interrupted
     """
     def __init__(self, address, *, hostname=None, pid=None, run_engine_id=None,
-                 loop=None, zmq=None, zmq_asyncio=None):
+                 loop=None, zmq=None, zmq_asyncio=None, prefix=None):
+        self._message_prefix = prefix
         if zmq is None:
             import zmq
         if zmq_asyncio is None:
@@ -227,13 +236,22 @@ class RemoteDispatcher(Dispatcher):
         self._socket.setsockopt_string(zmq.SUBSCRIBE, "")
         self._task = None
 
-        def is_our_message(_hostname, _pid, _RE_id):
+        def is_our_message(_hostname, _pid, _RE_id, message_prefix=None):
             # Close over filters and decide if this message applies to this
             # RemoteDispatcher.
-            return ((hostname is None or hostname == _hostname)
-                    and (pid is None or pid == _pid)
-                    and (run_engine_id is None or run_engine_id ==
-                         run_engine_id))
+            if message_prefix is not None:
+                return ((hostname is None or hostname == _hostname)
+                        and (pid is None or pid == _pid)
+                        and (run_engine_id is None or run_engine_id ==
+                            run_engine_id)
+                        and message_prefix == self._message_prefix
+                       )
+            else:
+                return ((hostname is None or hostname == _hostname)
+                        and (pid is None or pid == _pid)
+                        and (run_engine_id is None or run_engine_id ==
+                            run_engine_id))
+
         self._is_our_message = is_our_message
 
         super().__init__()
@@ -242,12 +260,16 @@ class RemoteDispatcher(Dispatcher):
     def _poll(self):
         while True:
             message = yield from self._socket.recv()
-            hostname, pid, RE_id, name, doc = message.split(b' ', 4)
+            if self.message_prefix is not None:
+                prefix, hostname, pid, RE_id, name, doc = message.split(b' ', 5)
+            else:
+                hostname, pid, RE_id, name, doc = message.split(b' ', 5)
+            prefix = prefix.decode()
             hostname = hostname.decode()
             pid = int(pid)
             RE_id = int(RE_id)
             name = name.decode()
-            if self._is_our_message(hostname, pid, RE_id):
+            if self._is_our_message(hostname, pid, RE_id, message_prefix=prefix):
                 doc = pickle.loads(doc)
                 self.loop.call_soon(self.process, DocumentNames[name], doc)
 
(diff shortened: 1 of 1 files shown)


bluesky/callbacks/zmq.py
```diff
@@ -242,12 +260,16 @@ def is_our_message(_hostname, _pid, _RE_id):
     def _poll(self):
         while True:
             message = yield from self._socket.recv()
-            hostname, pid, RE_id, name, doc = message.split(b' ', 4)
+            if self.message_prefix is not None:
+                prefix, hostname, pid, RE_id, name, doc = message.split(b' ', 5)
+            else:
+                hostname, pid, RE_id, name, doc = message.split(b' ', 5)
```
mrakitin: I think it should be as in the old code, i.e. `message.split(b' ', 4)`

bluesky/callbacks/zmq.py
```diff
@@ -43,8 +46,13 @@ def __init__(self, address, *, RE=None, zmq=None):
         self.hostname = socket.gethostname()
         self.pid = os.getpid()
         url = "tcp://%s:%d" % self.address
-        self._prefix = b'%s %d %d ' % (self.hostname.encode(),
-                                       self.pid, id(RE))
+        if prefix is not None:
+            self._prefix = b'%s %s %d %d ' % (prefix, self.hostname.encode(),
+                                              self.pid, id(RE))
+        else:
+            self._prefix = b'%s %s %d %d ' % (self.hostname.encode(), self.pid,
```
mrakitin: You don't need two `%s` in this case since you have 3 elements to format.

bluesky/callbacks/zmq.py
```diff
@@ -43,8 +46,13 @@ def __init__(self, address, *, RE=None, zmq=None):
         self.hostname = socket.gethostname()
         self.pid = os.getpid()
         url = "tcp://%s:%d" % self.address
-        self._prefix = b'%s %d %d ' % (self.hostname.encode(),
-                                       self.pid, id(RE))
+        if prefix is not None:
```
danielballan: `prefix` will never be `None` because it was set to `''` above. I think the simplest approach is to make `prefix=''` the default in the `__init__` signature and then remove this if block.

bluesky/callbacks/zmq.py
```diff
@@ -227,13 +236,22 @@ def __init__(self, address, *, hostname=None, pid=None, run_engine_id=None,
         self._socket.setsockopt_string(zmq.SUBSCRIBE, "")
         self._task = None
 
-        def is_our_message(_hostname, _pid, _RE_id):
+        def is_our_message(_hostname, _pid, _RE_id, message_prefix=None):
```
danielballan: Probably simpler to make the default `''` and remove the if block here too.

jrmlhermitte:
thanks guys, this was a quick transition from "let me ask slack" -> "wait let me just make an issue" -> "wait this can just be a PR for the same amount of time"

If no one objects, I'll clean it up, thanks!

bluesky/callbacks/zmq.py
jrmlhermitte: ah yes thanks! There is just one issue with the message format. It should probably be made homogeneous, have a delimiter. This is because the other end can't know what the format will be, and we probably don't want to mix different formats.
I'd vote for "foo\x00" (the null character)
In the docstring, we should mention the null character cannot be used in the prefix.

(and of course, all messages will have to come from matching versions so this will be incompatible with previous versions)

bluesky/callbacks/zmq.py
jrmlhermitte: thanks

bluesky/callbacks/zmq.py
jrmlhermitte: yep thanks!

CJ-Wright:
How does this change downstream code?

jrmlhermitte:
It will make it incompatible.
The Proxy is unaffected but the `RemoteDispatcher` would have to receive messages from a `Publisher` of the same version.

usage
```py3
# Proxy machine
./bluesky-0MQ-proxy 5556 5557

# the dispatcher 
dis_address = "localhost:5557"
dis = RemoteDispatcher(dis_address, prefix="apples")
dis.subscribe(print)
dis.start()

# the other dispatcher 
dis_address = "localhost:5557"
dis = RemoteDispatcher(dis_address, prefix="oranges")
dis.subscribe(print)
dis.start()

# the publisher
pub_address = "localhost:5556"
pub = Publisher(pub_address, prefix="apples")
doc = {"foo" : "bar"}
pub('start', doc)

# the publisher
pub_address = "localhost:5556"
pub = Publisher(pub_address, prefix="oranges")
doc = {"foo" : "bar"}
pub('start', doc)
```

CJ-Wright:
Should we pull out `pickle` and default it? So we can slot in other serialization tools? This way the ZMQ system would support the 4/5 split (serialization/transport).

jrmlhermitte:
I like @CJ-Wright 's approach. Backwards compatible and clean. What do you think @danielballan ?

The only counter argument I'd have for this is that topic is such a simple feature that we might want to just implement it for users (having o write your own serializer/deserializer, although probably just 2-3 lines of code, would be a little annoying for a topic.)

danielballan:
Seems plausible. Let's see what it looks like.

CJ-Wright:
What do you mean by

> having to write your own serializer/deserializer, although probably just 2-3 lines of code, would be a little annoying for just adding a topic

I'd suggest having pickle as the default
```python
def __init__(self, ..., serializer=None):
    if serializer is None:
        import pickle
        self.serializer = pickle
...
```

danielballan:
It's probably worth diving briefly into dask to see how they support this. Do they have one serializer/deserializer object? Or separate params? I read this part of dask once but don't recall the details anymore.

CJ-Wright:
@danielballan I think these docs discuss dask's serialization approach.
http://distributed.readthedocs.io/en/latest/serialization.html

jrmlhermitte:
@CJ-Wright I assumed by splitting into serialization/transport you meant that we'd push the "pop on/off topic from message" logic into the serialzier/deserializer.

i.e. the user would have to supplement pickle:
```py
def mynewserializer(doc):
    return  b"apples " + pickle.dumps(doc)

def mynewserializer2(bytes):
    return b"oranges " + pickle.dumps(doc)

def mynewdeserializer(bytes):
    topic, docs = bytes.split(b" ", 1)
    if topic == 'apples':
        return topic, pickle.loads(doc)
    else:
        return {}
```

or something of the sort.

In either case, if you did/did not suggest this, I :+1: the move to support multiple serializers.

Thanks @danielballan looks like they specify it explicily (see [here](http://distributed.readthedocs.io/en/latest/serialization.html) )
```py
from dask.distributed import Client
client = Client('tcp://scheduler-address:8786',
                serializers=['dask', 'pickle'],
                deserializers=['dask', 'msgpack'])
```

we could add `serializer=`, and `deserializer=`.

@CJ-Wright can you move this to an issue or PR? I thikn this is a little different from the discusion here.

danielballan:
Agreed we can tackle customizable (de)serialization in a separate issue. Let's make sure both these changes land in the same bluesky release though.

CJ-Wright:
xref: https://github.com/NSLS-II/bluesky/issues/1065

jrmlhermitte:
xreef #1069 for the serializer PR

jrmlhermitte:
From my point of view this is good to go. I'd say there are two major changes here:
- I have changed the delimiter for the messages in the prefix from " " to "0x00" (null character). This is in general I think safer. This will make this backwards incompatible with previous versions. This should be okay so long as we make sure the versions of bluesky on both the sender/receiver are the same.
- We can add an optional prefix to filter messages by. This doesn't reqiure much code and is a nice convenience to add (for the little amount of change it makes)

danielballan:
In a line comment that got swallowed up by GitHub, I wondered aloud whether it was worth removing my provisional hard-coded message prefix --- host + pid + id(RE) --- and _just_ exposing a user-configurable one. It would make the code simpler and faster and the interface more obvious.

danielballan:
Big :+1: on the null character. Much better than my original.

jrmlhermitte:
I prefer the user configurable one and vote for that as well.
I think most of the time a user remembering the topic they made is much easier than the hostname, PID and runengine PID.

So we remove everything and just leave the user prefix? I'm okay with that.

tacaswell:
No.

danielballan:
To elaborate, we are doing bug-fix releases this cycle, and doing this in the way we've agreed to do this requires an API break.

CJ-Wright:
In the meantime what do you suggest for "ZMQ multiplexing"

danielballan:
> In the meantime what do you suggest for "ZMQ multiplexing"

Throw a commit at this PR implementing [this comment](https://github.com/NSLS-II/bluesky/pull/1064#issuecomment-405698490), vendor in some xpd* package for this cycle with the expectation that you can rip it out and import it from bluesky after the next minor bluesky release.

danielballan:
Superceded by #1109, which adds the `prefix` _and removes_ and old way.

Issue #1065, opened 2018-07-06
Separate transport callbacks from serialization


CJ-Wright:
It would be nice to separate the transport of data (ZMQ, Kafka, etc.) from the serialization of it (pickle, dill, etc.)

Dask does this to good effect, see: http://distributed.readthedocs.io/en/latest/serialization.html

## Expected Behavior
Users can (optionally) specify how to serialize the data going through a transport layer.

## Current Behavior
ZMQ uses pickle

## Possible Solution
Take in the serialization functions (and their inverses) at init time on the transport layers with reasonable defaults.

jrmlhermitte:
Again, I :+1: a simple `serializer=`/`deserializer=` flag but no more complex.

Dask is slightly more complex in that it allows for specifying a list of serializers and successively try them until [one works](http://distributed.readthedocs.io/en/latest/serialization.html#use):

> Dask uses the serializers ['dask', 'pickle'] by default, trying to use dask custom serializers (described below) if they work and then falling back to pickle/cloudpickle.

(thanks @danielballan for that suggestion)

PR #1069, opened 2018-07-10
Add Serializer to zmq callback

jrmlhermitte:
For #1065 atn @CJ-Wright 

Note the prefix is not serialized using the serializer. This piece is interpreted internally by the zmq `Publisher` and `RemoteDispatcher`.

### DIFF
--- a/bluesky/callbacks/zmq.py
+++ b/bluesky/callbacks/zmq.py
@@ -33,7 +33,7 @@ class Publisher:
     >>> RE = RunEngine({})
     >>> publisher = Publisher('localhost:5567', RE=RE)
     """
-    def __init__(self, address, *, RE=None, zmq=None, serializer=None):
+    def __init__(self, address, *, RE=None, zmq=None, serializer=pickle.dumps):
         if zmq is None:
             import zmq
         if isinstance(address, str):
@@ -50,8 +50,6 @@ class Publisher:
         self._socket.connect(url)
         if RE:
             self._subscription_token = RE.subscribe(self)
-        if serializer is None:
-            serializer = pickle.dumps
         self._serializer = serializer
 
     def __call__(self, name, doc):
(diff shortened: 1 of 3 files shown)


bluesky/callbacks/zmq.py
```diff
@@ -33,7 +33,7 @@ class Publisher:
     >>> RE = RunEngine({})
     >>> publisher = Publisher('localhost:5567', RE=RE)
     """
-    def __init__(self, address, *, RE=None, zmq=None):
+    def __init__(self, address, *, RE=None, zmq=None, serializer=None):
```
CJ-Wright: Could we default this to `pickle.dumps` directly?

CJ-Wright:
Awesome!

bluesky/callbacks/zmq.py
jrmlhermitte: sounds great! I like it, changed.

jrmlhermitte:
codecov is complaining because there are lines run in a [separate thread](https://github.com/NSLS-II/bluesky/pull/1069/files#diff-3085c64224f496e2b3be30df9ed58a70R221) and hence the coverage is unknown to it

dev-requirements.txt
```diff
@@ -24,3 +24,4 @@ streamz
 sphinx==1.6.4
 sphinx_rtd_theme
 tifffile
+cloudpickle
```
danielballan: Can you move this into alphabetical order? Makes it easier to quickly check if a given entry is present.

bluesky/callbacks/zmq.py
```diff
@@ -33,7 +33,7 @@ class Publisher:
     >>> RE = RunEngine({})
     >>> publisher = Publisher('localhost:5567', RE=RE)
     """
-    def __init__(self, address, *, RE=None, zmq=None):
+    def __init__(self, address, *, RE=None, zmq=None, serializer=pickle.dumps):
```
danielballan: Can you add this parameter to the docstring?

bluesky/callbacks/zmq.py
```diff
@@ -204,13 +205,15 @@ class RemoteDispatcher(Dispatcher):
     >>> d.start()  # runs until interrupted
     """
     def __init__(self, address, *, hostname=None, pid=None, run_engine_id=None,
-                 loop=None, zmq=None, zmq_asyncio=None):
+                 loop=None, zmq=None, zmq_asyncio=None,
+                 deserializer=pickle.loads):
```
danielballan: Docstring update here too

jrmlhermitte:
I think this is good to go unless there are objections?
The failed test is a bluesky magics test also failing on #1068

PR #1076, opened 2018-08-02
Implement new callbacks mechanism

danielballan:
**Do not review yet -- this is still a moving target.**

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -3,7 +3,7 @@ Useful callbacks for the Run Engine
 """
 from itertools import count
 import warnings
-from collections import deque, namedtuple, OrderedDict, ChainMap
+from collections import defaultdict, deque, namedtuple, OrderedDict, ChainMap
 import time as ttime
 
 from datetime import datetime
@@ -120,7 +120,7 @@ class RunRouter(CallbackBase):
     """
     def __init__(self, callback_factories):
         self.callback_factories = callback_factories
-        self.callbacks = {}  # start uid -> callback
+        self.callbacks = defaultdict(list)  # start uid -> callbacks
         self.descriptors = {}  # descriptor uid -> start uid
         self.resources = {}  # resource uid -> start uid
 
@@ -131,16 +131,15 @@ class RunRouter(CallbackBase):
         except KeyError:
             # The belongs to a run that we are not interested in.
             return
-        cb = self.callbacks[start_uid]
-        return cb
+        return self.callbacks[start_uid]
 
     def event(self, doc):
-        cb = self._event_or_bulk_event(doc)
-        cb('event', doc)
+        for cb in self._event_or_bulk_event(doc):
+            cb('event', doc)
 
     def bulk_event(self, doc):
-        cb = self._event_or_bulk_event(doc)
-        cb('bulk_event', doc)
+        for cb in self._event_or_bulk_event(doc):
+            cb('bulk_event', doc)
 
     def _datum_or_bulk_datum(self, doc):
         resource_uid = doc['resource']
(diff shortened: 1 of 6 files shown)


bluesky/callbacks/core.py
```diff
@@ -50,6 +78,125 @@ def stop(self, doc):
...
+        for cb in cbs:
+            cb('resource', doc)
+
+    def start(self, doc):
+        for callback_factory in self.callback_factories:
+            cb = callback_factory(doc)
+            if cb is None:
+                # The callback_factory is not interested in this run.
+                continue
+            self.callbacks[doc['uid']].append(cb)
```
CJ-Wright: Would it be better to be explicit here, and after initializing all the callbacks to then pass the start document in?

bluesky/callbacks/core.py
```diff
@@ -50,6 +78,125 @@ def stop(self, doc):
...
+        if not cbs:
+            # This belongs to a run we are not interested in.
+            return
+        self.resources[doc['uid']] = start_uid
+        for cb in cbs:
+            cb('resource', doc)
+
+    def start(self, doc):
+        for callback_factory in self.callback_factories:
+            cb = callback_factory(doc)
```
CJ-Wright: Do we need a mechanism for pushing more information into the `callback_factory`? One could imagine wanting to pass information into the callback at init time which is not in the document. For instance, if I build a `callback_factory` which returns a `LiveScatter` how do I change the colormap of the scatter plot for the next scan?

CJ-Wright: Maybe the `callback_factory` signatures could be `cf(start_doc, **kwargs)` and we pass on **kwargs from the `RunRouter` itself? This would require the factory to take in arbitrary kwargs and pick and choose among them but that might be ok.

bluesky/callbacks/core.py
```diff
@@ -50,6 +78,125 @@ def stop(self, doc):
...
+    callback_factories : list
+        A list of callables with the signature:
+
+            callback_factory(start_doc)
+
+        which should return ``None`` or another callable with this signature:
+
+            callback(name, doc)
+    """
+    def __init__(self, callback_factories):
```
CJ-Wright: would it be possible to accept `**kwargs` to be passed to the callback factories?

danielballan:
Rebased on master. I have begun reworking this in light of our 'page' and 'bulk' work in event-model. I have also started implementing a two-layer system in RunRouter, whereby sometimes the decision of whether a given subsystem is applicable is made by _run_ and other times by _stream_, developed in conversations with @tacaswell and @CJ-Wright (separately).

There are two PRs from @awalter-bnl and @ke-zhang-rd respectively into this branch, which I will circle back to once those changes are in place here.

bluesky/callbacks/mpl_plotting.py
```diff
@@ -7,14 +7,242 @@
...
+                    break
+        else:
+            if plt.fignum_exists(fig_name):
+                # Generate a unique name by appending a number.
+                for number in itertools.count(2):
+                    new_name = '{} {}'.format(fig_name, number)
+                    if not plt.fignum_exists(new_name):
+                        fig_name = new_name
+                        break
+        ndims = len(dim_fields)
```
CJ-Wright: Should we be interested in the shape of the dims as well?

bluesky/callbacks/mpl_plotting.py
```diff
@@ -7,14 +7,242 @@
...
+
+            _, self.dim_stream = dimensions[0]
             return self
-        # Decide what plot(s) we need for this stream.
+
+        # Decide which plots we want for this stream.
+
+        self._descriptors[doc['uid']] = doc
+        stream_name = doc.get('name', 'primary')  # fall back for old docs
+        columns = hinted_fields(doc)
```
CJ-Wright: Should we filter this by things that are plot-able?

danielballan: We check `ndim` further down. Is that what you mean?

bluesky/callbacks/mpl_plotting.py
danielballan: I believe that each of these corresponds to a single axis and is therefore always 1D.

bluesky/callbacks/mpl_plotting.py
CJ-Wright: strings aren't plottable

bluesky/callbacks/callback_factories.py
```diff
@@ -0,0 +1,142 @@
...
+                     for fields, stream_name in hints['dimension']
+                     for field in fields]  # find all dimension field names.
+
+    # define some required parameters for setting up the grid plot.
+    # NOTE: THIS NEEDS WORK, in order to allow for plotting of non-grid type
+    # scans the following parameters need to be passed down to here from the RE
+    # This is the minimum information required to create the grid plot.
+    dim_names = [fields[0]
+                 for fields, stream_name in hints['dimensions']]
+    I_names = [c for c in hinted_fields(start_doc)
```
CJ-Wright: `hinted_fields` takes in a `descriptor` not a `start_doc`?

bluesky/callbacks/mpl_plotting.py
CJ-Wright: I'm not certain if that assumption will hold up when we have spectra. A spectra as a series of temperature readings would have temperature and energy hinted as independent variables, which I think would show up here in `dim_fields`.

bluesky/callbacks/core.py
```diff
@@ -1,17 +1,45 @@
...
+
+    def __init__(self, start_doc=None):
+        # Callbacks meant to be used with a RunRouter may *require* a start_doc
+        # at __init__ time by making the start_doc argument required.
+        # Others may not.
+        #
+        # If a start_doc is given, it is passed to the `start` method
+        # immediately, here. If one is not given, it is up to the caller to
+        # call `start(...)` later once the RunStart document is available.
+        if start_doc is not None:
```
CJ-Wright: Why is this needed? Shouldn't we this handled by the Callback factory explicitly (by calling `start` after the init).

stan-dot:
@danielballan are the goals of this PR still relevant? 

action suggestion:
This can be closed.
If the goals are not relvant anymore, that is the end.
if the goals are still relevant, a new issue can be made and linked to this PR, so that a fresh PR can be attempted at a later date following ideas from the PR possibly copying some code.

Issue #1095, opened 2018-09-18
Requested updates to LiveGrid


awalter-bnl:
# NOTE: This is being covered in PR #1098  
# Outline of scope
A number of new features/updates to LiveGrid have been requested. I have started this project in order to develop a plan to implement as much as possible. Listed below are the new features/upgrades to be added and some possible additional features (feel free to add others I may have missed). This is intended to start a discussion, and lead to a group conclusion on the what/when/how of implementing these changes.

## Additional features to be added:

1. Add the ability to define which scanned value is the x -axis and which is the y-axis.
    - This was requested by SMI in order to be able to swap which axis the 'inner motor' represents in grid scans.
     - We may want to us this feature to add some 'limited' multi dimensional plotting. (i.e. let the user select an x, a y and a z axis from a multi-dimensional scan to plot as a live grid plot.)
2. Add the actual trajectory on top of the image, see bluesky issue #1039 .
     - When a trajectory is known in advance we can overlay it onto the LiveGrid image.
     - It has been suggested (by @awalter-bnl) that we could use a colour change to show past/future points in the trajectory or remove already plotted points.
     - We could also update the readback position in the trajectory for all past points, overwriting the setpoint position.
3. Add the ability to switch between relative units and actual units for relative scans, see bluesky issue #1039.
     - At present it defaults to relative units as this is what is used to define extents.
4. Add the ability to define  arbitrary trajectories, see , see bluesky issue #1003, #1001 and #1061 
     - This will allow for square_spiral, and other arbitrary path scans that also result in a series of grid points, to use LiveGrid instead of LiveScatter.
     - @tacaswell suggested doing this by simply using the x and y co-ordinates from the event-source at each step to locate where to add the point on the grid.
     - Averaging if the point in the grid already contains a value will also allow all paths, regardless of if they result in a grid or not, to use LiveGrid.
     - This would also require passing in a 'raster shape' to indicate the size of the grid to be plotted against if the plan does not specify it.

## Possible extensions:

1. _Should we /How do we_ implement these proposed additions to BEC, I see two options (please add more if there are any) and they are not mutually exclusive:
    - through the inclusion of the optional kwargs in the metadata of the plan.
       - extents, shape, gridding, Ikey are all currently passed into Live_Grid in BEC through the start document in this way.
       - Has the benefit that the pre-defined plot parameters are stored in meta-data which allows for easy re-plotting later.
[...]

Issue #1061, opened 2018-06-27
scan_nd will not make heatmaps for two dimensional scans with Best Effort Callback


teddyrendahl:
## Expected Behavior

Both outer and inner product scans should give us the same BestEffortCallback

## Current Behavior

If you run `grid_scan` with two motors and `BestEffortCallback` we get two very nice heatmaps. Using the multi-axis syntax for `scan` does not  produce the same result. This is unfortunate as I want these plots but for an "inner product" instead of a grid.

## Possible Solution

The `hints` that are in `grid_scan` that alert `BestEffortCallback` should be added to the base `scan_nd`.

tacaswell:
The hints in grid scan tell us that it is a grid that we can stuff into a nice heat map.  We use dead reckoning based on the sequence number to know how to fill in the heat map (which maybe should changed?), but in general we don't know enough about a nd scan to know how to fill in a heat map.

awalter-bnl:
We have discussed changing this to allow square-spiral plans to also work with grid_scans by allowing rasters that have non-trivial orderings([bluesky issue 1003](https://github.com/NSLS-II/bluesky/issues/1003)). This may also help with this issue.

ambarb:
Naive question.  Does this requested change prohibit having `scan_nd()` output 1D plots in `LivePlot`?  Can the default remain 1D?

stan-dot:
@coretl this sounds relevant to scanspec, right? does scanspec solve this already?

Issue #1039, opened 2018-05-24
add actually trajectory to grid scan plots


tacaswell:
We set the extents of the image by looking at the start document (this has the surprising to me, but reported to be desired by @awalter-bnl , effect of making the tick labels for `rel_grid_scan` in relative units) and then we fill in the pixel be dead-reckonining on the sequence number.

We should plot the actual x-y trajectory on top of the image.

@awalter-bnl also suggest that we initialize the plot with the intended trajectory and then update it where it _actually_ went.  Use color to indicate past / future.

danielballan:
All sound like good ideas to me. Seems like we should make the absolute/relative thing configurable too.

ambarb:
I really like both suggestions(trajectory and configure absolute/relative for `rel_grid_scan`).

Could you add optional probe size argument and make the data point opaque for the future path??

tacaswell:
in progress in the BEC re-write, moving milestone.

stan-dot:
@coretl another scanspec related item

coretl:
yes

Issue #1003, opened 2018-03-08
ENH: teach LiveRaster how to deal with rasters that have non-trivial orderings


mrakitin:
See discussion in #1001.

## Expected Behavior

`LiveRaster` does not accept arbitrary trajectories at the moment, need to make it smarter.

## Current Behavior

See https://github.com/NSLS-II/bluesky/pull/1001#issuecomment-371602976.

## Possible Solution

Update the `LiveRaster` code to accept the sequence of coordinates as a new argument.

## Steps to Reproduce (for bugs)

See comparison of the `spiral_square` and `grid_scan` plots in https://github.com/NSLS-II/bluesky/pull/1001#issuecomment-371602976. The relevant code snippet is provided there as well.

## Context

#1001.

## Your Environment

Dev conda environment with Python 3.6:
```py
In [3]: import bluesky, ophyd, databroker

In [4]: bluesky.__version__
Out[4]: '1.2.0+42.g91c4ad6.dirty'

In [5]: ophyd.__version__
Out[5]: '1.0.0.post47+g4f93901'

In [6]: databroker.__version__
Out[6]: '0.10.0+3.g02424cf4'
```

mrakitin:
Attn: @danielballan, @awalter-bnl.

stan-dot:
@coretl another scanspec related item

coretl:
yes

PR #1098, opened 2018-09-27
Replacement to LiveGrid

awalter-bnl:
WORK IN PROGRESS, REQUIRES #1076 BEFORE IT CAN BE TESTED/ USED.
This PR will add a new `Grid` callback, a new `grid_factory` callback factory and potentially a new `path` callback. 

## Description
This will add a replacement to LiveGrid, using the new callback API being added in #1076 and incorporating the changes requested to LiveGrid as detailed in #1095.

## Motivation and Context
It primarily creates a new grid plot that works with the new callbacks mechanism found in #1076. In addition it adds several features requested across the issues #1095, #1039 , #1003, #1001 and #1061. These include: 

- allowing for arbitrary paths through the grid space.
    - This is done through the new function part of the callbacks that allow for more flexibility in the conversion between the data in the event doc and where that is included on the Grid. 
- adding greater flexibility to modify how images are plotted.
    - This is done via the new callback factories and the grid callback API, which allow for any matplotlib kwarg to be passed through to grid
- including the ability to swap axes directions, relative/absolute units and other plotting paths
    - This is done through the new functions for converting between the information in the event docs and where on the grid the data is to be placed.

## How Has This Been Tested?
This has yet to be tested, once PR #1076 is complete then it can be tested and it shouldn't be merged prior to that.

### DIFF
--- a/bluesky/callbacks/callback_factories.py
+++ b/bluesky/callbacks/callback_factories.py
@@ -1,4 +1,4 @@
-from .mpl_plotting import Grid
+from .mpl_plotting import Grid, Path
 import numpy as np
 
 
@@ -94,7 +94,14 @@ def grid_factory(start_doc):
                        extent[0][0] - y_step / 2,
                        extent[0][1] + y_step / 2]
 
+    # This section is where the scan path is defined, if the x_path and y_path
+    # are 'None' it is not overlayed.
+    # NOTE: we need to decide how to pass this info down to here.
+    x_path = None  # This should be able to take in the path info here.
+    y_path = None  # This should be able to take in the path info here.
+
     for I_name, ax in zip(I_names, axes):
+        # This section defines the function for the grid callback
         def func(self, bulk_event):
             '''This functions takes in a bulk event and returns x_coords,
             y_coords, I_vals lists.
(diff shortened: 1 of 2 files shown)


awalter-bnl:
git issues have been resolved

awalter-bnl:
I have resolved all of the changes requested by @danielballan

awalter-bnl:
@danielballan I was able to change the target branch to NSLS-II/one-run-callbacks but so far am unable to change it to danielballan/one-run-callbacks. The google search I did indicated that this was only possible between branches in the same repo, not between repos. I would rather not open a new PR and loose the comment history here, but if you insist on this going into danielballan/one-run-callbacks I will.

danielballan:
Before I forget, maybe we should call `Path` `Trajectory` or something because we might want to reserve "Path" for https://matplotlib.org/api/path_api.html

awalter-bnl:
The Travis errors are unrelated to this PR as none of them touch the specific files that have been changed here. (they appear to be test errors deep in the run-engine and associated with None-type objects from what I can tell).

awalter-bnl:
closing this as I have opened it against  'danielballan:one-run-callbacks'under [PR # 13](https://github.com/danielballan/bluesky/pull/13)

PR #1099, opened 2018-09-29
try to build a callback to retrieve stored data live

### LINKED ISSUE #1079: Doc/question: Adaptive scans with area detectors
Reading through the docs for single process [adaptive scans](http://nsls-ii.github.io/bluesky/tutorial.html#adaptive-logic-in-a-plan) everything looks great! However, does the same logic work for area detectors which save files? I'm guessing that the results that the detector responds with will be the filename and/or other information that is passed to the databroker for it to properly register the data. If this is the case what changes are needed to that scan to work with file saving detectors?

CJ-Wright:
Add a callback which keeps track of resource and datum documents, and has a retrieve method which can be used to turn datum ids into actual data.

closes #1079

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -379,7 +379,7 @@ class Retrieve(CallbackBase):
         self.resources[resource['uid']] = resource
         handler = self.handler_reg[resource['spec']]
 
-        key = (str(resource['uid']), handler.__name__)
+        key = (str(resource['uid']), handler.__module__, handler.__qualname__)
 
         kwargs = resource['resource_kwargs']
         rpath = resource['resource_path']
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -1,6 +1,7 @@
 """
 Useful callbacks for the Run Engine
 """
+import copy
 from itertools import count
 import warnings
 from collections import deque, namedtuple, OrderedDict, ChainMap
@@ -10,6 +11,9 @@ import os
 from datetime import datetime
 import numpy as np
 import logging
+
+from databroker._core import _sanitize
+
 from ..utils import ensure_uid
 logger = logging.getLogger(__name__)
 
@@ -355,7 +359,7 @@ class LiveTable(CallbackBase):
 
 
 class Retrieve(CallbackBase):
-    def __init__(self, root_map=None, handler_reg=None):
+    def __init__(self, handler_reg, root_map=None):
         if root_map is None:
             root_map = {}
         if handler_reg is None:
@@ -389,9 +393,37 @@ class Retrieve(CallbackBase):
     def datum(self, doc):
         self.datums[doc['datum_id']] = doc
 
-    def retrieve(self, datum_id):
+    def retrieve_datum(self, datum_id):
         doc = self.datums[datum_id]
         resource = self.resources[doc['resource']]
         handler_class = self.handler_reg[resource['spec']]
         key = (str(resource['uid']), handler_class.__name__)
         return self.handlers[key](**doc['datum_kwargs'])
+
+    def fill_event(self, event, fields=True, inplace=True):
+        if fields is True:
+            fields = set(event['data'])
+        elif fields is False:
+            # if no fields, we got nothing to do!
+            # just return back as-is
+            return event
+        elif fields:
+            fields = set(fields)
+
+        if not inplace:
+            ev = _sanitize(event)
+            ev = copy.deepcopy(ev)
+        else:
+            ev = event
+        data = ev['data']
+        filled = ev['filled']
+        for k in set(data) & fields:
+            # Try to fill the data
+            try:
+                v = self.retrieve_datum(data[k])
+                data[k] = v
+                filled[k] = True
+            # If retrieve fails keep going
+            except (ValueError, KeyError):
+                pass
+        return event
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -392,6 +392,6 @@ class Retrieve(CallbackBase):
     def retrieve(self, datum_id):
         doc = self.datums[datum_id]
         resource = self.resources[doc['resource']]
-        handler_class = self.handler_reg[resource]
+        handler_class = self.handler_reg[resource['spec']]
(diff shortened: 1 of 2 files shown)


bluesky/callbacks/core.py
```diff
@@ -351,3 +352,46 @@ def start(self, doc):
     def _print(self, out_str):
         self._rows.append(out_str)
         self._out(out_str)
+
+
+class Retrieve(CallbackBase):
+    def __init__(self, root_map=None, handler_reg=None):
```
danielballan: Why is the handler registry optional?

bluesky/callbacks/core.py
```diff
@@ -351,3 +352,46 @@ def start(self, doc):
...
+    def start(self, doc):
+        self.resources = {}
+        self.handlers = {}
+        self.datums = {}
+
+    def resource(self, resource):
+        self.resources[resource['uid']] = resource
+        handler = self.handler_reg[resource['spec']]
+
+        key = (str(resource['uid']), handler.__name__)
```
danielballan: I think I see what you're getting at here, but I'm not sure this is the best way to achieve that. What else could we key on?

bluesky/callbacks/core.py
```diff
@@ -351,3 +352,46 @@ def start(self, doc):
...
+        if handler_reg is None:
+            handler_reg = {}
+        self.resources = None
+        self.handlers = None
+        self.datums = None
+        self.root_map = root_map
+        self.handler_reg = handler_reg
+
+    def start(self, doc):
+        self.resources = {}
```
danielballan: These are never cleared. Should they be?

bluesky/callbacks/core.py
CJ-Wright: I pulled this out of the databroker code.

bluesky/callbacks/core.py
CJ-Wright: What do you mean they are never cleared? I guess I only wanted to deal with data from the run in question, so I cleared things out.

CJ-Wright:
This most likely should also implement a `fill_event` method.

bluesky/callbacks/core.py
CJ-Wright: Again copied out of the databroker code. I can make it required though.

bluesky/callbacks/core.py
CJ-Wright: https://github.com/NSLS-II/databroker/blob/master/databroker/assets/base_registry.py#L249

CJ-Wright:
I think this is ready for another round of review.

CJ-Wright:
It might be good to have `LiveImage` subclass from this. That way a) we remove the databroker dep and b) we can scrub the images, since we can cache the events and an index and then submit them as the "new" event to the update method, at which point lookup/retrieval will happen.

klauer:
Seems like a useful callback handler to have. 👍 (Can't comment on the implementation details just yet, though - still getting back into the whole NSLS-II stack)

danielballan:
Power-cycling...

bluesky/callbacks/core.py
danielballan: If it is meant to be a long-lived instance, the `resources` can grow large.

CJ-Wright:
Should this return documents, so that it is more like `databroker.documents(fill=True)`

bluesky/callbacks/core.py
danielballan: OK, then let's improve it there as well. I'm concerned that this isn't unique enough. What do you think about:

```py
(str(resource['uid']), handler.__module__, handler.__qualname__)
```

The `__module__` is the one I'm worried about omitting, but we might as well upgrade to `__qualname__` while we're here. Context: https://www.python.org/dev/peps/pep-3155/

CJ-Wright: I like `__qualname__`

bluesky/callbacks/core.py
CJ-Wright: Hmm. Is making the dict anew with every start not enough to clear the resources? Or should this be cleared more frequently? If so, how often?

danielballan: You're right. Now I have no idea what I was thinking with that original comment.

danielballan:
This point has come up a couple times, so we must be talking past each other. What do you mean when you say "return"? It's a callback, so `return` is a dead end.

bluesky/callbacks/core.py
danielballan: Fair. But yes let's make it required.

CJ-Wright:
The return could be convenient for the `LiveImage` rewrite, as we could have `event = super().event(doc)` where the new event would be filled. This would also be helpful for data processing pipelines of any kind since filling events is an important part of the data processing.

bluesky/callbacks/core.py
```diff
@@ -351,3 +356,74 @@ def start(self, doc):
...
+        self.handlers = {}
+        self.datums = {}
+
+    def resource(self, resource):
+        self.resources[resource['uid']] = resource
+        handler = self.handler_reg[resource['spec']]
+
+        key = (str(resource['uid']), handler.__name__)
```
CJ-Wright:
```suggestion
        key = (str(resource['uid']), handler.__module__, handler.__qualname__)
```

CJ-Wright:
This code should really poke the resources and datums at a lower level.

danielballan:
I think we should aim for something like [this](https://gist.github.com/danielballan/cd16936debeb03a674ac1671fe476a0b#file-sketch_analysisstore-py-L33-L72), added to the `event_model` package -- used by both databroker 1.0 and any applicable callbacks in bluesky -- for implementing filling. Maybe that's what you have in mind as well.

CJ-Wright:
Sorry this was the wrong PR to close. I think having both the document and datum access is useful.

danielballan:
Not sure what that means, can you explain?

CJ-Wright:
Having this method would be helpful: https://github.com/NSLS-II/bluesky/pull/1099/files#diff-692eb92ad1a097aaa8ea5c83893ad2e6R396

CJ-Wright:
I think this exists now

PR #1104, opened 2018-10-29
MNT: make Matplotlib actually optional

tacaswell:


pavoljuhas:
I suggest to remove ImportError handling, because it can mask legitimate errors in `mpl_plotting`.
How about delaying matplotlib imports until they are truly needed - 
https://github.com/NSLS-II/bluesky/compare/master...pavoljuhas:defer_matplotlib_import

As an added benefit, this would not import matplotlib with `bluesky.callbacks` when installed, but not otherwise used.

awalter-bnl:
Also looks good to me, sans the seemingly unrelated Travis failures

tacaswell:
There is no way to use the `mpl_plotting` module without having matplotlib installed, I think it is reasonable to have top-level imports.

Put the local imports into a `else` block like we do in the `__init__.py` file.

pavoljuhas:
> There is no way to use the mpl_plotting module without having matplotlib installed, I think it is reasonable to have top-level imports.

Yes, it will then fail as soon as someone instantiates any of its classes.

If you prefer a top-level import in mpl_plotting a proper way is to only import live plotting objects from `callbacks.mpl_plotting` and move them out of `callbacks`, `callbacks.core`.
If they must be there for the sake of backward compatibility, it should rather be thin wrappers with deprecation warning.

tacaswell:
This introduces circular imports that were previously hidden by the bad typo.  Might as will just nuke them as no one has complained about the back-compatibility change yet.

pavoljuhas:
> Might as will just nuke them as no one has complained about the back-compatibility change yet.

Done in #1105.  Compared to master there is an [extra error](https://travis-ci.org/pavoljuhas/bluesky/jobs/448419207#L1549) but that seems related to some invalid plot argument.

tacaswell:
I think we should leave them at the top level, it was just in `.core` that we get the circular imports.

pavoljuhas:
> I think we should leave them at the top level, it was just in .core that we get the circular imports.

I'd like to be able to import the `CallbacksBase` class without importing matplotlib.  As it is setting up a bluesky callback tests for databroker-elasticsearch requires [starting x-window server](https://github.com/NSLS-II/databroker-elasticsearch/blob/c96f712d9cd2a7f2de048d33f9487263eec649d1/.travis.yml#L25-L26).
On my OS X Anaconda the local pytest needs to be run as `python.app /path/to/pytest` because matplotlib osx backend does not start with plain Python.

If you want plotting callbacks at the top level, I very much suggest to move matplotlib import to their `__init__` methods - pavoljuhas/bluesky/defer_matplotlib_import@7be20d09f09f69f66ceb4e03d28592e3b0d4a9fc .

danielballan:
> I'd like to be able to import the CallbacksBase class without importing matplotlib.

I agree with @pavoljuhas on this. We have a similar problem in databroker: I trip over `$DISPLAY` issues often when I'm just trying to use the basic functionality.

> There is no way to use the mpl_plotting module without having matplotlib installed, I think it is reasonable to have top-level imports.

I agree with @tacaswell that a module full of matplotlib-related functionality should require matplotlib at import time. Better to fail early. I see imports buried in `__init__` as a last resort.

See #1107 for my proposal.

bluesky/tests/test_callbacks.py
```diff
@@ -449,3 +451,24 @@ def test_plotting_hints(RE, hw, db):
...
+
+class MyFinder(Finder):
+    def find_module(self, name, path):
+        if name.startswith('matplotlib'):
+            raise ModuleNotFoundError
+
+sys.meta_path.insert(0, MyFinder())
+
+from bluesky.callbacks import CallbackBase
+sys.exit(int('matplotlib' is sys.modules))
```
pavoljuhas: Should be `... in sys.modules`.

This does not help with the case when matplotlib is installed, 
but there is no DISPLAY and no plotting is intended.

tacaswell: yeah, should be '`in'`.  The no display case 'just works' on Matplotlib 3 and is a configuration error on the server for older Matplotlib versions.

danielballan:
Closing in favor of #1105, which I believe is less problematic than it may have seemed at first. Please see [this comment](https://github.com/NSLS-II/bluesky/pull/1105#pullrequestreview-170922815) if you have concerns.

PR #1109, opened 2018-11-01
API: Simplify zmq filtering.

danielballan:
It is possible to have multiple `Publisher` instances pushing documents to the same `Proxy`. A `RemoteDispatcher` can filter which Publisher(s) it is interested in by filtering on a prefix. Previously, this prefix was automatically cobbled together from the Publisher's hostname, process ID (pid), and the Python `id()` of its `RunEngine` instance. This PR replacing that automatic scheme was a single, user-configurable bytestring.

While we are breaking the API, I also propose deprecating automatic subscription. See commit message for rationale.

@Sasaank and @CJ-Wright -- I found it easier to start with passing tests (on the master branch) and make the simplest possible change to the tests, rather than heavily reworking them. I apologize for not building on your PR, which would have been my preference and usual way of operating. I found the test changes are too difficult to follow. Would you like to add a commit to this branch testing the `prefix` functionality specifically?


CJ-Wright:
Thank you very much for this! I'll add a PR to your PR.

danielballan:
Power-cycled...

CJ-Wright:
Thank you!

PR #1116, opened 2018-11-15
Revisit logging.

danielballan:
In the course or writing caproto, I now have better-informed opinions about how libraries should interact with Python's logging module.

This PR:

* includes a nice default logging handler, vendored from caproto, which is turn is a vendor/variation on Tornado's built-in handler
* avoids imposing log levels at import time
* adds DEBUG-level logging next to the `msg_hook`, which imposes a 4% performance hit on message throughput (when DEBUG logging is off) which seems worth paying to me

Example:

```python
In [5]: RE(count([det]))
Out[5]: ('daa5f363-1c44-4fd1-8023-c3289e7c8f81',)

In [6]: import logging

In [7]: logging.getLogger('bluesky').setLevel('DEBUG')

In [8]: RE(count([det]))
[I 10:15:28.478 run_engine:82] Change state on <bluesky.run_engine.RunEngine object at 0x7f76b49cc278> from 'idle' -> 'running'
[D 10:15:28.478 run_engine:1184] stage: (SynGauss(name='det', value=1.0, timestamp=1542294919.6508715)), (), {}
[D 10:15:28.478 run_engine:1184] open_run: (None), (), {'detectors': ['det'], 'num_points': 1, 'num_intervals': 0, 'plan_args': {'detectors': ["SynGauss(name='det', value=1.0, timestamp=1542294919.6508715)"], 'num': 1}, 'plan_name': 'count', 'hints': {'dimensions': [(('time',), 'primary')]}}
[D 10:15:28.478 run_engine:1356] Starting new with uid '423c2d6e-690c-4904-a87a-ee82fb08b33f'
[D 10:15:28.481 run_engine:1378] Emitted RunStart (uid='423c2d6e-690c-4904-a87a-ee82fb08b33f')
[D 10:15:28.481 run_engine:1184] checkpoint: (None), (), {}
[D 10:15:28.481 run_engine:1184] trigger: (SynGauss(name='det', value=1.0, timestamp=1542294919.6508715)), (), {'group': 'trigger-7921d8'}
[D 10:15:28.481 run_engine:1999] The object SynGauss(name='det', value=1.0, timestamp=1542294928.4819157) reports trigger is done with status True.
[D 10:15:28.482 run_engine:1184] wait: (None), (), {'group': 'trigger-7921d8'}
[D 10:15:28.482 run_engine:1184] create: (None), (), {'name': 'primary'}
[D 10:15:28.482 run_engine:1184] read: (SynGauss(name='det', value=1.0, timestamp=1542294928.4819157)), (), {}
[D 10:15:28.482 run_engine:1184] save: (None), (), {}
[D 10:15:28.485 run_engine:1693] Emitted Event Descriptor with name 'primary' containing data keys dict_keys(['det']) (uid='3bfb3fd9-dd26-4cf5-b7ab-11223380e760')
[D 10:15:28.486 run_engine:1725] Emitted Event with data keys dict_keys(['det']) (uid='2a308700-548b-4c33-8063-2bfa2c460d7b')
[D 10:15:28.486 run_engine:1184] close_run: (None), (), {'exit_status': None, 'reason': None}
[D 10:15:28.486 run_engine:1411] Stopping run '423c2d6e-690c-4904-a87a-ee82fb08b33f'
[D 10:15:28.487 run_engine:1436] Emitted RunStop (uid='2524a956-1d95-44c3-bf7c-f8b3b0016939')
[D 10:15:28.487 run_engine:1184] unstage: (SynGauss(name='det', value=1.0, timestamp=1542294928.4819157)), (), {}
[I 10:15:28.487 run_engine:82] Change state on <bluesky.run_engine.RunEngine object at 0x7f76b49cc278> from 'running' -> 'idle'
Out[8]: ('423c2d6e-690c-4904-a87a-ee82fb08b33f',)
```

This suggests some obvious additional improvements. (For example, we should give `RunEngine` a nice repr.) But I propose keeping the scope of this PR limited and iterating from there if we have time. We could do a _lot_ more but I want to make sure this lands in the next release.

### DIFF
--- a/bluesky/log.py
+++ b/bluesky/log.py
@@ -155,6 +155,21 @@ def set_handler(file=sys.stdout, datefmt='%H:%M:%S', color=True):
     -------
     handler : logging.Handler
         The handler, which has already been added to the 'bluesky' logger.
+
+    Examples
+    --------
+    Log to a file.
+
+    >>> set_handler(file=open('/tmp/what_is_happening.txt'))
+
+    Include the date along with the time. (The log messages will always include
+    microseconds, which are configured separately, not as part of 'datefmt'.)
+
+    >>> set_handler(datefmt="%Y-%m-%d %H:%M:%S")
+
+    Turn off ANSI color codes.
+
+    >>> set_handler(color=False)
     """
     global current_handler
     handler = logging.StreamHandler(file)


bluesky/log.py
```diff
@@ -0,0 +1,157 @@
...
+            record.exc_text = self.formatException(record.exc_info)
+        if record.exc_text:
+            formatted = '{}\n{}'.format(formatted.rstrip(), record.exc_text)
+        return formatted.replace("\n", "\n    ")
+
+
+plain_log_format = "[%(levelname)1.1s %(asctime)s.%(msecs)03d %(module)s:%(lineno)d] %(message)s"
+color_log_format = ("%(color)s[%(levelname)1.1s %(asctime)s.%(msecs)03d "
+                    "%(module)s:%(lineno)d]%(end_color)s %(message)s")
+log_date_format = "%H:%M:%S"
```
mrakitin: Can the date format be configurable here?

bluesky/run_engine.py
```diff
@@ -222,10 +222,9 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
 
         # Make a logger for this specific RE instance, using the instance's
         # Python id, to keep from mixing output from separate instances.
-        logger_name = "{name}_id{id}".format(name=__name__, id=id(self))
+        logger_name = "bluesky.RE.{id}".format(id=id(self))
         self.log = logging.getLogger(logger_name)
-        self.log.setLevel(logging.DEBUG)
-        self.verbose = False  # a convenience property, setting log.disabled
+        self.verbose = True  # a convenience property, setting log.disabled
```
mrakitin: the comment should be `a convenience property, setting log.enabled`, right?

danielballan: No, there is no `log.enabled`. This sets `log.disabled` (to `False`).

bluesky/log.py
danielballan: I don't think there is a nice way to expose that option. This module code adds a handler to the 'bluesky' logger at import time. Some function that _mutates_ that handler (`set_log_date_format` or something) after the fact feels dirty to me. Better for the user to write their own formatter and handler, reusing some of the pieces here if they like -- in my opinion.

danielballan: OK, I am convinced this is actually a really good suggestion. Stand by for commit. :- D

danielballan:
OK, I added a new function `set_handler` to replace `color_logs`, which allows the user to aim the logs at a file (other than the default `sys.stdout`), customize `datefmt`, or turn off ANSI color codes. For anything more invasive, like using `logging.handlers.RotatingFileHandler` instead of a simple `StreamHandler`, the user should just add their own handler. I consider `set_handler` a pragmatic convenience that is more user-focused than dev-focused, because devs can just use the normal Python `logging` API, and anything we add would just be more confusing.

danielballan:
This will be of interest to @lcdesilva.

PR #1117, opened 2018-11-15
ENH: Add optional file param to ts_msg_hook.

danielballan:
This will be passed through to print, and it has the same default value
is as print.

This allows usages like:

```
file = open('my_log_file', 'a')
import functools
func = functools.partial(ts_msg_hook, file=file)
RE.msg_hook = func
```

Somewhat moot once we can do with via the logger proposed in #1116, but
still probably worth doing.

### DIFF
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -974,14 +974,14 @@ def ensure_uid(doc_or_uid):
         return doc_or_uid
 
 
-def ts_msg_hook(msg):
+def ts_msg_hook(msg, file=sys.stdout):
     t = '{:%H:%M:%S.%f}'.format(datetime.datetime.now())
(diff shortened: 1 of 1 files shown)


stuartcampbell:
> This allows usages like:
> 
> ```
> file = open('my_log_file', 'a')
> import functools
> func = functools.partial(ts_msg_hook, file=file)
> RE.msg_hook = func
> ```

Is it worth adding this snippet somewhere to the documentation, as I think the use of things like partial do not spring to the general users mind.

danielballan:
Good suggestion @stuartcampbell. I added a commit illustrating this using a normal function. People who know about `functools.partial` will figure out for themselves that they can do that if they want to.

PR #1120, opened 2018-11-20
DOC: We need to pass a motor to scans

tacaswell:


doc/source/tutorial.rst
```diff
@@ -641,12 +641,12 @@ why.) Examples:
 
     from bluesky.plans import scan
 
-    def coarse_and_fine(detectors, start, stop):
+    def coarse_and_fine(detectors, mtr, start, stop):
```
danielballan: The vowel keys on your keyboard are broken again. :-D

PR #1121, opened 2018-11-20
Bluesky proxy verbose

ke-zhang-rd:
## Description

Add --verbose and --logfile feature to bluesky-proxy
## Motivation and Context

User friendly 
## How Has This Been Tested?

In terminal, run ipython
$ ipython --matplotlib=qt5
In ipython, run scirpts which will start proxy and RemoteDispatcher, in_port and out_port needed
%run -i bluesky-0MQ-proxy 5577 5578
If we add --verbose and --logfile=temp.log
%run -i bluesky-0MQ-proxy 5577 5578 --verbose --logfile=temp.log

### DIFF
--- a/scripts/bluesky-0MQ-proxy
+++ b/scripts/bluesky-0MQ-proxy
@@ -7,6 +7,13 @@ import sys
 from bluesky.callbacks.zmq import RemoteDispatcher
 
 def start_dispatcher(logfile):
+    """The dispatcher function
+    Parameters
+    ----------
+    logfile : string
+        string come from user command. ex --logfile=temp.log
+        logfile will be "temp.log". logfile could be empty.
+    """
     d = RemoteDispatcher('localhost:'+str(out_port))
     if logfile:
         logger = logging.getLogger('logfile')
@@ -16,7 +23,14 @@ def start_dispatcher(logfile):
         logger.addHandler(hdlr)
         logger.setLevel(logging.INFO) #Only write to file with logger.info or worse
         def log_writer(name, doc):
+            """logger's wrapper function
+                This function will be use to fit .subscribe() method.
+                It has two arguments as .subscribe expect. Inside, it
+                call logger.info to write doc which is a dict as a str
+                into logfile
+            """
             logger.info(str(doc))
+            print("Log",name ,"doc to", logfile)
         d.subscribe(log_writer) # Subscribe log writer
     else:
         d.subscribe(print)
@@ -41,10 +55,10 @@ if __name__ == "__main__":
         print("in_port:", in_port)
         print("out_port:", out_port)
         if args.verbose > 2:
-            obj = threading.Thread(target=start_dispatcher,
-                                    args=(args.logfile,),
-                                    daemon = True).start()  # Set daemon to all ipython exit
-                                                            # kill all threads
+            threading.Thread(target=start_dispatcher,
+                                args=(args.logfile,),
+                                daemon = True).start()  # Set daemon to all ipython exit
+                                                        # kill all threads
     print("Loading...")
     from bluesky.callbacks.zmq import Proxy  # this takes a couple seconds
     print("Connecting...")
--- a/scripts/bluesky-0MQ-proxy
+++ b/scripts/bluesky-0MQ-proxy
@@ -7,6 +7,10 @@ import sys
 from bluesky.callbacks.zmq import RemoteDispatcher
 from bluesky.log import set_handler
 
+
+logger = logging.getLogger('bluesky')
+
+
 def start_dispatcher(logfile):
     """The dispatcher function
     Parameters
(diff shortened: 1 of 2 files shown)


scripts/bluesky-0MQ-proxy
```diff
@@ -1,5 +1,40 @@
...
+    """The dispatcher function
+    Parameters
+    ----------
+    logfile : string
+        string come from user command. ex --logfile=temp.log
+        logfile will be "temp.log". logfile could be empty.
+    """
+    d = RemoteDispatcher('localhost:'+str(out_port))
+    if logfile:
+        logger = logging.getLogger('logfile')
```
danielballan: The logger name is more than a label for humans; it defines a kind of "pipeline" for log messages. All the messages logged to a `'bluesky'` or `'bluesky.dispatcher'` or `'bluesky.run_engine.msg'` will "flow" to the `'bluesky'` logger because they begin with `'bluesky.*`.

At import time, we get the `'bluesky'` logger and we add a handler with a nice formatter:

https://github.com/NSLS-II/bluesky/blob/abd8552881e17d3e80313922ec040269c9780586/bluesky/log.py#L189-L190

Therefore, when messages are logged to _any_ logger than begins with `bluesky.*`, those messages will be processed by that handler handler. We only need to set up one handler at the "top" and then all of the different loggers can just use that.

So, in this case, we can do something like:

```
from bluesky import set_handler

logger = logging.getLogger('bluesky.dispatcher')
logger.setLevel('DEBUG')
if logfile:
    # Direct the logging output to our file.
    set_handler(file=logfile)
# else use the default handler, configured automatically at import time,
# which will write to the stdout
def log_write(name, doc):
    ...  # same as you have written
```

As a separate point, notice that I have suggested 'DEBUG' not 'INFO'. Since this will log _the full document_ of everything that we see, I think we should call it DEBUG logging because I can't imagine anything more verbose that we could be doing.

scripts/bluesky-0MQ-proxy
```diff
@@ -1,5 +1,40 @@
...
+
+def start_dispatcher(logfile):
+    """The dispatcher function
+    Parameters
+    ----------
+    logfile : string
+        string come from user command. ex --logfile=temp.log
+        logfile will be "temp.log". logfile could be empty.
+    """
+    d = RemoteDispatcher('localhost:'+str(out_port))
```
danielballan: Another minor thing: it's more readable to write this as `f'localhost:{out_port}'`. That employs a feature only available in Python 3.6 or higher, but since ophyd just dropped Python 3.5 support, I see no reason that bluesky can't do the same.

scripts/bluesky-0MQ-proxy
```diff
@@ -9,10 +44,21 @@ if __name__ == "__main__":
                         help='port that RunEngines should broadcast to')
     parser.add_argument('out_port', type=int, nargs=1,
                         help='port that subscribers should subscribe to')
+    parser.add_argument('--verbose', '-v', action='count',
+                        help="Show more log messages. (Use -vvv for even more.)")
+    parser.add_argument('--logfile', type=str,
+                        help="Write logfile ")
```
danielballan: Mention that if it's not set, `--verbose` logs to the screen.

scripts/bluesky-0MQ-proxy
ke-zhang-rd: My point here is when user doesn't offer --logfile option which means we just "output" to screen. Could I just use `print()` to avoid create any obj related logging which is `else` part here.

scripts/bluesky-0MQ-proxy
ke-zhang-rd: I see

scripts/bluesky-0MQ-proxy
ke-zhang-rd: When --logfile is not set, `logfile` is `None` which will jump to `else`. `else` is using `print` as default to screen

scripts/bluesky-0MQ-proxy
```diff
@@ -1,6 +1,78 @@
...
+
+def start_dispatcher(logfile, poison_pill):
+    """The dispatcher function
+    Parameters
+    ----------
+    logfile : string
+        string come from user command. ex --logfile=temp.log
+        logfile will be "temp.log". logfile could be empty.
+    """
+    d = RemoteDispatcher(f'localhost:{out_port}')
```
mrakitin: Should we make the machine configurable, rather having `localhost` hard-coded?
Also, one-letter vars is not considered a good practice, consider changing it to something more verbose.

scripts/bluesky-0MQ-proxy
```diff
@@ -1,6 +1,78 @@
...
+        print("self._logfile")
+        print(self._logfile)
+        if self._logfile:
+            print("inside log")
+            logger = logging.getLogger('logfile')
+            hdlr = logging.FileHandler(self._logfile)
+            formatter = logging.Formatter('%(asctime)s %(levelname)s %(message)s')
+            hdlr.setFormatter(formatter)
+            logger.addHandler(hdlr)
+            logger.setLevel(logging.INFO) #Only write to file with logger.info or worse
```
mrakitin: stylistic comment - there should be a blank space after `#`, and 2 blank spaces between the end of the code string and the comment.

bluesky/callbacks/zmq.py
```diff
@@ -242,13 +242,25 @@ def __init__(self, address, *, prefix=None,
         self._socket.connect(url)
         self._socket.setsockopt_string(zmq.SUBSCRIBE, "")
         self._task = None
-
+        self._poision_pill = poison_pill
```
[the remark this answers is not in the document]
ke-zhang-rd: correct! this is a typo. Thanks

scripts/bluesky-0MQ-proxy
danielballan: In general the user can point `RemoteDispatcher` at whatever they like. _This_ `RemoteDispatcher`'s whole job is to listen to the `Proxy` that we are about to start below (so, definitely on `localhost`) and generate log messages.

PR #1122, opened 2018-11-20
Make `CallbackBase` composable

CJ-Wright:
The `CallbackBase` has returned to it's home repo of Bluesky...

This allows subclasses of `CallbackBase` to return mutated documents.
This could be very useful for things like live databroker exporting
which would mutate the resource and datum documents for the new
filepaths to the copied files.

This seems more flexable and has some nice usecases for very simple
pipelines/document mutation

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -42,7 +42,13 @@ LiveMesh = _deprecate_import_name("LiveMesh")
 class CallbackBase:
     def __call__(self, name, doc):
         "Dispatch to methods expecting particular doc types."
-        return getattr(self, name)(doc)
+        # if callback returns it will return a new dict document
+        ret = getattr(self, name)(doc)
+        if ret:
+            return name, ret
+        # if the return is None (very likely for most callbacks)
+        # give back the original name, doc pair
+        return name, doc
 
     def event(self, doc):
         pass
--- a/bluesky/tests/test_callbacks.py
+++ b/bluesky/tests/test_callbacks.py
@@ -456,11 +456,11 @@ def test_plotting_hints(RE, hw, db):
 
 
 def test_callback_return(RE, hw):
-    class c(CallbackBase):
+    class C(CallbackBase):
         def start(self, doc):
             doc.update({'hello': 'world'})
             return doc
     L = []
-    cc = c()
+    cc = C()
     RE(stepscan(hw.det, hw.motor), {'all': lambda *x: L.append(cc(*x))})
     assert L[0][1]['hello'] == 'world'
--- a/bluesky/tests/test_callbacks.py
+++ b/bluesky/tests/test_callbacks.py
@@ -453,3 +453,14 @@ def test_plotting_hints(RE, hw, db):
     RE(grid_scan([hw.det], hw.motor1, -1, 1, 2, hw.motor2, -1, 1, 2,
                  True, hw.motor3, -2, 0, 2, True))
     assert dc.start[-1]['hints'] == hint
+
+
+def test_callback_return(RE, hw):
+    class c(CallbackBase):
+        def start(self, doc):
+            doc.update({'hello': 'world'})
+            return doc
+    L = []
+    cc = c()
+    RE(stepscan(hw.det, hw.motor), {'all': lambda *x: L.append(cc(*x))})
+    assert L[0][1]['hello'] == 'world'


bluesky/callbacks/core.py
```diff
@@ -42,7 +42,13 @@ def f(*args, **kwargs):
 class CallbackBase:
     def __call__(self, name, doc):
         "Dispatch to methods expecting particular doc types."
-        return getattr(self, name)(doc)
```
danielballan: During our discussion on Slack where I encouraged this PR, I didn't realize that we were already returning the output of the method. How forward-looking of us! :-D

Given that, I don't understand why any modification is needed. Can't the subclass's methods return whatever they want, including `(name, doc)`. Adding something extra here just strikes me as potentially confusing magic.

CJ-Wright: 1. It seems to be more efficient to have the name handled at the top level, where we have actual access rather than in the lower level where we need to copy the method name again.
2. This means that every subclass needs to be a no-op for more than one document router to be put together. That seems a bit counter productive.

danielballan:
I didn't intend my comment to close debate. I am still presently against merging but for more debate.

bluesky/callbacks/core.py
danielballan: 1. You mean like:

```py
class ScienceDoer(CallbackBase):
    def start(self, doc):
        ...
        return 'start', new_doc
```

as opposed to

```py
class ScienceDoer(CallbackBase):
    def start(self, doc):
        ...
        return super().start(new_doc) .   # base class will return ('start', new_doc)
```

2. Can you give me a very succinct example?

CJ-Wright: Returning from super might be ok.

```python
ls = LiveScatter(...)
cb2 = OtherCallback(...)
L = []
RE.subscribe(lambda *x: L.append(cb2(*ls(*x))))
```
Does not work if LiveScatter doesn't send out a stream of documents.

danielballan: > Returning from super might be ok.

Wait...I'm showing you how I think your PR would work, not an alternative. How else would you involve the base class other than to call `super()`?

danielballan: I'm illustrating that adding complexity to the base class does not actually make the subclass simpler. Sure you don't have to type the method name again, as you say, but I think you would need to call `super()` which is more verbose and, in my opinion, also more opaque / harder to follow. See what I mean? Am I wrong about what the subclass would work? Maybe you should type on your ideal version of `ScienceDoer`'s return statement.

CJ-Wright: In this PR the subclass version of start returns a dict. That dict then gets associated with a name in the call method. Most of the document routers so far haven't overrided the call method.

```python
class ScienceDoer(CallbackBase):
    def start(self, doc):
        doc = doc.update(hello='world')
        return doc
```
This modifies the start document, it then ships out the new document, which when call finishes looks like a proper name, document pair

danielballan:
@CJ-Wright Reflecting on this, you've convinced me that there's no harm and some benefit to the change. I have some quibbles about the exact logic that I will leave in-line, but I think there is a path to merging.

bluesky/callbacks/core.py
```diff
@@ -42,7 +42,13 @@ def f(*args, **kwargs):
...
+        # if callback returns it will return a new dict document
+        ret = getattr(self, name)(doc)
+        if ret:
```
danielballan: This `if` statement raises a red flag for me. I would like to `return name, getattr(self, name)(doc)` in all cases. If a callback was not designed for a pipeline (and thus returns `None`) it should be up to the caller to say, "OK then pass through the original `doc`."

bluesky/callbacks/core.py
danielballan: What I was missing here is that the _caller_ of `start(doc)` is typically accessing it indirectly via `__call__(name, doc)`. It follows that no call to `super()` is needed: the subclass can just return the `doc` and rely on `__call__` to do the rest.

bluesky/callbacks/core.py
tacaswell: I agree with @danielballan from an explicitness PoV.  But I also agree with @CJ-Wright from a 'gentle migration PoV' this is a reasonable path forward.

Maybe

```python
ret = getattr(self, name)(doc)
return name, ret if ret is not None else doc
```

danielballan: Is that smoother change from the current behavior? How so?

danielballan: Is that smoother change from the current behavior? How so?

tacaswell: This is an API change no matter what (as now we are returning a tuple with `name` rather than what ever was happening before), however doing "the right thing" if the user returns `None` means that we do not have to go through and add dummy `return doc` to every callback (the 'gentle migration' is on the produces side).  This is a pattern we use in caproto too, returning `None` means "no opinion" letting the outer function have opinions on what to do next (or conversely letting the inner function not have opinions about what to do next).

Testing on `None` seems more defensive than testing on truth.  We don't have any valid documents that are currently 'falsy' that may not always be true (a 0 row page?) or maybe we want to let non-document things flow through this (should `stop` be allowed to return fit values?).

If some pipeline wants a 'stop propogating' flag it can either use `False` or it's own sentinel. 

On the other-other hand, maybe we want to have that discussion here?

danielballan: To avoid breakage, let's leave this alone and make a new thing, which we can use in the callbacks rewrite.

danielballan: There is more than one good reason to do this. See https://github.com/NSLS-II/event-model/pull/29

danielballan:
Sure, that seems easier than managing an API change cycle.

PR #1124, opened 2018-11-29
DOC: Label units on delay parameter.

danielballan:


awalter-bnl:
thanks for the quick update @danielballan, this is a minor doc change so merging it quickly

PR #1125, opened 2018-12-07
FIX: improve how we interact with the asyncio event loop

### LINKED ISSUE #1062: improve how we interact with the asyncio event loop
We currently start/stop the asyncio event loop in `__call__` and `resume` which is at best rude and at worst bad-practice.   This prevents us from running the RE inside of jupyter with tornado >= 5.  

This is a sketch of a path to using a running event loop and still getting the blocking behavior we need it `__call__`.  In the IPython case we will need to spawn a event loop on a background thread and then start it, in the case of jupyter we will have to bind the existing loop to the RE and use that. 

This also solves the problem of how we use `caproto.asyncio` with the RE (as it also just uses the background loop for the monitors)  

The kicker integration will need to move into `__call__` but I think I see a path to that that will end up being better anyway (basically re-use the logic that ipython already does it to start the event loop and then kill it on cue).  I suspect this will also fix the issue of open the modal window pausing the scan!

```python
import asyncio
import threading

loop = asyncio.new_event_loop()

def runner():
    loop.run_forever()

th = threading.Thread(target=runner)
th.start()

async def sleepy_worker(n):
    for j in range(n):
        print(f'sleeping loop {j} of {n}')
        await asyncio.sleep(1)
    return n

def sync_worker2(n):
    print('about to create task')    
    h = asyncio.run_coroutine_threadsafe(sleepy_worker(n), loop=loop)
    print('created task')        
    h.add_done_callback(lambda _: print('finished, kill kicker here'))
    if not h.done():
        # start kicker
        # qt_event_loop.exec_()
        pass
    print('about to waiting')
    return h.result()
    print('done waiting')


# sync_worker2(5)

```

### LINKED ISSUE #1087: Matplotlib import errors
Having resolved https://github.com/NSLS-II/bluesky/issues/1086 I now have a different Matplotlib error. In the Jupyter notebook I hit:
```python
$ install_kicker()

NotImplementedError                       Traceback (most recent call last)
<ipython-input-4-7e8b6286ffe4> in <module>()
      7 # Make plots update live while scans run.
      8 from bluesky.utils import install_kicker
----> 9 install_kicker()

~/bluesky-tutorial/lib/python3.6/site-packages/bluesky/utils.py in install_kicker(loop, update_rate)
    768     else:
    769         raise NotImplementedError("The matplotlib backend {} is not yet "
--> 770                                   "supported.".format(backend))
    771 
    772 

NotImplementedError: The matplotlib backend module://ipykernel.pylab.backend_inline is not yet supported.
```

and see the message:
```python
(bluesky-tutorial) Robins-Air:Bluesky robincole$ jupyter notebook --matplotlib=qt5
[W 13:48:41.567 NotebookApp] Unrecognized alias: '--matplotlib=qt5', it will probably have no effect.
```

In ipython run with `ipython --matplotlib=qt5` I hit:
```
(bluesky-tutorial) Robins-MacBook-Air:~ robincole$ ipython --matplotlib=qt5
Python 3.6.5 |Anaconda, Inc.| (default, Apr 26 2018, 08:42:37)
IPython 6.5.0 -- An enhanced Interactive Python. Type '?' for help.
[TerminalIPythonApp] WARNING | Eventloop or matplotlib integration failed. Is matplotlib installed?
ImportError                               Traceback (most recent call last)
~/bluesky-tutorial/lib/python3.6/site-packages/IPython/core/interactiveshell.py in enable_matplotlib(self, gui)
   3047                 gui, backend = pt.find_gui_and_backend(self.pylab_gui_select)
   3048
-> 3049         pt.activate_matplotlib(backend)
   3050         pt.configure_inline_support(self, backend)
   3051

~/bluesky-tutorial/lib/python3.6/site-packages/IPython/core/pylabtools.py in activate_matplotlib(backend)
    308     matplotlib.rcParams['backend'] = backend
    309
--> 310     import matplotlib.pyplot
    311     matplotlib.pyplot.switch_backend(backend)
    312

~/bluesky-tutorial/lib/python3.6/site-packages/matplotlib/pyplot.py in <module>()
    113 ## Global ##
    114
--> 115 _backend_mod, new_figure_manager, draw_if_interactive, _show = pylab_setup()
    116
    117 _IP_REGISTERED = None

~/bluesky-tutorial/lib/python3.6/site-packages/matplotlib/backends/__init__.py in pylab_setup(name)
     60     # imports. 0 means only perform absolute imports.
     61     backend_mod = __import__(backend_name, globals(), locals(),
---> 62                              [backend_name], 0)
     63
     64     # Things we pull in from all backends

~/bluesky-tutorial/lib/python3.6/site-packages/matplotlib/backends/backend_qt5agg.py in <module>()
     13
     14 from .backend_agg import FigureCanvasAgg
---> 15 from .backend_qt5 import (
     16     QtCore, QtGui, QtWidgets, _BackendQT5, FigureCanvasQT, FigureManagerQT,
     17     NavigationToolbar2QT, backend_version)

~/bluesky-tutorial/lib/python3.6/site-packages/matplotlib/backends/backend_qt5.py in <module>()
     17     _Backend, FigureCanvasBase, FigureManagerBase, NavigationToolbar2,
     18     TimerBase, cursors, ToolContainerBase, StatusbarBase)
---> 19 import matplotlib.backends.qt_editor.figureoptions as figureoptions
     20 from matplotlib.backends.qt_editor.formsubplottool import UiSubplotTool
     21 from matplotlib.figure import Figure

~/bluesky-tutorial/lib/python3.6/site-packages/matplotlib/backends/qt_editor/figureoptions.py in <module>()
     18 import matplotlib
     19 from matplotlib import cm, colors as mcolors, markers, image as mimage
---> 20 import matplotlib.backends.qt_editor.formlayout as formlayout
     21 from matplotlib.backends.qt_compat import QtGui
     22

~/bluesky-tutorial/lib/python3.6/site-packages/matplotlib/backends/qt_editor/formlayout.py in <module>()
     52
     53 from matplotlib import colors as mcolors
---> 54 from matplotlib.backends.qt_compat import QtGui, QtWidgets, QtCore
     55
     56

~/bluesky-tutorial/lib/python3.6/site-packages/matplotlib/backends/qt_compat.py in <module>()
    156         break
    157     else:
--> 158         raise ImportError("Failed to import any qt binding")
    159 else:  # We should not get there.
    160     raise AssertionError("Unexpected QT_API: {}".format(QT_API))
```

Sorry about this..!

danielballan:
Do not merge. Needs work.

Closes #1062
Closes #1087

TODO:
 - [x] rename `run_matplotlib_qApp`, it also handles the non Qt cases (semi-) gracefully
 - [x] make stop, request_pause, abort, halt use the same pattern as request_suspend
 - [x] make sure it actually works in a jupyter notebook and figures are responsive
 - [x] discuss `wait_for` api changes
 - [x] remove `_delayed_partial` which is a poorly implemented version of `threading.Timer`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,6 @@ install:
   - pip install -r requirements.txt
   - pip install -r dev-requirements.txt
   - pip install -U ophyd --pre
-  - pip install pyqt5
   - pip install codecov
   - python setup.py install
   # Need to clean the python build directory (and other cruft) or pytest is
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,10 @@ matrix:
     - os: osx
       language: generic
       env: PYTHON=3.7
+    - os: linux
+      python: nightly
+  allow_failures:
+    - python: nightly
 
 before_install:
   - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ dist: xenial
 language: python
 sudo: false
 
-services:
-  - xvfb
-
 env:
   global:
     # Doctr deploy key for NSLS-II/NSLS-II.github.io
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -2,22 +2,32 @@ from collections import ChainMap
 from cycler import cycler
 import numpy as np
 import warnings
+import functools
 from .core import CallbackBase, get_obj_fields
-from matplotlib.backends.qt_compat import QtCore
 
 
-class QtAwareCallback(CallbackBase):
+# use function + LRU cache to hide Matplotib import until needed
+@functools.lru_cache(maxsize=1)
+def _get_teleporter():
+    from matplotlib.backends.qt_compat import QtCore
 
     class Teleporter(QtCore.QObject):
         name_doc_escape = QtCore.Signal(str, dict, bool)
+    return Teleporter
 
-    def __init__(self, *args, maybe_use_teleporter=True, **kwargs):
-        self.__teleporter = self.Teleporter()
-        self.__teleporter.name_doc_escape.connect(self.__call__)
+
+class QtAwareCallback(CallbackBase):
+    def __init__(self, *args, use_teleporter=True, **kwargs):
+        if use_teleporter:
+            Teleporter = _get_teleporter()
+            self.__teleporter = Teleporter()
+            self.__teleporter.name_doc_escape.connect(self.__call__)
+        else:
+            self.__teleporter = None
         super().__init__(*args, **kwargs)
 
     def __call__(self, name, doc, escape=False):
-        if not escape:
+        if not escape and self.__teleporter is not None:
             self.__teleporter.name_doc_escape.emit(name, doc, True)
         else:
             return CallbackBase.__call__(self, name, doc)
@@ -388,8 +398,8 @@ class LiveGrid(QtAwareCallback):
                             extent=extent, aspect=self.aspect,
                             origin='lower')
 
-        # make sure the 'positive direction' of the axes matches what is defined in
-        #axes_positive
+        # make sure the 'positive direction' of the axes matches what
+        # is defined in axes_positive
         xmin, xmax = self.ax.get_xlim()
         if ((xmin > xmax and self.x_positive == 'right') or
                 (xmax > xmin and self.x_positive == 'left')):
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -17,7 +17,10 @@ def _get_teleporter():
 
 
 class QtAwareCallback(CallbackBase):
-    def __init__(self, *args, use_teleporter=True, **kwargs):
+    def __init__(self, *args, use_teleporter=None, **kwargs):
+        if use_teleporter is None:
+            import matplotlib
+            use_teleporter = 'qt' in matplotlib.get_backend()
         if use_teleporter:
             Teleporter = _get_teleporter()
             self.__teleporter = Teleporter()
@@ -75,7 +78,7 @@ class LivePlot(QtAwareCallback):
     def __init__(self, y, x=None, *, legend_keys=None, xlim=None, ylim=None,
                  ax=None, fig=None, epoch='run', **kwargs):
         import matplotlib.pyplot as plt
-        super().__init__(use_teleporter=kwargs.pop('use_teleporter', True))
+        super().__init__(use_teleporter=kwargs.pop('use_teleporter', None))
         if fig is not None:
             if ax is not None:
                 raise ValueError("Values were given for both `fig` and `ax`. "
@@ -216,7 +219,7 @@ class LiveScatter(QtAwareCallback):
     """
     def __init__(self, x, y, I, *, xlim=None, ylim=None,
                  clim=None, cmap='viridis', ax=None, **kwargs):
-        super().__init__(use_teleporter=kwargs.pop('use_teleporter', True))
+        super().__init__(use_teleporter=kwargs.pop('use_teleporter', None))
         import matplotlib.pyplot as plt
         import matplotlib.colors as mcolors
         if ax is None:
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -24,21 +24,19 @@ def _maybe_use_teleporter(cls):
     @functools.wraps(orig_init)
     def __init__(self, *args, maybe_use_teleporter=True, **kwargs):
         orig_init(self, *args, **kwargs)
-        self._orig_init = orig_init
-        self._orig_call = orig_call
+        _orig_call = orig_call
+        if not hasattr(self, '__teleporters'):
+            self.__teleporters = {}
+
         if Teleporter is not None and maybe_use_teleporter:
-            if not hasattr(self, '__teleporters'):
-                self.__teleporters = {}
 
             def inner_func(name, doc, *, target=self):
-                self._orig_call(target, name, doc)
+                _orig_call(target, name, doc)
 
             teleporter = Teleporter()
             teleporter.name_doc.connect(inner_func)
 
             self.__teleporters[cls_name] = teleporter
-        else:
-            self.__teleporter = None
 
     @functools.wraps(orig_call)
     def __call__(self, name, doc):
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -30,8 +30,8 @@ def _maybe_use_teleporter(cls):
             if not hasattr(self, '__teleporters'):
                 self.__teleporters = {}
 
-            def inner_func(name, doc):
-                orig_call(self, name, doc)
+            def inner_func(name, doc, *, target=self):
+                self._orig_call(target, name, doc)
 
             teleporter = Teleporter()
             teleporter.name_doc.connect(inner_func)
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -17,12 +17,9 @@ class QtAwareCallback(CallbackBase):
         super().__init__(*args, **kwargs)
 
     def __call__(self, name, doc, escape=False):
-        print(f'qta {type(self)}, {id(self)}, {escape}')
         if not escape:
-            print('about to emit')
             self.__teleporter.name_doc_escape.emit(name, doc, True)
         else:
-            print('calling up the stack')
             return CallbackBase.__call__(self, name, doc)
 
 
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -2,10 +2,58 @@ from collections import ChainMap
 from cycler import cycler
 import numpy as np
 import warnings
-
+import functools
 from .core import CallbackBase, get_obj_fields
 
 
+try:
+    from matplotlib.backends.qt_compat import QtCore
+
+    class Teleporter(QtCore.QObject):
+        name_doc = QtCore.Signal(str, dict)
+except ImportError:
+    Teleporter = None
+
+
+def _maybe_use_teleporter(cls):
+    orig_init = cls.__init__
+    orig_call = cls.__call__
+
+    cls_name = cls.__name__
+
+    @functools.wraps(orig_init)
+    def __init__(self, *args, maybe_use_teleporter=True, **kwargs):
+        orig_init(self, *args, **kwargs)
+        self._orig_init = orig_init
+        self._orig_call = orig_call
+        if Teleporter is not None and maybe_use_teleporter:
+            if not hasattr(self, '__teleporters'):
+                self.__teleporters = {}
+
+            def inner_func(name, doc):
+                orig_call(self, name, doc)
+
+            teleporter = Teleporter()
+            teleporter.name_doc.connect(inner_func)
+
+            self.__teleporters[cls_name] = teleporter
+        else:
+            self.__teleporter = None
+
+    @functools.wraps(orig_call)
+    def __call__(self, name, doc):
+        if hasattr(self, '__teleporters'):
+            self.__teleporters[cls_name].name_doc.emit(name, doc)
+        else:
+            return orig_call(self, name, doc)
+
+    cls.__init__ = __init__
+    cls.__call__ = __call__
+
+    return cls
+
+
+@_maybe_use_teleporter
 class LivePlot(CallbackBase):
     """
     Build a function that updates a plot from a stream of Events.
@@ -156,6 +204,7 @@ class LivePlot(CallbackBase):
         super().stop(doc)
 
 
+@_maybe_use_teleporter
 class LiveScatter(CallbackBase):
     """Plot scattered 2D data in a "heat map".
 
@@ -278,6 +327,7 @@ class LiveMesh(LiveScatter):
         super().__init__(*args, **kwargs)
 
 
+@_maybe_use_teleporter
 class LiveGrid(CallbackBase):
     """Plot gridded 2D data in a "heat map".
 
@@ -430,6 +480,7 @@ class LiveRaster(LiveGrid):
         super().__init__(*args, **kwargs)
 
 
+@_maybe_use_teleporter
 class LiveFitPlot(LivePlot):
     """
     Add a plot to an instance of LiveFit.
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -2,65 +2,31 @@ from collections import ChainMap
 from cycler import cycler
 import numpy as np
 import warnings
-import functools
 from .core import CallbackBase, get_obj_fields
+from matplotlib.backends.qt_compat import QtCore
 
 
-try:
-    from matplotlib.backends.qt_compat import QtCore
+class QtAwareCallback(CallbackBase):
 
     class Teleporter(QtCore.QObject):
-        name_doc = QtCore.Signal(str, dict)
-except ImportError:
-    Teleporter = None
+        name_doc_escape = QtCore.Signal(str, dict, bool)
 
-
-def _maybe_use_teleporter(cls):
-    orig_init = cls.__init__
-    orig_call = cls.__call__
-
-    cls_name = cls.__name__
-
-    @functools.wraps(orig_init)
     def __init__(self, *args, maybe_use_teleporter=True, **kwargs):
-        orig_init(self, *args, **kwargs)
-        _orig_call = orig_call
-        if not hasattr(self, '__teleporters'):
-            self.__teleporters = {}
-        if not hasattr(self, '__slots'):
-            self.__slots = {}
+        self.__teleporter = self.Teleporter()
+        self.__teleporter.name_doc_escape.connect(self.__call__)
+        super().__init__(*args, **kwargs)
 
-        if Teleporter is not None and maybe_use_teleporter:
-
-            def make_slot(target):
-
-                def inner_func(name, doc):
-                    _orig_call(target, name, doc)
-
-                return inner_func
-
-            teleporter = Teleporter()
-            slot = make_slot(self)
-            teleporter.name_doc.connect(slot)
-
-            self.__teleporters[cls_name] = teleporter
-            self.__slots[cls_name] = slot
-
-    @functools.wraps(orig_call)
-    def __call__(self, name, doc):
-        if hasattr(self, '__teleporters'):
-            self.__teleporters[cls_name].name_doc.emit(name, doc)
+    def __call__(self, name, doc, escape=False):
+        print(f'qta {type(self)}, {id(self)}, {escape}')
+        if not escape:
+            print('about to emit')
+            self.__teleporter.name_doc_escape.emit(name, doc, True)
         else:
-            return orig_call(self, name, doc)
-
-    cls.__init__ = __init__
-    cls.__call__ = __call__
-
-    return cls
+            print('calling up the stack')
+            return CallbackBase.__call__(self, name, doc)
 
 
-@_maybe_use_teleporter
-class LivePlot(CallbackBase):
+class LivePlot(QtAwareCallback):
     """
     Build a function that updates a plot from a stream of Events.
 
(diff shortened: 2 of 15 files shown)


danielballan:
This is essentially bait left to nerd-snipe @tacaswell into finishing the Qt integration sketched in #1062.

danielballan:
Status update:

Basic scanning and live-updating plots now works in both IPython and Jupyter (including with tornado > 5). Pause/resume is currently broken.

danielballan:
Now that this PR has rendered `install_qt_kicker()` unnecessary, it would be nice to do the same for `install_nb_kicker()`. One way to do this would be to check the matplotlib backend and automatically install the kicker on the "bluesky event loop" if it's `nbAgg`.

bluesky/utils.py
```diff
@@ -1298,3 +1270,41 @@ def my_name(obj):
...
+    This makes it possible for plots that use matplotlib's Qt backend to update
+    live during data acquisition.
+
+    It solves the problem that Qt must be run from the main thread.
+    If matplotlib and a known Qt binding are already imported, run the
+    matplotlib qApp until the task completes. If not, there is no need to
+    handle qApp: just wait on the task.
+    """
+
+    if any(p in sys.modules for p in ['PyQt4', 'pyside', 'PyQt5']):
```
tacaswell:
```suggestion
    if any(p in sys.modules for p in ['PyQt4', 'pyside', 'pyside2, 'PyQt5']):
```

danielballan: `'pyside2` -> `'pyside2'`

danielballan:
This is I think, now basically correct. Several tests that stopped/restarted the event loop need to be updated, and then we're done.

To surface some in-person comments by @tacaswell --- his work on https://github.com/NSLS-II/mily has surfaces several other ways that we could consider architecting this. The current approach may not be the best one in the long term, though it's unclear yet which approach is best.

This is detail-oriented work that requires focus that is in short supply at the moment, and I propose completing and merging the current approach. Our incompatibility with `tornado>=5` is a known stumbling point for users who are evaluating the library.

stuartcampbell:
If this PR makes it possible to work with `tornado>5` and is a good 'first attempt' then I would be in favour of merging and then revisiting / refactoring when we are clearer what the best approach is.

tacaswell:
I can locally sometimes get the test to run through without hanging or segfaulting.

tacaswell:
force-pushed to fix ci typo.

tacaswell:
rebased to account for conflicts in .travis.yml.

Will likely amend and force-push the last commit as well.

tacaswell:
The justification for the API change to `wait_for` is in these two snippets:

Old way:
```python
import asyncio
import time
start = time.monotonic()


async def f1():
    loop = asyncio.get_event_loop()
    ev = asyncio.Event()

    print('launch the task', time.monotonic() - start)
    tsk = loop.create_task(f2(ev.wait()))
    loop.call_later(2, ev.set)
    print('sleep for a bit (1s)', time.monotonic() - start)
    await asyncio.sleep(1)
    print('cancel the task', time.monotonic() - start)
    tsk.cancel()
    print('sleep for a bit', time.monotonic() - start)
    print('really cancelled?: ', tsk.cancelled(), time.monotonic() - start)
    ex = await tsk
    print('victory!, should be ~2s', time.monotonic() - start)
    if ex:
        raise ex


async def f2(fut):
    while True:
        try:
            print('  try to wait', time.monotonic() - start)
            (t,), _ = (await asyncio.wait([fut]))
            print('  and escape later', time.monotonic() - start)
            return t.exception()
        except Exception as ex:
            print('  Should fail once', time.monotonic() - start)
            print(repr(ex))




loop = asyncio.get_event_loop()
loop.run_until_complete(f1())


```

New way:
```python
import asyncio
import time
start = time.monotonic()


async def f1():
    loop = asyncio.get_event_loop()
    ev = asyncio.Event()

    print('launch the task', time.monotonic() - start)
    tsk = loop.create_task(f2(ev.wait))
    loop.call_later(2, ev.set)
    print('sleep for a bit (1s)', time.monotonic() - start)
    await asyncio.sleep(1)
    print('cancel the task', time.monotonic() - start)
    tsk.cancel()
    print('sleep for a bit', time.monotonic() - start)
    print('really cancelled?: ', tsk.cancelled(), time.monotonic() - start)
    ex = await tsk
    print('victory!, should be ~2s', time.monotonic() - start)
    if ex:
        raise ex


async def f2(fut):
    while True:
        try:
            print('  try to wait', time.monotonic() - start)
            (t,), _ = (await asyncio.wait([fut()]))
            print('  and escape later', time.monotonic() - start)
            return t.exception()
        except Exception as ex:
            print('  Should fail once', time.monotonic() - start)
            print(repr(ex))




loop = asyncio.get_event_loop()
loop.run_until_complete(f1())

```

----

The old way "works" on py3.6+ in that the timing is correct, but the inner task still sets a `RuntimeError` due to trying to re-await a coroutine.  In 3.5 this fails immediately (which is why the tests were failing) but it is no less wrong on 3.6+ and there is no garuntee that it won't re-break in the future.  The new way works everywhere.  I strongly suspect that there is no non-test use of `'wait_for'` in the wild (it is under documented and any use of it would likely be in a plan that is less than ideal _anyway_ as I don't see how to make use of `'wait_for'` and still have a plan that you can pass through `list` with no side-effects).

tacaswell:
However 

```python
import asyncio
import time
from asyncio.futures import Future
start = time.monotonic()


async def f1():
    loop = asyncio.get_event_loop()
    fut = Future()

    print('launch the task', time.monotonic() - start)
    tsk = loop.create_task(f2(fut))
    loop.call_later(2, fut.set_result, 3)
    print('sleep for a bit (1s)', time.monotonic() - start)
    await asyncio.sleep(1)
    print('cancel the task', time.monotonic() - start)
    tsk.cancel()
    print('sleep for a bit', time.monotonic() - start)
    print('really cancelled?: ', tsk.cancelled(), time.monotonic() - start)
    ex = await tsk
    print('victory!, should be ~2s', time.monotonic() - start)
    if ex:
        raise ex


async def f2(fut):
    while True:
        try:
            print('  try to wait', time.monotonic() - start)
            (t,), _ = (await asyncio.wait([fut]))
            print('  and escape later', time.monotonic() - start)
            return t.exception()
        except Exception as ex:
            print('  Should fail once', time.monotonic() - start)
            print(repr(ex))




loop = asyncio.get_event_loop()
loop.run_until_complete(f1())

```

Works correctly on all versions so the problem is passing on co-routines _as_ futures.  We should have a bit of a think about what to change this API to.

.travis.yml
```diff
@@ -33,6 +43,7 @@ install:
   - pip install -r requirements.txt
   - pip install -r dev-requirements.txt
   - pip install -U ophyd --pre
+  - pip install pyqt5
```
danielballan: Should this be added to `dev-requirements.txt`?

bluesky/run_engine.py
```diff
@@ -936,16 +1013,25 @@ def request_suspend(self, fut, *, pre_plan=None, post_plan=None,
 
         """
         if not self.resumable:
+            # TODO fix race condition here!
```
danielballan: Flagging so we don't forget it. Is this an "open an issue" problem or a "fix before merging" problem?

bluesky/run_engine.py
```diff
@@ -143,6 +190,19 @@ class RunEngine:
         Expected signature: f(md)
         Expected return: updated scan_id value
 
+    during_task : callable, optional
```
danielballan: :+1: very clear

bluesky/run_engine.py
```diff
@@ -288,6 +359,7 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
         self._exit_status = 'success'  # optimistic default
         self._reason = ''  # reason for abort
         self._task = None  # asyncio.Task associated with call to self._run
+        self._task_fut = None  # asyncio.Task associated with call to self._run
```
danielballan: This comment and the comment in the line above it are identical.

bluesky/run_engine.py
```diff
@@ -1237,6 +1399,24 @@ def _run(self):
...
                 except asyncio.CancelledError as e:
+                    if self._state == 'pausing':
+                        # if we got a CancelledError and we are in the
+                        # 'pausing' state clear the run permit and
+                        # bounce to the top
+                        self._run_permit.clear()
+                        continue
+                    if self._state in ('halting', 'stopping', 'aborting'):
+                        # if we got this while just keep going in tear-down
+                        exception_map = {'halting': PlanHalt,
```
danielballan: :+1:

doc/source/api_changes.rst
```diff
@@ -2,6 +2,94 @@
...
+
+  def blocking_func(ev : Threading.Event) -> None:
+      "Returns when ev is set"
+
+
+The default hook will handle the case of the Matplotilb Qt backend and
+the case of Matplotlib not being imported.
+
+
+``'wait_for'`` Msg now expects callables rather than futures
```
danielballan: :+1:

bluesky/run_engine.py
tacaswell: ah, this was confusing.  We used to get the task back directly from the asyncio call, but now we get back a future that knows about the task and acts as a proxy.  When you call `tsk.cancel()` it is left in a "please cancel the next time you get a chance" state (as it arranges to throw the `CancelledError` down into the wrapped co-ro).  However, when you call `future.cancel()` the proxy forwards cancel call and uncontitionally marks itself as cancelled.  Meanwhile, the underlying task may have delayed (or entirely eaten as we do) the cancellation and still be a going concern.   This effectively de-couples the future from the task which breaks things.

I may report this as a bug upstream, but need to write a ah, more minimal, example of what is going on.

tacaswell:
https://github.com/bluesky/bluesky/issues/1191#issuecomment-493305063 @prjemian 

That comment reminded me that interrupters exist and that this PR may break them.  Would you be able to manually test this locally?

We really need to move "interuptors" upstream.

tacaswell:
force-pushed to hide a dumb mistake

mrakitin:
Please note, I proposed to add OSX support for builds via a PR to this PR branch: https://github.com/danielballan/bluesky/pull/18.

danielballan:
In Jupyter, no plots appear. If you merge and release, please keep the `tornado` pin for now, unfortunately.

tacaswell:
There are still some rare race conditions in here related to pausing


```diff
diff --git a/bluesky/tests/test_run_engine.py b/bluesky/tests/test_run_engine.py
index 47017eca..e2320d52 100644
--- a/bluesky/tests/test_run_engine.py
+++ b/bluesky/tests/test_run_engine.py
@@ -392,7 +392,12 @@ def test_bad_call_args(RE):
     assert RE.state == 'idle'
 
 
-def test_record_interruptions(RE):
+@pytest.fixture(params=[f'iter{i}' for i in range(1000)])
+def multi_iterations(request):
+    return request.param
+
+
+def test_record_interruptions(RE, multi_iterations):
     docs = defaultdict(lambda: [])
 
     def collect(name, doc):

```

is enough to reliably reproduce it locally for me.

tacaswell:
Luckily that race condition was not too hard to fix, see the commit message on c28c8e1 for details

bluesky/run_engine.py
```diff
@@ -1308,46 +1557,44 @@ def _run(self):
...
+           def fut_fac() -> awaitable:
+               'This must work multiple times'
 
-        Where ``obj`` and **kwargs are the position and keyword-only arguments
-        for ``asyncio.await``
         """
         futs, = msg.args
-        yield from asyncio.wait(futs, loop=self.loop, **msg.kwargs)
+        futs = [f() for f in futs]
+        await asyncio.wait(futs, loop=self.loop, **msg.kwargs)
```
klauer: https://docs.python.org/3/library/asyncio-task.html#asyncio.wait
`The loop argument is deprecated and scheduled for removal in Python 3.10`

bluesky/utils.py
```diff
@@ -183,45 +189,40 @@ def __init__(self, RE):
         self.num_sigints_processed = 0  # count SIGINTs processed
 
     def __enter__(self):
-        self.RE.loop.call_later(0.1, self.check_for_signals)
         return super().__enter__()
 
-    def check_for_signals(self):
+    def handle_signals(self):
         # Check for pause requests from keyboard.
+        # TODO, there is a possible race condition between the two
```
klauer: For future reference, I had pasted this into our slack chat to explain multiple signals being ignored:
https://www.gnu.org/software/libc/manual/html_node/Checking-for-Pending-Signals.html
`Remember that if there is a particular signal pending for your process, additional signals of that same type that arrive in the meantime might be discarded. For example, if a SIGINT signal is pending when another SIGINT signal arrives, your program will probably only see one of them when you unblock this signal.`

bluesky/utils.py
```diff
@@ -771,48 +772,20 @@ def install_kicker(loop=None, update_rate=0.03):
 def install_qt_kicker(loop=None, update_rate=0.03):
     """Install a periodic callback to integrate Qt and asyncio event loops.
 
-    If a version of the Qt bindings are not already imported, this function
-    will do nothing.
-
-    It is safe to call this function multiple times.
+    DEPRECATED: This functionality is now handled automatically by default and
```
klauer: 🎆

bluesky/utils.py
```diff
@@ -1298,3 +1271,150 @@ def my_name(obj):
...
+                        os.read(rfd, 4096)
+                    except OSError as inst:
+                        print('failed to read wakeup fd: %s\n' % inst)
+
+                    wakeupsn.setEnabled(True)
+
+                wakeupsn.activated.connect(handleWakeup)
+
+            else:
+                # On Windows, non-blocking anonymous pipe or socket is
```
klauer: Has any interactive testing been done on Windows with this?

doc/source/api_changes.rst
```diff
@@ -2,12 +2,97 @@
...
+There is a a `Threading.Event` on the `.RunEngine` that will be set
+when the task for `.RunEngine._run` in completed, however we can not
+simple wait on that event as that would again cause the Qt windows to
+freeze.  We also do not want to bake a Matplotlib / Qt dependency
+directly into the `.RunEngine` so we added a hook, set at init time,
+for a function that passed the `~Thread.Event` and is responsible for
+blocking until it is set.  This function can do other things (such as
+run the Qt event loop) during that time.  The required signature is ::
+
+  def blocking_func(ev : Threading.Event) -> None:
```
klauer:
```suggestion
  def blocking_func(ev: Threading.Event) -> None:
```

bluesky/run_engine.py
```diff
@@ -602,44 +679,38 @@ def request_pause(self, defer=False):
...
+
+        def end_cb(fut):
+            coro_event.set()
+
+        def start_task():
+            nonlocal task
+            task = self.loop.create_task(self._request_pause_coro(defer))
+            task.add_done_callback(end_cb)
+
+        self.loop.call_soon_threadsafe(start_task)
```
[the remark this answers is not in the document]
tacaswell: yes.

bluesky/run_engine.py
```diff
@@ -1129,18 +1351,24 @@ def _run(self):
                     # This sleep has to be inside of this try block so
                     # that any of the 'async' exceptions get thrown in the
                     # correct place
-                    yield from asyncio.sleep(0, loop=self.loop)
+                    await asyncio.sleep(0, loop=self.loop)
```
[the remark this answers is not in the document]
tacaswell: I know, but not sure what the minimum version that works without it, deferring ripping this out until we sort that out.

bluesky/utils.py
tacaswell: added do docstring of base class, I think the race condition here is that the defer and hard pause tasks may both run at the same time.

tacaswell:
The mac test failures are very confusing as they are `UnbouldLocalError` s from inside of a list comprehension that works on linux.  Going to restart the 3.7 mac on the last passing build.

bluesky/run_engine.py
```diff
@@ -72,14 +107,20 @@ def __init__(self, machine_type):
...
+        with obj._state_lock:
+            super().__set__(obj, value)
+        value = self.__get__(obj, own)
         obj.log.info("Change state on %r from %r -> %r",
                      obj, old_value, value)
         if obj.state_hook is not None:
             obj.state_hook(value, old_value)
 
+    def __get__(self, instance, owner):
+        with instance._state_lock:
```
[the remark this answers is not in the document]
tacaswell: handled but in a way that did not explicitly touch this line.

bluesky/run_engine.py
```diff
@@ -1021,20 +1151,29 @@ def abort(self, reason=''):
         :meth:`RunEngine.halt`
         :meth:`RunEngine.stop`
         """
+        return self.__interpterer_helper(self._abort_coro(reason))
```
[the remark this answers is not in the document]
tacaswell: should have been `__interrupter_helper`

bluesky/run_engine.py
```diff
@@ -1101,14 +1274,63 @@ def _run(self):
...
             while True:
+                if self._state in ('pausing', 'suspending'):
+                    if not self.resumable:
+                        self._run_permit.set()
+                        stashed_exception = FailedPause()
+                        for task in self._status_tasks:
+                            task.cancel()
+                        self._state = 'aborting'
+                        continue
+                # currently only using this
```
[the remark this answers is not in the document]
tacaswell: expanded this comment to be a full sentance.

bluesky/utils.py
tacaswell: this was adapted from tortioushg which does get heavy windows use so I am optimistic.

.travis.yml
```diff
@@ -33,10 +37,6 @@ matrix:
     - os: osx
       language: generic
       env: PYTHON=3.7
-    - os: linux
-      python: nightly
-  allow_failures:
-    - python: nightly
```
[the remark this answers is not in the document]
tacaswell: I put it back, but because nightly is failing on building scipy and not providing any useful signal.

bluesky/callbacks/mpl_plotting.py
```diff
@@ -331,7 +362,8 @@ class LiveGrid(CallbackBase):
     def __init__(self, raster_shape, I, *,
                  clim=None, cmap='viridis',
                  xlabel='x', ylabel='y', extent=None, aspect='equal',
-                 ax=None, x_positive='right', y_positive='up'):
+                 ax=None, x_positive='right', y_positive='up', **kwargs):
+        super().__init__(**kwargs)
```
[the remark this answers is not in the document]
tacaswell: because in the other cases `kwargs` was already there and passed through to the plotting, here I added kwargs and can just pass it to `super()`.

.travis.yml
```diff
@@ -1,6 +1,10 @@
 dist: xenial
 language: python
 sudo: false
+
+services:
+  - xvfb
+
```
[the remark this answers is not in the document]
tacaswell: Pushed a commit to see if it works (but am skeptical).

setup.py
```diff
@@ -16,14 +14,13 @@
...
     url="https://github.com/NSLS-II/bluesky",
     packages=setuptools.find_packages(),
-    package_data={'bluesky': ['schema/*.json']},
     scripts=glob.glob('scripts/*'),
     python_requires='>=3.6',
     install_requires=requirements,
     classifiers=[
         "Development Status :: 3 - Alpha",
-        "Programming Language :: Python :: 3.4",
         "Programming Language :: Python :: 3.5",
```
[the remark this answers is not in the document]
tacaswell: done.

tacaswell:
The issue was due to pytest trying to un-roll `assert all(...)`  which they are in the process of reverting https://github.com/matplotlib/matplotlib/pull/14403 .

.travis.yml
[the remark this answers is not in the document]
mrakitin: Sadly true. Thanks anyway.

bluesky/callbacks/mpl_plotting.py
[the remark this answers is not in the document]
mrakitin: I see, makes sense.

.travis.yml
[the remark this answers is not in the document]
mrakitin: It seems to pass: https://travis-ci.org/bluesky/bluesky/builds/540895701?utm_source=github_status&utm_medium=notification, so it should be OK.

bluesky/run_engine.py
tacaswell: Same comment as the other one, want to deal with this in a follow up PR as we need to sort out what the minimum version of python that correctly works without `loop` is.

tacaswell:
I think this is ready to merge!

Issue #1045, opened 2018-05-31
Add a test of flyers + collect_asset_documents


tacaswell:
See https://github.com/NSLS-II/bluesky/pull/1044

May need to be coupled with changes to ophyd.sim

danielballan:
Excited to see @ke-zhang-rd on here. :- D

tacaswell:
Closed by #1127

PR #1127, opened 2018-12-11
TST: collect_asset_documents_with_flyer

ke-zhang-rd:

### DIFF
--- a/bluesky/tests/test_run_engine.py
+++ b/bluesky/tests/test_run_engine.py
@@ -137,14 +137,14 @@ def test_stop_motors_and_log_any_errors(RE, hw):
 
 
 @requires_ophyd
-def test_collect_asset_documents_with_flyer(RE):
+def test_flyer_with_collect_asset_documents(RE):
 
-    from ophyd.sim import det, flyer1, flyer2
+    from ophyd.sim import det, new_trivial_flyer, trivial_flyer
     from bluesky.preprocessors import fly_during_wrapper
+    assert hasattr(new_trivial_flyer, 'collect_asset_docs')
+    assert hasattr(trivial_flyer, 'collec_asset_docs') == False
+    RE(fly_during_wrapper(count([det], num=5), [new_trivial_flyer, trivial_flyer]))
 
-    RE(fly_during_wrapper(count([det], num=5), [flyer1, flyer2]))
-    for name, d in flyer1.collect_asset_docs():
-        print(name)
 
 @requires_ophyd
 def test_collect_uncollected_and_log_any_errors(RE):
--- a/bluesky/tests/test_run_engine.py
+++ b/bluesky/tests/test_run_engine.py
@@ -137,6 +137,16 @@ def test_stop_motors_and_log_any_errors(RE, hw):
 
 
 @requires_ophyd
+def test_collect_asset_documents_with_flyer(RE):
+
+    from ophyd.sim import det, flyer1, flyer2
+    from bluesky.preprocessors import fly_during_wrapper
+
+    RE(fly_during_wrapper(count([det], num=5), [flyer1, flyer2]))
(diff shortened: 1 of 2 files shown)


danielballan:
Let me expand on the goal since #1045 is pretty terse. In #1044 we fixed a crucial typo. Because we discovered the typo in the middle of a deployment, we didn't slow down to figure out why this typo wasn't covered by existing tests. Now we're circling back to add  tests that would have caught that typo and will prevent any similar mistakes in the future.

It's clear that no existing tests call `collect_asset_docs()` on a device, and we need to add a test that does that.

This PR uses `fly_during_wrapper` and then calls `collect_asset_docs()` directly. That's not right because `fly_during_wrapper` should cause the _RunEngine_ to do that.

The logic goes:
1. `fly_during_wrapper` yields a 'collect' Msg
2. RunEngine consumes that message and calls `RunEngine._collect`.
3. `RunEngine._collect` calls `collect_asset_docs()`.

That's why we shouldn't need to call `collect_asset_docs()` directly.

But then... why aren't existing flyer tests already causing `collect_asset_docs()` to be called? The answer is that `RunEngine._collect` can operate in two modes, for historical reasons. In older versions of bluesky and ophyd, there was no `collect_asset_docs()` method. Asset documents (which means Datum and Resource documents) were never seen by bluesky. Instead, ophyd inserted the documents directly into databroker, bypassing bluesky. In May, with the releases of bluesky 1.3.0 and ophyd 1.2.0, we added a new, improved* way of handling asset documents. Ophyd Devices could be updated to implement a new method -- `collect_asset_documents()`. `RunEngine._collect` looks for this method and, if it found that the Device implemented it, it uses it to retrieve Datum and Resource documents from the Device. If `RunEngine._collect` find that the Device does _not_ implement `collect_asset_docs()`, it assumes that the Device is old and has not been updated. Since old Devices inserted their Datum and Resource documents directly into databroker, the RunEngine does not have to do anything, so it moves on.

Now let's return to my question, "Why  aren't existing flyer tests already causing `collect_asset_docs()` to be called?" The [TrivialFlyer](https://github.com/NSLS-II/ophyd/blob/39811f6350917a81188cfeb471e14ecc3085f018/ophyd/sim.py#L492-L517) in `ophyd.sim` does not implement `collect_asset_docs()`; it is an "old Device" that has not been updated. Therefore, `RunEngine._collect` skips over it, and it never reaches the codepath that caused the bug addressed by #1044. If we add `collect_asset_docs()` to `TrivialFlyer` in ophyd and then run the bluesky tests, they will hit that codepath.

Therefore, I think this is what needs to be done:

* Ophyd should provide two simulate flyers: `TrivialFlyer` and `OldStyleTrivialFlyer` so that we can test both the new-style and old-style usages.
* Bluesky should have a test of each, and the tests should include `assert hasattr(TrivialFlyer, 'collect_asset_docs')` and `assert not hasattr(OldStyleTrivialFlyer, 'collect_asset_docs')` to ensure that `RunEngine._collect` is being made to exercise both of its modes of operation.

*Why is this "improved"? It makes all documents flow through the RunEngine in sequence, instead of having some document types flow through the RunEngine and others go directly to databroker. This means that _all_ subscribers get access to all document types. It also removes the necessity of ophyd to insert data into databroker, which is a big simplification.

Closes #1045

ke-zhang-rd:
Instead of use `TrivialFlyer` and `OldStyleTrivialFlyer`, `NewTrivialFlyer` and `TrivialFlyer` are used here which allow `TrivialFlyer` keep unchanged meaning. This would avoid updating any previous code use `TrivialFlyer`

danielballan:
Power-cycling to re-run tests against new ophyd release, which contains a new simulated flyer used by these tests.

danielballan:
Thanks, @ke-zhang-rd! I will sleep easier know that we have decent test coverage on this codepath now.

PR #1132, opened 2018-12-20
DOC: add warning about adjusting .PREC

### LINKED ISSUE #1115: Document common issue with badly configure IOC PREC.
Users often see that LiveTable displays 0 where it should be display some fractional number because the IOC is not set to display a sufficient number of significant digits. We discussed automatically switching to exponential notation but [concluded](https://github.com/NSLS-II/bluesky/pull/1046#issuecomment-436807672) that this didn't fully solve the problem and had down-sides. We decided to instead just document the problem. People probably won't find that documentation on their own, but at least we'll have something to link to, and the institutional knowledge will eventually spread.

I propose putting this somewhere near the LiveTable documentation.

tacaswell:
closes #1115

### DIFF
--- a/doc/source/callbacks.rst
+++ b/doc/source/callbacks.rst
@@ -194,6 +194,14 @@ field.
 
 .. autoclass:: bluesky.callbacks.LiveTable
 
+.. warning
+
+   The data in the table is formatted according to its type and the
+   precision reported by the control system.  If you are seeing too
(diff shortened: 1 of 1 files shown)


tacaswell:
self merged due to blessing from both @danielballan and @prjemian 

@prjemian feel free to push the big green button!

PR #1133, opened 2018-12-26
FIX: fix mpl callbacks

mrakitin:
## Description

Add backwards-compatibility when using the mpl callbacks the old way.
**Note**: the block of code in this PR had to be moved down after `CallbackBase` and `get_obj_fields` were defined, so that they are available when importing `mpl_plotting`.

## Motivation and Context

This was observed at NSLS-II QAS with @lcdesilva when the `LivePlot` failed to be subclassed with the master bluesky ([link](https://github.com/NSLS-II-QAS/profile_collection/blob/e5bbfb1135c1e482b82f371b507ca1c63f66963c/startup/05-aux-classes.py#L2)).

## How Has This Been Tested?

Successfully tested locally.

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -12,34 +12,6 @@ from ..utils import ensure_uid
 logger = logging.getLogger(__name__)
 
 
-# deprecate callbacks moved to mpl_plotting ----------------------------------
-
-def _deprecate_import_name(name):
-    wmsg = (
-        "In a future version of bluesky, {} will not be importable from "
-        "bluesky.callbacks.core or bluesky.callbacks. Instead, import it from "
-        "bluesky.callbacks.mpl_plotting. This change allows other callbacks, "
-        "unrelated to matplotlib, to be imported and used without importing "
-        "matplotlib.pyplot or configuring a DISPLAY."
-    ).format(name)
-    def f(*args, **kwargs):
-        # per bluesky convention use UserWarning instead of DeprecationWarning
-        warnings.warn(wmsg, UserWarning)
-        from . import mpl_plotting
-        cls = getattr(mpl_plotting, name)
-        return cls(*args, **kwargs)
-    f.__name__ = name
-    return f
-
-LiveScatter = _deprecate_import_name("LiveScatter")
-LivePlot = _deprecate_import_name("LivePlot")
-LiveGrid = _deprecate_import_name("LiveGrid")
-LiveFitPlot = _deprecate_import_name("LiveFitPlot")
-LiveRaster = _deprecate_import_name("LiveRaster")
-LiveMesh = _deprecate_import_name("LiveMesh")
-
-# ----------------------------------------------------------------------------
-
 class CallbackBase:
     def __call__(self, name, doc):
         "Dispatch to methods expecting particular doc types."
@@ -142,6 +114,32 @@ def get_obj_fields(fields):
             string_fields.extend(field_list)
     return string_fields
 
+
+# deprecate callbacks moved to mpl_plotting ----------------------------------
+
+def _deprecate_import_name(name):
+    wmsg = (
+        "In a future version of bluesky, {} will not be importable from "
+        "bluesky.callbacks.core or bluesky.callbacks. Instead, import it from "
+        "bluesky.callbacks.mpl_plotting. This change allows other callbacks, "
+        "unrelated to matplotlib, to be imported and used without importing "
+        "matplotlib.pyplot or configuring a DISPLAY."
+    ).format(name)
+    # per bluesky convention use UserWarning instead of DeprecationWarning
+    warnings.warn(wmsg, UserWarning)
+    from . import mpl_plotting
+    cls = getattr(mpl_plotting, name)
+    cls.__name__ = name
+    return cls
+
+LiveScatter = _deprecate_import_name("LiveScatter")
+LivePlot = _deprecate_import_name("LivePlot")
+LiveGrid = _deprecate_import_name("LiveGrid")
+LiveFitPlot = _deprecate_import_name("LiveFitPlot")
+LiveRaster = _deprecate_import_name("LiveRaster")
+LiveMesh = _deprecate_import_name("LiveMesh")
+
+# ----------------------------------------------------------------------------
 
 class CollectThenCompute(CallbackBase):
 


danielballan:
Fine by me as long as CI passes. Might be worth adding a test so we don’t accidentally break this usage again.

mrakitin:
Thanks, @danielballan! I added a test to cover this situation. It should fail in the current master branch:
```py
$ pytest test_callbacks.py -vvvv --pdb
============================================================================================= test session starts =============================================================================================
platform darwin -- Python 3.6.7, pytest-3.9.3, py-1.7.0, pluggy-0.8.0 -- /Users/mrakitin/anaconda3/envs/bluesky/bin/python
cachedir: .pytest_cache
rootdir: /Users/mrakitin/src/mrakitin/DAMA/bluesky, inifile:
collected 20 items

test_callbacks.py::test_all PASSED                                                                                                                                                                      [  5%]
test_callbacks.py::test_raising_ignored_or_not PASSED                                                                                                                                                   [ 10%]
test_callbacks.py::test_subs_input PASSED                                                                                                                                                               [ 15%]
test_callbacks.py::test_subscribe_msg PASSED                                                                                                                                                            [ 20%]
test_callbacks.py::test_unknown_cb_raises PASSED                                                                                                                                                        [ 25%]
test_callbacks.py::test_table_warns PASSED                                                                                                                                                              [ 30%]
test_callbacks.py::test_table PASSED                                                                                                                                                                    [ 35%]
test_callbacks.py::test_table_external PASSED                                                                                                                                                           [ 40%]
test_callbacks.py::test_live_fit PASSED                                                                                                                                                                 [ 45%]
test_callbacks.py::test_live_fit_multidim PASSED                                                                                                                                                        [ 50%]
test_callbacks.py::test_live_plot_from_callbacks FAILED                                                                                                                                                 [ 55%]
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

    def test_live_plot_from_callbacks():
        from bluesky.callbacks import LivePlot as LivePlotFromCallbacks
        from bluesky.callbacks.core import LivePlot as LivePlotFromCore
>       assert LivePlotFromCallbacks == LivePlot
E       assert <function _deprecate_import_name.<locals>.f at 0x119e9f9d8> == LivePlot

test_callbacks.py:308: AssertionError
```

mrakitin:
The test fails due to this check: https://github.com/NSLS-II/bluesky/blob/2f555a4496e6598992253b498145b76bebab29bb/.travis.yml#L44. Not sure what to do to allow mpl-free `LivePlot`, etc. If we plan do deprecate the old usage, we still should allow users to import the classes directly from `bluesky.callbacks` for a release or two. Thoughts?

danielballan:
FYI @pavoljuhas, @mrakitin realized that the solution in #1104 broke the ability to subclass. The solution here is more like your original proposal. It's our preference to do module-level imports, but this is the most pragmatic approach that achieves the desired result without breaking existing code.

bluesky/callbacks/core.py
```diff
@@ -143,6 +115,31 @@ def get_obj_fields(fields):
     return string_fields
 
 
+# deprecate callbacks moved to mpl_plotting ----------------------------------
+
+def _deprecate_import_name(name):
+    wmsg = (
```
danielballan: I think we have to remove this warning. Won't it be tripped at import time every time when the function is called just below? It's not so important to have it, anyway.

tacaswell:
I don't think we need the shims in `callbacks.core`.  Before #1104 they were not in the namespace so I don't think we need to worry about putting them back with deprecation.

bluesky/callbacks/core.py
mrakitin: Yes, indeed it is called at the import time for 6 times. I'll remove it.

bluesky/callbacks/__init__.py
```diff
@@ -8,7 +8,6 @@
                    get_obj_fields, CollectThenCompute, LiveTable)
 from .fitting import LiveFit
 
-# for back-compatibility import deprecated plotting callbacks here
-# TODO remove from here and also from core.py
-from .core import (LiveScatter, LivePlot, LiveGrid,
-                   LiveFitPlot, LiveRaster, LiveMesh)
+# TODO: remove later as it's deprecated
```
danielballan: I am OK with supporting these top-level imports forever. The important thing was not importing matplotlib at the top level, and that problem is solved.

bluesky/tests/test_callbacks.py
```diff
@@ -302,6 +302,19 @@ def gaussian(x, y, A, sigma, x0, y0):
         assert np.allclose(cb.result.values[k], v, atol=1e-6)
 
 
+def test_live_plot_from_callbacks():
+    import bluesky.callbacks.core
+    # We don't want the shims in callbacks.core, see #1133 for discussion
+    assert not hasattr(bluesky.callbacks.core, 'LivePlot')
+    # We still want the shims in callbacks.__init__
+    from bluesky.callbacks import LivePlot as LivePlotFromCallbacks
+    assert LivePlotFromCallbacks == LivePlot
```
pavoljuhas: it is better to use `is` instead of `==`

pavoljuhas:
LGTM to replace shim functions with delayed matplotlib imports.

pyflakes report some issues in the touched sources - can this be fixed?

```sh
$ git diff master... --name-only | xargs pyflakes
bluesky/callbacks/mpl_plotting.py:455: 'matplotlib.pyplot as plt' imported but unused
bluesky/tests/test_callbacks.py:13: 'bluesky.callbacks.CallbackBase' imported but unused
bluesky/tests/test_callbacks.py:15: 'signal' imported but unused
bluesky/tests/test_callbacks.py:16: 'threading' imported but unused
bluesky/tests/test_callbacks.py:17: 'time' imported but unused
```

bluesky/callbacks/__init__.py
mrakitin: It's addressed in https://github.com/NSLS-II/bluesky/pull/1133/commits/2f491f0bb725ce429502eefda52d376fc848e75a. Thanks for the review!

mrakitin:
Thanks @pavoljuhas, @tacaswell and @danielballan for reviews and help!

mrakitin:
Hooray!

PR #1139, opened 2019-01-09
MNT: Add Elasticsearch logger

ke-zhang-rd:
## Motivation and Context

https://github.com/NSLS-II/nslsii/issues/48
## How Has This Been Tested?

Elastic server is on nsls2 network, need find a way to test on control network


danielballan:
Per group discussion in Slack, this PR will be reopened against http://github.com/NSLS-II/nslsii to avoid adding a new dependency (`requests`) to bluesky. If other facilities are interested in using an ElasticSearch Python logger, we can come up with a non-facility-specific place to put it at that time.

Issue #1016, opened 2018-04-08
print_summary did not catch md typo


ambarb:
`yield from rel_scan([qem07],m1.pit,-30,30,31, md = {'reason':{'checking m1 before each cff'}})`

This was not interpreted by `print_summary()`.  At least it was the first scan of my plan, and I did not need to wait hours for it to start (thermal equilibration).

I changed to what it should be and then the RE was able to proceed without issue.
`yield from rel_scan([qem07],m1.pit,-30,30,31, md = {'reason'}:{'checking m1 before each cff'})`

tacaswell:
The first code snippet looks right and the second one looks wrong...

```python
In [20]: f = lambda md: md

In [21]: f({'reaon'}: {'bar'})
  File "<ipython-input-21-94fafaaa18b7>", line 1
    f({'reaon'}: {'bar'})
               ^
SyntaxError: invalid syntax


In [22]: f(md={'reaon'}: {'bar'})
  File "<ipython-input-22-24fc0efd222b>", line 1
    f(md={'reaon'}: {'bar'})
                  ^
SyntaxError: invalid syntax

```

I would have expected this to be

```python
yield from rel_scan([qem07],m1.pit,-30,30,31, md = {'reason':'checking m1 before each cff'})
```

I think the issue here is that `print_summary` does not attempt to validate the md, it just makes sure you python is valid and prints where it will move the motors.

tacaswell:
@ambarb Which beamline was this at?

ambarb:
@tacaswell sorry, I missed all the other comments.  This was at 2ID (SIX).

ambarb:
We were using the latest official release at that time.  Recently, we switched to the environment needed fro centroiding (XIP plugin).

awalter-bnl:
FYI I have been able to reproduce this problem using simulated hardware. Attached is a .py file that has 2 plans, one with the correct syntax and one with the incorrect syntax as stated by @ambarb above. 

The file loads without issue using %run, and performing summarize_plan/print_summary on either plan produces no error and correctly indicates the steps in the plan. running the scan using a run engine also works for both plans, unless a databroker.insert is subscribed at which point a `JSONDecodeError` exception is raised as it does not know how to interpret the 'value' in `md={'reason':{'attempt to cause issue'}}` as it is not a valid way to define a dictionary.

[...]

PR #1143, opened 2019-01-25
ENH: This adds a check of the meta-data for json compatibility to sumarize_plan

awalter-bnl:
## Description
The changes are made to `print_summary_wrapper`, when an `open_run` message is received it adds a check of the message.kwargs for json compatibility. If an issue is found it further attempts to isolate the cause. Finally it raises an exception with as much information about the issue as it can find.

## Motivation and Context
This resolves issue #1016 

## How Has This Been Tested?
This was tested using the file attached to issue #1016

### DIFF
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -294,15 +294,10 @@ def print_summary_wrapper(plan):
             try:
                 json.dumps(msg.kwargs)
             except TypeError:
-                error_keys=[]
-                for key, value in msg.kwargs.items():
-                    try:
-                        json.dumps(value)
-                    except TypeError:
-                        error_keys.append(key)
-                if error_keys:
+                bad_keys = _recursive_validate_json_dict(msg.kwargs)
+                if bad_keys:
                     raise TypeError(
-                        f'The values for the keys in "{error_keys}" '
+                        f'The values for the keys in "{bad_keys}" '
                         f'from the start document are not valid')
                 else:
                     raise TypeError(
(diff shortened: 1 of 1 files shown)


bluesky/preprocessors.py
```diff
@@ -289,6 +290,24 @@ def print_summary_wrapper(plan):
         cmd = msg.command
         if cmd == 'open_run':
             print('{:=^80}'.format(' Open Run '))
+            # This next bit checks the metadata for json compatibility
+            try:
+                json.dumps(msg.kwargs)
+            except TypeError:
+                error_keys=[]
```
tacaswell: If we are going to go down the route of sorting out what is failing, might as well write this as a recursive helper function.  The start document can be arbitrarily nested so the error checking should be too.

bluesky/preprocessors.py
```diff
@@ -289,6 +290,24 @@ def print_summary_wrapper(plan):
...
+                    try:
+                        json.dumps(value)
+                    except TypeError:
+                        error_keys.append(key)
+                if error_keys:
+                    raise TypeError(
+                        f'The values for the keys in "{error_keys}" '
+                        f'from the start document are not valid')
+                else:
+                    raise TypeError(
```
danielballan: Is it possible to hit this codepath?

tacaswell: Yes, the keys (not the values) could be what is failing to serialize.

awalter-bnl: or the combination of 'keys' and 'values' could be failing to serailize (i.e it is not written as a valid dict !-) )

bluesky/preprocessors.py
awalter-bnl: I will make this change, it makes sense. I just have to think about how to keep track fo the nesting when reporting back.

awalter-bnl:
I have made the validation recursive

danielballan:
@tacaswell Can you expand on what you meant by "recursive"? (This is not it.)

awalter-bnl:
I am guessing that this is no longer required, as there has been no movement on it since March, closing it out.

PR #1145, opened 2019-02-05
MNT: stop using deprecated API in matplotlib

tacaswell:
This fails with AttributeError on Matplotlib master and will fail on
3.1 (coming next month).


licode:
I also used this in pyxrf. I will fix it.

PR #1147, opened 2019-02-16
Add dev-requirements for test

stuartcampbell:
The tests fail due do some packages not being installed.  As these are in the `dev-requirements.txt` I am just installing them here.  I am happy to trim down from this to specific packages if we want to explicitly state them.  But as the travis build just includes them all upfront anyway and doesn't differentiate, I'm not sure we care.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -29,6 +29,7 @@ jobs:
 
   - script: |
       pip install pytest
+      pip install -r dev-requirements.txt
       pytest bluesky/tests --doctest-modules --junitxml=junit/test-results.xml
     displayName: 'pytest'
 


azure-pipelines.yml
```diff
@@ -23,12 +23,18 @@ jobs:
...
+      /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16
+      export DISPLAY=:99.0
+    displayName: 'Start Xvfb'
+    
   - script: python -m pip install --upgrade pip && pip install -r requirements.txt
     displayName: 'Install dependencies'
 
   - script: |
       pip install pytest
+      pip install -r dev-requirements.txt
```
mrakitin: pytest is already there in `dev-requirements.txt`, so the previous line can be safely removed.

azure-pipelines.yml
```diff
@@ -23,12 +23,18 @@ jobs:
     inputs:
       versionSpec: '$(python.version)'
       architecture: 'x64'
-
+      
+  - script: |
+      /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16
+      export DISPLAY=:99.0
```
tacaswell: The exports do not survive between scripts.

stuartcampbell: Thank you, I will do it in a more robust way

stan-dot:
@stuartcampbell  is this still a requested feature?

PR #1148, opened 2019-02-22
ENH: use the ISO8601 delimiters for date

prjemian:
ISO8601 date delimiter is "-", not "/"

space between date and time is now accepted as modification of ISO8601 standard so no change there

## Description
Existing: `Transient Scan ID: 17     Time: 2019/02/22 14:22:52`

Revised: `Transient Scan ID: 17     Time: 2019-02-22 14:22:52`

## Motivation and Context
Cosmetics:  why not use ISO standard format here?


prjemian:
@mrakitin Thanks for the review!

Issue #1144, opened 2019-01-28
Generalize list_scan and log_scan same as scan.


danielballan:
As of a couple releases ago, `bluesky.plans.scan` accepts an unlimited sequence of `motor, start, stop`, which is moves together ("inner product"). It would be nice if `bluesky.plans.list_scan` did the same. Currently, the user has to escalate to using a `cycler` to perform an inner product scan with custom step sizes.

danielballan:
This can also result in depreacting `one_1d_step`, as all scans will now be ND in general.

danielballan:
Current:
```
list_scan(dets, motor, [...])  # supports one motor only
scan(dets, motor, start, stop, motor2, start2, stop2, num_steps)  # "inner product" over motors
```

Proposed:

```python
list_scan(dets, motor, [...], motor2, [...])  # If list lengths are unequal, use the shortest one.
list_grid_scan(dets, motor, [...], motor2, [...])
```

Issue #929, opened 2017-12-13
Per step for `count`


CJ-Wright:
Should we have a `per_step` for count? (I think this was mentioned somewhere but I don't know/can't find where so I'm putting it here)

## Expected Behavior
Count takes in a `per_step`. When the plan is executed it does the per_step in the count's place.
Eg `count(..., per_step=take_dark)` would take a dark before taking the light image.
## Current Behavior
Count does not take a `per_step`.

## Possible Solution
Put in a `per_step` kwarg. Interestingly this raises a bit more general problem. The current `per_step` signature is `(detector, motor, step)`. However, there is no internal motor for this signature to be called upon. One thing we could do is to have two kwargs;  `per_step` and `per_step_args`. The args would be fed into `per_step` as required. This fixes some of the issues where accessing motors not in the plan (eg the shutter) is done by pulling them from the globals and hoping that they will be there. Additionally this allows for `per_steps` which don't need all/any of those arguments.

## Context
It would be nice to be able to take darks in the same start as the lights, being able to have a count `per_step` which closed the shutter and took a dark would be a nice way to handle this.

Attn: @chiahaoliu

danielballan:
This has been raised a couple times, and I've been shooting it down because it feels like overworking `count` -- too much complexity for not much code reuse gained. I would rather this be implemented as a separate plan that just uses `trigger_and_read`. I think that would be more clear.

I'm open to dissenting arguments....

CJ-Wright:
@tacaswell thoughts? (I think I remember a gchat discussion where you had suggested this)

teddyrendahl:
Feels to me like a bandaid for the real problem. `count` really feels like a `plan_stubs` to me. `trigger_and_read` is a nice alternative, but reimplementing it with a for loop every time (especially with all the nice logic regarding delays e.t.c) is a pain. I don't know @CJ-Wright specific use case, but in my head I imagine that instead of `per_step` hook on count you could invert it and use `count` as the `per_step` hook on a different plan?

```python
plan = scan([det], mot, -5, 5, per_step=stub_count([det], num=10, delay=1))
```

I think the proper solution might be adding a stub plan that does the same thing as count, then have the full `bluesky.plans.count` just call this with the correct metadata. However, if you wanted to try this now you could always use the `stub_wrapper` on `count`

ZLLentz:
[...]

PR #1149, opened 2019-02-27
ENH: adding the `per_step` kwarg to the `count` plan

awalter-bnl:
Adds the `per_step` kwarg to count

## Description
I have added the `per_step` kwarg to the plan `count` mirroring the way it is done for other built-in plans like `scan`. To do this I also had to add the new default `per_step` function `coutn_step` to `plan_stubs`.

## Motivation and Context
At one of the beamlines at NSLS-II they are interested in taking one, or more, 'spectra' at each step in a plan, which can be defined as a scan over a single axis. To do this for  the plan `scan` I created a `per_step` function that collects the `spectra` at each point. To be able to use the same `per_step` function with 'count' plans it required this change.

## How Has This Been Tested?
It has been manually tested, as well as getting the existing pytests to pass locally. I will ensure that the tests also pass on Travis prior to merging.

### DIFF
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -900,6 +900,28 @@ def caching_repeater(n, plan):
         yield from (m for m in lst_plan)
 
 
+def count_step(detectors, step, pos_cache):
+    """Inner loop of a count scan
+
+    This is the default function for ``per_step`` in count plans.
+
+    Parameters
+    ----------
+    detectors : iterable
+        devices to read
+    step : dict
+        maps motors to positions in this step. Not used, included for API
+        compatibility with ``bluesky.plan_stubs.one_nd_step``, and similar
+        custom ``per_step`` functions only.
+    pos_cache : dict
+        maps motors to their last-set positions. Not used, included for API
+        compatibility with ``bluesky.plan_stubs.one_nd_step``, and similar
+        custom ``per_step`` functions only.
+    """
+
+    yield from trigger_and_read(list(detectors))
+
+
 def one_1d_step(detectors, motor, step):
     """
     Inner loop of a 1D step scan
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -928,8 +928,13 @@ def one_1d_step(detectors, motor, step, take_reading=trigger_and_read):
         The motor to move
     step : Any
         Where to move the motor to
-    take_reading : Callable[List[OphydObj]] -> Generator[Msg], optional
-        function to do the actual acquisition.
+    take_reading : plan, optional
+        function to do the actual acquisition ::
+
+           def take_reading(dets, name='primary'):
+                yield from ...
+
+        Callable[List[OphydObj], Optional[str]] -> Generator[Msg], optional
 
         Defaults to `trigger_and_read`
     """
@@ -981,8 +986,15 @@ def one_nd_step(detectors, step, pos_cache, take_reading=trigger_and_read):
         mapping motors to positions in this step
     pos_cache : dict
         mapping motors to their last-set positions
-    take_reading : Callable[List[OphydObj]] -> Generator[Msg], optional
-        function to
+    take_reading : plan, optional
+        function to do the actual acquisition ::
+
+           def take_reading(dets, name='primary'):
+                yield from ...
+
+        Callable[List[OphydObj], Optional[str]] -> Generator[Msg], optional
+
+        Defaults to `trigger_and_read`
     """
     motors = step.keys()
     yield from move_per_step(step, pos_cache)
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -900,10 +900,10 @@ def caching_repeater(n, plan):
         yield from (m for m in lst_plan)
 
 
-def count_step(detectors, take_reading=trigger_and_read):
-    """Inner loop of a count scan.
+def one_shot(detectors, take_reading=trigger_and_read):
+    """Inner loop of a count.
 
-    This is the default function for ``per_step`` in count plans.
+    This is the default function for ``per_shot`` in count plans.
 
     Parameters
     ----------
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -900,7 +900,7 @@ def caching_repeater(n, plan):
         yield from (m for m in lst_plan)
 
 
-def count_step(detectors):
+def count_step(detectors, take_reading=trigger_and_read):
     """Inner loop of a count scan.
 
     This is the default function for ``per_step`` in count plans.
@@ -909,9 +909,19 @@ def count_step(detectors):
     ----------
     detectors : Iterable[OphydObj]
         devices to read
+
+    take_reading : plan, optional
+        function to do the actual acquisition ::
+
+           def take_reading(dets, name='primary'):
+                yield from ...
+
+        Callable[List[OphydObj], Optional[str]] -> Generator[Msg], optional
+
+        Defaults to `trigger_and_read`
     """
     yield Msg('checkpoint')
-    yield from trigger_and_read(list(detectors))
+    yield from take_reading(list(detectors))
 
 
 def one_1d_step(detectors, motor, step, take_reading=trigger_and_read):
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -914,7 +914,7 @@ def count_step(detectors):
     yield from trigger_and_read(list(detectors))
 
 
-def one_1d_step(detectors, motor, step, per_step=trigger_and_read):
+def one_1d_step(detectors, motor, step, take_reading=trigger_and_read):
     """
     Inner loop of a 1D step scan
 
@@ -928,7 +928,7 @@ def one_1d_step(detectors, motor, step, per_step=trigger_and_read):
         The motor to move
     step : Any
         Where to move the motor to
-    per_step : Callable[List[OphydObj]] -> Generator[Msg], optional
+    take_reading : Callable[List[OphydObj]] -> Generator[Msg], optional
         function to do the actual acquisition.
 
         Defaults to `trigger_and_read`
@@ -940,7 +940,7 @@ def one_1d_step(detectors, motor, step, per_step=trigger_and_read):
         yield Msg('wait', None, group=grp)
 
     yield from move()
-    return (yield from per_step(list(detectors) + [motor]))
+    return (yield from take_reading(list(detectors) + [motor]))
 
 
 def move_per_step(step, pos_cache):
@@ -967,7 +967,7 @@ def move_per_step(step, pos_cache):
     yield Msg('wait', None, group=grp)
 
 
-def one_nd_step(detectors, step, pos_cache, per_step=trigger_and_read):
+def one_nd_step(detectors, step, pos_cache, take_reading=trigger_and_read):
     """
     Inner loop of an N-dimensional step scan
 
@@ -981,12 +981,12 @@ def one_nd_step(detectors, step, pos_cache, per_step=trigger_and_read):
         mapping motors to positions in this step
     pos_cache : dict
         mapping motors to their last-set positions
-    per_step : Callable[List[OphydObj]] -> Generator[Msg], optional
+    take_reading : Callable[List[OphydObj]] -> Generator[Msg], optional
         function to
     """
     motors = step.keys()
     yield from move_per_step(step, pos_cache)
-    yield from per_step(list(detectors) + list(motors))
+    yield from take_reading(list(detectors) + list(motors))
 
 
 def repeat(plan, num=1, delay=None):
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -900,23 +900,15 @@ def caching_repeater(n, plan):
         yield from (m for m in lst_plan)
 
 
-def count_step(detectors, step, pos_cache):
+def count_step(detectors):
     """Inner loop of a count scan.
 
     This is the default function for ``per_step`` in count plans.
 
     Parameters
     ----------
-    detectors : iterable
+    detectors : Iterable[OphydObj]
         devices to read
-    step : dict
-        maps motors to positions in this step. Not used, included for API
-        compatibility with ``bluesky.plan_stubs.one_nd_step``, and similar
-        custom ``per_step`` functions only.
-    pos_cache : dict
-        maps motors to their last-set positions. Not used, included for API
-        compatibility with ``bluesky.plan_stubs.one_nd_step``, and similar
-        custom ``per_step`` functions only.
     """
 
     yield from trigger_and_read(list(detectors))
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -910,7 +910,7 @@ def count_step(detectors):
     detectors : Iterable[OphydObj]
         devices to read
     """
-
+    yield Msg('checkpoint')
     yield from trigger_and_read(list(detectors))
 
 
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -901,7 +901,7 @@ def caching_repeater(n, plan):
 
 
 def count_step(detectors, step, pos_cache):
-    """Inner loop of a count scan
+    """Inner loop of a count scan.
 
     This is the default function for ``per_step`` in count plans.
 
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -914,11 +914,24 @@ def count_step(detectors):
     yield from trigger_and_read(list(detectors))
 
 
-def one_1d_step(detectors, motor, step):
+def one_1d_step(detectors, motor, step, per_step=trigger_and_read):
     """
     Inner loop of a 1D step scan
 
     This is the default function for ``per_step`` param in 1D plans.
+
+    Parameters
+    ----------
+    detectors : iterable
+        devices to read
+    motor : Settable
+        The motor to move
+    step : Any
+        Where to move the motor to
+    per_step : Callable[List[OphydObj]] -> Generator[Msg], optional
+        function to do the actual acquisition.
+
+        Defaults to `trigger_and_read`
     """
     def move():
         grp = _short_uid('set')
@@ -927,7 +940,7 @@ def one_1d_step(detectors, motor, step):
         yield Msg('wait', None, group=grp)
 
     yield from move()
-    return (yield from trigger_and_read(list(detectors) + [motor]))
+    return (yield from per_step(list(detectors) + [motor]))
 
 
 def move_per_step(step, pos_cache):
@@ -954,7 +967,7 @@ def move_per_step(step, pos_cache):
     yield Msg('wait', None, group=grp)
 
 
-def one_nd_step(detectors, step, pos_cache):
+def one_nd_step(detectors, step, pos_cache, per_step=trigger_and_read):
     """
     Inner loop of an N-dimensional step scan
 
@@ -968,10 +981,12 @@ def one_nd_step(detectors, step, pos_cache):
         mapping motors to positions in this step
     pos_cache : dict
         mapping motors to their last-set positions
+    per_step : Callable[List[OphydObj]] -> Generator[Msg], optional
+        function to
     """
     motors = step.keys()
     yield from move_per_step(step, pos_cache)
-    yield from trigger_and_read(list(detectors) + list(motors))
+    yield from per_step(list(detectors) + list(motors))
 
 
 def repeat(plan, num=1, delay=None):
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -22,7 +22,7 @@ from . import preprocessors as bpp
 from . import plan_stubs as bps
 
 
-def count(detectors, num=1, delay=None, *, md=None):
+def count(detectors, num=1, delay=None, *, per_step=None, md=None):
     """
     Take one or more readings from detectors.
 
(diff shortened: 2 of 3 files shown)


bluesky/plan_stubs.py
```diff
@@ -892,6 +892,28 @@ def caching_repeater(n, plan):
         yield from (m for m in lst_plan)
 
 
+def count_step(detectors, step, pos_cache):
+    """Inner loop of a count scan
```
mrakitin:
```suggestion
    """Inner loop of a count scan.
```

mrakitin: I remember @danielballan was mentioning this in some previous reviews, that it should be a complete sentence ending with the dot.

awalter-bnl:
@tacaswell the error in Travis is in relation to the missing `set_draggable` attribute on the `Legend` object of the plot. Did the fix for this issue already make it in to bluesky master? Just wanting to make sure if I need to make a change to this test to resolve an issue resulting from that fix or if it will be resolved by that fix.

When running this test locally I get the error
`AttributeError: 'Legend' object has no attribute 'set_draggable'`

I am assuming on Travis that this is giving a warning which results in the assert on line 108 on `bluesky.tests.test_bec` not passing.

bluesky/plans.py
```diff
@@ -36,6 +36,17 @@ def count(detectors, num=1, delay=None, *, md=None):
         If None, capture data until canceled
     delay : iterable or scalar, optional
         Time delay in seconds between successive readings; default is 0.
+    per_step : callable, optional
```
danielballan: I don't think this signature should expect `step` and `pos_cache`. Notice the signatures of the existing `per_step` callables are suited to their purposes:

```
one_1d_step(detectors, motor, step)
one_nd_step(detectors, step, pos_cache)
```

I think this should default to plain `trigger_and_read` if `None`, with no need for a special `count_step` function necessary, and accept anything that has the same signature as `trigger_and_read`.

awalter-bnl: @danielballan the reason I have included `step` and `pos_cache` as arguments is so that I can use the _same_  `per_step` function in `scan` or `count` as outlined in the motivation and context above. This is fundamental to the current issue that this PR is looking to resolve.

FYI. I think all of the built in plans, with the exception of `list_scan`, use `one_nd_step` as the default 'per_step' function therefore I would propose that we modify `one_1d_step` to use the same signature as this and `one_nd_step`.

awalter-bnl:
FYI the failed test is in `bluesky.tests.test_bec.test_plot_int` I cannot replicate the failure locally and the test is written in a way that makes sorting out the issue next to impossible (I would need to know what the warnings are but have no way to extract them from Travis). Any ideas on how to resolve this are welcome.

bluesky/plans.py
awalter-bnl: In fact after looking in more detail, one_1d_step appears to be a specific case of one_nd_step so I think we could deprecate one_1d_step entirely.

danielballan: . Yes, once https://github.com/NSLS-II/bluesky/issues/1144 is done we can deprecate `one_1d_step`. I can see your point that having one `per_step` interface is advantageous and generally in keeping with the spirit of limiting the number of interfaces. I wonder a little if, as a user of `count`, I would wonder why my `per_step` function has to accept two parameters that it doesn't actually use. Perhaps we should use signature inspection to accept either `one_nd_step`-like functions or `trigger_and_read`-like functions. See `scan` for an example of this. Thoughts?

awalter-bnl: are you suggesting that we use the system we use in `scan` (and `grid_scan`) in order to accept an arbitrary number of `motor, start, stop` arguments? 

If so I am not sure that is the best path, that would mean that every `per_step` function would need complicated logic in order to deal with various numbers of parameters being passed in. The solution I am proposing here is to say that the `plans` know what parameters are useful and provide empty dictionaries for those that are not useful. This seems like the most simple solution, and because the 'extra' parameters both relate to moving motors passing in empty dictionaries is equivalent to saying 'move nothing'.

In fact passing empty `dicts` for `step` and `scan` in `count` to `one_nd_scan` would also work, so that would also negate the need for a custom default `per_step` for count.

danielballan: Not exactly. I'm suggesting that the `per_step` in `count` specifically could accept the standard `one_nd_step` interface or the standard `trigger_and_read` interface. Everything else would require the `one_nd_step` interface (and also, just for back-compat reasons, `one_1d_step`).

I'm not sure I like that better than what you have proposed here -- just brainstorming what our options are. Any opinions, @ZLLentz @teddyrendahl?

awalter-bnl: ok so now I think I understand, you are suggesting having some logic in `count` that works out what parameters to send to the `per_step` function based on what the function is. is that it?

I still think it is cleaner and neater to pass in the emtpy `dicts`, but as always will follow the consensus.

danielballan: Yes, that is he possibility I want to consider. I’m not sure I disagree with your view to have only one “per_step” signature, just want to consider it.

tacaswell: This is making the assumption that all per-step will be happy with empty dictionaries as input.

I think that it is clearer here if `count` has it's own `per_step` signature.  If you can re-use another `per_step` function, it is easy enough to use `partial` to project out the extra inputs that `count` does not provide (or make them have default values).

If we set the precedent that all `per_step` must have the same signature, what do we do if we come up with a plan that requires a _more_ complex signature?

On the other hand, a `per_step` that is only meant for `count` (ex, https://github.com/NSLS-II/bluesky/pull/1149/files#diff-ef3b7c7e4a0df009f8b37ec3d866bfbbR895) that will happily go into the `one_1d_step` slot but very much not work.   It is possible to defensively write your count-only per_step to raise if the second two arguments are not empty, but it seems clearer to let cpython raise a `TypeError` and clearly explain what the problem is.

awalter-bnl:
> I think that it is clearer here if count has it's own per_step signature. If you can re-use another per_step function, it is easy enough to use partial to project out the extra inputs that count does not provide (or make them have default values).

I think the exact opposite is true, why introduce complexity (.i.e. a user has to know that for plan 'a' you need this signature but for plan 'b' this different signature, or for plan 'c' ........). Which I think complicates things to much.

Aside from user confusion when we have to build these plans in factories from code (i.e. behind the scenes of a potential GUI) then this also becomes increasingly complicated as you have to consider each case individually.....

> If we set the precedent that all per_step must have the same signature, what do we do if we come up with a plan that requires a more complex signature?

This is a fair point, although I guess we could add extra info via `kwargs` (i.e. add `**kwargs` to the signature) which would be selectively ignored by some `per_step` functions but not `others`.

> On the other hand, a per_step that is only meant for count (ex, https://github.com/NSLS-II/bluesky/pull/1149/files#diff-ef3b7c7e4a0df009f8b37ec3d866bfbbR895) that will happily go into the one_1d_step slot but very much not work.

@danielballan and I had this conversation above, I think we don't need `count_step` (or `one_1D_step` for that matter anymore as `one_nd_step` will work in all our use cases as is (it would require some changes to the plans that currently use `one_1D_step` but that is already planned as mentioned by @danielballan above).
If someone insists on writing a `count_only` `per_step` plan then I think the onus is on them to ensure that it raises the correct error, especially when making it universal (in terms of the plans we currently have) is as simple as adding the one line `yield from move_per_step(step, pos_cache)` as is found here https://github.com/NSLS-II/bluesky/blob/226d74f97ae34963733bf725b2eed5d2790ffe88/bluesky/plan_stubs.py#L951

bluesky/plans.py
awalter-bnl: > I think that it is clearer here if count has it's own per_step signature. If you can re-use another per_step function, it is easy enough to use partial to project out the extra inputs that count does not provide (or make them have default values).

I think the exact opposite is true, why introduce complexity (.i.e. a user has to know that for plan 'a' you need this signature but for plan 'b' this different signature, or for plan 'c' ........). Which I think complicates things to much.

Aside from user confusion when we have to build these plans in factories from code (i.e. behind the scenes of a potential GUI) then this also becomes increasingly complicated as you have to consider each case individually.....

> If we set the precedent that all per_step must have the same signature, what do we do if we come up with a plan that requires a more complex signature?

This is a fair point, although I guess we could add extra info via `kwargs` (i.e. add `**kwargs` to the signature) which would be selectively ignored by some `per_step` functions but not `others`.

> On the other hand, a per_step that is only meant for count (ex, https://github.com/NSLS-II/bluesky/pull/1149/files#diff-ef3b7c7e4a0df009f8b37ec3d866bfbbR895) that will happily go into the one_1d_step slot but very much not work.

@danielballan and I had this conversation above, I think we don't need `count_step` (or `one_1D_step` for that matter anymore as `one_nd_step` will work in all our use cases as is (it would require some changes to the plans that currently use `one_1D_step` but that is already planned as mentioned by @danielballan above).
If someone insists on writing a `count_only` `per_step` plan then I think the onus is on them to ensure that it raises the correct error, especially when making it universal (in terms of the plans we currently have) is as simple as adding the one line `yield from move_per_step(step, pos_cache)` as is found here https://github.com/NSLS-II/bluesky/blob/226d74f97ae34963733bf725b2eed5d2790ffe88/bluesky/plan_stubs.py#L951

danielballan: > If we set the precedent that all per_step must have the same signature, what do we do if we come up with a plan that requires a more complex signature?

I think this is convincing.

> I think we don't need count_step ... one_nd_step will work in all our use cases as is

I think this assumes too much about what other types of plans might be added in the future, unforeseen at this point.

> we could add extra info via kwargs... which would be selectively ignored by some per_step functions but not others

This pattern is one we avoid at all costs. It falls under the category of "silently ignoring user input" which can lead to very confusing traps.

> I think it [would be less clear to give `count`'s `per_step` its own signature], why introduce complexity (.i.e. a user has to know that for plan 'a' you need this signature but for plan 'b' this different signature, or for plan 'c' ........). Which I think complicates things to much.

I can see how the "complication" argument can cut both ways. Looking at the library as a whole, "one per_step interface" has a simplicity to it, as you argue. But approach it as a user of `count` trying to write my own `per_step` funciton, I would wonder why my function needs to accept `step` and `pos_cache` ("What even are those?") only to ignore them. (Apologies for repeating myself here.)

In summary given that
) complication can cut both ways
* silently ignoring user inputs is known to create deeply confusing "bugs"
( we can't be sure that the `one_nd_step` signature is general enough to cover all future plans

I think there is a strong case for making `count`'s `per_step` expect the signature of `trigger_and_read`.

I am still interested in input from @ZLLentz and @teddyrendahl if you guys have time to weigh in.

awalter-bnl: I reiterate my stance above that I think having different signatures for `per_step` is _more_ confusing than having some which_ignore_ some parameters. I am completely not convinced of the arguments above. 

This suggestion is moving a small level of complexity, in ensuring that the `per_step` functions all have the same signature, to ensuring that both users and code which wants to work with all plans and a `per_step` function have to generate a multitude of `per_step` functions and know which plan to send them too.

danielballan: Consider the logical extension of that argument. We can't just have one `bluesky.plans.plan` that takes every possible argument and ignores ones that are not applicable.

danielballan: It might be worth spelling out the alternative @tacaswell and I are now advocating, just to make sure we are all on the same page.

There would be two `per_step` signatures in bluesky (plus the one deprecated one, supported for back-compat) with room to grow more. I can't picture us growing a _multitude_ of more signatures, because there aren't that many fundamentally different kinds of plan likely to be added to bluesky, but I could picture a couple more in a long term.

Thus, the use cases and associated user experiences would be:

1. I need to customize `count`. I check the docstring. It says, "`per_step` accepts a plan with the signature `f(detector)`. The default is `bluesky.plan_stubs.trigger_and_read`." I do a customization like:

    ```py
    def my_trigger_and_read(detectors):
        yield from bluesky.plan_stubs.trigger_and_read(detector)
        print('hi')
    ```

2. I need to customize a 1D or ND scan.  I check the docstring. It says, "`per_step` accepts a plan with the signature `f(detectors, step, pos_cache)`. The default is `bluesky.plan_stubs.one_nd_step`." I I do a customization like:
    ```py
    def my_one_nd_step(detectors, step, pos_cache):
        yield from bluesky.plan_stubs.one_nd_step(detectors, step, pos_cache)
        print(f'finished step {step}')

3. But what I want to use `my_one_nd_step` on `count`? The thing is, this customization makes explicit use of `step` so I _can't_ just plug it into `count` as is. It does something that doesn't make sense to do for `count` because `count` doesn't scan movable things in steps.

4. OK, what if I want to use `my_trigger_and_read` in a scan? That one makes more sense: you have to copy `one_nd_step` and replace the `trigger_and_read` call in its inner loop for `my_trigger_and_read`.

In summary, making one signature to rule them all doesn't make the user experience simpler; it just buries the real difference where they are harder to see.

ZLLentz: Some rambling thoughts, note that I don't have strong opinions about this:
- I agree with Dan that it isn't necessary to require extra parameters that are effectively always receiving empty dictionaries. The important thing is that the parameters have the same names when they are relevant.
- I think the problem with not being able to plug `my_one_nd_step` into `count` is highlighting a problem I have with `scan_nd`'s `per_step` parameter: in general what we really want is just to modify or append to the `trigger_and_read` step. Any `per_step` I've written simply passes the `step` and `pos_cache` args directly into `move_per_step`, so if there weren't compelling use cases for these args I'd actually prefer to have `per_step(detectors)` be the norm everywhere. By using `scan_nd` I'm implying that I want to do the `move_per_step` step scan. But there are compelling use-cases here, so it's fine.
- In general, I don't like `per_step` options at all because they make it very easy to make plans with misleading metadata. We'll have plans named `count` that aren't really counting... Plans named `scan_nd` that aren't really scanning. etc.

danielballan:
> If this is the case, I'd vote no `per_step` on `count` at all. If you are getting to the point where you want to start making more complex scans `bps.repeat` has all the functionality while still being approachable.

Now that you mention this @teddyrendahl I recall that [adding `per_step` to `count` has been proposed and rejected before](https://github.com/NSLS-II/bluesky/issues/929). @ZLLentz, @teddyrendahl, @prjemian, and I came to what reads to me like a strong consensus to instead adopt [a different proposal](https://github.com/NSLS-II/bluesky/pull/936) that implemented [`repeat`](https://nsls-ii.github.io/bluesky/generated/bluesky.plan_stubs.repeat.html#bluesky.plan_stubs.repeat).

I propose mentioning `repeat` prominently in the docstring for `count` and explicitly pointing out that it is the recommended way to achieve `per_step`-like functionality that we provide in `scan_nd` and its derivatives.

bluesky/plans.py
tacaswell: @teddyrendahl the reason for passing in the the positions is if you need to implement a scan that has complex "these can move in parallel, these must move in series" constraints.

tacaswell:
I do not think that `repeat` and `count(..., per_step=my_perstep)` are equivalent.  The big difference is that `count` will take care of generating your start/stop messages, but repeat will not.   In many cases where you want to use `per_step` you are still fundamentally doing what ever the outer scan is, just with a slightly more complex inner loop (examples are flashing a shutter or rocking a crystal during the acquisition) so the default meta-data is still correct.

If you want 1 start document, with `count` 's pre_step you would do

```python
count(dets, ..., per_step=my_per_step)
```

if you wanted the same with `repeat` you would have to do something like

```python
stage_wrapper(run_wrapper(repeat(lambda : my_per_step(dets)), ...), md={...}), dets)
```

which is eliding the complexity of generating the 'standard' metadata.

Thus, I disagree that we should just make note of `repeat` in the docstring.

-----

@ZLLentz 's suggestion that `per_step` on the ND scans should be simpler got me thinking and maybe the solution there is to put `per_step` (with the same signature as `count` on the `one_nd_step` and `one_1d_step`?

----

There seems to be consensus that the API of the count `per_step` should be `def f(dets)`.


---


I have push a PR making all of these changes https://github.com/awalter-bnl/bluesky/pull/3

bluesky/plans.py
ZLLentz: @tacaswell at the risk of bringing this comment chain too far off-topic, I see the use cases but still find the API of "`scan_nd` decides what points we use based on my args, and then I also pass a `per_step` to re-interpret them" to be less intuitive than "I write a new scan that decides what complex set of points I use"

Back on-topic: I agree that count's `per_step` should just have the `detectors` argument. We should think of `scan_nd` as "adding" arguments to `per_step` rather than thinking of `count` as "removing" arguments from `per_step`.

awalter-bnl:
I agree with @tacaswell that `count` is not equivalent to `repeat`. Further I think that `count` is an essential part of the list of plans, as it is the moist convenient solution to acquiring a scan with 1 'step' or a time-series scan.

I like the idea of making the base `per_step` function `trigger_and_read` as this really is the simplest possible task to do at each 'step' in a scan where as the `move_per_step` part is what hapens 'between' steps IMO.

So long story short I am in favour of @tacaswell's proposed changes. We just need to be a little bit careful on release of this change as I know of 3 custom `per_step` plans used at beamlines at NSLS-II that will need to be modified to prevent them causing an issue, as this is a breaking change.

danielballan:
> We just need to be a little bit careful on release of this change as I know of 3 custom per_step plans used at beamlines at NSLS-II that will need to be modified to prevent them causing an issue, as this is a breaking change.

If I understand correctly, this is actually not a breaking change. It adds `per_step` to `count` and it adds an _additional_ hook to customized just the `trigger_and_read` part of `one_nd_step` and `one_1d_step`. The existing `per_step` hooks in `scan_nd` (and friends) work the same as before.

awalter-bnl:
OK so the part I was missing (which I made an assumption about) is how do we, for all the scans that use `scan_nd` as the base plan, actually change the value of `take_reading`.

In the current PR by @tacaswell there is no mechanism to do this, I assumed ( and now see that it isn't the case) that `per_step` would actually be passed down and instead of replacing `one_nd_step` it would replace `take_reading`. I strongly suggest this be done, but even if it isn't there is no way that I an see to _ask_ scan_nd to have a different `take_reading` value.

As an aside, can I suggest that we make `take_reading` also include the `name` kwarg for changing the stream_name. i.e. make it have the same signature as `trigger_and_read`.

awalter-bnl:
OK so I just spoke in person with @danielballan about this and he mentioned the way to do it is
`scan(... , per_step=functools.partial(one_nd_step( take_reading=my_trigger_read())),...)`

While this is a little complicated, it will work for what I want to do as the same approach can also be used for `count`

awalter-bnl:
So after discussing with @tacaswell, I asked him to push to this branch with his suggested changes.

One further change I would like to suggest (which was actually @tacaswell's sugesstion) is to make `step` and `pos_cache` optional in `one_nd_step` with defaults being `None`. That would then get around the issue of passing in empty dictionaries in the case of `count` but still give the flexibility of using the same `per-step` for all scans.

This would also remove a complication when using `functools.partial` to add a different `take_reading` function where it is still necessary to _know_ which default `per_step` function applies to the scan. This is still the main issue I see with the current suggestion.

tacaswell:
I on one hand I would like to defer the discussion of changing the defaults to later, but if we do that we would also want to deprecate `count_step` as it will be redundant with `one_*d_step` so it is better to do here.

The pro is a simplification in terms of having fewer functions, but the con is a complexification in that the smaller number of functions will each be more complex which makes them harder to reason about.

For example, we need to take care of the fact that `motor` and `value` 

```python

def one_1d_step(dets, motor=None, value=None, take_reading=trigger_and_read):
     if (motor is None) ^ (value is None):
        raise("coupled kwaaaaargs")
    if motor:
        yield from do_move()
   yield from take_reading
```

Which requires documenting that it moves things, unless you don't pass anything etc.  This also shows at least one major anti-pattern (coupled arguments) and maybe another one (very different behavior of the functions based on inputs).

awalter-bnl:
>I on one hand I would like to defer the discussion of changing the defaults to later

As this would directly impact how I would use this I think we should resolve it now.

> The pro is a simplification in terms of having fewer functions, but the con is a complexification in that the smaller number of functions

We would not be removing the complexity but moving it to each of the _custom_ per step functions as they would have to do:
```
if scan_type is in [scan, grid_scan, ...]:
    # use the signature with step and pos_cache
elif scan_type in [count,.... ]:
    # use the signature without step and pos_cache
else:
    ...

or we would not be able to use the same `per_step` with all of the plans.

**Comment by danielballan** [comment]
Discussion on call:

@tacaswell: You would never write an `if` block like the one in the comment above because you already know what kind of scan you have and what signature it needs.

@ZLLentz: Give `per_step` a different name in `count` because you aren't really "stepping" (maybe stepping in _time_) and it has a different signature. Maybe `per_shot`?

@tacaswell: Independent of what we do with the stock one, you could write a `per_step` that works will all of them, which is why I'm OK with the names being the name.

@danielballan Given that, @awalter-bnl can write a `per_step` that works with both, and we can defer whether we want to make `one_nd_step` work with `count` as well as `scan_nd` to a separate conversation/PR.

**Comment by ZLLentz** [comment]
I thought about this a bit more and my opinion shifted since the call. I think this PR is already exactly as it should be.

If we want to use the same `per_step` logic with both `count` and `scan`, the preferred approach should be to refactor the `detectors` part out of our `per_step` stubs, rather than add complexity to the `per_step`. Very roughly:

```
def my_take_reading(detectors):
    yield from special_task()
    yield from trigger_and_read(detectors)

def my_per_step(detectors, step, pos_cache):
    yield from my_step_logic(step, pos_cache)
    yield from my_take_reading(detectors)

my_scan = scan(*my_scan_args, per_step=my_per_step)
my_count = count(*my_count_args, per_step=my_take_reading)
```

Is there a situation where it is important to pass exactly the same stub plan into all scans? If not, then no undue complexity is being shifted to the user's implementation.

danielballan:
Revisiting this, I agree with @ZLLentz's last comment. I move to close debate and merge as is.

I have pushed commits to update the documentation and to implement the name change (proposed by @ZLLentz above) of calling new the argument to `count` `per_shot` because it has a different expected signature than `per_step` in the scans and there is no "step" involved here. Correspondingly, I have changed the name of the default `per_shot` plan to `one_shot`, matching the existing `per_step` functions `one_nd_step` and (deprecated) `one_1d_step`.

stuartcampbell:
I think we should move on and just merge this PR.

danielballan:
I agree. Thanks for the initial work @awalter-bnl and the input @teddyrendahl and @ZLLentz.

awalter-bnl:
Thanks for pushing this through @danielballan, it is good to see it make its way through to merge.

PR #1151, opened 2019-03-08
ENH: adding a couple of plan_stubs which set an EpicsMotor user offset

awalter-bnl:
Correctly setting the user offset value on an Epics Motor requires values for other parameters to be set first. I have added 2 plan_stubs which create plans for executing these changes.

# Description
I have added 2 new plan_stubs that should be used with EpicsMotors to correctly set the user_offset to a new value. for a detailed description of why this is needed see the section under Calibration-related fields at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html

## Motivation and Context
This is required to make setting the user offset easier in user defined plans, it is mainly used after runnign some 'alignment scans' so that any previous motor values are referenced to a known position.

### DIFF
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -1028,7 +1028,7 @@ def update_epicsmotor_user_offset(epicsmotor, val):
     There is a complex set of steps involved in correctly updating the user
     offset value as the effect of changing its value depends on he values of
     other signals, this plan_stub genreates the instructions to perform those
-    steps. See the section on Calibration-related fileds at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html>>>>>>>ccb1ced23ad740f7c7c0e9a5a312643f125c1599
+    steps. See the section on Calibration-related fileds at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html
     for a detailed description of the how this is meant to work.
 
     Parameters
@@ -1052,7 +1052,7 @@ def set_current_epicsmotor_value(epicsmotor, val):
     There is a complex set of steps involved in correctly setting the current
     user value to a new value this has the effect of changing the user_offset
     value, this plan_stub genreates the instructions to perform those
-    steps. See the section on Calibration-related fileds at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html>>>>>>>ccb1ced23ad740f7c7c0e9a5a312643f125c1599
+    steps. See the section on Calibration-related fileds at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html
     for a detailed description of the how this is meant to work.
 
     Parameters
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -1020,3 +1020,50 @@ def repeat(plan, num=1, delay=None):
                     yield Msg('sleep', None, d)
 
     return (yield from repeated_plan())
+
+
+def update_epicsmotor_user_offset(epicsmotor, val):
+    '''Follows the correct procedure to update an EpicsMotor user offset.
+
+    There is a complex set of steps involved in correctly updating the user
+    offset value as the effect of changing its value depends on he values of
+    other signals, this plan_stub genreates the instructions to perform those
+    steps. See the section on Calibration-related fileds at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html>>>>>>>ccb1ced23ad740f7c7c0e9a5a312643f125c1599
+    for a detailed description of the how this is meant to work.
+
+    Parameters
+    ----------
+    epicsmotor : ``ophyd.EpicsMotor`` instance
+        This is an EpicsMotor instance whose user-offset is to be changed
+    val : float
+        The new value for the user offset.
+    '''
+    if hasattr(epicsmotor, 'user_offset') and hasattr(epicsmotor,
+                                                      'offset_freeze_switch'):
+        initial_offset_freeze = epicsmotor.offset_freeze_switch.position
+        yield from mv(epicsmotor.offset_freeze_switch, 0)
+        yield from mv(epicsmotor.user_offset, val)
+        yield from mv(epicsmotor.offset_freeze_switch, initial_offset_freeze)
+
+
+def set_current_epicsmotor_value(epicsmotor, val):
+    '''Follows the correct procedure to reset the EpicsMotor user value to val.
+
+    There is a complex set of steps involved in correctly setting the current
+    user value to a new value this has the effect of changing the user_offset
+    value, this plan_stub genreates the instructions to perform those
+    steps. See the section on Calibration-related fileds at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html>>>>>>>ccb1ced23ad740f7c7c0e9a5a312643f125c1599
+    for a detailed description of the how this is meant to work.
+
+    Parameters
+    ----------
+    epicsmotor : ``ophyd.EpicsMotor`` instance
+        This is an EpicsMotor instance whose user-offset is to be changed
+    val : float
+        The new value for the current motor position.
+    '''
+    if hasattr(epicsmotor, 'set') and hasattr(epicsmotor, 'set_use_switch'):
+        initial_set_use_switch = epicsmotor.set_use_switch.position
+        yield from mv(epicsmotor.set_use_switch, 0)
+        yield from mv(epicsmotor, val)
+        yield from mv(epicsmotor.set_use_switch, initial_set_use_switch)


danielballan:
Definitely needs discussion. Bluesky does not currently contain _any_ plans or plan stubs that assume specific hardware. This cuts against the abstraction line between bluesky and ophyd.

danielballan:
I will open a PR into ophyd when I get a chance to make a counterproposal of how this could be done on that side. This seems to be exactly the kind of hardware-specific detail that should not leak into plans.

awalter-bnl:
I am happy to do it either way (as a plan in bluesky or in ophyd). After a discussion with @tacaswell we said probably bluesky is a better place and it was quick enough to write here that this is where I put it.

teddyrendahl:
This a good discussion point for the `action` API Ken came up with to specify device specific `bluesky` routines in `Ophyd` as well

klauer:
Following @teddyrendahl's comment... is this perhaps a good time to discuss the merits of having a hard bluesky dependency in ophyd?

danielballan:
Yes.

bluesky/plan_stubs.py
```diff
@@ -1020,3 +1020,50 @@ def repeated_plan():
...
     return (yield from repeated_plan())
+
+
+def update_epicsmotor_user_offset(epicsmotor, val):
+    '''Follows the correct procedure to update an EpicsMotor user offset.
+
+    There is a complex set of steps involved in correctly updating the user
+    offset value as the effect of changing its value depends on he values of
+    other signals, this plan_stub genreates the instructions to perform those
+    steps. See the section on Calibration-related fileds at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html>>>>>>>ccb1ced23ad740f7c7c0e9a5a312643f125c1599
```
tacaswell: This url does to look right, assume 

```suggestion
    steps. See the section on Calibration-related fileds at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html
```

bluesky/plan_stubs.py
```diff
@@ -1020,3 +1020,50 @@ def repeated_plan():
...
+        yield from mv(epicsmotor.offset_freeze_switch, initial_offset_freeze)
+
+
+def set_current_epicsmotor_value(epicsmotor, val):
+    '''Follows the correct procedure to reset the EpicsMotor user value to val.
+
+    There is a complex set of steps involved in correctly setting the current
+    user value to a new value this has the effect of changing the user_offset
+    value, this plan_stub genreates the instructions to perform those
+    steps. See the section on Calibration-related fileds at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html>>>>>>>ccb1ced23ad740f7c7c0e9a5a312643f125c1599
```
tacaswell:
```suggestion
    steps. See the section on Calibration-related fileds at https://epics.anl.gov/bcda/synApps/motor/R4-1/motorRecord.html
```

bluesky/plan_stubs.py
```diff
@@ -1020,3 +1020,50 @@ def repeated_plan():
...
+    for a detailed description of the how this is meant to work.
+
+    Parameters
+    ----------
+    epicsmotor : ``ophyd.EpicsMotor`` instance
+        This is an EpicsMotor instance whose user-offset is to be changed
+    val : float
+        The new value for the user offset.
+    '''
+    if hasattr(epicsmotor, 'user_offset') and hasattr(epicsmotor,
```
tacaswell: I would not put these behind a conditional, if you pass in a non-epics motor to this (or the other one) they _should_ blow up as they are not going to do what you think they will do.  

For example if you pass a mover in, expecting it to reset the origin to where ever you currently are (say 1000) and then tell it to move to 1 expecting it to move one unit forward but instead it moves 999 units the other way.

bluesky/plan_stubs.py
```diff
@@ -1020,3 +1020,50 @@ def repeated_plan():
...
+                                                      'offset_freeze_switch'):
+        initial_offset_freeze = epicsmotor.offset_freeze_switch.position
```
tacaswell: please do this with `bps.read`

bluesky/plan_stubs.py
```diff
@@ -1020,3 +1020,50 @@ def repeated_plan():
...
+
+    Parameters
+    ----------
+    epicsmotor : ``ophyd.EpicsMotor`` instance
+        This is an EpicsMotor instance whose user-offset is to be changed
+    val : float
+        The new value for the current motor position.
+    '''
+    if hasattr(epicsmotor, 'set') and hasattr(epicsmotor, 'set_use_switch'):
+        initial_set_use_switch = epicsmotor.set_use_switch.position
```
tacaswell: ditto on doing this with `bps.read`

tacaswell:
I am not convinced that this as qualitatively different than `mov` requiring the `set` method.  

We could make this more general by taking in a signal or name for the switches, something like

```python
def maybe_read(obj):
    ret = yield from bsp.read(obj)
    if ret is None:
        return ret
    h = obj.hints.get('fields', [])
    k = h[0] if len(h) else obj.name  # heuristic needs a bit of work
    return ret[k]['value']

def set_stuff_protected_by_a_switch(parent, val, switch, victim, off_value):
   if isinstance(switch, str):
       switch = getattr(parent, switch)
   if victim = None:
        victim = parent
   if isinstance(victim, str):
       victim = getattr(parent, victim)
  
   init_value = yield from maybe_read(switch)
   def do_the_thing()
       yield from bps.mov(switch, off_value)
       yield from bps.mov(victim, val)

   yield from finalize_wrapper(do_the_thing(),
                               yield from bps.mov(switch, init_value))

```

which now makes almost no hardware specific assumptions and supports the general case of where we need to flip a switch, change something, and then flip the switch back to where ever it was.  This may not actually be general enough (and hiding the motor-record specific aspects of this is just an intellectual vanity project).

In anycase, we should use finalize_wrapper to make sure that in the case of failure / user interuption, etc we re-set the switch (if the user uses `halt` there is still a chance to escape without re-setting the switch, but if they have called halt things are going _very_ sideways and likely need BL staff to fix things anyway).

teddyrendahl:
> I am not convinced that this as qualitatively different than mov requiring the set method.

Confused by this comment. "Set" is a pretty general verb, can be applied to a lot of devices and signals reasonably (probably the reason you didn't got with "move" for the `BlueskyInterface`). Expecting anything but an `EpicsMotor` to have something `user_offset_freeze_switch`.

Also why does this have to be `bluesky` plan at all? This is like three lines of code just as a convenience method on `EpicsMotor` itself. Don't see this function benefitting much from going through a `RunEngine`.

prjemian:
Changing the motor record's offset field requires 3 channel access put calls.  

1. place the record in SET mode
1. define the new offset
1. place the motor in USE mode

These operations would block if not written as a plan.

tacaswell:
> Also why does this have to be `bluesky` plan at all? This is like three lines of code just as a convenience method on `EpicsMotor` itself. Don't see this function benefitting much from going through a `RunEngine`.

Because we want to use this in plans

```python
def my_plan():
    yield from some_stuff()
    motor.reset_the_zero_to_where_the_motor_currently_is()
    yield from some_other_stuff()
```

is going to do surprising things if you do

```python
print_summary(my_plan())
```

In general, if you want to take advantage of any of the preview functionality the plans need to be 'safe' to iterate through which more-or-less means that they should only touch the hardware via the RE.

We either need to do this as a plan or need to start growing a greater number of `Msg` types to access non-standard methods on devices.  Or add some blind call messages to RE

```python
def _call_meth(self, msg):
    return msg.obj(*msg.args, **msg.kwargs)

def _acall_meth(self, msg):
    return (await msg.obj(*msg.args, **msg.kwargs)
```

Then you could do things like

```python
def my_plan():
    yield from some_stuff()
    yield Msg('call_meth', motor.reset_the_zero_to_where_the_motor_currently_is)
    yield from some_other_stuff()
```

I don't recall if we ever talked about this idea before..... 

--------
> probably the reason you didn't got with "move" for the `BlueskyInterface`

naming thing is hard.

----------

> > I am not convinced that this as qualitatively different than mov requiring the set method.
> 
> Confused by this comment.

In the sense that not all objects can be slotting into all positions of all plans.  I admit it is a bit of an extreme position.

prjemian:
Example for a continuous rotation tomography stage, [sets rotation axis to angle modulo one full rotation](https://github.com/NSLS-II/ophyd/pull/700).  (This code allows any modulo number but we all know it is only for modulo=360.)

tacaswell:
@prjemian I don't think that link went where you wanted it to...

teddyrendahl:
Okay. I stand convinced of why you want it in plan form. I do think I'd like to support this generally in `ophyd`. Here is a gist @klauer made prototyping the idea. API pending this would look like:

```python
class EpicsMotor(Device):
      ...

      @action
      def set_offset(self, value):
          try: 
              yield from set(self.user_offset_freeze_switch, 0)
               yield from set(self.user_offset, value)
          except:
               logger.exception(...)
           finally:
                yield from set(self.user_offset_freeze_switch, 1)
```
You could then use this in a plan by calling `motor.set_offset.plan()` or call it stand-alone without an RE. I just don't like the idea of keeping device specific plans in a completely separate repository than the devices themselves. I also think this is the more sane way than trying to create all necessary permutations for `Msg` verbiage.

We have a large need for this with devices that have `insert` and `remove` positions but could also be `set` with a floating point number. Nice to have access to them all in plans and just in an IPython shell.

prjemian:
@tacaswell You're correct.  Fixed it.
@teddyrendahl Does that `yield from set()` yield up a `Msg` object (or list of 'em)?  A plan is a generator of Msg objects for handling by the RunEngine.

prjemian:
Also, this a series of operations the process must first put motor in SET mode, *then* change the motor offset, *then* put the motor back into USE mode.  The bps.mv is the call that always waits for completion.  The bps.abs_set does not wait unless an additional keyword is passed telling it to wait.

prjemian:
Finally, what specific exceptions are expected?  Can the except clause be narrowed to one type of exception?

tacaswell:
@prjemian I assume `set` was short-hand for `bps.abs_set(..., wait=True)`.   Any exceptions that are raised by while executing the `Msg` are passed back through to be re-raised in the plan so it can in principle be anything.  Additionally there are a handful of 'control flow' exception from the RE (such as a 'cancel the plan' exception and `GeneratorExit`)

@teddyrendahl We need to use `finalize_wrapper` or handle `GeneratorExit` on it's own.  Once that has been thrown in (which cpython does to tell a generator "you will never have `next` called on you again" you are not allowed to `yield` / `yield from`.  If you do, it will result in a `RuntimeError` (I think on the principle that it is better to blow up and tell you it isn't going to do what you think rather than silently ignore you).

teddyrendahl:
@tacaswell Point heard about the the `finalize_wrapper`. Probably should have avoided writing out the actual code, don't think it helped my argument. I am interested about the possibility of this living on the `ophyd` object itself though. I think we had discussed the possibility of this sort of thing for stuff like `stage_plan`

danielballan:
>  I am interested about the possibility of this living on the ophyd object itself though.

I am interested in that too. From a bluesky point of view, setting the user offset should work the same on an EpicsMotor, PVPositioner, or some non-EPICS ophyd object.

The boundary between what is a "device-specific detail" is bound to be fluid, and it's normal to have some _ad hoc_ plans that are explicitly designed to work with specific devices. But when we see those plans moving into bluesky itself, as proposed by this PR, it's time to rethink whether the ophyd side can be holding more of those details so that plans can remain generalized.

In a couple places --- staging being the most obvious one --- we are seeing that these device-specific details are often multi-step operations that could naturally reuse Msg semantics and benefit from things like pause/rewind/resume.

awalter-bnl:
So a basic summary of the conversation (numberred/lettered below so we can more easily refer) each issue  so far (correct me if I am wrong).
1. As this is device specific we should do it in ophyd.
2. This is best done as a plan so we can utilize all of the built-in (error, pause-resume ...) `RunEngine` functionality.
3. We should think about using this same mechanism for other cases (i.e. adding multi-state positioner capability to existing motors, that does stuff like motor.some_attr.set('position_x') _moves motor to position 'x'_) +100% on this for me

Which if we maintain our strict non-dependency of bluesky/ophyd seems to be somewhat incompatible. The proposed solutions I see so far I think are:

A. move this to Ophyd and create a bluesky dependency.
B. move this to Ophyd and not make use of the built-in features of the `RunEngine`
C. leave this in Bluesky.

I really am not tied strongly to any of these solutions, and feel free to add any others (prefereably including the letter reference at the start). Is there any consensus on which way we should head?

awalter-bnl:
As there is no consensus on how to implement this I am closing it out.

klauer:
I think we'll grab this and run if we ever manage to get bluesky plans into ophyd.

PR #1158, opened 2019-03-10
FIX: Do not break support for mpl 2.x.

danielballan:


danielballan:
Similar to but not exactly the same as the approach in https://github.com/NSLS-II/bluesky/pull/1149 which is tied up in discussion

danielballan:
I think it might be worth rolling a new micro release for this.

PR #1165, opened 2019-03-18
remove np sanitize from ZMQ publisher

CJ-Wright:

### DIFF
--- a/bluesky/callbacks/zmq.py
+++ b/bluesky/callbacks/zmq.py
@@ -6,7 +6,6 @@ import socket
 import time
 import warnings
 from ..run_engine import Dispatcher, DocumentNames
-from ..utils import apply_to_dict_recursively, sanitize_np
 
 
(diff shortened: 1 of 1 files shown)


CJ-Wright:
@NSLS-II/dama ready for review!

CJ-Wright:
We need to remove it because it would mutate the docs which is a no-no with the ``doct`` returns from the databroker. I'm not certain why it was there in the first place

tacaswell:
It is the serializer's problem to deal with numpy arrays.

This is here as a legacy thing before we had really sorted out who was responsible for dealing converting numpy arrays -> bytes.   If we convert to lists pre-emptively it means that serializers can not be smart if they want to be.

danielballan:
Thanks @CJ-Wright for the improvement, @mrakitin for asking good questions that prompted a better public record of this change, and @tacaswell for providing context.

PR #1170, opened 2019-04-05
MNT: add more helpful error message if obj.read returns None

tacaswell:
This would have lead to a much clearer traceback at a beamline.

Ran into this working with @EliotGann If `read` returned `None` you got an exception in `save` when we try to collate the results.  This moves the failure to much when it happens and provides better exception (I ended up doing a binary search on a long list of baseline devices to find the problem).

If other also think this is a good idea I'll throw a test onto it.


danielballan:
Good usability improvement! Thanks!

PR #1175, opened 2019-05-03
ENH: provide a way to exclude scan_id

mrakitin:
## Description

Provides a way to exclude the `scan_id` field from metadata.

## Motivation and Context

@prjemian's [snapshot](https://aps-bluesky-tools.readthedocs.io/en/latest/snapshot.html) tool allows to get PV data into a Beamline's databroker, however, it's not aware of the latest `scan_id`, and starts from scratch (`scan_id=1`):
```py
In [8]: db[-1].start
Out[8]:
{'time': 1556900218.8359325,
 'uid': '<REDACTED>',
 'plan_description': 'archive snapshot of ophyd Signals (usually EPICS PVs)',
 'purpose': 'archive a set of EPICS PVs',
 'plan_type': 'generator',
 'plan_name': 'snapshot',
 'software_versions': {'python': '3.6.7 |Anaconda, Inc.| (default, Oct 23 2018, 19:16:44) \n[GCC 7.3.0]',
  'PyEpics': '3.3.3',
  'bluesky': '1.5.2',
  'ophyd': '1.3.3',
  'databroker': '0.12.2',
  'apstools': '1.1.0'},
 'hostname': '<REDACTED>',
 'username': 'mrakitin',
 'scan_id': 1,
 'login_id': '<REDACTED>',
 'iso8601': '2019-05-03 12:16:58.835718',
 'hints': {}}
```
This PR attempts to prepare a way for the tool (and not only for it) to not have the `scan_id` in such scans at all.

## How Has This Been Tested?

To be tested.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1353,7 +1353,11 @@ class RunEngine:
         self._run_start_uids.append(self._run_start_uid)
 
         # Run scan_id calculation method
-        self.md['scan_id'] = self.scan_id_source(self.md)
+        if self.scan_id_source is not None:
+            self.md['scan_id'] = self.scan_id_source(self.md)
+        else:
+            if 'scan_id' in self.md:
+                del self.md['scan_id']
 
         # For metadata below, info about plan passed to self.__call__ for.
         plan_type = type(self._plan).__name__


bluesky/run_engine.py
```diff
@@ -1353,7 +1353,11 @@ def _open_run(self, msg):
         self._run_start_uids.append(self._run_start_uid)
 
         # Run scan_id calculation method
-        self.md['scan_id'] = self.scan_id_source(self.md)
+        if self.scan_id_source is not None:
+            self.md['scan_id'] = self.scan_id_source(self.md)
```
danielballan: My concern with this is that it would have the side-effect of resetting the `scan_id` to `1` in the HistoryDict, which might not be what the user wants.

Could it be useful provide the `scan_id_source` some way of saying, "Based on the metadata, this scan doesn't get a `scan_id`" without also reseting `scan_id` to `1`? Example sketch to explain what I mean:

Something like

```py
stateful_md = self.md.copy()
if scan_id_source is not None:
    scan_id = scan_id_source(stateful_md)
else:
    scan_id = None
if scan_id is None:  # could be because scan_id_source is not set or because it is set and returned None
    del stateful_md['scan_id']
else:
    self.md['scan_id'] = scan_id
    stateful_md['scan_id'] = scan_id
...
# Combine metadata, in order of decreasing precedence:
md = ChainMap(self._metadata_per_call,  # from kwargs to self.__call__	        md = ChainMap(self._metadata_per_call,  # from kwargs to self.__call__
                      msg.kwargs,  # from 'open_run' Msg	                      msg.kwargs,  # from 'open_run' Msg
                      {'plan_type': plan_type,  # computed from self._plan	                      {'plan_type': plan_type,  # computed from self._plan
                       'plan_name': plan_name},	                       'plan_name': plan_name},
                      stateful_md)  # stateful, persistent metadata	                      self.md)  # stateful, persistent metadata
```

mrakitin: Looks great, I'll port this snippet to the PR.

mrakitin:
Please note, I added a test to check that scenario:
- normal RE scan with `scan_id`
- custom RE scan without `scan_id`
- normal RE scan with `scan_id` again <-- here we should see the incremented `scan_id` from the first scan.

bluesky/tests/test_scan_id_source.py
```diff
@@ -0,0 +1,25 @@
+import bluesky.plans as bp
+
+
+def test_relative_pseudo(hw, RE, RE_no_scan_id, db):
```
danielballan: It might be worth giving this a more descriptive name. (Looks like it was copy/pasted from a test with a different goal that this one.)

mrakitin: Good catch! Indeed, I copied over the test from another module and forgot to change the name.

mrakitin: Updated to be `test_scan_id_source` via https://github.com/NSLS-II/bluesky/pull/1175/commits/7dddd6f38dd656219ac5f294274131f8c9481d6e.

danielballan:
I think my sketch missed something important. Can you check whether the `scan_id` is incremented between scans using the default `scan_id_source`? My guess is that it is not.

danielballan:
Ah, nevermind. [This line](https://github.com/NSLS-II/bluesky/pull/1175/files#diff-918281cac085be7995578c4c2c57cd4eR1366) handles that.

mrakitin:
Thanks for the review, @danielballan. I added one more minor assertion in the test. If anybody else have a chance to review, please do so.

bluesky/run_engine.py
```diff
@@ -1353,7 +1353,18 @@ def _open_run(self, msg):
...
         # Run scan_id calculation method
-        self.md['scan_id'] = self.scan_id_source(self.md)
+        stateful_md = self.md.copy()
+        if self.scan_id_source is not None:
+            scan_id = self.scan_id_source(stateful_md)
+        else:
+            scan_id = None
+
+        if scan_id is None:
+            if 'scan_id' in stateful_md:
```
tacaswell: you can use `stateful_md.pop('scan_id', None)` which will do the same thing with one less conditional.

tacaswell:
I don't understand why we want this complexity in handling this?  If the user starts up a RE with a) `HistoryDict` backing the RE b) no scan_id incrementer I think that is a clear indication that the user does not want `scan_id` included or persistent.  I am less worried about the confusion of using the same `HistoryDict` across multiple sessions with different `scan_id` incrementers accidentally re-setting the scan_id to 1 and more concerned with "I see 'scan_id' in `RE.md` WHY ISN"T IT IN MY HEADER".

The first is a maybe annoying consequence of user action but is something that follows directly from the rules ("RE.md is **splated into the start document", "__call__ may mutate RE.md", "if you are using a `HistoryDict` as RE.md it is persisted to disk", "I use a scan_id_source=None" -> "scan_id is removed from my on-disk persistent state so everything is consistent").

On the other hand, this behavior changes the rules by adding a couple of "UNLESS ..." special cases.  Long term I think that is much more confusing.

@mrakitin In those tests each RE has a normal dictionary backing the `md` so there is no coupling in any case.  A better test wolud be to mutate the `scan_id_source` on  RE in between calls.

This also adds the ability for `scan_id_source` to return `None` to opt-out of the scan_id, is that intentional?

This needs documentation.

bluesky/run_engine.py
```diff
@@ -1353,7 +1353,18 @@ def _open_run(self, msg):
         self._run_start_uids.append(self._run_start_uid)
 
         # Run scan_id calculation method
-        self.md['scan_id'] = self.scan_id_source(self.md)
+        stateful_md = self.md.copy()
```
tacaswell: It is a bit pedantic, but I find this variable name confusing, we are using it to specifically make sure that our changes are _not_ stateful!

To that end `local_md = dict(self.md)` may be safer (incase the `copy` method of the stateful thing returns a copy if it's self in a way that is still tied to what ever we are trying to seperate it from.  By using `dict()` we are sure we have a plain python dict).

mrakitin:
> @mrakitin In those tests each RE has a normal dictionary backing the `md` so there is no coupling in any case. A better test wolud be to mutate the `scan_id_source` on RE in between calls.

@tacaswell, the current implementation of the test (https://github.com/NSLS-II/bluesky/pull/1175#issuecomment-489230939) covers the case of @prjemian's snapshot tool, which uses a separate instance of RE internally via CLI, which creates records in the same collection databroker.

I am fine with adding one more test to mutate the `scan_id_source` between a single RE calls, however I think the current test should exist as well.

----

> This also adds the ability for `scan_id_source` to return `None` to opt-out of the scan_id, is that intentional?

I think it was your idea from our in-person conversation and in Slack. Did I miss something?

bluesky/run_engine.py
mrakitin: The "stateful" name comes from this comment: https://github.com/NSLS-II/bluesky/blob/ebc912a9962e7796d3d1b03f2d5da08804d345fd/bluesky/run_engine.py#L1367

I'm fine with making it `local_md` as suggested. +1 for the `dict()` approach.

mrakitin: Implemented via https://github.com/NSLS-II/bluesky/pull/1175/commits/d4b0aa4ab8b3cf02c7223024e3e31d803f8828ef.

mrakitin:
@tacaswell, I addressed your concern about the function returning `None` via https://github.com/NSLS-II/bluesky/pull/1175/commits/b6bb35e7e94565bba248c3fbdc7bbd0d60a5c485. Now a `RuntimeError` exception will be raised in case it returns `None`, and we have a test for it.

I think the only item we are missing is the documentation update. I'll let you know when it's done, so another round of review can be kicked-off. I suggest @danielballan to re-re-review it as well 😃 once done.

mrakitin:
The docstring was updated via https://github.com/NSLS-II/bluesky/pull/1175/commits/0ac7a7b8cb89d606dbdf36d840626a0ef0e4211d. It's ready for another round of review.

danielballan:
> Mid-term we probably want to deprecate this whole thing and do it either via a pre-processor (or a general md mutation function) so should do the simplest thing here to make the deprecating path as simple as possible.

I agree with this.

mrakitin:
@tacaswell, do you mind to reiterate on your concerns with the implementation here (for short-term usage and to unblock https://github.com/BCDA-APS/apstools/pull/133 to allow the snapshot feature at NSLS-II)?

prjemian:
We closed BCDA-APS/apstools#133 as `wontfix` since this PR has stalled.  A resolution of how to exclude `scan_id` from a run's metadata would be useful.  As our [comment notes](https://github.com/BCDA-APS/apstools/pull/133#issuecomment-923116415), this capability would be useful when documenting (in a separate run) about a long sequence of run's to be acquired.  The APS USAXS [does this](https://apstools.readthedocs.io/en/latest/source/_plans.html#apstools._plans.doc_run.documentation_run) on a routine basis.

tacaswell:
Re-reading this, I stand by my concern about dropping `'scan_id'` from the start document if it is in `RE.md`.  I think the right path is to if there is not `scan_id` source available, do not update it, but if it is there, it is there 🤷🏻 .

You could can also argue that mixing user controlled meta-data and RE controlled mete-data in `RE.md` was a mistake, but then we would have had to persist 2 things instead of one!

stan-dot:
the repo mentioned as motivation has moved on, possibly already includes the functionality in question. The usefulness of this feature has been emphasized in 2021 as often used. (https://github.com/bluesky/bluesky/pull/1175#issuecomment-923120382)

@prjemian what is the case in 2024?


Also this might clear up once some metadata-related issues /PRs are settled

prjemian:
tl;dr: `scan_id` is searchable metadata of interest to experimenters.  Keep it in one place. I'm expecting it as a key in `RE.md`.  Don't use any other name for it.

The APS has been dark for the past year with its upgrade so our experience is a year old now.  The common use case at USAXS has not presented realistic problems for us.  The data processing client ignores the _runs_ that document the next series of scans.  

The snapshot tool was discarded from apstools because users were not interested in this capability.

We resolve the problem of using `scan_id=1` by [overriding](https://github.com/BCDA-APS/bluesky_training/blob/31f30e9747cd0e5a59fb6f8e8d836e5fbff1db50/bluesky/instrument/framework/initialize.py#L119) `RE.scan_id_source`.

stan-dot:
not sure if I understand completely, would you say that we can close this PR @prjemian ?

prjemian:
What's the disposition of `scan_id` now?  Is it a key of the `RE.md` dict or is it an attribute of a run object (whatever that is)?  I'm OK if it is one of these, as long as it is a searchable term in the catalog.

Also, there are a couple merge conflicts to be resolved.

tacaswell:
Currently
 - `'scan_id'` is a special key in the evend-model schemas (https://github.com/bluesky/event-model/blob/145118722a2e2f15abf32010e3e6b71506398140/event_model/schemas/run_start.json#L342-L346) that says that if it is present, then it must be an integer and is the "Scan ID number, not globally unique".  
 - `'scan_id'` is special in the RE as we have code to update it `_open_run` (https://github.com/bluesky/bluesky/blob/997eb2289420da244ba243b65816611403931273/src/bluesky/run_engine.py#L1843-L1844) so that every start document it produces _does_ have one.
 - We do offer a hook to normalize the metadata on the way out https://github.com/bluesky/bluesky/blob/997eb2289420da244ba243b65816611403931273/src/bluesky/run_engine.py#L1864

Due to the schema we can not put `None` in (or the document validation will fail on the way out) and can not avoid adding the key to `RE.md` due to the way `_open_run` is implemented.

Currently the only way to strip `'scan_id'` from a start document is to use a `md_normalizer` to return a copy without the key (which you can do today with out any changes to the RE).

The ask here is to give the ability for the RE to produces start documents that do not include `'scan_id'` in a more natural way.  This current implementation will pop the key out of `RE.md` in some cases which is not great.

I have an idea for an alternative implementation, I'll open a PR shortly.

tacaswell:
You can also currently override `'scan_id'` via `RE(..., scan_id=0)` or `yield from bps.open_run(..., scan_id=0)`.  The scan_id in `RE.md` will still be incremented, but the higher-priority metadata source will be what shows up in the output.

PR #1187, opened 2019-05-15
added ophyd and bluesky version to RunEngine self.md

gwbischof:


gwbischof:
https://github.com/bluesky/bluesky/issues/1050

bluesky/tests/test_metadata.py
```diff
@@ -0,0 +1,14 @@
+import bluesky
+import ophyd
+from bluesky import RunEngine
+
+def test_blueskyversion():
```
danielballan: Please use the RunEngine pytest fixture.

```python
def test_blueskyversion(RE):
    ...
```

bluesky/tests/test_metadata.py
```diff
@@ -0,0 +1,14 @@
...
+    re = RunEngine()
+    print(re.md.get('BLUESKY_VERSION'))
```
danielballan: It's usually better to remove debug prints from tests once they are working, just to avoid noise when someone needs to run `pytest -s` to debug another test.

bluesky/run_engine.py
```diff
@@ -228,6 +228,17 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
         if md is None:
             md = {}
         self.md = md
+
+        try:
+            import ophyd
+            self.md['OPHYD_VERSION'] = ophyd.__version__
+        except ImportError:
+            self.log.exception("Failed to import __version__ from ophyd.")
```
danielballan: In my opinion this is too aggressive. At most this should log at DEBUG level. If I'm using the RunEngine without ophyd, I don't want to see loud ERROR-level tracebacks.

bluesky/tests/test_metadata.py
gwbischof: updated

bluesky/tests/test_metadata.py
gwbischof: updated

bluesky/run_engine.py
gwbischof: updated

bluesky/run_engine.py
```diff
@@ -228,6 +228,18 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
         if md is None:
             md = {}
         self.md = md
+        self.md['versions'] = {}
```
tacaswell: It may be better to use `md.setdefault('version', {})` which will not overwrite if the user passes something in like:

```python
RE = RunEngine({'version': {'my_other_lib': 5}})
```

```suggestion
        self.md.setdefault('versions', {})
```

```

File: bluesky/run_engine.py [hunk 5]
```diff
@@ -228,6 +228,18 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
         if md is None:
             md = {}
         self.md = md
+        self.md.setdefault('versions', {})
+
+        try:
+            import ophyd
+            self.md['versions']['ophyd'] = ophyd.__version__
+        except ImportError:
+            self.log.debug("Failed to import __version__ from ophyd.")
```
mrakitin: I'd be more precise here:
```suggestion
            self.log.debug("Failed to import ophyd.__version__.")
```

bluesky/run_engine.py
```diff
@@ -228,6 +228,18 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
...
+
+        from ._version import get_versions
```
mrakitin: Does not `import .__version__` work?

bluesky/tests/test_metadata.py
```diff
@@ -0,0 +1,9 @@
+import bluesky
+import ophyd
+
+def test_blueskyversion(RE):
+    assert RE.md['versions'].get('bluesky') == bluesky.__version__
+
+def test_ophydversion(RE):
+    assert RE.md['versions'].get('ophyd') == ophyd.__version__
+
```
mrakitin:
```suggestion
```

bluesky/run_engine.py
gwbischof: import .__version__ doesn't seem to work.

mrakitin: OK, thanks for the clarification. Never tried the versioneer from within a package.

bluesky/run_engine.py
```diff
@@ -228,6 +228,18 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
         if md is None:
             md = {}
         self.md = md
+        self.md.setdefault('versions', {})
+
+        try:
+            import ophyd
+            self.md['versions']['ophyd'] = ophyd.__version__
+        except ImportError:
+            self.log.debug("Failed to import ophyd.__version__.")
```
mrakitin: Ah, we actually have to say that importing `ophyd` itself failed.
```suggestion
            self.log.debug("Failed to import ophyd.")
```

bluesky/run_engine.py
danielballan: I think that would have to be `from . import __version__`. And this, as is, is probably slightly safer because it will avoid circular imports.

mrakitin: Great, I like @danielballan's suggestion.

mrakitin: @gwbischof, do you still plan to apply the suggested changes?

danielballan: Sorry, @mrakitin, I did not mean to mark this as resolved just now—that was a misclick.

But I think I phrased my suggestion poorly. I am suggesting to leave this as is because to import from the top level would create a circular import. To explain why: the top-level __init__.py imports from this already and with your change this would import from the top-level __init__.py. As written now there is no circular import.

mrakitin: @danielballan, OK, it's a reasonable point.

mrakitin:
I think we all agreed on the approach, however the tests are failing for Python 3.5 and nightly. Should we ignore the failures?

danielballan:
Yes, it looks like

```
bluesky/tests/test_bec.py::test_plot_ints 
```

has become frequently flaky. Nothing to do with this.

Issue #959, opened 2018-01-24
Make 'state' externally read-only


jrmlhermitte:
As @mrakitin pointed out, state should be property not just a string:

https://github.com/NSLS-II/bluesky/blob/master/bluesky/run_engine.py#L517

mrakitin:
Thanks for documenting it!

tacaswell:
I thought the state was controlled by a super state machine instance....

tacaswell:
https://github.com/NSLS-II/bluesky/blob/4bedbe7f8659beece0c24716c2d885f673dd8031/bluesky/run_engine.py#L184-L186

jrmlhermitte:
but should it be settalbe by user?
ie.
```py
_state = LoggingPropertyMachine(RunEngineStateMachine)

@property
def state(self):
    return self._state
```

danielballan:
This and #960 might be a good enough occasion to rip out `super_state_machine` and replace it with something simpler.

danielballan:
Closed by #1188

PR #1188, opened 2019-05-15
Make 'state' externally read-only #959

jklynch:
This PR addresses issue #959.

## Description

change RunEngine.state to RunEngine._state
create a state property


mrakitin:
@jklynch, congrats on your first contribution! Do you also mind to create a test to actually check that the `RE.state` is read-only via a separate PR? Thanks.

Issue #1185, opened 2019-05-15
Bluesky new logging framework


ke-zhang-rd:
## Expected Behavior

Control flow on handler level 
## Current Behavior

Control flow on logger level
## Possible Solution

new log.py file copy paster most from caproto log [here](https://github.com/caproto/caproto/blob/master/caproto/_log.py) 
## Steps to Reproduce (for bugs)

1.
2.
3.

danielballan:
## Important use cases we want to support

* Log each `Msg` before it is processed.
* When a document is emitted to the callback registry, log its type and uid.
* Log which status objects we are waiting on.
* Log when status objects finish.

# List of loggers we need
* `bluesky`
* `bluesky.emit_document`
* `bluesky.msg`
* `bluesky.status`

## List of extras we need (where applicable)

* `uid`
* `repr(msg)`

ke-zhang-rd:
What do you think could be tags for bluesky log, or do we need them @danielballan

danielballan:
I listed the tags I had in mind under "List of extras" above. Seems like we have similar ideas. What do you mean by "old_status" vs "new_status"? Are you referring to `RE.state`?

danielballan:
For anyone else following along, @ke-zhang-rd has updated the comment at the top of this issue in place to reflect our current plan.

danielballan:
I think the idea of labeling the msg-related log lines is good, but I would prefer less SHOUT-y capitalization and `[]`s. Something like:

```
[D 17:46:32.405      run_engine: 1188] msg save: (None), (), {}
```

danielballan:
On the train yesterday, @tacaswell and @ke-zhang-rd went back and forth about whether a log record should be generated and, if so, how much information should be in that record. Proposals:

1. Do not log the documents at all. We have `RE.subscribe` for that.
2. Log the whole document and let the handler/formatter decide how much of it to incorporate.
3. Log just the document type.
4. Log the document type and the uid of the current Run.

One rejected possibility was to log the document type and document uid. This is not tenable for EventPages and DatumPages, which have a column of uids, not one uid.

Currently, I have a slight preference is (4) but I think this needs more thought and discussion.

ke-zhang-rd:
> On the train yesterday, @tacaswell and @ke-zhang-rd and I went back and forth about whether a log record should be generated and, if so, how much information should be in that record. Proposals:
> 
> 1. Do not log the documents at all. We have `RE.subscribe` for that.
[...]

PR #1193, opened 2019-05-16
ENH: New logging framework for bluesky

ke-zhang-rd:
## Description

From issue #1185 
## Motivation and Context
Logging flow control on handler level and new logging Filter feature

## How Has This Been Tested?
Right now, I'm using a simple script
```
import logging
from ophyd.sim import det1, det2
from bluesky import RunEngine
from bluesky.log import set_handler
hd = set_handler(level='DEBUG')
RE = RunEngine({})
dets = [det1, det2]
RE(count(dets))
```

As I expect
```
[D 00:48:56.449      run_engine: 1734] [event] [d1e59f21-1e1c-4f13-ae2e-725cfbfc1d94] Emitted Event with data keys dict_keys(['det1', 'det2']) (uid='d1e59f21-1e1c-4f13-ae2e-725cfbfc1d94')
```

## checklist

- [x] Test script
- [ ] `.rst` Document update 
- [x] logging message simplify and message pattern optimize


bluesky/run_engine.py
```diff
@@ -78,8 +79,11 @@ def __set__(self, obj, value):
         old_value = self.__get__(obj)
         super().__set__(obj, value)
         value = self.__get__(obj)
-        obj.log.info("Change state on %r from %r -> %r",
-                     obj, old_value, value)
+        tags = {'old_status': old_value,
+                'direction': '--->>>',
```
danielballan: I see that this was inspired by 'direction' in caproto. I don't think we need it here, because it always changes from old to new.

Please name these tags `old_state` and `new_state`. The word "status" usually refers to ophyd Status objects, so this could be confusing.

bluesky/run_engine.py
```diff
@@ -222,7 +226,7 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
 
         # Make a logger for this specific RE instance, using the instance's
         # Python id, to keep from mixing output from separate instances.
-        logger_name = "bluesky.RE.{id}".format(id=id(self))
+        logger_name = 'bluesky.RE.{id}'.format(id=id(self))
```
danielballan: As we did with `caproto.threading.client.Context`, we should remove the templated portion `.{id}` and use just one logger.

bluesky/run_engine.py
```diff
@@ -1372,7 +1376,7 @@ def _open_run(self, msg):
 
         doc = dict(uid=self._run_start_uid, time=ttime.time(), **md)
         yield from self.emit(DocumentNames.start, doc)
-        self.log.debug("Emitted RunStart (uid=%r)", doc['uid'])
+        doc_logger.debug("Emitted RunStart (uid=%r)", doc['uid'], extra={'doc_name': 'Start', 'doc_uid': doc['uid']})
```
danielballan:
```suggestion
        doc_logger.debug("Emitted RunStart (uid=%r)", doc['uid'],
        extra={'doc_name': 'start', 'doc_uid': doc['uid']})
```

bluesky/run_engine.py
```diff
@@ -1430,7 +1434,8 @@ def _close_run(self, msg):
                    num_events=num_events)
         self._clear_run_cache()
         yield from self.emit(DocumentNames.stop, doc)
-        self.log.debug("Emitted RunStop (uid=%r)", doc['uid'])
+        doc_logger.debug("Emitted RunStop (uid=%r)", doc['uid'],
+                         extra={'doc_name': 'Stop', 'doc_uid': doc['uid']})
```
danielballan:
```suggestion
                         extra={'doc_name': 'stop', 'doc_uid': doc['uid']})
```

bluesky/run_engine.py
```diff
@@ -1691,9 +1696,10 @@ def _save(self, msg):
                        configuration=config, name=desc_key,
                        hints=hints, object_keys=object_keys)
             yield from self.emit(DocumentNames.descriptor, doc)
-            self.log.debug("Emitted Event Descriptor with name %r containing "
-                           "data keys %r (uid=%r)", desc_key,
-                           data_keys.keys(), descriptor_uid)
+            doc_logger.debug("Emitted Event Descriptor with name %r containing "
+                             "data keys %r (uid=%r)", desc_key,
+                             data_keys.keys(), descriptor_uid,
+                             extra={'doc_name': 'Descriptor', 'doc_uid': descriptor_uid})
```
danielballan:
```suggestion
                             extra={'doc_name': 'descriptor', 'doc_uid': descriptor_uid})
```

bluesky/run_engine.py
```diff
@@ -1724,8 +1730,8 @@ def _save(self, msg):
                    time=ttime.time(), data=data, timestamps=timestamps,
                    seq_num=seq_num, uid=event_uid, filled=filled)
         yield from self.emit(DocumentNames.event, doc)
-        self.log.debug("Emitted Event with data keys %r (uid=%r)", data.keys(),
-                       event_uid)
+        doc_logger.debug("Emitted Event with data keys %r (uid=%r)", data.keys(),
+                         event_uid, extra={'doc_name': 'Event', 'doc_uid': event_uid})
```
danielballan:
```suggestion
                         event_uid, extra={'doc_name': 'event', 'doc_uid': event_uid})
```

bluesky/log.py
```diff
@@ -125,22 +136,200 @@ def format(self, record):
...
+    """
+    warnings.warn(f"The function color_logs is deprecated. "
+                  f"Use `config_bluesky_logging(color={color})` instead.")
+    config_bluesky_logging(color=color)
+
+
 logger = logging.getLogger('bluesky')
+doc_logger = logging.getLogger('bluesky.emit_document')
+msg_logger = logging.getLogger('bluesky.msg')
+status_logger = logging.getLogger('bluesky.status')
```
danielballan: Please rename to `state_logger` and `'bluesky.RE.state'`

bluesky/log.py
```diff
@@ -125,22 +136,200 @@ def format(self, record):
...
+    If False, do the opposite.
+    """
+    warnings.warn(f"The function color_logs is deprecated. "
+                  f"Use `config_bluesky_logging(color={color})` instead.")
+    config_bluesky_logging(color=color)
+
+
 logger = logging.getLogger('bluesky')
+doc_logger = logging.getLogger('bluesky.emit_document')
+msg_logger = logging.getLogger('bluesky.msg')
```
danielballan: I think this should be bluesky.RE.msg. But bluesky.emit_document can stay as it is because the document-emitting code may someday move outside the RunEngine.

bluesky/log.py
```diff
@@ -35,22 +38,16 @@ def _stderr_supports_color():
...
-    (unless ``--logging=none`` is used).
-    Color support on Windows versions that do not support ANSI color codes is
-    enabled by use of the colorama__ library. Applications that wish to use
-    this must first initialize colorama with a call to ``colorama.init``.
-    See the colorama documentation for details.
-    __ https://pypi.python.org/pypi/colorama
-    .. versionchanged:: 4.5
-       Added support for ``colorama``. Changed the constructor
-       signature to be compatible with `logging.config.dictConfig`.
+    * Includes extra record attributes (pv, our_address, their_address,
```
danielballan: This needs to be updated.

bluesky/log.py
```diff
@@ -106,7 +103,21 @@ def __init__(self, fmt=DEFAULT_FORMAT, datefmt=DEFAULT_DATE_FORMAT,
             self._normal = ''
 
     def format(self, record):
-        record.message = record.getMessage()
+        message = []
+        if hasattr(record, 'old_status'):
```
danielballan: I don't know that we need any special additions to the formatter in this case. It looks like the log messages themselves are enough. I think the value of `extra` will just be in providing structured information to handlers that accept structured data, like logstash.

bluesky/log.py
```diff
@@ -125,22 +136,200 @@ def format(self, record):
...
+    if isinstance(levelno, int):
+        return levelno
+    else:
+        raise ValueError("Your level is illegal, please use one of python logging string")
 
-current_handler = None  # overwritten below
 
+class DocFilter(logging.Filter):
+    '''
+    Block any message that is lower than certain level except it related to target
```
danielballan: Something I missed in the caproto review: the first section of a docstring must be one line followed by a blank line, or else Sphinx will not format it properly. I suggest:

Filter records below a given level by document type.

bluesky/log.py
```diff
@@ -125,22 +136,200 @@ def format(self, record):
...
+    documents. You have option to choose whether env, config and misc message is exclusive
+    or not.
+
+    Parameters
+    ----------
+    names : string or list of string
+        Name of document type list which will be filtered in, one of 'Start', 'Descriptor', 'Event' and 'Stop'.
```
danielballan: The strings should be lowercase, as in the convention everywhere else in bluesky. And datum and resource should be added.

bluesky/log.py
```diff
@@ -125,22 +136,200 @@ def format(self, record):
...
+        If True, records for which this filter is "not applicable" (i.e. the
+        relevant extra context is not present) will be blocked. False by
+        default.
+
+    Returns
+    -------
+    passes : bool
+    '''
+
+    def __init__(self, *names, level='WARNING', exclusive=False):
```
danielballan: This should be `commands`.

bluesky/log.py
```diff
@@ -1,8 +1,9 @@
 # The LogFormatter is adapted light from tornado, which is licensed under
 # Apache 2.0. See other_licenses/ in the repository directory.
-
+import fnmatch
```
danielballan: I don't think we need support partial-matching in any of the filters here. All usages of `fnmatch` may be removed.

bluesky/run_engine.py
ke-zhang-rd: Without arrow, I feel [state1] [state2] looks unclear. Probably, [Old: <state1>] [New: <state2>]?

bluesky/run_engine.py
ke-zhang-rd: That was my initial thinking. I was thinking you probably want to have id to do some fancy thing

bluesky/run_engine.py
ke-zhang-rd: If so, tags of document and tags of Msg are not esay to figure out

bluesky/log.py
ke-zhang-rd: My motivation is let each logging message easy to be figured whether they are create by docuemnt_emit, state_chang or msg command. Put tags in front of message might be useful?

bluesky/run_engine.py
danielballan: I think in this case the log message makes it clear. Elsewhere I suggested that we do not need to include the extras in the Formatter the way we did with caproto.

danielballan: I think in this case the log message makes it clear. Elsewhere I suggested that we do not need to include the extras in the Formatter the way we did with caproto.

ke-zhang-rd: Also, if we don't have that tag, it isn't very clear how Filter work. not clear which one is filtered out.

danielballan: I don't understand how having a tag `'direction'` that always has the same value is helpful.

bluesky/log.py
danielballan: I don't think that most users will want to think about it in those terms. A message saying "The state changed from X to Y," or "A __ document has been emitted (uid=__)" is sufficiently clear and nicely readable.

The slightly different approach in caproto was serving a different purpose, where there were many overlapping tags, and a more engineering-focused "audience".

bluesky/run_engine.py
stuartcampbell: @ke-zhang-rd This is the sort of thing we are trying to avoid by having the design discussions on the issue before investing work in the code.

danielballan:
Superseded by #1293.

PR #1190, opened 2019-05-15
modifying list_scan to accept multiple motors

awalter-bnl:
modifying list_scan to accept multiple motor/list pairs

## Description
Current API:
```
list_scan(dets,` motor, [...])  # supports one motor only
scan(dets, motor, start, stop, motor2, start2, stop2, num_steps)  # "inner product" over motors
```
Proposed API:
```
list_scan(dets, motor, [...], motor2, [...])  # If list lengths are unequal, blow up.
list_grid_scan(dets, motor, [...], motor2, [...], motor3, [...], snaking=True|False|tuple)
```
## Motivation and Context
fixes #1144 

## How Has This Been Tested?
has not yet been tested

PR #1194, opened 2019-05-17
multi-motor log_scans

### LINKED ISSUE #1144: Generalize list_scan and log_scan same as scan.

awalter-bnl:
modifying log_scan to accept multiple motor/list pairs

## Description
Current API:
```
log_scan(dets, motor, start, stop, num )  # supports one motor only
scan(dets, motor, start, stop, motor2, start2, stop2, num_steps)  # "inner product" over motors
```
Proposed API:
```
log_scan(dets, motor1, start1, stop1, motor2, start2, stop2, num, ...)  #
log_grid_scan(dets, motor,start1, stop1, num1, motor2, start2,stop2, num2, ..., snaking=True|False|tuple)
```
Also make similar changes to `rel_log_scan` and make `rel_log_grid_scan`

## Motivation and Context
fixes #1144 

## How Has This Been Tested?
tests of the new features have been added, they won't pass until PR #1190 is merged however

## NOTE: this relies on bluesky PR# [1190](https://github.com/bluesky/bluesky/pull/1190) being merged first which adds similar funcitonality to `list_scan`

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -17,7 +17,7 @@ from . import plan_patterns
 
 from . import utils
 from .utils import Msg
-
+import math
 from . import preprocessors as bpp
 from . import plan_stubs as bps
 
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -258,22 +258,40 @@ def _rel_scan_1d(detectors, motor, start, stop, num, *, per_step=None,
     return (yield from inner_relative_scan())
 
 
-def log_scan(detectors, motor, start, stop, num, *, per_step=None, md=None):
+def log_scan(detectors, *args, num, base=10, per_step=None, md=None):
     """
-    Scan over one variable in log-spaced steps.
+    Scan over one, or more, variable(s) in log-spaced steps simultaneously
+    (inner product).
 
     Parameters
     ----------
     detectors : list
         list of 'readable' objects
-    motor : object
-        any 'settable' object (motor, temp controller, etc.)
-    start : float
-        starting position of motor
-    stop : float
-        ending position of motor
+    *args :
+        For one dimension, ``motor, start, stop``.
+        In general:
+
+        .. code-block:: python
+
+            motor0, start1, stop1,
+            motor1, start2, start2,
+            ...,
+            motorN, startN, stopN
+
+        Motors can be any 'settable' object (motor, temp controller, etc.)
+
+        The values from above are defined as:
+
+            motor : object
+                any 'settable' object (motor, temp controller, etc.)
+            start : float
+                starting position for motor.
+            stop : float
+                ending position for motor.
     num : int
         number of steps
+    base : float
+        base of log scale, defaults to 10
     per_step : callable, optional
         hook for customizing action of inner loop (messages per step)
         Expected signature: ``f(detectors, motor, step)``
@@ -283,18 +301,36 @@ def log_scan(detectors, motor, start, stop, num, *, per_step=None, md=None):
     See Also
     --------
     :func:`bluesky.plans.rel_log_scan`
+    :func:`bluesky.plans.log_grid_scan`
+    :func:`bluesky.plans.rel_log_grid_scan`
     """
+
+    if len(args) % 3 != 0:
+        raise ValueError('The list of arguments handed to ``log_scan()`` must '
+                         'contain multiples of the 3 arguments `motor`, '
+                         '`start`, `stop`. However what was recieved is of '
+                         'length {len(args)} which is not divisble by 3.'
+                         'The full set of arguments are {}.'.format(len(args),
+                                                                    args))
+
+    list_args = []
+    repr_args = []
+    for motor, start, stop in partition(3, args):
+        if start <= 0 or stop <= 0:  # resolve 'math.log' error if necesary
+            offset = -1*min(start, stop)+1
+        else:
+            offset = 0
+
+        list_args.extend([motor, np.logspace(math.log(start+offset, base),
+                                             math.log(stop+offset, base), num,
+                                             base=base)-offset])
+        repr_args.extend([repr(motor), start, stop])
+
     _md = {'detectors': [det.name for det in detectors],
-           'motors': [motor.name],
-           'num_points': num,
-           'num_intervals': num - 1,
-           'plan_args': {'detectors': list(map(repr, detectors)), 'num': num,
-                         'start': start, 'stop': stop, 'motor': repr(motor),
+           'plan_args': {'detectors': list(map(repr, detectors)),
+                         'args': repr_args, 'num': num, 'base': base,
                          'per_step': repr(per_step)},
            'plan_name': 'log_scan',
-           'plan_pattern': 'logspace',
-           'plan_pattern_module': 'numpy',
-           'plan_pattern_args': dict(start=start, stop=stop, num=num),
            'hints': {},
            }
     _md.update(md or {})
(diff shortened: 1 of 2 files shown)


awalter-bnl:
I have opened this to get feedback on the proposed structure of a change to `log_scan` to allow multi-axes. all feedback is welcomed !-). Once we get general agreement on the structure I will flesh it out, add tests and confirm the existing test work.

danielballan:
I have received feedback in the past that log-spaced scanning is not actually useful. We added it for the sake of symmetry, as `numpy` provides `linspace` and `logspace`. Are we aware that anyone has used it?

awalter-bnl:
@danielballan I don't disagree that it is not generally useful and don't know if anyone uses it. but it is not adding much overhead to update/maintain them as it will just be a wrapper around `list_scan` and it's variants.

awalter-bnl:
This PR is now ready to review, note that it requires PR #1190 (the list scan counterpart) to be merged in order for the tests to pass. But I have tested them locally and it appears to work.

danielballan:
CI is failing on this.

awalter-bnl:
Cycling to rerun the CI, which needed the new list scan PR to pass.

awalter-bnl:
The CI failure is weird to me, it is failing in building the tutorial docs, but they weren't touched in this PR at all. Will need some time to investigate.

>[31;01mException occurred:[39;49;00m
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/IPython/sphinxext/ipython_directive.py", line 583, in process_input
    raise RuntimeError('Non Expected warning in `{}` line {}'.format(filename, lineno))
RuntimeError: Non Expected warning in `/home/travis/build/bluesky/bluesky/doc/source/tutorial.rst` line 599
[31;01mThe full traceback has been saved in /tmp/sphinx-err-i5r40ay0.log, if you want to report the issue to the developers.[39;49;00m
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
Makefile:61: recipe for target 'html' failed
make: *** [html] Error 1
make: Leaving directory '/home/travis/build/bluesky/bluesky/doc'

awalter-bnl:
cycling to try to get tests to pass again

awalter-bnl:
This is now passing, thanks @danielballan for the fix

bluesky/plans.py
```diff
@@ -17,7 +17,7 @@
 
 from . import utils
 from .utils import Msg
-
+import math
```
danielballan: Might as well use `numpy.log` here (and in the tests) for consistency and to avoid paying the overload of importing `math`.

bluesky/plans.py
```diff
@@ -283,18 +301,36 @@ def log_scan(detectors, motor, start, stop, num, *, per_step=None, md=None):
...
+
+    list_args = []
+    repr_args = []
+    for motor, start, stop in partition(3, args):
+        if start <= 0 or stop <= 0:  # resolve 'math.log' error if necesary
+            offset = -1*min(start, stop)+1
+        else:
+            offset = 0
+
+        list_args.extend([motor, np.logspace(math.log(start+offset, base),
```
danielballan: I think `offset` is not necessary and this should be simply `np.logspace(start, stop, num, base=base`)`, it was before. This change seems to be an API break unless I am misunderstanding.

From `numpy.logspace`:
```
Parameters
----------
start : array_like
    ``base ** start`` is the starting value of the sequence.
stop : array_like
    ``base ** stop`` is the final value of the sequence, unless `endpoint`
    is False.  In that case, ``num + 1`` values are spaced over the
    interval in log-space, of which all but the last (a sequence of
    length `num`) are returned.
```

awalter-bnl: @danielballan I should have documented this correctly.

I don't think this is an API break as per the current documentation below start and stop are the start postion and strop position, as they are for all other plans we have:
>def log_scan(detectors, motor, start, stop, num, *, per_step=None, md=None):
    """
    Scan over one variable in log-spaced steps.
    Parameters
    ----------
    detectors : list
        list of 'readable' objects
    motor : object
        any 'settable' object (motor, temp controller, etc.)
    start : float
        starting position of motor
    stop : float
        ending position of motor
    num : int
        number of steps
    per_step : callable, optional
        hook for customizing action of inner loop (messages per step)
        Expected signature: ``f(detectors, motor, step)``
    md : dict, optional
        metadata
    See Also
    --------
    :func:`bluesky.plans.rel_log_scan`
    """

Adding `offset` in here allows for <=0 start  and/or stop positions, it is set to zero for all other cases. So this is an extension not an API break.

I am however happy for this to wait until detailed review.

bluesky/plans.py
awalter-bnl: @danielballan `numnpy.log` does not have the 'base' kwarg that is required here, hence why i used `math.log` instead.

bluesky/plans.py
danielballan: I agree offset is an extension. Focusing on the other part, I think it is possible to interpret changing start and stop as either:

* a bug fix — it changes the behavior to match the documentation and to be consistent with the other plans, or
* an API change — typing the same arguments now moves the motors to a different place than before.

I mostly agree that this PR would be an improvement, but it does seem a potentially dangerous to change a plan in a way that causes it to put the motors in very different positions than it did before. Anyone who is using it might be caught by surprise by the change.

Radical proposal: if we agree that the current implementation is not good, we might consider addressing that by deprecating log_scan entirely rather than fixing it in a way that could break user code in a surprising way. We agree that usage is rare, possibly nonexistent. Maybe it’s better to let the individual beamline scientist choose the parametrization that makes sense to their users by way of a custom plan that wraps list_scan and/or list_grid_scan.

awalter-bnl: Thanks for the feedback @danielballan 

>an API change — typing the same arguments now moves the motors to a different place than before.

I am just now realizing that the current `log_scan` behaviour is not consistent with it's docstring. I agree that this could be then viewed as an API change, and that it is probably a bad move.

>Radical proposal: if we agree that the current implementation is not good, we might consider addressing that by deprecating log_scan entirely rather than fixing it in a way that could break user code in a surprising way.

As a counter-proposal how about we rename the scans in this PR as 'logstep_scan' , ....
This has the benefits:
- it removes the potential API break, and introduces the API consistent with existing plans.
- it's name matches more closely to it's purpose (to scan in log-spaced steps over one or more axis/axes).

We could then consider deprecating `log_scan` long term.

danielballan: I don't have strong feelings either way, but I would lean toward not adding two new plans if we think that are rarely/never used. Might be worth pulling this back up to a GH issue for wider consideration.

danielballan:
Summary of discussion in Monday meeting:

We agree that current signature of `log_scan`, where `start` and `stop` mean what they mean in `numpy.logspace` and not the actual starting and stopping positions, is potentially confusing.

My view is that we should deprecate `log_scan` and not add any new log-scaled plans to the library because:
* In four years, we don't know of _any_ times that `log_scan` has actually been used in the wild.
* The potential uses for it that we can imagine would need something a little bit different than `log_scan` or the variation proposed here. The user would have to build on `list_scan`, which is easy enough.
* A user building their own thing on `list_scan` can choose a signature that is clear and meaningful for their application.
* A "mesh log scan" seems like a _very_ rare beast to me. More likely one would want to scan with log spacing on some dimensions and linear spacing on others, which again can be done easily enough with `list_grid_scan`.
* Merging something "just in case we need it" crowds the documentation and adds to the maintenance burden.

@awalter-bnl's view is that we should deprecate `log_scan` and add new versions with new names like `log_step_scan` and `log_step_grid_scan` because:
* They have already been written in this PR. If we decline to merge and then discover a need a year or more from now, the work will likely need to be redone due to bit rot.

Neither of us feels strongly enough about our position to carry the day. Would others weigh in?

awalter-bnl:
am closing this as the lack of recent discussion indicates it is no longer required.

PR #1195, opened 2019-05-17
add test_state_is_readonly

jklynch:
A unit test verifying RunEngine.state is not assignable.

## Description
def test_state_is_readonly(RE):
    with pytest.raises(AttributeError):
        RE.state = 'running'

## Motivation and Context
This is in support of #1188.


tacaswell:
I took the liberty of making it non-draft and merging it.

Issue #1197, opened 2019-05-18
Use new jsonschema types API.


danielballan:
We are using a part of the jsonschema API that is no longer supported and therefore forced to pin to `jsonschema<3`. We should fix that so we are compatible with the latest releases of `jsonschema`.

danielballan:
Specifically the `types` argument used here

https://github.com/bluesky/bluesky/blob/72588b76ed3f7af053a8f5fd97524d05efde1d5c/bluesky/run_engine.py#L28

has been deprecated.

danielballan:
It looks like [this](https://python-jsonschema.readthedocs.io/en/stable/validate/#validating-types) is the new way of handling be tolerant about types.

jklynch:
The 'new way' might look something like this:
```
def is_array(checker, instance):
    return (
        jsonschema.validators.Draft7Validator.TYPE_CHECKER.is_type(instance, 'array') or
        isinstance(instance, tuple)
    )

array_type_checker = jsonschema.validators.Draft7Validator.TYPE_CHECKER.redefine(
    'array',
    is_array
)

Validator = jsonschema.validators.extend(
    jsonschema.validators.Draft7Validator,
    type_checker=array_type_checker
)

```
A validator object must be instantiated with a schema so what currently looks like this:
```
_validate(doc, schemas[DocumentNames.event])
```
could be replaced with one line like this:
```
Validator(schema=schemas[DocumentNames.event]).validate(doc)
```
It might be a little nicer to cache the validators:
```
schema_validators[DocumentNames.event].validate(doc)
```

jklynch:
Also is there a place all that new code could go? The event-model project seems like the best place but maybe you would rather leave that alone?

danielballan:
I think you are right that a dictionary `schema_validators` would be handy to have in the event-model package.

danielballan:
Closed by #1200

PR #1200, opened 2019-05-22
New jsonschema types api #1197

jklynch:
This PR depends on https://github.com/bluesky/event-model/pull/79.
Replace calls to deprecated jsonschema.validate() with sanctioned jsonschema.Validator.validate().

## Description
Use a dict of jsonschema.Validator imported from event-model.
Remove jsonschema<3 pin from requirements.txt.

## Motivation and Context
This change is simply to get away from the deprecated jsonschema.validate().

## How Has This Been Tested?
Existing tests pass.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 cycler
 event-model>=1.5.0,!=1.8.0
 historydict
-jsonschema
 numpy
 super_state_machine
 toolz
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 cycler
-event-model>=1.5.0,!=1.8.0
+event-model==1.10.0rc1
 historydict
 numpy
 super_state_machine
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 cycler
-event-model>=1.10.0rc1
+event-model>=1.10.0
 historydict
 numpy
 super_state_machine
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 cycler
 event-model>=1.5.0,!=1.8.0
 historydict
-jsonschema<3
+jsonschema
```
danielballan: I believe that, with this change, `jsonschema` is longer a direct dependency of bluesky (i.e. we don't import it anymore) and can therefore be removed from this list. It will come in through our `event-model` dependency and we can manage the `jsonschema` pinning there, which is easier than keeping both libraries in sync.

jklynch: I did not notice that.

jklynch:
PR https://github.com/bluesky/event-model/pull/79 will have to be merged before this one will pass

requirements.txt
```diff
@@ -1,7 +1,6 @@
 cycler
-event-model>=1.5.0,!=1.8.0
+event-model==1.10.0rc1
```
tacaswell:
```suggestion
event-model>=1.10.0rc1
```

Use a lower bound so that when we do an event-model final we don't have to re-pin this!

jklynch:
event-model 1.10.0rc1 setup.py specifies minimum Python version 3.6, so the bluesky Python 3.5 CI build fails to install event-model

mrakitin:
Didn't we decide to abandon Python 3.5?

PR #1204, opened 2019-05-27
BLD: Add python_requires.

danielballan:
This field in `setup(...)`, now part of our standard cookiecutter for
new projects, was never added to bluesky.


jklynch:
Reviewed and tested against 3.6.

PR #1206, opened 2019-05-28
small update to spiral_fermat and rel_spiral_fermat to add ellipticty

### LINKED ISSUE #648: elliptical spiral and fermat spiral scans
Can we add the ability for step sizes to be different in the x and y direction (elliptical rather than circular) in spiral and fermat_spiral scans?

awalter-bnl:
It appears that elliptical spiral scan capability was added to both ``spiral_pattern`` and ``spiral_fermat_pattern`` but only then propagated up to the ``spiral`` plan. This PR adds the required ``kwarg`` to ``spiral_fermat`` to complete this fix.

## Description
Added the ``kwarg`` `dr_y=None`` to  ``rel_spiral``, ``spiral_fermat`` and ``rel_spiral_fermat`` which is passed down to to the associated plan or plan_pattern (where it was already included). It also makes a few 'pep-8' style changes.

## Motivation and Context
This adds elliptical fermat spiral capability to the ``spiral_fermat`` plan
fixes #648 

## How Has This Been Tested?
This has been tested manually and passed (locally) the CI testing.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1101,7 +1101,8 @@ def tweak(detector, target_field, motor, step, *, md=None):
 
 
 def spiral_fermat(detectors, x_motor, y_motor, x_start, y_start, x_range,
-                  y_range, dr, factor, *, tilt=0.0, per_step=None, md=None):
+                  y_range, dr, factor, *, dr_y=None, tilt=0.0, per_step=None,
+                  md=None):
     '''Absolute fermat spiral scan, centered around (x_start, y_start)
 
     Parameters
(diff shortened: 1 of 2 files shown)


bluesky/plans.py
```diff
@@ -1124,6 +1125,9 @@ def spiral_fermat(detectors, x_motor, y_motor, x_start, y_start, x_range,
         delta radius
     factor : float
         radius gets divided by this
+    dr_y : float, optional
+        Delta radius along the major axis of the ellipse, if not specifed
+        defaults to dr.
```
mrakitin: I don't see it's assigned `dr` anywhere in the code below. Only `dr_y=dr_y` is used. Could you please point me to the place where this default is processed?

awalter-bnl: It is done in a for loop in the `spiral_fermat` plan pattern [here](https://github.com/bluesky/bluesky/blob/99eb69cd2d55c42103607765c4439833fac8ee2e/bluesky/plan_patterns.py#L242) in a round about way. we actually use `dr_aspect` (`dr/dr_y`) to create the ellipticity in the plan pattern and set it to 1 (i.e `dr_y` = `dr` ) if `dr_y` is `None`.

Does this clarify your question @mrakitin?

mrakitin: Thanks for the explanation, sounds good!

PR #1212, opened 2019-06-06
FIX: Include 'extents' in list_grid_scan md.

danielballan:
The ``LiveGrid`` callback (and presumably other consumers yet to be
written) makes use of the ``'extents'`` metadata from ``grid_scan``. The
new plan ``list_grid_scan`` should include ``'extents'`` as well.

Follow up on #1190.


danielballan:
#1213 has happened twice in a row. Rerunning the build again...

Issue #1207, opened 2019-05-28
Change default serialization in bluesky.callbacks.zmq to msgpack-numpy


danielballan:
~We currently JSON-serialize the data pushed into `bluesky.callbacks.zmq`. Clearly, serializing to a binary format would be faster.~ In fact, we moved from JSON to pickle some time ago.

As we buy into msgpack more and more across the project, it seems like a good default choice. Note that the serializer is _configurable_ so users who want to use a different serializer may still do so.

See https://github.com/bluesky/suitcase-msgpack/pull/2#issuecomment-492295894 for a usage example.

Additionally, [this logging code](https://github.com/bluesky/bluesky/blob/99eb69cd2d55c42103607765c4439833fac8ee2e/scripts/bluesky-0MQ-proxy#L25-L35) relies of the `doc` being serialized in a human-readable text format. That should be pulled up from the script `bluesky-0MQ-proxy` into [the `Proxy` class itself](https://github.com/bluesky/bluesky/blob/master/bluesky/callbacks/zmq.py#L82) so that the deserializer is configurable for the purposes of showing readable log messages.

danielballan:
See #1214 for benchmarks. Using msgpack-numpy isn't faster, and pickle is built in. For this lightweight message bus, using pickle feels right. Closing.

PR #1214, opened 2019-06-07
change default serialization for ZMQ to msgpack

jklynch:
This PR sets default serialization for ZMQ to msgpack. **This change breaks two unit tests because msgpack does not distinguish lists from tuples.** Deserialized data can have lists or tuples but not both.

The default value for bluesky.callbacks.zmq.Publisher is set to 
 ```python
serializer=partial(
    msgpack.packb,
    use_bin_type=True,
    default=msgpack_numpy.encode)
```
The default value for bluesky.callbacks.zmq.RemoteDispatcher is set to
```python
deserializer=partial(
    msgpack.unpackb,
    use_list=True,
    raw=False,
    object_hook=msgpack_numpy.decode)
```

## Motivation and Context
This change addresses issue #1207. 

## How Has This Been Tested?
All bluesky unit tests pass except test_zmq() and test_zmq_prefix().

This is the pytest output of the problematic deserialization done both ways.

First with `msgpack.unpackb(..., use_list=True)`:
```python
E         [('start',
E         {'detectors': ['det'],
E         -    'hints': {'dimensions': [[['time'], 'primary']]},   <- output from msgpack.unpackb
E         ?                             ^^      -           ^
E         +    'hints': {'dimensions': [(('time',), 'primary')]},  <- input to msgpack.packb
E         ?                             ^^       ++          ^
E         'num_intervals': 0,
E         'num_points': 1,
```
Now with `msgpack.unpackb(..., use_list=False)`.
```python
E         Full diff:
E         [('start',
E         -   {'detectors': ('det',),  <- output from msgpack.unpackb
E         ?                 ^      --
E         +   {'detectors': ['det'],   <- input to msgpack.packb
E         ?                 ^     +
E         -    'hints': {'dimensions': ((('time',), 'primary'),)},  <- output from msgpack.unpackb
E         ?                            ^                      ^^
E         +    'hints': {'dimensions': [(('time',), 'primary')]},   <- input to msgpack.packb
E         ?                            ^                      ^
E         'num_intervals': 0,
E         'num_points': 1,
```

### DIFF
--- a/bluesky/callbacks/zmq.py
+++ b/bluesky/callbacks/zmq.py
@@ -1,9 +1,6 @@
 import asyncio
 import copy
-import multiprocessing
 import pickle
-import socket
-import time
 import warnings
 from ..run_engine import Dispatcher, DocumentNames
 
--- a/bluesky/callbacks/zmq.py
+++ b/bluesky/callbacks/zmq.py
@@ -2,7 +2,15 @@ import asyncio
 import copy
 import pickle
 import warnings
+
+import msgpack
+import msgpack_numpy
+
 from ..run_engine import Dispatcher, DocumentNames
+
+
+def msgpack_serialize(document)
+    return msgpack.packb(document, default=msgpack_numpy.encode)
 
 
 class Publisher:
@@ -37,7 +45,7 @@ class Publisher:
     >>> RE.subscribe(publisher)
     """
     def __init__(self, address, *, prefix=b'',
-                 RE=None, zmq=None, serializer=pickle.dumps):
+                 RE=None, zmq=None, serializer=msgpack_serialize):
         if RE is not None:
             warnings.warn("The RE argument to Publisher is deprecated and "
                           "will be removed in a future release of bluesky. "
--- a/bluesky/callbacks/zmq.py
+++ b/bluesky/callbacks/zmq.py
@@ -1,16 +1,12 @@
 import asyncio
 import copy
-import pickle
+from functools import partial
 import warnings
 
 import msgpack
 import msgpack_numpy
 
 from ..run_engine import Dispatcher, DocumentNames
-
-
-def msgpack_serialize(document)
-    return msgpack.packb(document, default=msgpack_numpy.encode)
 
 
 class Publisher:
@@ -45,7 +41,10 @@ class Publisher:
     >>> RE.subscribe(publisher)
     """
     def __init__(self, address, *, prefix=b'',
-                 RE=None, zmq=None, serializer=msgpack_serialize):
+                 RE=None, zmq=None,
+                 serializer=partial(msgpack.packb,
+                                    use_bin_type=True,
+                                    default=msgpack_numpy.encode)):
         if RE is not None:
             warnings.warn("The RE argument to Publisher is deprecated and "
                           "will be removed in a future release of bluesky. "
@@ -224,7 +223,10 @@ class RemoteDispatcher(Dispatcher):
     """
     def __init__(self, address, *, prefix=b'',
                  loop=None, zmq=None, zmq_asyncio=None,
-                 deserializer=pickle.loads):
+                 deserializer=partial(msgpack.unpackb,
+                                      use_list=True,
+                                      raw=False,
+                                      object_hook=msgpack_numpy.decode)):
         if isinstance(prefix, str):
             raise ValueError("prefix must be bytes, not string")
         if b' ' in prefix:


danielballan:
I evidently mis-remembered the current state of bluesky. In #1207 I wrote, mistakenly:

> We currently JSON-serialize the data pushed into bluesky.callbacks.zmq.

I am sure that at one point we _did_ use JSON, but now I see that we _currently_ use pickle, thanks I believe to some work by @CJ-Wright that I had forgotten about. Pickle is already a step in the right direction---being binary, is would be faster and more compact than JSON---and has the advantage of being built-in. Msgpack has the reputation for being faster and more compact. Is that speed and compactness worth introducing a dependency?

@jklynch, do you think it would be worthwhile to compare the size (i.e. `len` of the encoded bytestream) and serialization/deserialization time (via `%timeit`) for realistic payloads?

jklynch:
@danielballan I will do that comparison. My instinct is that there will be very little difference between JSON, pickle, and msgpack but I would like to find out for certain.

CJ-Wright:
@danielballan The move to pickle is because the detectors use `OrderedDict` for their configuration in descriptors, the ast handling was not able to figure it out and we moved to pickle.

danielballan:
Thanks for the reminder @CJ-Wright. I think we weren't actually doing JSON serialization (that wouldn't involve `ast`, right?) but rather we were doing this:

https://github.com/bluesky/bluesky/blob/ada5f6f832c2a2526fbbb7e762cca152b11397bf/bluesky/callbacks/zmqpub.py#L49

which looking back seems to me now like an embarrassing hack! If we did `json.dumps((name, doc), cls=event_model.NumpyEncoder)` I think JSON would work fine. My money is on pickle and msgpack-numpy being significantly faster than JSON---we'll see!

mrakitin:
> I am sure that at one point we _did_ use JSON, but now I see that we _currently_ use pickle, thanks I believe to some work by @CJ-Wright that I had forgotten about. Pickle is already a step in the right direction---being binary, is would be faster and more compact than JSON---and has the advantage of being built-in. Msgpack has the reputation for being faster and more compact. Is that speed and compactness worth introducing a dependency?

I think it's worth it if the benchmark tests show a significant speed-up.

CJ-Wright:
Would it be possible for the benchmarks include 0, 1, 2, and 3D data? I'm particularly interested in 2k square and 4k square images with both uint16 and float32 dtypes.

jklynch:
@CJ-Wright  Is there a simulated detector that will generate that kind of data, or some other good way to do it?

danielballan:
```python
from bluesky import RunEngine
from ophyd.sim import SynSignal
import numpy
RE = RunEngine({})
det = SynSignal(func=lambda: numpy.random.randint(0, 255, (2000, 2000), dtype=numpy.uint16), name='det')
output = []  # will contain (name, doc) pairs
RE(count([det]), lambda name, doc: output.append((name, doc)))
# Now serialize `output`
```

danielballan:
This simple test is also suggestive:

```python
In [18]: arr = numpy.random.randn(4000, 4000)

In [19]: %timeit json.dumps(arr, cls=event_model.NumpyEncoder)
9.87 s ± 171 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [21]: %timeit pickle.dumps(arr)
91.5 ms ± 635 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [23]: %timeit msgpack.packb(arr, use_bin_type=True, default=msgpack_numpy.encode)
92.6 ms ± 830 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

 is still worthwhile to look at more realistic payloads (dicts, dicts of numpy arrays) as suggested above.

jklynch:
I stand corrected:
 - pickle.dumps is fastest, has smallest output
 - msgpack.packb is a close second
 - json.dumps is dead last
(on fairly realistic payloads)

2000x2000 uint16
json.dumps:
times: [3.34, 3.10, 3.12]
len: 201079620
pickle.dumps:
times: [0.02, 0.02, 0.02]
len: 80007017
msgpack.packb:
times: [0.05, 0.05, 0.05]
len: 88007745

2000x2000 float32
json.dumps
times: [22.3, 21.6, 21.7]
len: 891886876
pickle.dumps
times: [0.12, 0.13, 0.12]
len: 160007320
msgpack.packb
times: [0.15, 0.14, 0.12]
len: 176008048

danielballan:
Great! Thanks for these numbers, @jklynch, and thanks also for your my patience with me having jumped the gun in suggesting you implement this before we all got the facts straight.

I can think of some advantages of msgpack over pickle _in general_ in our stack:

* It safe to decode msgpack from an untrusted source, whereas decoding pickle can execute arbitrary code.
* Pickle is explicitly intended for transient inter-process or inter-host communication but not for long-term on-disk storage.
* Msgpack can be read by [many languages](https://msgpack.org/index.html) whereas pickle is a Python thing.

I don't see that any of those advantages apply to our lightweight tool for pushing documents between two bluesky-aware components. But I think msgpack-numpy as the default encoding for _Kafka_ might make more sense, as it is easier to imagine consumers of that stream written in other languages that can't decode pickle but can decode msgpack. We have also discussed msgpack as a straightforward archival (i.e. tape) format.

Given that, I'm inclined to leave the 0MQ stuff as is, and focus on Kafka. Thoughts?

mrakitin:
Agree on having 2 approaches - 0MQ and kafka.

jklynch:
I do like msgpack for being language-agnostic, and opening up the Kafka messages to languages other than Python seems like a good idea.

jklynch:
Decision is to leave pickle as default ZMQ serialization.

PR #1215, opened 2019-06-11
TST: make three-hit-sigint tests less prone to race conditions

tacaswell:

PR #1216, opened 2019-06-13
Clean up sigint test and skip on OSX on travis

tacaswell:
It was noted that we would sometimes see rouge SIGINTs escaping from
the test suite which would pre-maturely exit the tests.  On further
investigation we found

 1. there was a race condition in the tests that gave a window for the
    killer threads to fire when our context manager was not in place
    and hence the SIGINT would raise a KeybroardIntreupt someplace
    else in the test (sometimes in the pytest / coverage machinery)
 2. the file descriptor wake-up machinery does not work on OSX on
    travis (see #1215 for attempts to debug this).  It works exactly
    as expected on linux on travis and locally and locally on OSX.

This PR cleans the test up and skips it on OSX on travis.

### DIFF
--- a/bluesky/tests/test_run_engine.py
+++ b/bluesky/tests/test_run_engine.py
@@ -578,7 +578,7 @@ def test_sigint_three_hits(RE, hw):
 
     pid = os.getpid()
 
-    def sim_kill(n=1):
+    def sim_kill(n):
         for j in range(n):
             time.sleep(.02)
             os.kill(pid, signal.SIGINT)
@@ -588,8 +588,6 @@ def test_sigint_three_hits(RE, hw):
 
     def self_sig_int_plan():
         threading.Timer(.05, sim_kill, (3,)).start()
-        threading.Timer(.1, sim_kill, (3,)).start()
-        threading.Timer(.2, sim_kill, (3,)).start()
         yield from abs_set(motor, 1, wait=True)
 
     start_time = ttime.time()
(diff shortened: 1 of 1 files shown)


tacaswell:
sigh, trying again with the correct pytest path.

tacaswell:
:tada: finally passed!

bluesky/tests/test_run_engine.py
```diff
@@ -571,34 +571,41 @@ def simple_plan(motor):
...
     end_time = ttime.time()
-    assert end_time - start_time < 0.2  # not enough time for motor to cleanup
+    # not enough time for motor to cleanup, but long enough to start
+    assert 0.05 < end_time - start_time < 0.2
     RE.abort()  # now cleanup
+
     done_cleanup_time = ttime.time()
-    assert done_cleanup_time - end_time > 0.3
+    # this should be 0.5 (the motor.delay) above, leave sloppy for CI
+    assert 0.6 > done_cleanup_time - end_time > 0.3
```
klauer:
```suggestion
    assert 0.3 < done_cleanup_time - end_time < 0.6
```

Does this make anyone else's head hurt? :(

bluesky/tests/test_run_engine.py
```diff
@@ -571,34 +571,41 @@ def simple_plan(motor):
...
 
     lp = RE.loop
     motor.loop = lp
-    start = time.monotonic()
-    threading.Timer(.05, sim_kill, (3,)).start()
-    threading.Timer(.1, sim_kill, (3,)).start()
-    threading.Timer(.2, sim_kill, (3,)).start()
+
+    def self_sig_int_plan():
+        threading.Timer(.05, sim_kill, (3,)).start()
```
klauer: :+1: Makes more sense to do it like this - start the timer only after the plan starts. Could also do it with a `threading.Event` or so.

tacaswell: Yeah, all of the sigint tests could be cleaned up.

PR #1228, opened 2019-07-22
Change recommended RE.md to PersistentDict.

### LINKED ISSUE #1227: Switch from historydict to zict.
By default, `RE.md` is a plain dict. Many users want `RE.md` to persistent between IPython sessions. To support this, we developed `historydict`, a dict-like object that syncs itself to a sqlite database on disk. We use it at NSLS-II and we recommend it in the bluesky docs.

There is a bad interaction between `historydict` and the event loop refactor. The sqlite connections are created on the main thread but then used on the RunEngine's background thread.

We had [previously considered](https://github.com/bluesky/bluesky/issues/924) switching from `historydict` to [zict](https://zict.readthedocs.io/en/latest/) a very lightweight library of useful dict-like objects that is part of the dask ecosystem. It is more flexible than historydict, and we would be using a community thing instead of our own thing. Perhaps the best solution to this issue to use this as a reason to switch to zict.

I propose to:

1. Add ~zarr~ zict as a required dependency. (Historydict is currently a required dependency too.)
2. Add a `PersistentDict` object to `bluesky.utils`. It would take a filepath as its argument and create dict that sync to that file---probably as a JSON using the numpy-safe JSON encoder from event-model.
3. In the docs, provide a migration script copy content out of existing historydict files.
4. Update the docs to recommend `PersistentDict`.
5. Deprecate `get_history`.
6. Leave it up to the user/facility to decide where to store the persistent md, instead of replicating the config search path that we put in `get_history`.
7. In a later release, remove `get_history` and the `historydict` dependency.

danielballan:
Closes #1227. See that issue for motivation and design discussion.

This is a critical fix that should go into 1.6.0.

### DIFF
--- a/bluesky/tests/test_persistent_dict.py
+++ b/bluesky/tests/test_persistent_dict.py
@@ -1,0 +1,34 @@
+import collections.abc
+
+from bluesky.utils import PersistentDict
+import numpy
+from numpy.testing import assert_array_equal
+
+
+def test_persistent_dict(tmp_path):
+    d = PersistentDict(tmp_path)
+    d['a'] = 1
+    d['b'] = (1, 2)
+    d['c'] = numpy.zeros((5, 5))
+    d['d'] = {'a': 10, 'b': numpy.ones((5, 5))}
+    expected = dict(d)
+    actual = PersistentDict(tmp_path)
+    recursive_assert_equal(actual, expected)
+
+    # Update a value and check again.
+    d['a'] = 2
+    expected = dict(d)
+    recursive_assert_equal(actual, expected)
+
+    # Smoke test the accessor and the __repr__.
+    assert d.directory == tmp_path
+    d.__repr__()
+
+
+def recursive_assert_equal(actual, expected):
+    assert set(actual.keys()) == set(expected.keys())
+    for key in actual:
+        if isinstance(actual[key], collections.abc.MutableMapping):
+            recursive_assert_equal(actual[key], expected[key])
+        else:
+            assert_array_equal(actual[key], expected[key])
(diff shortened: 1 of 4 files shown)


prjemian:
Basically, replace sqlite3 persistence with PersistentDict.
Any gotchas?

I've no particular experience with PersistentDict and our beam lines are not tied to any specific implementation of persistence.   Where's the documentation for PersistentDict?

This should work like plumbing, hope this will not be user-visible.

danielballan:
> Is this OS-independent (looks like it is)?

Yes.

> Thought you said zict was not a good fit. Can you comment more about how msgpack overcomes this fit?

Yes, I was originally imagining a solution that let us use a single file. Upon further consideration, using a directory is a better choice because it can be faster and mitigates contention issues. Using msgpack for serialization overcomes my concerns with JSON (which has to convert numpy arrays to lists) and pickle (which raises security concerns when applied to long-term storage).

> Good find. I think that would also be a defensible choice; it is a well-documented and actively maintained project. I'm personally happier with a thin wrapper (`PersistentDict`) of a thin wrapper (`zict`) of a `MutableMapping`. It looks like `PersistentMapping` introduces additional abstractions (a cache and a "data manager") that are beyond the scope of our needs.

danielballan:
Thanks for the good questions, @prjemian!

bluesky/tests/test_persistent_dict.py
```diff
@@ -0,0 +1,34 @@
+import collections.abc
+
+from bluesky.utils import PersistentDict
+import numpy
+from numpy.testing import assert_array_equal
```
mrakitin: Would you mind to use the sorting approach for the imports we discuss before (such as native Python libs go first, then community packages, and then our development)?

mrakitin:
Thanks for adding the test with RE!

jklynch:
Hurray for numpy.testing.assert_array_equal! I have had trouble testing msgpack because it turns lists into tuples, or the other way, and this handles it.

danielballan:
Thanks, all!

PR #1229, opened 2019-07-22
Fix: BestEffortCallback - table is always printed

dmgav:
**Fix to BestEffortCallback function.** 

As a result of recent changes, Bluesky stopped printing the table during acquisition. For example, the following code (part of the basic Bluesky tutorial) would not print the table if run with python or ipython:

```
from bluesky import RunEngine
RE = RunEngine({})

bec = BestEffortCallback()
RE.subscribe(bec)

from databroker import Broker
db = Broker.named('temp')
RE.subscribe(db.insert)

from ophyd.sim import det1, det2

from bluesky.plans import count, scan
dets = [det1, det2]

RE(count(dets, num=5, delay=1))
```

After the proposed change, the table printing code is executed even if there is no useful data to print.

### DIFF
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -174,173 +174,176 @@ class BestEffortCallback(QtAwareCallback):
                 fixed_dim_fields.extend(fields)
             self.dim_fields = fixed_dim_fields
 
-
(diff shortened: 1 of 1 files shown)


awalter-bnl:
good catch @dmgav, when I first saw this issue I assumed it was intentional !-) .

PR #1230, opened 2019-07-23
Apply isort.

danielballan:
Please do not merge until further discussion in #1231. This is an illustration.


danielballan:
Verdict from group discussion is: maybe someday; there are trade-offs; not today!

PR #1232, opened 2019-07-23
Fix: segmentation fault while plotting from BestEffortCallback

dmgav:
** Fix for segmentation fault while plotting with BestEffortCallback in IPython **

If run in IPython, the following code would cause segmentation fault during execution of BestEffortCallback. It was found that the Qt backend is not identified by QtAwareCallback.__init__(). As a result, Teleporter object is not created and Qt is called directly from background thread. Converting Qt backend name ('Qt5Add') to lowercase ```use_teleporter = 'qt' in matplotlib.get_backend().lower()``` fixed the problem. 

```
from bluesky import RunEngine
RE = RunEngine({})

bec = BestEffortCallback()
RE.subscribe(bec)

from databroker import Broker
db = Broker.named('temp')
RE.subscribe(db.insert)

from ophyd.sim import det, det1, det2, motor

from bluesky.plans import count, scan
dets = [det1, det2]
det1.kind = "hinted"  # This will cause BestEffortCallback to plot data from 'det1'

RE(count(dets, num=5, delay=1))

RE(scan(dets, motor, -2, 2, 5))

from bluesky.callbacks.mpl_plotting import LivePlot
lp = LivePlot('det', 'motor')
RE.subscribe(lp)
RE(scan([det], motor, -5, 5, 30))
```

### DIFF
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -23,7 +23,7 @@ from .fitting import PeakStats
 
 class BestEffortCallback(QtAwareCallback):
     def __init__(self, *, fig_factory=None, table_enabled=True, **kwargs):
-        super().__init__(use_teleporter=kwargs.pop('use_teleporter', None))
+        super().__init__(**kwargs)
         # internal state
         self._start_doc = None
         self._descriptors = {}
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -23,7 +23,8 @@ from .fitting import PeakStats
 
 class BestEffortCallback(QtAwareCallback):
     def __init__(self, *, fig_factory=None, table_enabled=True, **kwargs):
-        super().__init__(**kwargs)
+        import matplotlib.pyplot as plt
+        super().__init__(use_teleporter=kwargs.pop('use_teleporter', None))
         # internal state
         self._start_doc = None
         self._descriptors = {}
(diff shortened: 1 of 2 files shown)


bluesky/callbacks/best_effort.py
```diff
@@ -23,7 +23,8 @@
 
 class BestEffortCallback(QtAwareCallback):
     def __init__(self, *, fig_factory=None, table_enabled=True, **kwargs):
-        super().__init__(**kwargs)
+        import matplotlib.pyplot as plt
```
danielballan: Is this import used?

dmgav: It is unused. Removed.

bluesky/callbacks/best_effort.py
```diff
@@ -23,7 +23,7 @@
 
 class BestEffortCallback(QtAwareCallback):
     def __init__(self, *, fig_factory=None, table_enabled=True, **kwargs):
-        super().__init__(**kwargs)
+        super().__init__(use_teleporter=kwargs.pop('use_teleporter', None))
```
tacaswell: I think this means we now accept kwargs but don't do anything with them (which can let typos silently pass).  Do we need to do this popping?

dmgav: Yes, I think you are right. I followed the pattern used in LivePlot, but BestEffortCallback treats the parameters differently.

danielballan: Indeed, the `kwargs` are not used anywhere inside the body of `BestEffortCallback`, which means we are just silently swallowing arbitrary parameters. Maybe while we're here we just remove `**kwargs` from the signature and go with:

```py
def __init__(self, *, fig_factory=None, table_enabled=True, use_teleporter=None):
    super().__init__(use_teleporter=use_teleporter)
```

Would that make sense?

danielballan: Ah, I just saw 7f943e7. That works too. Merging!

PR #1236, opened 2019-07-30
DOC: remove duplicate import in PersistentDict example

mrakitin:
Minor correction to the example from #1228.


mrakitin:
Thanks @awalter-bnl!

PR #1244, opened 2019-08-21
Provide better exceptions for down-stream consumers

### LINKED ISSUE #1240: More helpful error messages when plan_stubs are run without enough context
Diverting discussion from #1238. Can we make this error message more specific or prescriptive?

```python
In [3]: RE(trigger_and_read())
<snipped>
IllegalMessageSequence: Cannot bundle readings without an open run. That is, 'create' must be preceded by 'open_run'.
```

awalter-bnl:
This PR attempts to make the exception exposed to the users easier to understand. 

## Description
This adds 3 new exceptions (splitting the existing 'IllegalMessageSequence' exception into : IllegalRunSequence, IllegalBundlingSequence and IllegalMonitorSequence. It also adds a new attribute ``RunEngine._exception_attrs`` (a dict mapping required exception attribute names to RunEngine attributes to extract the values from) and a new method ``RunEngine._add_attrs_to_exception(exception, msg, obj=None)`` that adds the attributes defined in the above dictionary and ``msg`` and ``obj`` attributes to the exception before returning it.

Finally this PR updates all of the existing instances of ``IllegalMessageSequence`` to use the above changes.

## Motivation and Context
A discussion of the motivation can be found in PR #1240 

## How Has This Been Tested?
It is yet to be tested, and is submitted as a Draft to gain feedback on the approach taken before completing it and adding tests.

closes #1240

### DIFF
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -16,7 +16,7 @@ except ImportError:
 
 
 from .utils import (separate_devices, all_safe_rewind, Msg, ensure_generator,
-                    short_uid as _short_uid)
+                    short_uid as _short_uid, IllegalMessageSequence)
 
 
 def create(name='primary'):
@@ -773,6 +773,7 @@ def trigger_and_read(devices, name='primary'):
     msg : Msg
         messages to 'trigger', 'wait' and 'read'
     """
+
     # If devices is empty, don't emit 'create'/'save' messages.
     if not devices:
         yield from null()
@@ -798,8 +799,13 @@ def trigger_and_read(devices, name='primary'):
         yield from save()
         return ret
     from .preprocessors import rewindable_wrapper
-    return (yield from rewindable_wrapper(inner_trigger_and_read(),
-                                          rewindable))
+    try:
+        return (yield from rewindable_wrapper(inner_trigger_and_read(),
+                                              rewindable))
+    except IllegalMessageSequence as excep:
+        excep_str = ('While performing a "trigger_and_read" the following '
+                     f'message sequence issue occurred: {excep!r}')
+        raise type(excep)(excep.msg, excep_str) from excep
 
 
 def broadcast_msg(command, objs, *args, **kwargs):
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -804,7 +804,7 @@ def trigger_and_read(devices, name='primary'):
                                               rewindable))
     except IllegalMessageSequence as excep:
         excep_str = ('While performing a "trigger_and_read" the following '
-                     f'message sequence issue occurred: {excep!r}')
+                     f'message sequence issue occurred:        {excep!r}')
         raise type(excep)(excep.msg, excep_str) from excep
 
 
@@ -910,8 +910,13 @@ def one_1d_step(detectors, motor, step):
         yield Msg('set', motor, step, group=grp)
         yield Msg('wait', None, group=grp)
 
-    yield from move()
-    return (yield from trigger_and_read(list(detectors) + [motor]))
+    try:
+        yield from move()
+        return (yield from trigger_and_read(list(detectors) + [motor]))
+    except IllegalMessageSequence as excep:
+        excep_str = ('While performing a "one_1d_step" the following message '
+                     f'sequence issue occured:        {excep!r}')
+        raise type(excep)(excep.msg, excep_str) from excep
 
 
 def move_per_step(step, pos_cache):
@@ -954,8 +959,13 @@ def one_nd_step(detectors, step, pos_cache):
         mapping motors to their last-set positions
     """
     motors = step.keys()
-    yield from move_per_step(step, pos_cache)
-    yield from trigger_and_read(list(detectors) + list(motors))
+    try:
+        yield from move_per_step(step, pos_cache)
+        yield from trigger_and_read(list(detectors) + list(motors))
+    except IllegalMessageSequence as excep:
+        excep_str = ('While performing a "one_nd_step" the following message '
+                     f'sequence issue occured:        {excep!r}')
+        raise type(excep)(excep.msg, excep_str) from excep
 
 
 def repeat(plan, num=1, delay=None):
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1651,7 +1651,7 @@ class RunEngine:
             current_run = self._run_bundlers[run_key]
         except KeyError as ke:
             raise IllegalRunSequence(msg, "Cannot bundle readings without an "
-                                     "open run. That is, 'create'must be "
+                                     "open run. That is, 'create' must be "
                                      "preceded by 'open_run'.") from ke
 
         return (await current_run.create(msg))
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -30,13 +30,30 @@ except ImportError:
     current_task = Task.current_task
     del Task
 
-from .utils import (CallbackRegistry, SigintHandler,
-                    normalize_subs_input, AsyncInput,
-                    NoReplayAllowed, RequestAbort, RequestStop,
-                    RunEngineInterrupted, IllegalMessageSequence,
-                    FailedPause, FailedStatus, InvalidCommand,
-                    PlanHalt, Msg, ensure_generator, single_gen,
-                    default_during_task)
+from .utils import (
+    AsyncInput,
+    CallbackRegistry,
+    FailedPause,
+    FailedStatus,
+    IllegalBundlingSequence,
+    IllegalMessageSequence,
+    IllegalMonitorSequence,
+    IllegalRunSequence,
+    InvalidCommand,
+    Msg,
+    NoReplayAllowed,
+    PlanHalt,
+    RequestAbort,
+    RequestStop,
+    RunEngineInterrupted,
+    SigintHandler,
+    default_during_task,
+    ensure_generator,
+    new_uid,
+    normalize_subs_input,
+    short_uid,
+    single_gen,
+)
 
 
 class _RunEnginePanic(Exception):
@@ -1559,9 +1576,9 @@ class RunEngine:
         # TODO extract this from the Msg
         run_key = _extract_run_key(msg)
         if run_key in self._run_bundlers:
-            raise IllegalMessageSequence("A 'close_run' message was not "
-                                         "received before the 'open_run' "
-                                         "message")
+            raise IllegalRunSequence(msg, "A 'close_run' message was not "
+                                     "received before the 'open_run' "
+                                     "message")
 
         # Run scan_id calculation method
         self.md['scan_id'] = self.scan_id_source(self.md)
@@ -1604,9 +1621,12 @@ class RunEngine:
         try:
             current_run = self._run_bundlers[run_key]
         except KeyError as ke:
-            raise IllegalMessageSequence("A 'close_run' message was not "
-                                         "received before the 'open_run' "
-                                         "message") from ke
+            raise IllegalRunSequence(msg, "A 'close_run' message was received "
+                                     "but there is no run open. If this "
+                                     "occurred after a pause/resume, add a "
+                                     "'checkpoint' message after the "
+                                     "'close_run' message.") from ke
+
         ret = (await current_run.close_run(msg))
         del self._run_bundlers[run_key]
         return ret
@@ -1630,9 +1650,10 @@ class RunEngine:
         try:
             current_run = self._run_bundlers[run_key]
         except KeyError as ke:
-            raise IllegalMessageSequence("Cannot bundle readings without "
-                                         "an open run. That is, 'create' must "
-                                         "be preceded by 'open_run'.") from ke
+            raise IllegalRunSequence(msg, "Cannot bundle readings without an "
+                                     "open run. That is, 'create'must be "
+                                     "preceded by 'open_run'.") from ke
+
         return (await current_run.create(msg))
 
     async def _read(self, msg):
@@ -1683,8 +1704,9 @@ class RunEngine:
         try:
             current_run = self._run_bundlers[run_key]
         except KeyError as ke:
-            raise IllegalMessageSequence("A 'monitor' message was sent but no "
-                                         "run is open.") from ke
+            raise IllegalRunSequence(msg, "A 'monitor' message was sent but "
+                                     "no run is open.") from ke
+
         await current_run.monitor(msg)
         await self._reset_checkpoint_state_coro()
 
@@ -1700,7 +1722,7 @@ class RunEngine:
         try:
             current_run = self._run_bundlers[run_key]
         except KeyError as ke:
-            raise IllegalMessageSequence(
+            raise IllegalRunSequence(msg,
                 "A 'unmonitor' message was sent but no "
                 "run is open.") from ke
         await current_run.unmonitor(msg)
@@ -1719,9 +1741,8 @@ class RunEngine:
         except KeyError as ke:
             # sanity check -- this should be caught by 'create' which makes
             # this code path impossible
-            raise IllegalMessageSequence(
-                "A 'save' message was sent but no " "run is open."
-            ) from ke
+            raise IllegalRunSequence(msg, "A 'save' message was sent but no "
+                                     "run is open.") from ke
 
         await current_run.save(msg)
 
@@ -1736,9 +1757,9 @@ class RunEngine:
         try:
             current_run = self._run_bundlers[run_key]
         except KeyError as ke:
-            raise IllegalMessageSequence(
-                "A 'drop' message was sent but no " "run is open."
-            ) from ke
+            raise IllegalRunSequence(msg, "A 'drop' message was sent but no "
+                                     "run is open.") from ke
+
         await current_run.drop(msg)
 
     async def _kickoff(self, msg):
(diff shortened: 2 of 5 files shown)


awalter-bnl:
I am pausing at this point to allow for feedback on what already exists. As the next step is to o through each of the 'combined plan_stubs' like ``trigger_and_read`` I would like agreement on this part first prior to making changes there to avoid redoing work later.

bluesky/run_engine.py
```diff
@@ -1602,9 +1615,11 @@ def fut_fac() -> awaitable:
         the RunStart document
         """
         if self._run_is_open:
-            raise IllegalMessageSequence("A 'close_run' message was not "
-                                         "received before the 'open_run' "
-                                         "message")
+            exception = IllegalRunSequence(
```
danielballan: It is normal for user-defined exceptions to take custom arguments in their `__init__`. See [documented examples](https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions). I didn't know that when I wrote the code in event-model that I presented as a model for this approach, and so I chose to monkey-patch instead, but setting things up at `__init__` time is always better if it's possible.

So this could be:

```py
IllegalRunSequence(
    "A 'close_run' message was not received before the 'open_run' "
    "message", msg, self._currently_open_runs)
```

which to my reading seems more self-contained / easy to follow. It removes `self._exception_attrs` and `_add_attrs_to_exception` entirely, and along with them the need to cross-reference anything to follow the action of the code. Each exception type can take in only those things which are useful/relevant. That is, some may want `self._currently_open_runs`; some way want `self._state`; and so on.

danielballan: I just opened a PR to fix event-model to follow this approach as well. The relevant commit is https://github.com/bluesky/event-model/pull/104/commits/bb9618ec639687bcf8e38587f5e40246972f3176.

awalter-bnl: good call @danielballan, this is a much neater solution. I will make this change this afternoon.

awalter-bnl: fixed

awalter-bnl:
@danielballan is this looking better?

danielballan:
Yep, the strucutre looks good to me. New thought: I wonder if `IllegalMessageSequence` (and its subclasses) should _always_ receive the `Msg` object and have `exception.msg` as an attribute. It seems generically useful for the exception should provide a reference the message that triggered the problem. And, with in `msg`, you'd get `msg.command` (e.g. `'create'`, `'monitor'`, ...) and `msg.obj` for free.

awalter-bnl:
@danielballan That suggestion was so obvious I feel stupid for not thinking of it myself. I have made the change.

danielballan:
Haha, well it didn't jump out at me in the first review either. :- )

I think is probably ready for the next phase, unless others have structural comments about this aspect.

I'm still feeling unsure about the names of the exception classes, but I can't think of anything obviously better, and we can always address that with Find/Replace late in the process if need be.

awalter-bnl:
FYI the failed tests is in python 3.6 for linux build and is an unrelated timeout error in ``tests/test_callbacks.py/test_broker_base``. This is unrelated to the current work, and is marked as ``xfail`` but is still causing this set of tests to fail.

awalter-bnl:
So I was about to start adding the custom messages to the ``plan_stubs`` that yield from other ``plan_stubs``, but happened upon a slightly different approach to what was previously suggested.

Instead of providing custom messages for each of the new exception classes What do people think of the following:

```
import bluesky.plan_stubs 
import bluesky 
from bluesky import RunEngine 
from bluesky.utils import IllegalMessageSequence 
from ophyd.sim import det 
RE = RunEngine() 
  
def improved_trigger_and_read(detectors): 
    try: 
        yield from bluesky.plan_stubs.trigger_and_read(detectors) 
    except IllegalMessageSequence as excep: `
         excep.args = (excep.args if excep.args else tuple()) + ("This error occured inside a 'trigger_and_read' plan_stub",) 
        raise  
```

which results in the traceback message being:
```IllegalRunSequence: ("Cannot bundle readings without an open run. That is, 'create' must be preceded by 'open_run'.", "This error occured inside a 'trigger_and_read' plan_stub")```

I think this has a few main benefits:
1. It requires just one 'except'  section for all the new error types (by excepting the common parent ``IllegalMessageSequence``), but still provides the extra information that this occurred inside another plan_stub, not at the message plan_stub.
1. It in principle allows for nesting of the traceback message, so at a glance we could see for instance that the error occurred inside a ``scan`` plan, where it yielded from a ``trigger_and_read``, which yielded from a ``create`` which generated the error.
1. It still allows for the more detailed error messages on a case by case basis.

Is this a solution worth pursuing?

**Comment by danielballan** [comment]
This looks to me like "exception chaining," which Python now has special syntax for.

```py
def improved_trigger_and_read(detectors): 
    try: 
        yield from bluesky.plan_stubs.trigger_and_read(detectors) 
    except IllegalMessageSequence as exc:
        raise PlanYieldedIllegalMessages("The `trigger_and_read` plan stub caused this error.") from exc
```

Now `raise ... from ...`, which will cause IPython to show both exceptions and explain, "This exception was the direct cause of...."

As you say, this does not prevent us from doing more specific things like:

```py
def improved_trigger_and_read(detectors): 
    try: 
        yield from bluesky.plan_stubs.trigger_and_read(detectors) 
    except IllegalRunSequence:
        raise PlanYieldedIllegalMessages("It looks like you used `trigger_and_read` without first opening a Run.")
```

later and/or in specific cases.

<hr />

I had in mind that `improved_trigger_and_read` was a shorthand construct for the purposes of conversation. From #1240:

> To refine this, I would make the try/except part of trigger_and_read itself rather than wrapping it as I did above.

Are we on the same page about that? I don't think we needs or want to add another layer in the already-deep stack over this.

awalter-bnl:
@danielballan you are correct, sort off, the benefit of my solution is one exception, chained messages where as you have chained exceptions which do not print to the command line the same way and hence are not as intuitive when looking at the trace back IMO.


Yes we are in the same page regarding implementing the try-except inside trigger_And_read. Like you I was just trying to quickly check how the exception looked from the command line

ZLLentz:
It's likely that the run ids, the exception type, and the offending message will be enough to cover most use cases here. I'm struggling to imagine a use for the run engine state, but maybe I'm just missing something. We were imagining bundling a bunch of information onto these, but I suspect it may become more useful to subclass the exceptions further so you can quickly pick out the relevant failure mode in an except block. There's probably some sort of balance to strike here, but it won't be obvious what needs to be added until the plan-specific error messages are explored.

I think this PR is good as-is, the only change I would suggest is removing the run engine state unless someone can propose a use for it, but I don't feel strongly about that. More info can be added to the exceptions or more subclasses can be made on an as-needed basis.

tacaswell:
Looking at #1248 some of these are going to have to get raised in objects that we may not want to have access to the RE state.  On a bit more thought, I would lean towards including only the offending msg and not the current state of the RE (at least until we settle how we want to modularize it).

danielballan:
> Can you clarify if you want to just remove the state attribute from exceptions or if you want the plan error messages added to this PR prior to merging.

I would personally be satisfied with either:
(1) The `IllegalMessageSequence` (and subclasses...) only get the `msg`, and with that change with merge this, tackling the changes to `bluesky.plans` in future PRs.
(2) We do `bluesky.plans` changes as part of this PR and see if we come up with a reason to need private state from the RE in the messages.

Given the subsequent conversation and also @tacaswell's desire to get this merged soon-ish, I think (1) is the better option. We can always add more things later.

awalter-bnl:
So I have stripped everything but the message from the IllegalMessageSequence attributes, and updated the tests and message functions to match.

I have also added the try-except around `trigger_and_read`, if there is consensus on this being the correct approach I will add the try-except to the rest of the plan-stubs that require it and add some tests for the plan_stubs errors.

For reference the traceback message for `trigger_and_read` will now look like:

```
In [1]: import bluesky.plan_stubs 
   ...: import bluesky 
   ...: from bluesky import RunEngine 
   ...: from bluesky.utils import IllegalMessageSequence 
   ...: from ophyd.sim import det 
   ...: RE = RunEngine() 
   ...:  
   ...: RE(bluesky.plan_stubs.trigger_and_read([det]))                                                              
[E 08:28:20.524 run_engine:1524] Run aborted
      File "/home/alwalter/NSLS-II/bluesky/bluesky/plan_stubs.py", line 804, in trigger_and_read
        rewindable))
      File "/home/alwalter/NSLS-II/bluesky/bluesky/preprocessors.py", line 681, in rewindable_wrapper
        return (yield from plan)
      File "/home/alwalter/NSLS-II/bluesky/bluesky/plan_stubs.py", line 794, in inner_trigger_and_read
        yield from create(name)
      File "/home/alwalter/NSLS-II/bluesky/bluesky/plan_stubs.py", line 41, in create
        return (yield Msg('create', name=name))
      File "/home/alwalter/NSLS-II/bluesky/bluesky/run_engine.py", line 1450, in _run
        new_response = await coro(msg)
      File "/home/alwalter/NSLS-II/bluesky/bluesky/run_engine.py", line 1712, in _create
        raise IllegalRunSequence(msg, "Cannot bundle readings without an "
    bluesky.utils.IllegalRunSequence: Cannot bundle readings without an open run. That is, 'create'must be preceded by 'open_run'.
    
    The above exception was the direct cause of the following exception:
    
      File "/home/alwalter/NSLS-II/bluesky/bluesky/run_engine.py", line 1369, in _run
        stashed_exception or resp)
      File "/home/alwalter/NSLS-II/bluesky/bluesky/plan_stubs.py", line 809, in trigger_and_read
        raise IllegalMessageSequence(excep.msg, excep_str) from excep
    bluesky.utils.IllegalMessageSequence: While performing a "trigger_and_read" the following message sequence issue occured: Cannot bundle readings without an open run. That is, 'create'must be preceded by 'open_run'.
IllegalRunSequence                        Traceback (most recent call last)
~/NSLS-II/bluesky/bluesky/plan_stubs.py in trigger_and_read(devices, name)
    803         return (yield from rewindable_wrapper(inner_trigger_and_read(),
--> 804                                               rewindable))
    805 

~/NSLS-II/bluesky/bluesky/preprocessors.py in rewindable_wrapper(plan, rewindable)
    680     else:
--> 681         return (yield from plan)
    682 

~/NSLS-II/bluesky/bluesky/plan_stubs.py in inner_trigger_and_read()
    793                 yield from wait(group=grp)
--> 794             yield from create(name)
    795             ret = {}  # collect and return readings to give plan access to them

~/NSLS-II/bluesky/bluesky/plan_stubs.py in create(name)
     40     """
---> 41     return (yield Msg('create', name=name))
     42 

~/NSLS-II/bluesky/bluesky/run_engine.py in _run(self)
   1449                         # raised
-> 1450                         new_response = await coro(msg)
   1451 

~/NSLS-II/bluesky/bluesky/run_engine.py in _create(self, msg)
   1711         if not self._run_is_open:
-> 1712             raise IllegalRunSequence(msg, "Cannot bundle readings without an "
   1713                                      "open run. That is, 'create'must be "

IllegalRunSequence: Cannot bundle readings without an open run. That is, 'create'must be preceded by 'open_run'.

The above exception was the direct cause of the following exception:

IllegalMessageSequence                    Traceback (most recent call last)
<ipython-input-1-6e0cf610d738> in <module>
      6 RE = RunEngine()
      7 
----> 8 RE(bluesky.plan_stubs.trigger_and_read([det]))

~/NSLS-II/bluesky/bluesky/run_engine.py in __call__(self, *args, **metadata_kw)
    834             self._task_fut.add_done_callback(set_blocking_event)
    835 
--> 836         self._resume_task(init_func=_build_task)
    837 
    838         if self._interrupted:

~/NSLS-II/bluesky/bluesky/run_engine.py in _resume_task(self, init_func)
    963                     if (exc is not None
    964                             and not isinstance(exc, _RunEnginePanic)):
--> 965                         raise exc
    966 
    967     def install_suspender(self, suspender):

~/NSLS-II/bluesky/bluesky/run_engine.py in _run(self)
   1523             self._reason = str(err)
   1524             self.log.exception("Run aborted")
-> 1525             raise err
   1526         finally:
   1527             # Some done_callbacks may still be alive in other threads.

~/NSLS-II/bluesky/bluesky/run_engine.py in _run(self)
   1367                         try:
   1368                             msg = self._plan_stack[-1].throw(
-> 1369                                 stashed_exception or resp)
   1370                         except Exception as e:
   1371                             # The current plan did not handle it,

~/NSLS-II/bluesky/bluesky/plan_stubs.py in trigger_and_read(devices, name)
    807         excep_str = ('While performing a "trigger_and_read" the following '
    808                      'message sequence issue occured: ' + str(excep))
--> 809         raise IllegalMessageSequence(excep.msg, excep_str) from excep
    810 
    811 

IllegalMessageSequence: While performing a "trigger_and_read" the following message sequence issue occured: Cannot bundle readings without an open run. That is, 'create'must be preceded by 'open_run'.
```

bluesky/plan_stubs.py
```diff
@@ -773,33 +773,40 @@ def trigger_and_read(devices, name='primary'):
...
+            yield from create(name)
+            ret = {}  # collect and return readings to give plan access to them
+            for obj in devices:
+                reading = (yield from read(obj))
+                if reading is not None:
+                    ret.update(reading)
+            yield from save()
+            return ret
+        from .preprocessors import rewindable_wrapper
+        return (yield from rewindable_wrapper(inner_trigger_and_read(),
```
danielballan: I think the `try:` block only needs to wrap this line.

bluesky/plan_stubs.py
```diff
@@ -773,33 +773,40 @@ def trigger_and_read(devices, name='primary'):
...
+                                              rewindable))
+
+    except IllegalMessageSequence as excep:
+        excep_str = ('While performing a "trigger_and_read" the following '
+                     'message sequence issue occured: ' + str(excep))
```
danielballan: Spelling fix, and also an exception's `repr` includes the type while the `str` just includes the message.

```suggestion
                     f'message sequence issue occurred: {excep!r}')
```

bluesky/plan_stubs.py
```diff
@@ -773,33 +773,40 @@ def trigger_and_read(devices, name='primary'):
...
+        raise IllegalMessageSequence(excep.msg, excep_str) from excep
```
danielballan: Do you want to raise a generic `IllegalMessageSequence` here or raise a `type(excep)`? We can be sure that `issubclass(type(excep), IllegalMessageSequence)`, but `type(excep)` may be a more specific subclass.

awalter-bnl:
Made the requested changes from @dallan, The new error message is:

```
IllegalRunSequence: While performing a "trigger_and_read" the following message sequence issue occurred: IllegalRunSequence("Cannot bundle readings without an open run. That is, 'create' must be preceded by 'open_run'.")
```

awalter-bnl:
@tacaswell, @dallan, @ZLLentz I think this is ready for another review, everything expected to be part of this PR is done. Below is an example of the exception message (minus the traceback) produced when running `RE(one_1d_step(...))`. I think this is a good outcome.

FYI I think there should be strong consideration to including this in all of our built-in plans (perhaps by a plan decorator?)  As this will then produce a very instructive message that is easy for the average user to at least work out _where_ the issue occurred. But that should probably be part of another discussion/issue/PR/


```
IllegalRunSequence: While performing a "one_1d_step" the following message sequence issue occured:
        IllegalRunSequence('While performing a "trigger_and_read" the following message sequence issue occurred:
        IllegalRunSequence("Cannot bundle readings without an open run. That is, \'create\' must be preceded by \'open_run\'.")')
```

gwbischof:
Looks good to me, I don't see any problems

awalter-bnl:
thanks for the review @mrakitin, I am waiting on feedback from the rest of the group regarding the 'assert(...)/ assert ...' comment.

tacaswell:
`assert` is not a function so if you write `assert (False, msg)` that is interpreted as `assert tuple((False, msg))` which is always True meaning we will never fail the assert.

awalter-bnl:
@tacaswell it is true that 

> assert(false, msg) 

Does not ever return false, however

> assert(false)

Does fail the assert, which is how it is used here. So as it is written it will fail the test when something doesn’t work. That being said if stylistically

> assert false

Is preferred I am happy to make the change, is this the case?

ZLLentz:
I'm not sure about style, but practically you want to reduce the likelihood of someone coming later and turning `assert(False)` into `assert(False, msg)` and thinking they fixed a bug.

danielballan:
Right... the use of the phrase "used as a function" and the phrase "does not ever return false" in comments above makes me think there is still confusion here. The `assert` is always a statement, not an expression. It doesn't return things; and it's not being called. The addition of the parens around `False` is doing the same thing as:

````
>>> 1 + (1)
2
```

The parens are superfluous and have no effect. You could maybe make a case for

```py
assert (some_long_expression_that_might_span_lines)
```

but `assert(False)` with no space suggests a function call and is likely to be confusing, as @ZLLentz suggests. For the simple expression under review here, let's please omit the parens.

awalter-bnl:
@danielballan there is no confusion on my end, it is clear that assert is not a function (I never sggested that it was). While my reference to 'returning false' was a little clumsy it was not meant to imply I thought it was a function.

Anyway, while both options are programmatically equivalent the consensus appears to be that the parens are potentially confusing so I have removed them.

tacaswell:
@awalter-bnl can you rebase this?

awalter-bnl:
thanks for doing the rebase @tacaswell I was hoping to get to it, but my other priorities haven't yet given me the time.

tacaswell:
fully rebased on master.

One slight change in the logic is that if you try to unmonitor without an open stream the `IllegalRunSequence` takes prescedence over the `IllegalMonitorSequence`.  Both technically apply but I think the run sequence is the bigger problem.

tacaswell:
@danielballan @ZLLentz I dismissed both of your reviews, please re-review.

PR #1245, opened 2019-08-21
MNT: Remove some warnings

tacaswell:
There are still some warnings in our test suite but this makes some
progress on clearing them.


tacaswell:
force-pushed the former last commit (trying to fix zmq) out of existence.  Those tests failed locally, but on me weird set up so was hoping it was just me.

PR #1246, opened 2019-08-21
ENH: add hooks to support a baton IOC

tacaswell:
There will be a partner PR into nslsii with the IOC + ophyd object


tacaswell:
If the name is confusing it can be renamed.  As @prjemian as asked/noted else where the origin is the analogy of a relay-race baton from track or the conductors baton from an orchestra (the RE that holds it is in charge!).

tacaswell:
The suspender path that @prjemian has in #1247 may be a better path as it achieves the same effect (maybe a bit better as it will suspend a plan if the baton is stolen) and requires no changes to the RE.

danielballan:
I agree.

PR #1248, opened 2019-08-24
Enh multiple runs

tacaswell:
This is the first pass at seeing how hard this will be (plus the commits from #1245 ).

Have all but 8 tests passing locally, suspect the remaining few are just a few details away.

Have not started to implement anyway for the plan to ask for a second open run and need to list the messages that are and are not routed to the `RunBundler`.


bluesky/bundlers.py
```diff
@@ -0,0 +1,579 @@
...
+                self._config_desc_cache[obj] = obj.describe_configuration()
+                self._cache_config(obj)
+
+            # check that current read collides with nothing else in
+            # current event
+            cur_keys = set(self._describe_cache[obj].keys())
+            for read_obj in self._objs_read:
+                # that is, field names
+                known_keys = self._describe_cache[read_obj].keys()
+                if set(known_keys) & cur_keys:
```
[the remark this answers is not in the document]
danielballan: Might as well add that while we are here.

bluesky/utils.py
```diff
@@ -50,11 +50,11 @@ class RunEngineControlException(Exception):
 
 
 class RequestAbort(RunEngineControlException):
-    pass
+    exit_status = 'abort'
```
danielballan: :+1: Slick.

bluesky/run_engine.py
```diff
@@ -703,28 +673,11 @@ def request_pause(self, defer=False):
         self._deferred_pause_requested = False
         self._interrupted = True
         self._state = 'pausing'
-
-        self._record_interruption('pause')
+        for run in self._run_bundlers.values():
+            run._record_interruption('pause')
```
danielballan: Should we make all the API on the Bundler than needs to be called by the RunEngine public?

tacaswell: I could go either way on this, on one hand that seems like the right Python thing to do, on the other hand, Bundler it self is a very internal thing and I don't think we want users to be actively poking at them (which is what public methods suggest you can do!).

I think the RE touching private methods on Bundler is ok based on an analogy to `friend` classes in c++ .

tacaswell:
To first order, this seems to work!

```
In [44]: from bluesky import preprocessors as bpp
    ...: from bluesky import plans as bp
    ...: from bluesky import plan_stubs as bps
    ...: from bluesky.preprocessors import define_run_wrapper as drw
    ...: from ophyd.sim import motor, det
    ...: 
    ...: 
    ...: def interlaced_plan():
    ...:     run_ids = list('abc')
    ...:     for rid in run_ids:
    ...:         yield from drw(bps.open_run(md={rid: rid}), run_id=rid)
    ...: 
    ...:     for j in range(5):
    ...:         for i, rid in enumerate(run_ids):
    ...:             yield from bps.mov(motor, j + 0.1 * i) 
    ...:             yield from drw(bps.trigger_and_read([motor, det]), run_id=rid)
    ...:         yield from bps.sleep(1)
    ...: 
    ...:     for rid in run_ids:
    ...:         yield from drw(bps.close_run(), run_id=rid)

In [45]: RE(interlaced_plan())
Out[45]: 
('4a960b8a-a2c8-40a5-a696-1c6a32c1ad50',
 '955ad826-7013-404a-99d6-3e4ce9a3cb46',
 'f90065b3-9db7-4322-aa58-9a0a7380d2da')

In [46]: db[-1].table()
Out[46]: 
                                 time  motor  motor_setpoint       det
seq_num                                                               
1       2019-09-04 09:10:26.830444098    0.2             0.2  0.980199
2       2019-09-04 09:10:27.844032526    1.2             1.2  0.486752
3       2019-09-04 09:10:28.853575468    2.2             2.2  0.088922
4       2019-09-04 09:10:29.866254568    3.2             3.2  0.005976
5       2019-09-04 09:10:30.876429558    4.2             4.2  0.000148

In [47]: db[-2].table()
Out[47]: 
                                 time  motor  motor_setpoint       det
seq_num                                                               
1       2019-09-04 09:10:26.825020075    0.1             0.1  0.995012
2       2019-09-04 09:10:27.840028763    1.1             1.1  0.546074
3       2019-09-04 09:10:28.852145195    2.1             2.1  0.110251
4       2019-09-04 09:10:29.862403154    3.1             3.1  0.008189
5       2019-09-04 09:10:30.874826193    4.1             4.1  0.000224

In [48]: db[-3].table()
Out[48]: 
                                 time  motor  motor_setpoint       det
seq_num                                                               
1       2019-09-04 09:10:26.819765568    0.0             0.0  1.000000
2       2019-09-04 09:10:27.835811615    1.0             1.0  0.606531
3       2019-09-04 09:10:28.849414349    2.0             2.0  0.135335
4       2019-09-04 09:10:29.858419657    3.0             3.0  0.011109
5       2019-09-04 09:10:30.872261286    4.0             4.0  0.000335

```

Still need an API to thread the run_id through at at least the stubs and maybe all of the built in plans?  At a minimum the built in plans should generate a uuid (or object?) for themselves.

The consequences of what this will to to BEC need to be considered.

tacaswell:
## Affected messages
 - open_run
 - close_run
 - create
 - read
 - monitor
 - unmonitor
 - save
 - drop
 - kickoff
 - complete
 - collect
 - configure

## Unaffected messages:
- install_suspender
- rewindable
- resume
- pause
- remove_suspender
- unsubscribe
- checkpoint
- set
- subscribe
- null
- stage
- trigger
- input
- sleep
- clear_checkpoint
- wait
- wait_for
- unstage
- stop

bluesky/bundlers.py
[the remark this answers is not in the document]
tacaswell: done, see code below.

bluesky/run_engine.py
tacaswell: These were made public.

Issue #1184, opened 2019-05-15
make sure BEC never kills a plan due to raising an exception


tacaswell:

CJ-Wright:
Should we advocate for putting non-breaking callbacks on a proxy?

tacaswell:
@CJ-Wright that is also a solution, but going from 0 to running proxy is not trivial on the floor.

danielballan:
Should we add an `ignore_exceptions` kwarg to `RunEngine.subscribe`? We have an instance-global switch `RunEngine.ignore_callback_exceptions` which is too broad (it would ignore `db.insert` failure too). Maybe the right place for that switch is on a per-subscription basis.

One reason _not_ to do this is that ignoring exceptions isn't always enough. In the case of pyolog, we have to also process the callbacks on a Thread because the pyolog sometimes fails by being slow.

danielballan:
Closed by #1249

PR #1249, opened 2019-08-30
Paranoid decorator

tacaswell:
This adds two decorators (one function level, one class level) to use to make callbacks "safe" and wrap all of our built-in callbacks with them.

We have had too many cases of bugs in callbacks crashing runs.  This should prevent that at the cost of debugging _why_ plots or tables are not working.

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -3,6 +3,8 @@ Useful callbacks for the Run Engine
 """
 from itertools import count
 import warnings
+import os
+
 from collections import deque, namedtuple, OrderedDict
 import time as ttime
 from functools import wraps as _wraps, partial as _partial
@@ -26,6 +28,8 @@ def make_callback_safe(func=None, *, logger=None):
         except Exception as ex:
             if logger is not None:
                 logger.exception(ex)
+            if os.environ.get('BLUESKY_DEBUG_CALLBACKS', False):
+                raise ex
 
     return inner
 
@@ -350,6 +354,8 @@ class LiveTable(CallbackBase):
             self._print('{{k:*^{self._min_width}}}'
                         .format(self=self)
                         .format(k=' failed to format row '))
+            if os.environ.get('BLUESKY_DEBUG_CALLBACKS', False):
+                raise ex
         super().event(doc)
 
     def stop(self, doc):
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -10,6 +10,34 @@ from functools import wraps as _wraps, partial as _partial
 from datetime import datetime
 import logging
 from ..utils import ensure_uid
+def make_callback_safe(func=None, *, logger=None):
+    if func is None:
+        return _partial(make_callback_safe, logger=logger)
+
+    @_wraps(func)
+    def inner(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as ex:
+            if logger is not None:
+                logger.exception(ex)
+
+    return inner
+
+
+def make_class_safe(cls=None, *, to_wrap=None, logger=None):
+    from event_model import DocumentNames
+
+    if cls is None:
+        return _partial(make_class_safe, to_wrap=to_wrap, logger=logger)
+
+    if to_wrap is None:
+        to_wrap = ['__call__']
+
+    for f_name in to_wrap:
+        setattr(cls, f_name, make_callback_safe(getattr(cls, f_name), logger=logger))
+
+    return cls
 
 
 class CallbackBase:
@@ -352,33 +380,3 @@ class LiveTable(CallbackBase):
     def _print(self, out_str):
         self._rows.append(out_str)
         self._out(out_str)
-
-
-def make_callback_safe(func=None, *, logger=None):
-    if func is None:
-        return _partial(make_callback_safe, logger=logger)
-
-    @_wraps(func)
-    def inner(*args, **kwargs):
-        try:
-            return func(*args, **kwargs)
-        except Exception as ex:
-            if logger is not None:
-                logger.exception(ex)
-
-    return inner
-
-
-def make_class_safe(cls=None, *, to_wrap=None, logger=None):
-    from event_model import DocumentNames
-
-    if cls is None:
-        return _partial(make_class_safe, to_wrap=to_wrap, logger=logger)
-
-    if to_wrap is None:
-        to_wrap = ['__call__']
-
-    for f_name in to_wrap:
-        setattr(cls, f_name, make_callback_safe(getattr(cls, f_name), logger=logger))
-
-    return cls
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -10,6 +10,11 @@ from functools import wraps as _wraps, partial as _partial
 from datetime import datetime
 import logging
 from ..utils import ensure_uid
+import logging
+
+logger = logging.getLogger(__name__)
+
+
 def make_callback_safe(func=None, *, logger=None):
     if func is None:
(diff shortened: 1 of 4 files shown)


danielballan:
Remembering that we've waffled a couple times on whether callback failures to usually kill the scan or not, I think it's worth careful consideration before merging.

The first approach we took was a global switch:

```py
RE.ignore_callback_exceptions = True
```

which I think we agree is too blunt. If data persistence fails (e.g. `db.insert`) we _want_ execution to stop so that the user is not misled into assuming that their data is successfully being saved. But if visualization fails, it would be better to issue a warning and let the user decide whether to interrupt acquisition or not.

This PR proposes making the decision between critical and non-critical at callback definition time. What would be the pros/cons of making it at subscription time instead? Something like:

```py
RE.subscribe(bec, errors='warn')
```

where the options might be 'ignore', 'warn', and 'raise' (or, if we follow `b''.decode()`, perhaps 'strict'). I see a benefit in deferring to the subscriber the decision of whether a given callback is critical.

tacaswell:
```python
RE.subscribe(bec, errors='warn')
```

harkens back to `subscribe` and `subscribe_lossless` from the very early days.  I saw a great quote the other day "problems can be 'solved', dilemmas can only be managed"  [[1]](https://www.johndcook.com/blog/2019/08/31/hopeless-task-of-unicode/)

There are some callbacks that we know at definition time should never under any circumstances kill data collection (ex plotting, olog) or under every circumstance should kill data collection (failure to record data).  I am having trouble thinking if a use case where we want to defer that decision to subscription time.  That seems like a high risk of failing to mark something as `strict=False` and then have a rare exception kill a scan or to give the users a giant foot cannon to ignore errors on data insert.

That said, I am not sure that this is an either/or, we may want the decorator _and_ a keyword argument.   We are also straddling the library vs application line here.

prjemian:
Since some (data storage) should always fail while others (visualization) should never fail, there is no best decision for all.  Can we defer this decision to the specific implementation?  A callback might have its own default setting which could be changed at the time the callback object is created.

tacaswell:
We have three suggestions for when we should decide if a given callback should be allowed to kill a running scan by failing:

 1. At class/function definition time (this PR) so all instances of that call back will have the same behavior.  This is technically the status-quo, the decorators are just helpers to make it easier to be very sure it will never raise
 2. At class init time (what @prjemian is suggesting) which means each callback will have to have the machinery to raise or not.  Do we want _all_ sub-classes of `CallbackBase` to have this knob?
 3. At subscription time (what @danielballan is suggesting).  This means no callback class has to carry any machinery, but I think runs a higher risk of miss-configuration.

danielballan:
> I am having trouble thinking if a use case where we want to defer that decision to subscription time.

Fair. I am not sure we need a switch _at subscription time_, but we need a switch. In the scenario where callbacks are failing, we want to option to make them fail loudly so we can `%debug`. (I recall  this was our motivation for flipping the default value of `ignore_callback_exceptions` to `True`.)This could be done via a parameter to `__init__` or an environment variable.

In the spirit of @prjemian's suggestion, I think the simplest way forward is to give `CallbackBase` (and its successor, `event_model.DocumentRouter`) a new parameter in `__init__`. Subclasses can specify a default, to mitigate the risk of misconfiguration, but users can easily override it at subscription time. This _could_ be implemented through a decorator that intercepts a specific parameter, but adding a parameter to the base class seems simpler.

It _also_ be useful to have an override that checks for environment variable (e.g. `BLUESKY_DEBUG_CALLBACKS`) that puts everything into strict mode.

tacaswell:
I don't think per-instance state is worth the complexity, both in the code and in the "is _this_ instance going to raise?" .  I put in a check on the suggested ENV which I think is the best of both worlds.  The callback author can decide how critical _this_ callback is and there is a way to make everything raise for debugging purposes.

tacaswell:
Closes #1184

bluesky/callbacks/core.py
```diff
@@ -3,15 +3,55 @@
...
+
+def make_callback_safe(func=None, *, logger=None):
+    if func is None:
+        return _partial(make_callback_safe, logger=logger)
+
+    @_wraps(func)
+    def inner(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except Exception as ex:
```
danielballan: No need to capture `ex` in a variable.
```suggestion
        except Exception:
```

bluesky/callbacks/core.py
```diff
@@ -3,15 +3,55 @@
...
+            if logger is not None:
+                logger.exception(ex)
```
danielballan: I believe this expects a string message (suggestion: “This exception occurred in a ‘safe’ callback and will be ignored to avoid interrupting acquisition.”) The exception is detected automatically from context and need not be passed in.

bluesky/callbacks/core.py
```diff
@@ -3,15 +3,55 @@
...
+            if os.environ.get('BLUESKY_DEBUG_CALLBACKS', False):
+                raise ex
```
danielballan:
```suggestion
                raise
```

bluesky/callbacks/core.py
```diff
@@ -3,15 +3,55 @@
...
 from ..utils import ensure_uid
+from event_model import DocumentNames
+
+
+logger = logging.getLogger(__name__)
+
+
+def make_callback_safe(func=None, *, logger=None):
+    if func is None:
+        return _partial(make_callback_safe, logger=logger)
```
danielballan: This is clever!

bluesky/callbacks/core.py
```diff
@@ -3,15 +3,65 @@
...
+            return func(*args, **kwargs)
+        except Exception:
+            debug_mode = os.environ.get("BLUESKY_DEBUG_CALLBACKS", False)
+            if logger is not None:
+                if debug_mode:
+                    msg = f"Exception in {func}"
+                else:
+                    msg = (
+                        f"An exception raised in the callback {func} "
+                        + "is being suppressed to not interupt plan "
```
danielballan:
```suggestion
                        + "is being suppressed to not interrupt plan "
```

bluesky/callbacks/core.py
```diff
@@ -3,15 +3,65 @@
...
+                        + "execution.  To investigate try setting the "
+                        + "BLUESKY_DEBUG_CALLBACKS env to '1'"
```
danielballan: ...or really to _anything_.  Should we make it so the `BLUESKY_DEBUG_CALLBACKS=0` turn has no effect?

danielballan: @mrakitin What do you think? Is it better to have this switch on set/unset (ignoring the value) or to provide a way to explicitly set this to off?

mrakitin: I think the first way (a switch) is better, as one does not have to worry about setting it to anything by default, and resetting `BLUESKY_DEBUG_CALLBACKS= <shell command>` will work if needed. Same was as https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUNBUFFERED works (I saw people use `1` or `x` to enable it).

mrakitin: Also, it seems we don't need `+`s here.

bluesky/tests/test_callbacks.py
```diff
@@ -476,3 +487,196 @@ def test_plotting_hints(RE, hw, db):
...
+    assert test_function(False) is False
+    assert logger.exception.call_count == 1
+
+
+@pytest.fixture
+def EvilBaseClass(request):
+    class MyError(RuntimeError):
+        ...
+
+    class EvilCallback(CallbackBase):
```
mrakitin: I like the name 😄

PR #1252, opened 2019-09-04
TST: don't set the register on test devices

tacaswell:
This will reduce the number of warnings from the test suite.

The deleted test is explicitly testing the old "internal mode" which
if we continue to test it, should be tested in ophyd or databroker,
but not in bluesky.

We want to push the test suite to be fail-on-warning.

It is changes to the tests


danielballan:
Thanks for cleaning these up, @tacaswell.

PR #1256, opened 2019-10-04
Support adaptive fly scans.

danielballan:
In step scans, the plans can access the data like:

```py
reading = yield from bps.read(obj)
```

This PR enables this for fly scans:

```py
payload = yield from bps.collect(obj)
```

In situations where the payload is too large to accumulate in memory,
this new option opts out of accumulation and causes the coroutine to
return ``None``.

```py
yield from bps.collect(obj, return_payload=False)
```


danielballan:
New commits look good, @tacaswell. Thanks.

danielballan:
Extra context for posterity: this was written for / tested with https://github.com/danielballan/adaptive-with-delay

tacaswell:
The very wrong docstring on `_collect` was copy-pasta errors from the bundler refactor, sorry about that.

danielballan:
"Copy-pasta errors" is a typo for the ages.

prjemian:
🍝

PR #1258, opened 2019-10-16
Unit test updates for ophyd simulated Devices

jklynch:
Two small changes to unit tests for compatibility with updated ophyd simulated hardware.
This is really the work of @tacaswell, in case that is not obvious.

## Motivation and Context
This PR corresponds to the larger PR for ophyd #762, which aligns simulated hardware classes with "real" ophyd objects. This resolves an issue in suitcase-utils, which currently depends on an ophyd branch.

## How Has This Been Tested?
The bluesky tests pass with ophyd PR #762.

### DIFF
--- a/bluesky/tests/test_callbacks.py
+++ b/bluesky/tests/test_callbacks.py
@@ -151,7 +151,7 @@ def test_table(RE, hw):
         hw.det.precision = 2
         hw.motor.precision = 2
         hw.motor.setpoint.put(0.0)  # Make dtype 'number' not 'integer'.
-        hw.det.trigger()  # Make dtype 'number' not 'integer'.
+        hw.det.trigger()
         assert hw.det.describe()['det']['precision'] == 2
         assert hw.motor.describe()['motor']['precision'] == 2
         assert hw.det.describe()['det']['dtype'] == 'number'
--- a/bluesky/tests/test_callbacks.py
+++ b/bluesky/tests/test_callbacks.py
@@ -151,7 +151,7 @@ def test_table(RE, hw):
         hw.det.precision = 2
         hw.motor.precision = 2
         hw.motor.setpoint.put(0.0)  # Make dtype 'number' not 'integer'.
-        hw.det.put(0.0)  # Make dtype 'number' not 'integer'.
+        hw.det.trigger()  # Make dtype 'number' not 'integer'.
         assert hw.det.describe()['det']['precision'] == 2
         assert hw.motor.describe()['motor']['precision'] == 2
         assert hw.det.describe()['det']['dtype'] == 'number'
--- a/bluesky/tests/test_new_examples.py
+++ b/bluesky/tests/test_new_examples.py
@@ -642,7 +642,7 @@ def test_no_rewind_device(hw):
 
 def test_monitor(RE, hw):
     from ophyd.sim import SynSignal
-    signal = SynSignal(name='signal')
+    signal = SynSignal(name='signal', value=0.0)
     RE(monitor_during_wrapper(count([hw.det], 5), [signal]))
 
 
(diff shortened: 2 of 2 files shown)


bluesky/tests/test_callbacks.py
```diff
@@ -151,7 +151,7 @@ def test_table(RE, hw):
         hw.det.precision = 2
         hw.motor.precision = 2
         hw.motor.setpoint.put(0.0)  # Make dtype 'number' not 'integer'.
-        hw.det.put(0.0)  # Make dtype 'number' not 'integer'.
+        hw.det.trigger()  # Make dtype 'number' not 'integer'.
```
danielballan:
```suggestion
        hw.det.trigger()
```
It seems the comment no longer applies.

bluesky/tests/test_new_examples.py
```diff
@@ -641,7 +641,9 @@ def test_no_rewind_device(hw):
 
 
 def test_monitor(RE, hw):
-    RE(monitor_during_wrapper(count([hw.det], 5), [hw.det1]))
+    from ophyd.sim import SynSignal
```
mrakitin: Can this import be moved to the top of the file?

danielballan: I think this is intentional. Since bluesky does not have an ophyd dependency, we have laid some track for making it possible run the subset of the bluesky tests that don't require ophyd. The `hw` fixture is a way of declaring, "This test needs ophyd."

mrakitin: Ah, I see. It makes sense even that we have it in https://github.com/bluesky/bluesky/blob/018d4bdd9e0f6d5365e75374f26247328763e1f6/dev-requirements.txt#L16

bluesky/tests/test_new_examples.py
```diff
@@ -641,7 +641,9 @@ def test_no_rewind_device(hw):
...
+    signal = SynSignal(name='signal')
```
tacaswell:
```suggestion
    signal = SynSignal(name='signal', value=0.0)
```

bluesky/tests/test_callbacks.py
jklynch: The `hw.det.trigger()` call is not necessary with the corresponding ophyd updates. Does it make sense to remove this call, @tacaswell?

mrakitin:
Is it actually "Ready for review" or still a draft?

bluesky/tests/test_new_examples.py
```diff
@@ -641,7 +641,9 @@ def test_no_rewind_device(hw):
 
 
 def test_monitor(RE, hw):
-    RE(monitor_during_wrapper(count([hw.det], 5), [hw.det1]))
+    from ophyd.sim import SynSignal
+    signal = SynSignal(name='signal', value=0.0)
```
tacaswell:
```suggestion
    signal = SynSignal(name='signal')
    signal.put(0.0)
```

Sigh.

prjemian:
Why "do not assign to value at initialization"?

jklynch:
Can you address this question, @tacaswell?

jklynch:
This does not answer the question, but I think `value=0.0` was added in hope that it would do everything `put(0.0)` does, so reverting to `put(0.0)` is going back to the way things were.

mrakitin:
We'll question @tacaswell in person next week why that `value=0.0` -> `put(0.0)` was needed. To me it looks like they should behave identically.

danielballan:
My guess is that `.put()` triggers a monitor update and initialization does not.

tacaswell:
You would _think_ that, but internal to synsignal it does `super().__init__(value=_func(), ..., **kwargs)` so you get "multiple values passed" errors.

Issue #1261, opened 2019-11-01
A no-op QtAwareCallback for 1.5.x?


danielballan:
We will need to add [`QtAwareCallback`](https://github.com/bluesky/bluesky/blob/1538bdfd316811dcbfdc66bad2faea1e764ed26f/bluesky/callbacks/mpl_plotting.py#L21) to any callbacks defined in beamline profiles for bluesky 1.6.x. Should we define a class by this name that does nothing in 1.5.x so that beamlines can switch environments without changing their profiles?

danielballan:
Closed by #1262

PR #1262, opened 2019-11-01
MNT: Add no-op-QtAwareCallback

ke-zhang-rd:
## Description

This is for issue #1261


danielballan:
Power-cycled to re-run CI on top of #1306.

PR #1264, opened 2019-11-07
submit_info plan

mrakitin:
## Description

This PR provides a plan that can be used to submit information to databroker with the customizable `plan_name`.

## Motivation and Context

I was asked about it by @lyang11973, and the implementation here is the extension of the original suggestion to use `bp.count([], md={...})`.

## How Has This Been Tested?

A pytest was added to the test suite.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1843,6 +1843,22 @@ def x2x_scan(detectors, motor1, motor2, start, stop, num, *,
         md=_md))
 
 
+def submit_info(info, plan_name='info', field_name='info'):
+    """
+    Submit info into the metadata of the start document.
+
+    Parameters
+    ----------
+    info : any serializable object
+        the information to submit
+    plan_name : str, optional
+        the name of the plan (useful for further searches)
+    field_name : str, optional
+        the name of the field under which the info will appear in the metadata
+    """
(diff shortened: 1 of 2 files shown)


tacaswell:
I can think of a couple of reasons this would be useful, but I am curious what problem @lyang11973 is trying to solve.

prjemian:
There is a [`documentation_run()`](https://github.com/BCDA-APS/apstools/blob/786d4f6a1b73dc92de7d342aacd824eeeaad5547/apstools/plans/doc_run.py#L46-L87) **plan** ([source](https://github.com/BCDA-APS/apstools/blob/786d4f6a1b73dc92de7d342aacd824eeeaad5547/apstools/plans/doc_run.py#L46-L87)) in apstools.  The `plan_name` could be set in the metadata when calling.

```python
RE(documentation_run("very simple demo", md=dict(stream="information", plan_name="submit_info")))
```

prjemian:
We use it to document a batch list (it describes the next series of runs).

stan-dot:
this is an interesting use case, @prjemian  is this still relevant given how the apstools have changed?

prjemian:
I'll check on this since it is older.

prjemian:
I'm not sure that I would still need to use this.

PR #1265, opened 2019-11-13
Accept kwargs in bps.mv

danielballan:
The `bps.abs_set` plan accepts `**kwargs` and passes them through to
`obj.set(..., **kwargs)`. This PR makes `bps.mv` do so as well. I
suspect that we might have omitted it this on purpose because, unlike
`abs_set`, `mv` accepts *multiple* devices and there is no way to
express specific `kwargs` for each device without making the signature
complex. However, for common `kwargs` like `timeout` this might be nice
to have.

### DIFF
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -214,7 +214,7 @@ def rel_set(obj, *args, group=None, wait=False, **kwargs):
         abs_set(obj, *args, group=group, wait=wait, **kwargs)))
 
 
-def mv(*args, **kwargs):
+def mv(*args, group=None, **kwargs):
     """
     Move one or more devices to a setpoint. Wait for all to complete.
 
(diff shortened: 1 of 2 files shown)


bluesky/plan_stubs.py
```diff
@@ -214,7 +214,7 @@ def rel_set(obj, *args, group=None, wait=False, **kwargs):
         abs_set(obj, *args, group=group, wait=wait, **kwargs)))
 
 
-def mv(*args):
+def mv(*args, **kwargs):
```
tacaswell: Should probably explicitly list `group` here and then do the `if is None...` dance or we will have 1 kwarg which will not work (and does make sense to broadcast across all of the motors.

tacaswell:
> I suspect that we might have omitted it this on purpose because, unlike
> `abs_set`, `mv` accepts _multiple_ devices and there is no way to
> express specific `kwargs` for each device without making the signature
> complex.

This is my recollection as well.

PR #1272, opened 2019-12-04
fixes #1270 - adding utils.plot_prune_fifo()

### LINKED ISSUE #1270: LivePlot window problem when legend too large
In https://github.com/APS-USAXS/ipython-usaxs/issues/309, @jilavsky reports

> When graphs are opened for too many scans, matPlotLib used to generate these has bug. Graphs become useless and there does not seem obvious way to fix this.

See the attached picture in that issue.

## Expected Behavior
For unattended, continuous operation over long periods of time (such as a weekend), the plot windows must be stable.

## Current Behavior
When the legend grows too large (@jilavsky, can you provide some numbers here?), problems happen (@jilavsky, can you be more specific?)

## Possible Solution
1. Add an option to set the maximum number of traces shown in a single window. 
1. Add an option to open new window after *n* number of traces.
1. other

prjemian:
In several issues (https://github.com/bluesky/bluesky-mpl/issues/4, https://github.com/APS-USAXS/ipython-usaxs/issues/309, and here in #1270), it has been noted that plotting too many traces in a LivePlot window causes the best effort callback to crash or the LivePlot window to become too small to view the plot.

## Description

This PR introduces a utility function `plot_prune_fifo()` (not a bluesky plan) which the user calls at their choice to replot the LivePlot window with only the last *n* traces shown.

If the user has chosen to display peak centers (pressing `P` in the plot window) as lines, this function will work around those lines.  If there is no plot window that matches the given axes, the function will return silently.

For this function to work, it is necessary for the user to pass in the BestEffortCallback object and the objects for y and x axes, in addition to the non-negative number of traces to keep when replotting.

## Motivation and Context
fixes #1270

## How Has This Been Tested?
This has been tested programmatically while repeating a set of lineup scans, seeking the position of some simulated diffraction peak.  Before each pass through the loop, a call to `plot_prune_fifo(bec, 4, noisy, m1)` is used to limit the number of graphs shown to a number known not to cause the problems reported.

In the testing loop, the peak parameters are first randomized, then a `bp.scan()` is run (over an absolute range known to contain the peak somewhere, typical of a beam line search) to assess `bec.peaks["cen"]` and adjust the positioner.  A second `bp.rel_scan()` is run between +/- `bec.peaks["fwhm"]` to refine the `bec.peaks`  findings.

## Screenshots:

during continuous running: 
Clipboard01

then, later:
Clipboard01

and then even later, still looping `while True`:  
Clipboard01


prjemian:
This [travis-ci failure](https://travis-ci.org/bluesky/bluesky/builds/620856496?utm_source=github_status&utm_medium=notification) is for some other problem not in this PR since the PR has not introduced any new tests nor should it influence any existing tests.

mrakitin:
Actually, the failure is not with TravisCI, but rather with the slight reduction of the test coverage.

prjemian:
Isn't that from codecov?  Four out of five travis-ci workers were green.  This one was red: https://travis-ci.org/bluesky/bluesky/jobs/620856501?utm_medium=notification&utm_source=github_status

mrakitin:
We seem to ignore the nightly failures as they can be caused by various unrelated changes in Python itself or other libraries.

bluesky/utils.py
```diff
@@ -1486,3 +1486,54 @@ def _rearrange_into_parallel_dicts(readings):
...
+        lp = liveplot.get(y.name)
+        if lp is None or lp.x != x.name or lp.y != y.name:
+            continue
+
+        # pick out only the traces that contain plot data
+        # skipping the lines that show peak centers
+        lines = [
+            tr
+            for tr in lp.ax.lines
+            if len(tr._x) != 2
```
mrakitin: Why not the `>` condition? Do we care about the len=0 or 1?

bluesky/utils.py
```diff
@@ -1486,3 +1486,54 @@ def _rearrange_into_parallel_dicts(readings):
...
+    find the plot with axes x and y and replot with only the last *n* lines
+
+    Note: this is not a bluesky plan.  Call it as normal Python function.
+
+    EXAMPLE::
+
+        plot_prune_fifo(bec, 1, noisy, m1)
+
+    PARAMETERS
+
```
mrakitin: Could you please change it to be consistent with the other code and be a valid sphinx doc? The EXAMPLE block may need that update too.

bluesky/utils.py
```diff
@@ -1486,3 +1486,54 @@ def _rearrange_into_parallel_dicts(readings):
         data[key] = payload['value']
         timestamps[key] = payload['timestamp']
     return data, timestamps
+
+
+def plot_prune_fifo(bec, n, y, x):
```
mrakitin: I'd prefer to have the arguments required to be specified with the kwarg names (put `*` after the `bec` argument). Also, having longer, self-explanatory names would greatly benefit the end users.

prjemian: None of these arguments are optional (kwargs) so wouldn't the `*` come after all of them?

But, to avoid having to provide the `bec` argument at all, this function might be better located as a method of BestEffortCallback.  Agree?

bluesky/utils.py
```diff
@@ -1486,3 +1486,54 @@ def _rearrange_into_parallel_dicts(readings):
...
+
+    Note: this is not a bluesky plan.  Call it as normal Python function.
+
+    EXAMPLE::
+
+        plot_prune_fifo(bec, 1, noisy, m1)
+
+    PARAMETERS
+    
+    bec : object
```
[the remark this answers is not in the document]
prjemian: That explains what I have just seen elsewhere.  Thanks for the catch. catches.

bluesky/utils.py
prjemian: We're trimming out the lines which mark the peak centers, specifically.  Other cases can take care of themselves, as usual.  The logic is correct for this intent.

prjemian: The lines which mark the peak centers have exactly two points and the X value is identical for each of those.  It is truly rare (although possible) to have measured data with exactly two points that have the same X.  Not sure I have any other means to identify.  Perhaps if I look at the Y values?  They might have special meaning, as well, and that special meaning could be used to identify them uniquely.

bluesky/utils.py
prjemian: Thanks for the review.  I'll make those changes.

bluesky/utils.py
prjemian: Sorry.  I can't see any of these as kwargs.  Since no other args are parsed or accepted, there is nothing to change.

bluesky/callbacks/best_effort.py
```diff
@@ -441,6 +441,55 @@ def clear(self):
...
+                tr
+                for tr in lp.ax.lines
+                if len(tr._x) != 2 
+                    or len(tr._y) != 2 
+                    or (len(tr._x) == 2 
+                        and tr._x[0] != tr._x[1])
+            ]
+            if len(lines) > n:
+                print(f"limiting LivePlot({y_signal.name}) to {num_lines} lines")
+                lp.ax.lines = lines[-num_lines:]
```
tacaswell: Please use the remove method on the `Line2D` object rather than directly touching the `ax.lines` list.

prjemian: Sounds good.

prjemian: Like this?

```
                keepers = lines[-num_lines:]
                for tr in lp.ax.lines:
                    if tr not in keepers:
                        tr.remove()
```

If plot centers are drawn, they will be recreated with the next scan.

bluesky/callbacks/best_effort.py
```diff
@@ -441,6 +441,53 @@ def clear(self):
...
+                tr
+                for tr in lp.ax.lines
+                if len(tr._x) != 2 
+                    or len(tr._y) != 2 
+                    or (len(tr._x) == 2 
+                        and tr._x[0] != tr._x[1])
+            ]
+            if len(lines) > n:
+                print(f"limiting LivePlot({y_signal.name}) to {num_lines} lines")
+                keepers = lines[-num_lines:]
```
danielballan: :100: variable name choice :- D

bluesky/callbacks/best_effort.py
```diff
@@ -441,6 +441,53 @@ def clear(self):
...
+            number of lines (plotted scans) to keep, must be >= 0
+        y_signal: object
+            instance of ophyd.Signal (or subclass), 
+            dependent (y) axis
+        x_signal: object
+            instance of ophyd.Signal (or subclass), 
+            independent (x) axis
+
+        """
+        assert num_lines >= 0, "num_lines must be 0 or greater"
```
danielballan: Minor Python quibble: `assert` shouldn't be used for validating user input because the language does not guarantee that asserts are run, such as when the interpreter is running in performance-optimized mode. Would you mind changing this to something like

```py
if ...:
    raise ValueError(...)
```
?

bluesky/utils.py
danielballan: I agree that making this a method is a good choice. Also, the new, more descriptive parameter names are much more clear.

To explain, I think @mrakitin is suggesting [keyword-only arguments](https://www.python.org/dev/peps/pep-3102/) here so that `plot_prune_fifo` could _only_ be called by spelling out the parameter names.  Personally I think it would be OK to leave the arguments as the normal positional-or-keyword kind in this case, but that's a matter of taste.

mrakitin: Right. Thanks for the reference, Dan. 

It would be much clearer when the args are spelled when the function is called, so we are not confused by the reverse order of x and y.

bluesky/callbacks/best_effort.py
```diff
@@ -441,6 +441,53 @@ def clear(self):
         self._buffer = StringIO()
         self._baseline_toggle = True
 
+    def plot_prune_fifo(self, num_lines, y_signal, x_signal):
```
mrakitin:
```suggestion
    def plot_prune_fifo(self, *, num_lines, y_signal, x_signal):
```

bluesky/callbacks/best_effort.py
```diff
@@ -441,6 +441,53 @@ def clear(self):
...
+        """
+        Find the plot with axes x_signal and y_signal.  Replot with only the last *num_lines* lines.
+
+        .. note:: This is not a bluesky plan.  Call it as a normal Python function.
```
mrakitin: This note may not be relevant anymore, as the function became a method of `BestEffortCallback`.

bluesky/callbacks/best_effort.py
```diff
@@ -441,6 +441,53 @@ def clear(self):
...
+        ----------
+        num_lines: int
+            number of lines (plotted scans) to keep, must be >= 0
+        y_signal: object
+            instance of ophyd.Signal (or subclass), 
+            dependent (y) axis
+        x_signal: object
+            instance of ophyd.Signal (or subclass), 
+            independent (x) axis
+
```
mrakitin:
```suggestion
```

bluesky/callbacks/best_effort.py
mrakitin: Wow, I didn't know about the performance mode. I used asserts a lot in various projects, probably worth to reassess their usefulness.

bluesky/callbacks/best_effort.py
```diff
@@ -441,6 +441,53 @@ def clear(self):
...
+            lines = [
+                tr
+                for tr in lp.ax.lines
+                if len(tr._x) != 2 
+                    or len(tr._y) != 2 
+                    or (len(tr._x) == 2 
+                        and tr._x[0] != tr._x[1])
+            ]
+            if len(lines) > n:
+                print(f"limiting LivePlot({y_signal.name}) to {num_lines} lines")
```
mrakitin:
```suggestion
                print(f"Limiting LivePlot({y_signal.name}) to {num_lines} lines")
```

bluesky/callbacks/best_effort.py
```diff
@@ -441,6 +441,53 @@ def clear(self):
...
+            # skipping the lines that show peak centers
+            lines = [
+                tr
+                for tr in lp.ax.lines
+                if len(tr._x) != 2 
+                    or len(tr._y) != 2 
+                    or (len(tr._x) == 2 
+                        and tr._x[0] != tr._x[1])
+            ]
+            if len(lines) > n:
```
mrakitin:
```suggestion
            if len(lines) > num_lines:
```

bluesky/callbacks/best_effort.py
```diff
@@ -441,6 +441,53 @@ def clear(self):
...
+
+        """
+        assert num_lines >= 0, "num_lines must be 0 or greater"
+        for liveplot in self._live_plots.values():
+            lp = liveplot.get(y_signal.name)
+            if lp is None or lp.x != x_signal.name or lp.y != y_signal.name:
+                continue
+
+            # pick out only the lines that contain plot data
+            # skipping the lines that show peak centers
```
mrakitin: Thanks for the explanation.

mrakitin:
I think testing this functionality with pytest is still possible, so we would at least hit the leftover `n` variable name.

bluesky/callbacks/best_effort.py
prjemian: can do

bluesky/callbacks/best_effort.py
prjemian: Ah, now I see [... as slowly he uses a feature he has been ignoring ...]

bluesky/callbacks/best_effort.py
prjemian: I agree, I'll pull it out.

bluesky/callbacks/best_effort.py
prjemian: Doesn't Sphinx need this blank line due to the indented content?

bluesky/callbacks/best_effort.py
prjemian: Now it's a complete sentence but it needs to end with a period.  I'll add that.

prjemian:
As for the order of arguments, y before x, that was my second choice and it was to match the same order when calling `bp.scan()`.  I'll switch that back since *x before y* is more natural.

bluesky/callbacks/best_effort.py
mrakitin: Not after the list of arguments in case we don't have the continuation of the docstring, as far as I know.

bluesky/callbacks/best_effort.py
mrakitin: Thanks!

prjemian:
still, this deserves at least some attention from unit testing ... working on that (while waiting for XPCS instrument)

prjemian:
While writing the unit test, the editor says the arguments to `bec.plot_prune_fifo()` **must** be specified by name.  I object to this, strongly.  We must not make this verbose.  I will revert back to the simpler named arguments by removing the `*`.

prjemian:
The new tests must be checked.  *Every* test in test_bec.py fails on my workstation.  I'm new to using `pytest`.

bluesky/tests/test_bec.py
```diff
@@ -109,3 +109,40 @@ def test_plot_ints(RE):
...
+
+    # add more lines to the LivePlot
+    RE(bps.repeater(num_pruned, scan,[hw.ab_det], hw.motor, 1, 5, 5))
+
+    # get the LivePlot object, again, in case the UUID was changed
+    liveplots_list = list(bec._live_plots.values())
+    assert len(liveplots_list) == 1
+    lp = liveplots_list[0].get(hw.ab_det.name)
+    assert lp is not None
+    assert len(lp.ax.lines) == num_pruned*2
```
mrakitin: Let's make potential flake8 tests happy.
```suggestion
    assert len(lp.ax.lines) == num_pruned * 2
```

bluesky/tests/test_bec.py
```diff
@@ -109,3 +109,40 @@ def test_plot_ints(RE):
...
+    lp = liveplots_list[0].get(hw.ab_det.name)
+    assert lp is not None
+    assert len(lp.ax.lines) == num_pruned
+
+    # prune the LivePlot (this is no-op since we have exact number to keep)
+    bec.plot_prune_fifo(num_pruned, hw.motor, hw.ab_det)
+    assert len(lp.ax.lines) == num_pruned
+
+    # add more lines to the LivePlot
+    RE(bps.repeater(num_pruned, scan,[hw.ab_det], hw.motor, 1, 5, 5))
```
mrakitin: pep8 here too.
```suggestion
    RE(bps.repeater(num_pruned, scan, [hw.ab_det], hw.motor, 1, 5, 5))
```

bluesky/tests/test_bec.py
```diff
@@ -109,3 +109,40 @@ def test_plot_ints(RE):
...
+
+
+def test_plot_prune_fifo(RE, hw):
+    bec = BestEffortCallback()
+    RE.subscribe(bec)
+
+    num_pruned = 2
+
+    # create the LivePlot
+    RE(bps.repeater(num_pruned, scan,[hw.ab_det], hw.motor, 1, 5, 5))
```
mrakitin:
```suggestion
    RE(bps.repeater(num_pruned, scan, [hw.ab_det], hw.motor, 1, 5, 5))
```

prjemian:
I don't understand why the Y axis is named different than expected.  Can anyone explain?

Here's where the unit test fails in [travis-ci](https://travis-ci.org/bluesky/bluesky/jobs/621258714?utm_medium=notification&utm_source=github_status)

```
        # get reference key for our detector
        det_name = next(iter(bec._live_plots[uuid]))
        # should be same as hw.ab_det.name, prove it now
>       assert det_name == hw.ab_det.name
E       AssertionError: assert 'det_a' == 'det'
E         - det_a
E         ?    --
E         + det
```

prjemian:
This was the line that started the scan

    RE(bps.repeater(num_pruned, scan, [hw.ab_det], hw.motor, 1, 5, 5))

and the name `det_a` is shown in the LiveTable output:

```
Transient Scan ID: 1     Time: 2019-12-05 19:22:54
Persistent Unique Scan ID: '466e42eb-6916-488a-aa56-b89fc9d8c03c'
New stream: 'primary'
|   seq_num |       time |      motor |      det_a |
|         1 | 19:22:54.2 |      1.000 |      0.734 |
|         2 | 19:22:54.3 |      2.000 |      0.538 |
|         3 | 19:22:54.3 |      3.000 |      0.653 |
|         4 | 19:22:54.4 |      4.000 |      0.935 |
|         5 | 19:22:54.4 |      5.000 |      0.967 |
generator scan ['466e42eb'] (scan num: 1)
```

mrakitin:
> I don't understand why the Y axis is named different than expected. Can anyone explain?
> 
> Here's where the unit test fails in [travis-ci](https://travis-ci.org/bluesky/bluesky/jobs/621258714?utm_medium=notification&utm_source=github_status)
> 
> ```
>         # get reference key for our detector
>         det_name = next(iter(bec._live_plots[uuid]))
>         # should be same as hw.ab_det.name, prove it now
> >       assert det_name == hw.ab_det.name
> E       AssertionError: assert 'det_a' == 'det'
> E         - det_a
> E         ?    --
> E         + det
> ```

I don't know where the `_a` is appended to the name of the detector, defined in https://github.com/bluesky/ophyd/blob/6414a0673793309ad7cd2a55d736e3bc49a8b6cb/ophyd/sim.py#L1355:
```py
    ab_det = ABDetector(name='det', labels={'detectors'})
```

@danielballan, @tacaswell, do you have any hint about it?

mrakitin:
Ah, I see now, it the name of the first `read_attr`:
```py
$ ipython
Python 3.6.7 |Anaconda, Inc.| (default, Oct 23 2018, 14:01:38)
IPython 7.5.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from ophyd.sim import hw

In [2]: hw = hw()

In [3]: hw.ab_det
Out[3]: ABDetector(prefix='', name='det', read_attrs=['a', 'b'], configuration_attrs=[])

In [4]: hw.ab_det.name
Out[4]: 'det'

In [5]: hw.ab_det.read_attrs
Out[5]: ['a', 'b']
```

bluesky/utils.py
mrakitin: Looks OK for now.

prjemian:
OK, seems a special case for *this* signal.  Thanks for the mining.

mrakitin:
It's not a special case, it seems to be a general construction method of the field names corresponding to the "hinted" component names used in the live plots/tables. Continuing the demo:
```py
In [6]: hw.ab_det.kind
Out[6]: <Kind.config|normal: 3>

In [7]: hw.ab_det.a.kind
Out[7]: <Kind.hinted: 5>

In [8]: hw.ab_det.b.kind
Out[8]: <Kind.normal: 1>
```
That's why we don't see the `det_b` in the table.

prjemian:
For the unit test, I'm going to add the `.a`.  Is this a practice the prune method should consider?  I avoided using str as parameters, instead taking the str from the signal.name because deciding what string to provide (dotted, underlined, something else) was one of the early, hard stumbles I had when learning ophyd.

mrakitin:
It seems like an off-by-one bug. I reproduced it locally with `pytest -vvvv -x --pdb`, and seeing the following:
```
X: motor
Y: det_a
before: [<matplotlib.lines.Line2D object at 0x127dc9208>, <matplotlib.lines.Line2D object at 0x127c80be0>, <matplotlib.lines.Line2D object at 0x127d75ba8>, <matplotlib.lines.Line2D object at 0x12805a278>]
Limiting LivePlot(det_a) to 2 lines.
after: [<matplotlib.lines.Line2D object at 0x127c80be0>, <matplotlib.lines.Line2D object at 0x127d75ba8>, <matplotlib.lines.Line2D object at 0x12805a278>]
```

prjemian:
Right.  But *how* is it off by one?  Time for temporary diagnostics in the plot_prune_fifo()

bluesky/callbacks/best_effort.py
```diff
@@ -441,6 +441,55 @@ def clear(self):
...
+                    or len(tr._y) != 2 
+                    or (len(tr._x) == 2 
+                        and tr._x[0] != tr._x[1])
+            ]
+            print(f"found these lines: {lines}")
+            if len(lines) > num_lines:
+                print(f"Limiting LivePlot({y_signal.name}) to {num_lines} lines.")
+                keepers = lines[-num_lines:]
+                print(f"keepers: {keepers}")
+                for tr in lp.ax.lines:
```
mrakitin: Should it be looping over `lines` instead of `lp.ax.lines`, as the center lines were removed at this point?

mrakitin: I changed it locally, and now it's passing!
```
$ pytest -vvvv -x --pdb bluesky/tests/test_bec.py::test_plot_prune_fifo
================================================= test session starts ==================================================
platform darwin -- Python 3.6.7, pytest-4.5.0, py-1.7.0, pluggy-0.12.0 -- /Users/mrakitin/anaconda3/envs/bluesky/bin/python
cachedir: .pytest_cache
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/mrakitin/src/mrakitin/DAMA/bluesky
plugins: faulthandler-1.6.0, benchmark-3.2.2
collected 1 item

bluesky/tests/test_bec.py::test_plot_prune_fifo PASSED                                                           [100%]

=============================================== 1 passed in 4.62 seconds ===============================================
```
I can push my change to your PR branch if you wish.

prjemian:
This [diagnostic](https://travis-ci.org/bluesky/bluesky/jobs/621332819?utm_medium=notification&utm_source=github_status) show that the `LivePlot.remove()` method has not been successful for one of the traces (`0x7f532810f7b8`) to be removed:

```
keepers: [<matplotlib.lines.Line2D object at 0x7f5328199e80>, <matplotlib.lines.Line2D object at 0x7f53281406d8>]
after: [<matplotlib.lines.Line2D object at 0x7f532810f7b8>, <matplotlib.lines.Line2D object at 0x7f5328199e80>, <matplotlib.lines.Line2D object at 0x7f53281406d8>]
```

prjemian:
Do it, thanks for following along.

prjemian:
As an aside, here's what happens when I try to run pytest locally:

```
(db-test) mintadmin@mint-vm:~/.../eclipse/bluesky$ pytest -vvvv -x --pdb bluesky/tests/test_bec.py::test_plot_prune_fifo
=========================================================== test session starts ============================================================
platform linux -- Python 3.7.5, pytest-5.3.1, py-1.8.0, pluggy-0.13.1 -- /home/mintadmin/Apps/anaconda/envs/db-test/bin/python
cachedir: .pytest_cache
rootdir: /home/mintadmin/Documents/eclipse/bluesky
collected 1 item                                                                                                                           

bluesky/tests/test_bec.py::test_plot_prune_fifo ERROR                                                                                [100%]
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

tmpdir = local('/tmp/pytest-of-mintadmin/pytest-3/test_plot_prune_fifo0')

    @pytest.fixture(scope='function')
    def hw(tmpdir):
        from ophyd.sim import hw
>       return hw(str(tmpdir))
E       TypeError: hw() takes 0 positional arguments but 1 was given

bluesky/tests/conftest.py:31: TypeError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/mintadmin/Documents/eclipse/bluesky/bluesky/tests/conftest.py(31)hw()
-> return hw(str(tmpdir))
(Pdb) 
```

mrakitin:
Yep, I had this problem as well. Reinstalling the latest pre-release of ophyd solved the issue.

bluesky/callbacks/best_effort.py
prjemian: I'm on a roll.  I'll do it from here.

prjemian: I'll do that from here

mrakitin: image

prjemian:
Looks like I'll want a bluesky-pytest environment

prjemian:
Yup.  Over. A. Barrel.

mrakitin:
It seems to be passing now (slowly...). I think the diagnostics/debugging prints can be removed now.

prjemian:
all diagnostics removed and travis passes four of five workers (nightly fails but not for this code)

Looks ready to merge.

danielballan:
Thanks a bunch for pushing this upstream, @prjemian.

Issue #1268, opened 2019-11-25
always construct live plots in the same thread as the RunEngine event loop


jklynch:
At BMM during testing with bluesky 1.6.0rc2 an issue was seen with LivePlots that looks like this:
```
QObject::startTimer: Timers can only be used with threads started with QThread
```

## Possible Solution
Construct all plots in the RunEngine event loop.

## Steps to Reproduce (for bugs)

1. install 2019C3.1.1rc3 environment:
```
xf06bm@xf06bm-ws1 [~] > CONDARC= conda create -p /tmp/new-stuff collection=2019C3.1.1rc3 -c https://repo.anaconda.com/pkgs/main -c https://conda.anaconda.org/nsls2forge --no-channel-priority --override-channels
```
2. Run bsui
3. RE(linescan(xafs_y,'i0', -2, 2, 5, force=True))

danielballan:
Closed by #1285

PR #1273, opened 2019-12-05
Construct Qt objects in the right thread

jklynch:
This PR addresses #1268.


danielballan:
Note to selves: We thought of a way to fix bluesky--Qt integration with a lighter touch, but this seems like it may be the best long-term approach, so let's keep this branch around for after the 1.6.0 final release.

danielballan:
Refacoring on `bluesky_mpl` when it's ready is still of interest. The immediate problem was solved by #1285. This branch should be kept around though.

PR #1276, opened 2019-12-19
FIX: stop was not correctly exiting the plan early

tacaswell:
RE.stop was not actually aborting the plans early when coming out of
the paused state.

Use the same logic used in abort and halt

Added a new test the verifies the correct exception type is raised out
of the paused point.


danielballan:
This looks like it could be an important test failure: https://travis-ci.org/bluesky/bluesky/jobs/627443576#L2149

tacaswell:
Restarted the mac jobs, those all looked like resource-starved race conditions.

PR #1280, opened 2020-01-15
CI: disable OSX testing on travis

tacaswell:
The workers we are getting on travis seem to be significantly throttled and
we are getting persistent failures due to some of the timing tests.

Also add linux py38

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,8 @@ matrix:
       env: PUBLISH_DOCS=1
     - os: linux
       python: 3.7
+    - os: linux
+      python: 3.8
     #     - os: osx
     #       language: generic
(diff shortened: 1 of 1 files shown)


danielballan:
I wish it weren’t so, but I don’t see another way forward here.

tacaswell:
The timings on the tests are a bit fuzzy to account for fluctuations in host performance, I suspect that we can tweak the tests to be more resilient to this variation, but I would put that low on the priority queue.

I am a bit worried about blindly making the timings looser as that could push us to getting false-positives (the tests pass when they should fail) rather than false negatives (the tests fail when they should pass).  False positives are a significatly worse failure mode as we don't know we are getting them.

PR #1297, opened 2020-02-14
FIX: setting run name for nested runs (`set_run_name_decorator')

dmgav:
This PR includes fix for the existing issue with management of run names in multirun plans. If the plan includes nested runs, the implementation of ``set_run_name_wrapper`` would always overwrite the plan name in the message (``Msg.run``) so that the message returned by the plan would always contain the name assigned to the outermost run. The modified wrapper overwrites the run name only if it has the default value. If the name is modified in the inner run, it passes through the wrapper of the outer run unmodified.

## Description

The default value for the run name is a fixed reference to an instance of ``object()``. The reference remains constant for the whole instance of the run engine. If the name is assigned to the run using ``set_run_name_wrapper`` or ``set_run_name_decorator``, it always has ``str`` type. The modified wrapper checks the type of the run name type (``Msg.run``) and replaces it with the new value only if the type is not ``str``. 

Additional changes:

-- ``define_run_wrapper`` and ``define_run_decorator`` were renamed into ``set_run_name_wrapper`` and ``set_run_name_decorator``.

-- the type of the new run name, passed to the wrapper as a parameter, is verified. If the type is not ``str``, ``ValueError`` is raised. This prevents users for setting run names to any values other than strings.

## Motivation and Context

The issue was discovered while testing the new features, which allows interlaced (and nested) runs in a single plan.

## How Has This Been Tested?

The changes are covered by tests. New tests: ``test_multirun_smoke_nested`` and ``test_multirun_run_name``. The existing tests were slightly modified for consistency.

### DIFF
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -1331,4 +1331,4 @@ def set_run_id_wrapper(plan, run):
 
     return (yield from msg_mutator(plan, _set_run_id))
 
-set_run_name_decorator = make_decorator(set_run_id_wrapper)
+set_run_id_decorator = make_decorator(set_run_id_wrapper)
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -1319,7 +1319,9 @@ def define_run_wrapper(plan, run):
         The run to set on each Msg
     """
     def _set_run_id(msg):
-        return msg._replace(run=run)
+        if not isinstance(msg.run, str):
+            msg = msg._replace(run=run)
+        return msg
 
     return (yield from msg_mutator(plan, _set_run_id))
 
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -1307,7 +1307,7 @@ class SupplementalData:
         return (yield from plan)
 
 
-def set_run_name_wrapper(plan, run):
+def set_run_id_wrapper(plan, run):
     """
     Add a run id to each message in wrapped plan
 
@@ -1328,4 +1328,4 @@ def set_run_name_wrapper(plan, run):
 
     return (yield from msg_mutator(plan, _set_run_id))
 
-set_run_name_decorator = make_decorator(set_run_name_wrapper)
+set_run_name_decorator = make_decorator(set_run_id_wrapper)
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -1307,7 +1307,7 @@ class SupplementalData:
         return (yield from plan)
 
 
-def set_run_name_wrapper(plan, run_name):
+def set_run_name_wrapper(plan, run):
     """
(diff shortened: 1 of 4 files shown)


bluesky/preprocessors.py
```diff
@@ -1307,20 +1307,25 @@ def __call__(self, plan):
...
-        The run to set on each Msg
+    run : str
+        The run name to set on each Msg
     """
+
+    if not isinstance(run, str):
+        raise ValueError(f"run name must be a string: passed value {run} is of type {type(run)}")
     def _set_run_id(msg):
-        return msg._replace(run=run)
+        if not isinstance(msg.run, str):
```
tacaswell: I would prefer that this be `if msg.run is None`.  It is _marginally_ more performant and means things will still work if the user provides some none-string input.  I think that, for things to work, we only really need this to be hashable.

The `object()` singleton is only droped in once the RE gets ahold of this.  On a bit more reflection, I think the logic in `_extract_run_key` is not needed.  I think it is a hold over from an earlier theory when we considered putting the run id in the kwargs.

dmgav: Yes, I think it can be done this way as well. In this case ``_extract_run_key`` should simply return ``msg.run``, or be completely removed. Then user provided run name should be only rejected if it is ``None``. Another question if the right name should be 'run name' or 'run ID'. The latter may be easily mixed with scan ID.

dmgav: Removing ``_extract_run_key`` completely may make it difficult to change the behavior in the future.

dmgav: I assume that if ``run_id`` may have any type but ``None``, the wrapper/decorator should be renamed into ``set_run_id``. The user documentation should recommend to use strings in manually created plans for better readability.

dmgav:
The changes to the PR include:
-- eliminating ``_extract_run_key`` (replicated commits from https://github.com/bluesky/bluesky/pull/1298). Now the default values for ``Msg.run`` is ``None`` and user defined Run ID may be of any type.

-- Renamed ``set_run_name_wrapper/decorator`` to ``set_run_id_wrapper/decorator``. (Any other suggestions about the names?)

-- Modified the unit tests to conform with the changes.

tacaswell:
The py37 failures looks like cython / cextension snafus.

tacaswell:
Looks like https://github.com/scikit-image/scikit-image/issues/4451

danielballan:
The term “Run ID” made me assume we were talking about RunStart uid. Since runs already have two ids (uid and scan_id) maybe it would be better to avoid that term. “Run name” doesn’t seem too bad to me; it suggests that it is intended to be human-settable and human-readable, which the uid is not.

danielballan:
“Run key” is also an option.

dmgav:
I will try changing it to `run_key`

PR #1299, opened 2020-02-18
ENH: add message to query the type of the RE

tacaswell:
Adds a new message 'RE_class' which defaults to being on the RE to
query what the class of the RE running the plan is.

This is useful in at least two cases:
 - identify if we are being listified
 - determine if we are running in a RE-subclass

Added a test.

This can be used to reliably determine if we are in "simulation mode"
or if the RunEngine has been sub-classed etc.


tacaswell:
The last commit of #1297 will fix the py37 failure.

PR #1300, opened 2020-02-18
Using existing callbacks with nested multirun plans

dmgav:
The change in the PR allow to use existing callbacks, including ``BestEffortCallback``, `LivePlot` and `LiveTable` with plans that contain nested runs. Following is the example of such scan, which starts and completes separate run `sim_plan_inner` from `sim_plan`: 
```
@bpp.set_run_key_decorator("run_inner")
@bpp.run_decorator(md={"run_key": "run_1"})
def sim_plan_inner():
    # ... some commands, e.g. use simulated 'hw.motor1', 'hw.motor2' and 'hw.det2'

@bpp.set_run_key_decorator("run_outer")
@bpp.run_decorator(md={"run_key": "run_1"})
def sim_plan(npts):
    # ... some commands, e.g. use simulated 'hw.motor', 'hw.det'
    yield from sim_plan_inner()
    # ... some commands, continue 'sim_plan'
```

Regular subscription to RunEngine can not be used with nested plans, since it provides no mechanism for routing data to different callbacks or different instances of the same callback based on ID of the run producing the data. Instead, the callbacks should be subscribed to the RunEngine via RunRouter (example for the plan above):
```
from event_model import RunRouter

def factory(name, doc):
    cb_list = []
    if doc["run_key"] == "run_1":
        # Pick a set of callbacks based on metadata in the start document
        cb_list.append(LiveTable([hw.motor, hw.det]))
        cb_list.append(LivePlot('det', x='motor'))
    elif doc["run_key"] == "run_2":
        cb_list.append(LiveTable([hw.motor1, hw.motor2, hw.det2]))
    for cb in cb_list:
        cb(name, doc)
    return cb_list, []

RE = RunEngine({})
rr = RunRouter([factory])
RE.subscribe(rr)
```

## Description

The changes include:

-- `CallbackBase` is now derived from `event_model.DocumentRouter`, which provides default implementations for the `event_page` and `datum_page` methods.

-- 'QT teleporter' is initialized from within the `RunEngine`, not during callback instatiation.

-- Blank separator lines are added before and after tables in `BestEffortCallback` and `LiveTable` for better visual presentation of nested tables. Additional parameter `separator_lines` is added `LiveTable` initalization arguments, that allows to disable additional separator lines. It is used when `LiveTable` is called from `BestEffortCallback` to avoid insertion of unnecessary blank lines.

-- Additional tests are added.
## Motivation and Context

PR addresses a problem of visualization of data from nested multirun scans.

## How Has This Been Tested?

Additional unit tests are included. The operation was also manually tested to check quality of visual presentation of output data.

### DIFF
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -400,10 +400,6 @@ class BestEffortCallback(QtAwareCallback):
             if peak_stats is not None:
                 peak_stats('event', doc)
 
-    #def event_page(self, doc):
-    #    for d in unpack_event_page(doc):
-    #        self.event(d)
-
     def stop(self, doc):
         if self._table is not None:
             self._table('stop', doc)
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -16,6 +16,7 @@ import threading
 import time
 from warnings import warn
 import weakref
+from event_model import unpack_event_page
 
 from .core import LiveTable, make_class_safe
 from .mpl_plotting import LivePlot, LiveGrid, LiveScatter, QtAwareCallback
@@ -398,6 +399,10 @@ class BestEffortCallback(QtAwareCallback):
             peak_stats = self._peak_stats.get(doc['descriptor'], {}).get(y_key)
             if peak_stats is not None:
                 peak_stats('event', doc)
+
+    def event_page(self, doc):
+        for d in unpack_event_page(doc):
+            self.event(d)
 
     def stop(self, doc):
         if self._table is not None:
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -143,7 +143,7 @@ class BestEffortCallback(QtAwareCallback):
         # Print heading.
         tt = datetime.fromtimestamp(self._start_doc['time']).utctimetuple()
         if self._heading_enabled:
-            print("Transient Scan ID: {0}     Time: {1}".format(
+            print("\n\nTransient Scan ID: {0}     Time: {1}".format(
                 self._start_doc.get('scan_id', ''),
                 time.strftime("%Y-%m-%d %H:%M:%S", tt)))
             print("Persistent Unique Scan ID: '{0}'".format(
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -400,9 +400,9 @@ class BestEffortCallback(QtAwareCallback):
             if peak_stats is not None:
                 peak_stats('event', doc)
 
-    def event_page(self, doc):
-        for d in unpack_event_page(doc):
-            self.event(d)
+    #def event_page(self, doc):
+    #    for d in unpack_event_page(doc):
+    #        self.event(d)
 
     def stop(self, doc):
         if self._table is not None:
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -356,7 +356,7 @@ class BestEffortCallback(QtAwareCallback):
         if stream_name == self.dim_stream:
             if self._table_enabled:
                 # plot everything, independent or dependent variables
-                self._table = LiveTable(list(self.all_dim_fields) + columns)
+                self._table = LiveTable(list(self.all_dim_fields) + columns, separator_lines=False)
                 self._table('start', self._start_doc)
                 self._table('descriptor', doc)
 
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -14,7 +14,6 @@ from contextlib import ExitStack
 import threading
 import weakref
 from .bundlers import RunBundler
-from .utils import initialize_backend
 
 import concurrent
 
@@ -195,12 +194,17 @@ class RunEngine:
         Expected signature: f(md)
         Expected return: updated scan_id value
 
-    during_task : callable, optional
-        Function to be run to block the main thread during `RE.__call__`
+    during_task : reference to an object of class DuringTask, optional
+        Class methods: ``initialize()`` to be run in ``RE.__init__``
+        to initialize the backend; ``block()`` to be to be run to block
+        the main thread during `RE.__call__`
 
-        The required signature is ::
+        The required signatures for the class methods ::
 
-              def blocking_func(ev: Threading.Event) -> None:
+              def initialize() -> None:
+                   "Returns once initialization code is completed"
+
+              def block(ev: Threading.Event) -> None:
                   "Returns when ev is set"
 
         The default value handles the cases of:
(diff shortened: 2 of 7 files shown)


bluesky/run_engine.py
```diff
@@ -341,6 +342,8 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
         self.record_interruptions = False
         self.pause_msg = PAUSE_MSG
```
danielballan: I hope we can find a way to solve this without introducing this coupling to matplotlib and a `sys.modules` check deeply into RunEngine. Would it work to do it in the __init__ of LivePlot (and similar)?

bluesky/callbacks/best_effort.py
```diff
@@ -399,6 +400,10 @@ def event(self, doc):
             if peak_stats is not None:
                 peak_stats('event', doc)
 
+    #def event_page(self, doc):
```
danielballan: Might as well delete these.

danielballan:
Other than the structural concern noted in line, the rest of the implementation looks solid. Thanks for tackling this.

bluesky/run_engine.py
tacaswell: I would feel better if we moved this to the code that sets up the background thread ?  It would still be run on init, but it would at least put all of the funny stuff together?

dmgav: By the code that sets up the background thread you mean `_ensure_event_loop_running` function?

Actually, if all Matplotlib/Qt related code should be moved to `default_during_task`, the function could be changed to a callable class which will have the methods `initialize` (to initializes the teleporter if the backend is Qt) and `__call__`  (the same code as current implementation of `default_during_task`). The method 'initalize' can be called from `RE.__init__()`.

danielballan: I had that same thought, in fact. If this has to be a two-step process for Qt, it might be a two-step process for other implementations as well. Establishing a two-part interface for it seems right.

danielballan: To get into names/details, maybe expect an object with methods `initialize` and `block`. We use `__call__` a lot of places, but we do that partly for brevity in user code and partly so that plain functions can be used in some cases (e.g. for callbacks). Neither of those reasons apply here, so an explicit method like `block` might be more clear.

bluesky/utils.py
```diff
@@ -1336,148 +1336,175 @@ def my_name(obj):
...
-                def exit_loop():
-                    blocking_event.wait()
-                    # If the above wait ends quickly, we need to avoid the race
-                    # condition where this thread might try to exit the qApp
-                    # before it even starts.  Therefore, we use QTimer, below,
-                    # which will not start running until the qApp event loop is
-                    # running.
-                    event_loop.exit()
+    def block(self, blocking_event):
+        pass
```
tacaswell:
```suggestion
        blocking_event.wait()
```

We should make this a usable implementation.

dmgav:
The latest changes include switching from callback function `during_task` to a class `DuringTask` implementing methods `initialize` and `block`. The class `DefaultDuringTask` inherits from `DuringTask`. The code in `initialize` method is initializing the Qt teleporter (if backend is Qt). The code from `default_during_task()` is moved to `DefaultDuringTask.block()'.

tacaswell:
I think that there is still some subtle ordering issues that can crop up:

```python
from bluesky import RunEngine
RE = RunEngine()  # <- does initialize the teleporter

# break 1
import matplotlib
bec = BestEffortCallback()

RE.subscribe(bec)
# break 2

RE(plan(...))
```

Will still lead the possible race conditions on the first scan.  This code is a bit contrived, but in the ipython profile you could imagine having the first and second blocks being in different files.  On the other hand, I think it is reasonable to expect that whoever sets those up to do things correctly.

Another option would be to call both initialize and block on every `RE.__call__` incantation (on right before creating the task, the other right after).  That way we are sure to offer the object a chance to run set up code on the main thread before we start the plan and a chance to do whatever it wants while it waits for the plan to finish.

dmgav:
@tacaswell Actually the example of the plan will still work fine, because `bec = BestEffortCallback()` will still call `_get_teleporter()`, which will try to initialize the teleporter. This is how initialization was always done. My first temporary solution was to create a 'dummy' instance of BEC before the plan is called. 

If subscription is done via `RunRouter`, BEC instances are created in the thread which runs the event loop, so initialization call in the BEC constructor doesn't work. In this case, if Matplotlib is not imported after RE is created, the plan fails. But it fails consistently on the development stage until the problem is fixed, whereas initialization in the `__call__` method may cause intermittent issues.

danielballan:
I think @dmgav is right. I don't see way to absolutely guarantee that `initialize()` is called after mpl is imported but before it is used. I can see an argument for making every effort not to fail and I can also see an argument for failing early/repeatably with a clear error message.

tacaswell:
Right, because I can always bury the mpl import inside of a callback inside of a router.

tacaswell:
If we are only going to call `initialize` once, I'm not convinced that we want to have this two part API.  We are already breaking it so we can drop `bluesky.utils.default_during_task`, change the default to `None` 

To be more concrete https://github.com/dmgav/bluesky/compare/bec_fix...tacaswell:bec_A?expand=1 or https://github.com/dmgav/bluesky/compare/bec_fix...tacaswell:bec_B?expand=1

danielballan:
I agree that if we are only calling it once we might as well call it when the default is set (at RE init time) and, for any non-default task, leave it up to the user to do any required initialization.

dmgav:
I changed the implementation according to 'bec_A' by @tacaswell.

bluesky/utils.py
```diff
@@ -1336,148 +1336,175 @@ def my_name(obj):
...
+                    # https://stackoverflow.com/a/4939113/380231
+                    kick_timer = QtCore.QTimer()
+                    kick_timer.timeout.connect(null)
+                    kick_timer.start(50)
+
+                    cleanup = kick_timer.stop
+
+                # we also need to make sure that the qApp never sees
+                # exceptions raised by python inside of a c++ callback (as
+                # it will segfault its self because due to the way the
```
mrakitin:
```suggestion
                # it will segfault itself because due to the way the
```

bluesky/utils.py
```diff
@@ -1336,148 +1336,175 @@ def my_name(obj):
...
+                # code is called there is no clear way to propgate that
```
mrakitin:
```suggestion
                # code is called there is no clear way to propagate that
```

bluesky/utils.py
```diff
@@ -1336,148 +1336,175 @@ def my_name(obj):
...
-                    except OSError as inst:
-                        print('failed to read wakeup fd: %s\n' % inst)
 
-                    wakeupsn.setEnabled(True)
+class DefaultDuringTask(DuringTask):
 
-                wakeupsn.activated.connect(handleWakeup)
+    def __init__(self):
+        """
+        Initialize backend. Currently on Qt backend is supported. The function
```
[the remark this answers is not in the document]
clintonroy: Currently only the Qt backend is supported, at a guess.

dmgav:
@mrakitin Thank you for the suggestions. All typos are fixed now.

mrakitin:
Done, thanks Dmitri!

PR #1301, opened 2020-02-18
ENH: add a message to reset the user position on a motor

tacaswell:
This came up at JPLS at NSLS-II.  Due to their geometry and the fact
that their sample is fluid and evaporates, they need to frequently
reset the user position.

attn @prjemian

### DIFF
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -1084,3 +1084,43 @@ def repeat(plan, num=1, delay=None):
                     yield Msg('sleep', None, d)
 
     return (yield from repeated_plan())
+
+
+def reset_user_position(obj, value, *, verbose=False):
+    """
+    Repeat a plan num times with delay and checkpoint between each repeat.
+
+    This is different from ``repeater`` and ``caching_repeater`` in that it
+    adds ``checkpoint`` and optionally ``sleep`` messages if delay is provided.
+    This is intended for users who need the structure of ``count`` but do not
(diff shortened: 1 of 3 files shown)


tacaswell:
This is currently registered onto the RE in the JPLS profile, I wanted to hoist this up here to make sure it did not get lost.  I'm on board with holding off merging this for a while.

The innards of `set_current_position` need to be atomic from the POV of the RunEngine so I don't think we can make it purely a plan.

Maybe we need to add a module-level registry of messages (currently the registry lives on the RE instance so if you want to be a third-party package that sources messages you need to get a hold of the RE that the user is using)?  Maybe entry points?

prjemian:
This is a three line plan in apstools (sending link later).  These steps

Put motor record into SET
write new position to VAL
put into USE

The signals are already past of EpicsMotor

What additional complexity is needed?

tacaswell:
What happens if the plan gets interrupted after it is put in SET mode but before it is put back into USE mode?

danielballan:
@prjemian: If interrupted mid-stream, human intervention is required. Automatic recover is not possible, too likely to result in undefined / unintended behavior. That would just be an unavoidable risk of using this `Msg` type in a plan.

tacaswell:
The point of adding this Msg type is to make the possibility of being left in a super weird state as low as possible.

stan-dot:
this is covered here now  https://github.com/bluesky/bluesky/pull/1778 , closing

PR #1302, opened 2020-02-20
FIX: regression in validating the per_step signature for 1d case

### LINKED ISSUE #1295: Old per_step scans now invalid
`per_step` scans have closely regulated signatures. The regulator changed.

## Expected Behavior

per_step scans that worked in `bluesky=1.5.5` should work in `bluesky=1.6.0` or have clear instructions to upgrade

## Current Behavior

The signatures of `one_nd_step` and `one_1d_step` were changed, but these signatures are being used to check if a user's `per_step` function is valid. https://github.com/bluesky/bluesky/blob/feb8e7951276be3fe7714035faf8756a5fb1f4f1/bluesky/plans.py#L922-L944 

Note how the error message states the old signature, but the code requires the new one.

## Possible Solution

Either the error message should be changed to reflect the new signature, or the old signature should still be valid.

1. Run any scan with a previously-working per_step arg

## Context

I want to upgrade to `bluesky=1.6.0` when available

## Your Environment

`bluesky=1.6.0rc4` installed in dev mode

tacaswell:
closes #1295

Fixes a regression in how we validate the per_step functions

Added tests of both passing and failing signatures

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -1,6 +1,6 @@
 import sys
 import inspect
-from itertools import chain
+from itertools import chain, zip_longest
 from functools import partial
 
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


danielballan:
New tests fail.

tacaswell:
```

E           AssertionError: Pattern
'per_step must be a callable with the signature <Signature (detectors, step, pos_cache)> or <Signature (detectors, motor, step)>.' not found in
'per_step must be a callable with the signature <Signature (detectors, step, pos_cache)> or <Signature (detectors, motor, step)>.'
```

sigh.

tacaswell:
Lets see if that works any better.

tacaswell:
The missing lines in the patch are the never-run bodies of the dead functions...

klauer:
So for this to work, is it required for the parameters in our per_step functions to have the exact names `['detectors', 'motor', 'step']`?

danielballan:
Yes, but that has always been true.

klauer:
Interesting - I (obviously) didn't recall that...

danielballan:
We checked for exactly signature equality.

https://github.com/bluesky/bluesky/blob/75bb50d4116560851bafe663d2a227348808701e/bluesky/plans.py#L717-L721

This is more relaxed because it allows for `per_step` to have additional optional parameters, which we will ignore.

danielballan:
Looks like we introduced the strict checking in v1.2.0 when `one_nd_step` and `one_1d_step` were unified. Before 1.2.0 there was no signature check.

klauer:
@danielballan I believe you 😄 - I just didn't remember that.

tacaswell:
I was also surprised, but went with preserving as much of the same functionality as before.

This may still be a bit _too_ permissive, but you will have to be obtuse to run into those issues.

PR #1309, opened 2020-02-25
Fix for CI in 1.5.x branch

dmgav:
Fixed CI for v1.5.x branch:
- removed `test_reg_reader` test (uses deprecated feature of Databroker);
- fixed `test_table` test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,6 @@ matrix:
       python: 3.6
     - os: linux
       python: 3.7
-    - os: linux
-      python: 3.8
     - python: nightly
 
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,10 @@ matrix:
   include:
     - python: 3.6
       env: BUILD_DOCS=1
+    - os: linux
+      python: 3.7
+    - os: linux
+      python: 3.8
     - python: nightly
   allow_failures:
     - python: nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ install:
   - export GIT_FULL_HASH=`git rev-parse HEAD`
   - pip install --upgrade pip
   - pip install --upgrade cython
+  - pip install --upgrade numpy
   - pip install --quiet --upgrade --global-option='--without-libyaml' pyyaml
   - pip install -r requirements.txt
   - pip install -r dev-requirements.txt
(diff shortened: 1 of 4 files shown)


danielballan:
Cherry-picked @tacaswell's commit from `master` that fixed the 3.7 build.

danielballan:
That helped, but all the 0MQ tests are failing.

dmgav:
The branch used to be tested only for Python 3.6. All tests are passing locally with Python 3.7.

danielballan:
I can reproduce the Python 3.7 `test_zmq.py` failures locally.

We pinned pyzmq to 16.0.4 in 552a1b0e8 when 17.0.0 broke our CI. (We never discovered what the exact problems was. Then we unpinned it in cfc9c2a59 "to fix py37" as it says in the commit message. I have pushed a new commit to this branch to unpin that.

danielballan:
OK, with that fix 3.7 passes. 3.8 has a slew of failures. Any objection to removing 3.8 from the 1.5.x test matrix. (The `master` branch tests on 3.8, so we'll support it for 1.6.0+.)

dmgav:
I removed tests with Python 3.8. Hopefully everyone will quickly migrate to v1.6.0.

.travis.yml
```diff
@@ -12,12 +12,10 @@ cache:
 matrix:
   fast_finish: true
   include:
-    - python: 3.6
-      env: BUILD_DOCS=1
```
mrakitin: I am concerned by the removal of this line to force it to build docs. Is it intended?

mrakitin: The commit message for https://github.com/bluesky/bluesky/pull/1309/commits/f526e3be4910acfc497be298456380701db92296 explains why it was done. It's OK to not build the docs for the maintenance branch.

danielballan:
OK. Supporting the current latest stable Python would normally be standard practice, but given the circumstances I think we should carry on and address 3.8 only if there is a need.

dmgav:
I removed the flag that enables building docs because it looks like it could overwrite docs on `blueskyproject.io`. If I am wrong and this is not the case, then I can enable document building.

danielballan:
Seems correct to me. Until recently we haven't had much need for back-port releases so, speaking for myself, I'm still sorting out how to do it properly.

PR #1310, opened 2020-02-26
Support for Python 3.8

dmgav:
This PR is expected to make the branch compatible with Python 3.8. The changes include:
- removal of use of deprecated feature (exactly the same change was applied in v1.6.0);
- replaced all occurrances of `@asyncio.coroutine` by `async def` (optional change, does not influence the build)

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -851,8 +851,7 @@ class RunEngine:
         self._suspenders.add(suspender)
         suspender.install(self)
 
-    @asyncio.coroutine
-    def _install_suspender(self, msg):
+    async def _install_suspender(self, msg):
         """
         See :meth: `RunEngine.install_suspender`
 
@@ -880,8 +879,7 @@ class RunEngine:
             suspender.remove()
         self._suspenders.discard(suspender)
 
-    @asyncio.coroutine
-    def _remove_suspender(self, msg):
+    async def _remove_suspender(self, msg):
         """
         See :meth: `RunEngine.remove_suspender`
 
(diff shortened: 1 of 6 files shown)


danielballan:
Nice. Thank you.

bluesky/run_engine.py
```diff
@@ -2399,8 +2367,7 @@ def subscribe(self, func, name='all'):
             self._token_mapping[public_token] = private_tokens
             return public_token
 
-        if name not in DocumentNames:
-            name = DocumentNames[name]
+        name = DocumentNames[name]
```
mrakitin: This condition didn't make sense before. Maybe we need to use the following?
```py
if name in DocumentNames:
    name = DocumentNames[name]
```

mrakitin:
We should also fix the `.travis.yml` file as we did for 1.6.x. Here are the warnings from TravisCI:
image

dmgav:
Still fails one test, which passes locally.

bluesky/run_engine.py
dmgav: This feature was deprecated before and it is removed from Python 3.8. The condition `name in DocumentNames` raises exception with the following message:
```
unsupported operand type(s) for 'in': 'str' and 'EnumMeta'
```
The condition was removed from v1.6.0 completely. I guess the purpose of this condition is to check if `name` is already of `DocumentName` enum type and perform the conversion if not. But the expected values of the 'name' parameter of the function are from the fixed list of strings, so we can perform conversion without verifying if it is a string. The exception will be raised if there is a problem.

dmgav:
The issue was compatibility with PyTest 5.2.1. The test used to pass successfully on the later PyTest 5.3.5. But it seems that using `match` parameter of `pytest.raises` to check the error message is a better way, since it accepts regular expressions.

PR #1313, opened 2020-03-07
ENH 'rd'

tacaswell:
## Description

This is the mirror of `mv` and for many devices we would expect
the value from `rd` to go into `mv` / `abs_set` correctly.

This idea was suggested by Claudio Mazzoli (@cmazzoli ) to Andi Barbour (@ambarb) and then
explained to me by Andi. 

I would like to try and sneak this in under the wire for 1.6 because due to investigating https://github.com/bluesky/ophyd/issues/814 it is clear there is a strong demand for this.  I think this would cover about a third of the usages of `obj.value (a third are covered by `bps.mvr`, and the last third by `@reset_position_decorator`).

## Motivation and Context

Frequently in plan users want to get "the value" from a Device.  The easy to get thing is the reading, which is a dictionary, which you need to know the right key to get out, but sometimes it comes back as None so you have to be careful.

```python
ret = ((yield from read(obj)) or {}).get('det', {}).get('value', default)

```
is probably the shortest way to write that logic _if_ you know the key already.  This lets you spell that as 

```python
ret = yield from bps.rd(obj, default_value=default)
```

which does not require the plan author to know the name and raises nice errors (even when only `list`ed) if we can not guess what key to use.

## How Has This Been Tested?

Added a number of tests to the test suite.

### DIFF
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -334,7 +334,7 @@ def rd(obj, *, default_value=0):
         The "single" value of the device
 
     """
-    hints = obj.hints.get("fields", [])
+    hints = getattr(obj, 'hints', {}).get("fields", [])
     if len(hints) > 1:
         msg = (
             f"Your object {obj} ({obj.name}.{obj.dotted_name}) "
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -337,23 +337,21 @@ def rd(obj, *, default_value=0):
     hints = getattr(obj, 'hints', {}).get("fields", [])
     if len(hints) > 1:
         msg = (
-            f"Your object {obj} ({obj.name}.{obj.dotted_name}) "
-            + f"has {len(hints)} items hinted ({hints}).  We "
-            + "do not know how to pick out a single value.  Please "
-            "adjust the hinting by setting the kind of the components of "
-            "this device or by rd ing one of it's components"
+            f"Your object {obj} ({obj.name}.{getattr(obj, 'dotted_name', '')}) "
+            f"has {len(hints)} items hinted ({hints}).  We do not know how to "
+            "pick out a single value.  Please adjust the hinting by setting the "
+            "kind of the components of this device or by rd ing one of it's components"
         )
         raise ValueError(msg)
     elif len(hints) == 0:
         if hasattr(obj, "read_attrs"):
             if len(obj.read_attrs) != 1:
                 msg = (
-                    f"Your object {obj} ({obj.name}.{obj.dotted_name}) "
-                    + f"and has {len(obj.read_attrs)} read attrs.  We "
-                    "do not know how to pick out a single value.  Please "
-                    "adjust the hinting/read_attrs by setting the kind of "
-                    "the components of this device or by rd ing one of it's "
-                    "components"
+                    f"Your object {obj} ({obj.name}.{getattr(obj, 'dotted_name', '')}) "
+                    f"and has {len(obj.read_attrs)} read attrs.  We do not know how to "
+                    "pick out a single value.  Please adjust the hinting/read_attrs by "
+                    "setting the kind of the components of this device or by rd ing one "
+                    "of it's components"
                 )
 
                 raise ValueError(msg)
@@ -378,12 +376,10 @@ def rd(obj, *, default_value=0):
         (data,) = ret.values()
     except ValueError as er:
         msg = (
-            f"Your object {obj} ({obj.name}.{obj.dotted_name}) "
-            f"and has {len(ret)} read values.  We "
-            "do not know how to pick out a single value.  Please "
-            "adjust the hinting/read_attrs by setting the kind of "
-            "the components of this device or by rd ing one of it's "
-            "components"
+            f"Your object {obj} ({obj.name}.{getattr(obj, 'dotted_name', '')}) "
+            f"and has {len(ret)} read values.  We do not know how to pick out a "
+            "single value.  Please adjust the hinting/read_attrs by setting the "
+            "kind of the components of this device or by rd ing one of it's components"
         )
 
         raise ValueError(msg) from er
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -15,8 +15,13 @@ except ImportError:
     from toolz import partition
 
 
-from .utils import (separate_devices, all_safe_rewind, Msg, ensure_generator,
-                    short_uid as _short_uid)
+from .utils import (
+    separate_devices,
+    all_safe_rewind,
+    Msg,
+    ensure_generator,
+    short_uid as _short_uid,
+)
 
 
 def create(name='primary'):
@@ -210,8 +215,12 @@ def rel_set(obj, *args, group=None, wait=False, **kwargs):
     :func:`bluesky.plan_stubs.wait`
     """
     from .preprocessors import relative_set_wrapper
-    return (yield from relative_set_wrapper(
-        abs_set(obj, *args, group=group, wait=wait, **kwargs)))
+
+    return (
+        yield from relative_set_wrapper(
+            abs_set(obj, *args, group=group, wait=wait, **kwargs)
+        )
+    )
 
 
 def mv(*args, group=None, **kwargs):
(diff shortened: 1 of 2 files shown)


danielballan:
I like it, and support sneaking it into 1.6.

I think it is not obvious enough that “bps.rd” does not add a reading to an Event the way that “bps.read” does. That important difference should he advertised more prominently in the docstring and perhaps made more obvious in the name; rd looks like consonant-free “read” and might be mistaken for an alias.

Maybe “get_val”?

bluesky/plan_stubs.py
```diff
@@ -295,6 +302,82 @@ def inner_mvr():
...
+            f"Your object {obj} ({obj.name}.{obj.dotted_name}) "
+            + f"has {len(hints)} items hinted ({hints}).  We "
+            + "do not know how to pick out a single value.  Please "
+            "adjust the hinting by setting the kind of the components of "
+            "this device or by rd ing one of it's components"
+        )
+        raise ValueError(msg)
+    elif len(hints) == 0:
+        if hasattr(obj, "read_attrs"):
+            if len(obj.read_attrs) != 1:
```
danielballan: We have so far avoided referencing `read_attrs` in bluesky. It is not part of the ophyd—bluesky interface; it is an ophyd implementation detail from blueksy’s POV. Would it be alright to do this check by looking at the length of the reading? It is slightly better to fail before we do the reading, but maybe not worth adding this to the interface (even the optional interface).

bluesky/plan_stubs.py
```diff
@@ -295,6 +302,82 @@ def inner_mvr():
...
+        The value to return when not running in a "live" RunEngine (when
+        ret = yield from obj.read() returns None)
+
+    Returns
+    -------
+    val : Any or None
+        The "single" value of the device
+
+    """
+    hints = obj.hints.get("fields", [])
```
danielballan: We normally access hints via describe(). I think this might be the first time we consult the attribute directly. It is definitely cheaper, so I think there is a case for taking this approach and effectively adding it as an optional attribute in the ophyd—bluesky interface.

It is strictly optional though—I think we have tests that ensure this—so this should be wrapped.
```suggestion
    hints = getattr(obj, 'hints', {}).get("fields", [])
```

tacaswell:
Fair point regarding not leaking the API, but it is currently optional (and falls back to failing later), and I think balanced by two things: a) failing when the plan is `list`'d or `print_summary`'d b) making sure that the things in a) do not do any network/controls system work.

tacaswell:
is `dotted_name` also optional on the ophyd side?

bluesky/plan_stubs.py
tacaswell: https://github.com/bluesky/bluesky/pull/1313#issuecomment-596443877 <- answered there, resolving this to not split the conversation (are lose it to eventual comment folding).

bluesky/plan_stubs.py
```diff
@@ -295,6 +302,91 @@ def inner_mvr():
...
+
+    # handle the no hint 1 field case
+    try:
+        (data,) = ret.values()
+    except ValueError as er:
+        msg = (
+            f"Your object {obj} ({obj.name}.{getattr(obj, 'dotted_name', '')}) "
+            f"and has {len(ret)} read values.  We do not know how to pick out a "
+            "single value.  Please adjust the hinting/read_attrs by setting the "
+            "kind of the components of this device or by rd ing one of it's components"
```
danielballan:
```suggestion
            "kind of the components of this device or by rd ing one of its components"
```

bluesky/plan_stubs.py
```diff
@@ -295,6 +302,91 @@ def inner_mvr():
...
+        raise ValueError(msg)
+    elif len(hints) == 0:
+        if hasattr(obj, "read_attrs"):
+            if len(obj.read_attrs) != 1:
+                msg = (
+                    f"Your object {obj} ({obj.name}.{getattr(obj, 'dotted_name', '')}) "
+                    f"and has {len(obj.read_attrs)} read attrs.  We do not know how to "
+                    "pick out a single value.  Please adjust the hinting/read_attrs by "
+                    "setting the kind of the components of this device or by rd ing one "
+                    "of it's components"
```
danielballan:
```suggestion
                    "of its components"
```

danielballan:
We have not documented `dotted_name` as part of the Bluesky Interface, so I think we should treat it as optional for now. Maybe we should add it.

tacaswell:
I briefly thought about adding a "should this be triggered?" flag, however decided not to because if you have a device that needs to be triggered, it is likely too complex for the target use-case of this plan.

bluesky/plan_stubs.py
```diff
@@ -295,6 +302,91 @@ def inner_mvr():
...
+    The devices is not triggered and this plan does not create any Events
+
+    Parameters
+    ----------
+    obj : Device
+        The device to be read
+
+    default_value : Any
+        The value to return when not running in a "live" RunEngine (when
+        ret = yield from obj.read() returns None)
```
ZLLentz: I'm not sure I understand what `default_value` does here. I wasn't aware that `read` behaves differently outside of a run engine. I haven't been following `bluesky` very closely so I suspect I've missed something.

danielballan: Oh, don't worry, this isn't new at all. It's just that when you do an adaptive pattern like this:

```py
def plan(...):
    ...
    reading = yield from bluesky.plan_stubs.read(...)
    ...
```

and then iterative over it like `list(plan)` then `reading` is `None`. Executing it with the `RunEngine` (or hypothetically anything that calls `send` to send a value back) gives you an actual reading.

ZLLentz: I see. I don't quite understand why `reading` becomes `None` in this case instead of a `Msg` object but I do understand why a value is needed.

ZLLentz: Either way, the docstring should have the example from your comment instead of the line it currently has, since `yield from obj.read()` is probably some sort of error in all cases (yielding from a dictionary)

danielballan: Good catch on the docstring.

The `reading` would never be a `Msg` object. Going step by step: the `Msg('read', ...)` yielded by `bluesky.plan_stubs.read(...)` on the right side of `=` goes to the RE, is dispatched by the command lookup table to `RunEngine._read(msg)`, which calls `reading = obj.read()`, and then calls `plan.send(reading)` to send that `reading` back to the plan, where it lands on the left side of the `=`.

If the plan is consumed by something that does not `.send`, then the left side of the `=` just gets `None`.

This only comes up in the context of adaptive plans. (Many plans do not capture the result of `yield from`.) The `None` can be interpreted as, “Whatever is consuming this plan apparently isn’t actually getting readings from hardware. It might be a simulator or summarizer. We can’t actually do adaptive logic without readings, so here is `None` as a placeholder.” This behavior has been in place since early days.

Am I addressing the point of confusion? Apologies if I’m talking past you.

tacaswell: oops, that is definitely a typo in the docs, sorry.

ZLLentz: To illustrate my confusion, here is an IPython session I just ran using your example:
```
In [1]: import bluesky.plan_stubs                                               

In [2]: def plan(): 
   ...:     reading = yield from bluesky.plan_stubs.read(object) 
   ...:     return reading 
   ...:                                                                         

In [3]: list(plan())                                                            
Out[3]: [read: (<class 'object'>), (), {}]
```
What am I missing?

ZLLentz: I had a misunderstanding about how `yield from` and generators worked. See the source of confusion:
```
In [1]: import bluesky.plan_stubs                                               

In [2]: def plan(): 
   ...:     reading = yield from bluesky.plan_stubs.read(object) 
   ...:     print(reading) 
   ...:     yield reading 
   ...:                                                                         

In [3]: list(plan())                                                            
None
Out[3]: [read: (<class 'object'>), (), {}, None]
```

danielballan: Oh, that is confusing. One thing this tells as, as an aside, in that our less "special" `Msg.__repr__` is an improvement in situations like this. On bluesky `master` this looks like:

```
In [5]: def plan(): 
   ...:     reading = yield from bluesky.plan_stubs.read(object) 
   ...:     print('reading', reading) 
   ...:     yield reading 
   ...:                                                                                                                                                                                       

In [6]: list(plan())                                                                                                                                                                          
reading None
Out[6]: [Msg('read', obj=<class 'object'>, args=(), kwargs={}, run=None), None]
```

ambarb:
Thanks for getting this done so quickly.  Looking forward to showing users.

PR #1314, opened 2020-03-07
MNT: passing `loop=` to asyncio.Event is deprecated in py38

tacaswell:
This shims it out in our core code, there is still some usage in the
tests that will be trickier to remove.

This will become critical for the first facility that moves to py38 (it is _super_ noisy).

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -298,13 +298,25 @@ class RunEngine:
         self._th = _ensure_event_loop_running(loop)
         self._state_lock = threading.RLock()
         self._loop = loop
-
+        if sys.version_info < (3, 8):
+            self._loop_for_kwargs = self._loop
+        else:
+            self._loop_for_kwargs = None
         # When set, RunEngine.__call__ should stop blocking.
         self._blocking_event = threading.Event()
 
         # When cleared, RunEngine._run will pause until set.
-        self._run_permit = asyncio.Event(loop=loop)
-        self.loop.call_soon_threadsafe(self._run_permit.set)
+        self._run_permit = None
+
+        setup_event = threading.Event()
+
+        def setup_run_permit():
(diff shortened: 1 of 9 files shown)


danielballan:
Power-cycled to see if this fixes the Python 3.8 + ophyd `master` failures we have seen on bluesky `master`.

bluesky/run_engine.py
```diff
@@ -298,13 +298,25 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
         self._th = _ensure_event_loop_running(loop)
         self._state_lock = threading.RLock()
         self._loop = loop
-
+        if sys.version_info < (3, 8):
+            self._loop_for_kwargs = self._loop
```
klauer, line 302: Random thought - what if you made `partial`s for the 3 different versions of the functions taking `loop=`?

tacaswell: Interesting idea.  That would put many more copies of this version check in the code base though?

I think we should also push this version check back to py37 (as that is when https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop came in which is what makes explicitly passing the loop kwarg around unneeded).

klauer: I meant more along the lines of:

```python
self._asyncio_event = functools.partial(asyncio.Event, loop=(self._loop if sys.version_info < (3, 7) else None))
```
Perhaps it'd be more confusing 🤷
3.7 makes sense, then.

tacaswell: Ah, that makes sense.  Seems 6 of one half-dozen of the other.

I don't think we support changing the loop that the RE is attached to?

klauer: Supporting that seems like a recipe for disaster.

PR #1317, opened 2020-03-12
TST: compatibility with event-model v1.14.0

dmgav:
The PR is fixing minor CI issue related to changes in event-model v1.14.0.


dmgav:
I guess `codecov` coverage dropped, because I removed explicit call to the BEC instance.

PR #1320, opened 2020-03-13
Rework which version of ophyd to use in CI.

danielballan:
Formerly we *always* used ophyd `--pre`. Intead, use ophyd stable for
most builds, and use ophyd `master` for one build.

Justification:

* `master` provides an earlier warning than `--pre` about any changes to
  ophyd that break bluesky tests
* We should have been testing against the latest stable release of ophyd
  to ensure bluesky is not relying on unreleased changes to ophyd.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,9 @@ jobs:
       python: 3.6
       env: PUBLISH_DOCS=1
     - os: linux
+      python: 3.6
+      env: OPHYD_MASTER=true
+    - os: linux
       python: 3.7
     - os: linux
       python: 3.8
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -66,7 +69,11 @@ install:
   - pip install --quiet --upgrade --global-option='--without-libyaml' pyyaml
   - pip install -r requirements.txt
   - pip install -r requirements-dev.txt
-  - pip install -U ophyd --pre
+  # Install ophyd master branch if build settings indicate to do so.
+  - |
+    if [ "$OPHYD_MASTER" = true ] ; then
+      pip install git+https://github.com/bluesky/ophyd@master
```
tacaswell: Does this need a `-U`?

danielballan: Apparently not (given the evidence of CI) but it wouldn't hurt to add it for robustness against future changes and for readability.

gwbischof:
Looks good to me

Issue #1315, opened 2020-03-09
Update docs to address multi-run functionality.


danielballan:

danielballan:
Closed by #1321

PR #1319, opened 2020-03-13
Rewrite logging docs.

danielballan:
This borrows substantially from caproto, but it's different enough
that it cannot practically share source with it.

PR #1321, opened 2020-03-13
Documentation on multi-run plans

dmgav:
-- Documentation on multi-run plans.

-- A number of minor corrections to documentation for compatibility with Sphinx 2.4.4

-- Addition test `test_push_start_document` to improve coverage an potentially eliminate `codecov` failure.


danielballan:
Thanks for all the sphinx cleanup here.

To avoid code duplication, would it be helpful to use `literalinclude` to embed the example scripts in the narrative text? See https://nsls-ii.github.io/scientific-python-cookiecutter/writing-docs.html#embedded-scripts

We do this, for example, in https://raw.githubusercontent.com/caproto/caproto/master/doc/source/iocs.rst

doc/source/multi_run_plans.rst
```diff
@@ -0,0 +1,511 @@
+Multi-Run Plans
+===============
+
+This section is a brief tutorial on multi-run plans (introduced in Bluesky v1.6.0).
+A traditional single-run plan contains a set of instructions for running only one scan,
```
danielballan: We try to avoid using the word "scan" to mean "run" because:

* Some experiments aren't really scanning anything, and people in the past have been confused on that point.
* Sometimes a run is scanning multiple axis, such as `grid_scan`, so wording like this can be confusing.

"Run" is a pretty vague term, but we have never come up with a better one that sufficiently conveys how general it is. This language the [the event-model documentation on our Data Model](https://blueskyproject.io/event-model/data-model.html) may be useful:

> The scope of a “run” is left to the instrument team or individual scientist. It is quite generic: a set of Documents generated by following a given sequence of instructions. This might encompass a single short measurement (say, “a count”) or a multi-step procedure such as a raster scan. A run should represent a set of individual measurements that will be collected or processed together.

I think every instance of the word "scan" here and below should be changed to "run" or "plan", depending on the situation.

doc/source/multi_run_plans.rst
```diff
@@ -0,0 +1,511 @@
...
+    # Using 'bps.open_run()' and 'bps.close_run()' stabs to define a scan
+
+    import bluesky.plan_stubs as bps
+    from bluesky import RunEngine
+
+    RE = RunEngine({})
+
+    def sample_plan():
+        ...
+        bps.open_run(md={})  # 'md' - metadata to be added to the 'start' document
```
danielballan: missing `yield from` here and below

doc/source/multi_run_plans.rst
```diff
@@ -0,0 +1,511 @@
...
+
+    def sample_plan():
+        bps.open_run(md={})  # Starting the first run
+        < code that controls execution of the scan >
+        bpp.count(<some arguments>)  # Attempting to run a preassembled plan from an open run
+        bps.close_run()
+
+    RE(sample_plan())
+
+While it is unusual to call preassembled plans directly from within a user-defined scan, one may encounter
```
danielballan: Is this unusual? User-defined plans that glue together preassembled plans into longer routines seem common.

danielballan:
Closes #1315 

Merge conflicts came in through #1319 in `requirements-dev.txt`.

doc/source/multi_run_plans.rst
dmgav: Thanks. Sorry I missed the obvious.

doc/source/multi_run_plans.rst
dmgav: I guess my wording is incorrect. Preassembled plans are typically glued together in sequential manner: once one is finished, then the other one is started. They are not put one inside of another. Should I call 'user-defined scan' as 'user-defined run'?

doc/source/multi_run_plans.rst
dmgav: It seems like 'run' is a vague term when speaking about the structure of data acquisition system. But technically from the point of view of Bluesky it is a sequence of messages that starts with `open_run` and ends with `close_run`. The data acquired during run execution will be kept together and could be identified by scan ID and UID. A plan may contain no defined run (for example just move motors). Would that be correct?

I was thinking if I should use 'scan' or 'run' and decided to use 'scan', but I guess 'run' is more accurate word.

danielballan: Yeah, I agree that "run" is a vague term. I think it's the right choice for internal consistency with the object names in our libraries (`open_run`) and the rest of the documentation, but I agree it's not ideal. I sometimes use the word "scan" when speaking to certain audiences, but for a general audience, "run" is more precise.

doc/source/multi_run_plans.rst
danielballan: No problem. An easy mistake to make---hopefully https://github.com/bluesky/bluesky/issues/1239 will help make it less easy to make.

doc/source/multi_run_plans.rst
danielballan: Your new phrasing here seems clearer to me. :+1:

dmgav:
Yes, I think the PR is ready to merge.

PR #1322, opened 2020-03-13
Move ophyd master build to 3.8.

danielballan:
For consideration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,12 +27,12 @@ jobs:
       python: 3.6
       env: PUBLISH_DOCS=1
     - os: linux
-      python: 3.6
-      env: OPHYD_MASTER=true
-    - os: linux
       python: 3.7
     - os: linux
       python: 3.8
+    - os: linux
+      python: 3.8
+      env: OPHYD_MASTER=true
     #     - os: osx
     #       language: generic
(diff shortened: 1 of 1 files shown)


danielballan:
Decided against adding 3 more builds, but it would be better to test against the latest Python. Most likely to find issues.

dmgav:
May be the test with Ophyd master should be allowed to fail?

danielballan:
I'm slightly against allowing failures. Those can sometimes pass unnoticed.

I think we let Python nightly builds fail because we don't _control_ Python nightly, and we don't want temporary upstream problems to block merging bluesky PRs. But we do control ophyd `master`, and it may be useful to know right away if we introduce an incompatibility between ophyd and bluesky.

Thoughts?

dmgav:
Yes, I think it can be done this way. But then it may be a good practice to manually check if Bluesky CI is still passing before merging Ophid PRs.

danielballan:
Yes, I hope this would be last-resort alert. Ideally any changes to ophyd that could affect bluesky (especially changes to `ophyd.sim`) should be run against the bluesky test suite locally. Added a note in https://github.com/bluesky/ophyd/issues/822

PR #1326, opened 2020-03-17
Flake8 Compliance

dmgav:
The changes in this PR include the following:

- `flake8` is added to CI tests;

- the code is manually modified to comply with `flake8`.

## Description

`flake8` configuration is defined in `.flake8` file in the root of the Bluesky project. `flake8` is configured to allow the maximum line length of 115 characters. This is the maximum line length currently selected in scientific cookiecutter. `flake8` is also set to ignore the examples in `doc/source/examples`, since those examples are included in the document text and are formatted primarily to optimize visual perception. To comply with `flake8`, the examples would have to be modified to include additional blank lines, which is undesirable.

Changes to the code include:

- addition of missing imports and removing unused imports. Unused imports that were supposed to stay (in `__init__.py` files were marked with `noqa`);

- rearrangement of import order;

- removal of unused local variables;

- replacement of bare `except` with `except BaseException`;

- removal of `color_logs` from `__all__` in `bluesky/log.py`;

- manual changes to code formatting to comply with `flake8` requirements.

The most problematic files: `bluesky/tests/interactive/best_effort_cb.py` and `bluesky/tests/test_scientific.py`.

### DIFF
--- a/bluesky/callbacks/olog.py
+++ b/bluesky/callbacks/olog.py
@@ -158,7 +158,7 @@ class OlogCallback(CallbackBase):
         from pyOlog import SimpleOlogClient
         self.client = SimpleOlogClient()
         # Check at init time we are in an IPython session.
-        from IPython import get_ipython
+        from IPython import get_ipython  # noqa: F401
 
     def start(self, doc):
         from IPython import get_ipython
--- a/bluesky/callbacks/zmq.py
+++ b/bluesky/callbacks/zmq.py
@@ -146,7 +146,7 @@ class Proxy:
                 out_port = backend.bind_to_random_port("tcp://*")
             else:
                 backend.bind("tcp://*:%d" % out_port)
-        except:
+        except BaseException:
             # Clean up whichever components we have defined so far.
             try:
                 frontend.close()
--- a/bluesky/commandline/zmq_proxy.py
+++ b/bluesky/commandline/zmq_proxy.py
@@ -55,10 +55,12 @@ def main():
         logger.setLevel('INFO')
         if args.verbose > 2:
             logger.setLevel('DEBUG')
+        # Set daemon to kill all threads upon IPython exit
         threading.Thread(target=start_dispatcher,
                          args=('localhost', out_port, args.logfile),
-                         daemon=True).start()  # Set daemon to all ipython exit
-        #                                        kill all threads
+                         # Set daemon to all ipython exit kill all threads
+                         daemon=True).start()
+
     print("Connecting...")
     proxy = Proxy(in_port, out_port)
     print("Receiving on port %d; publishing to port %d." % (in_port, out_port))
--- a/bluesky/commandline/zmq_proxy.py
+++ b/bluesky/commandline/zmq_proxy.py
@@ -32,7 +32,7 @@ def start_dispatcher(host, port, logfile):
             logger.info("%s: %r", name, doc)
         else:
             logger.debug("%s: %r", name, doc)
-    dispatcher.subscribe(log_writer) # Subscribe log writer
+    dispatcher.subscribe(log_writer)  # Subscribe log writer
     dispatcher.start()
 
 
@@ -58,7 +58,7 @@ def main():
         threading.Thread(target=start_dispatcher,
                          args=('localhost', out_port, args.logfile),
                          daemon=True).start()  # Set daemon to all ipython exit
-                                               # kill all threads
+        #                                        kill all threads
     print("Connecting...")
     proxy = Proxy(in_port, out_port)
     print("Receiving on port %d; publishing to port %d." % (in_port, out_port))
--- a/bluesky/commandline/zmq_proxy.py
+++ b/bluesky/commandline/zmq_proxy.py
@@ -58,7 +58,6 @@ def main():
         # Set daemon to kill all threads upon IPython exit
         threading.Thread(target=start_dispatcher,
                          args=('localhost', out_port, args.logfile),
-                         # Set daemon to all ipython exit kill all threads
                          daemon=True).start()
 
     print("Connecting...")
--- a/bluesky/tests/test_run_engine.py
+++ b/bluesky/tests/test_run_engine.py
@@ -166,7 +166,7 @@ def test_flyer_with_collect_asset_documents(RE):
     from ophyd.sim import det, new_trivial_flyer, trivial_flyer
     from bluesky.preprocessors import fly_during_wrapper
     assert hasattr(new_trivial_flyer, 'collect_asset_docs')
-    assert hasattr(trivial_flyer, 'collec_asset_docs') is False
+    assert not hasattr(trivial_flyer, 'collec_asset_docs')
     RE(fly_during_wrapper(count([det], num=5), [new_trivial_flyer, trivial_flyer]))
 
 
--- a/bluesky/tests/test_run_engine.py
+++ b/bluesky/tests/test_run_engine.py
@@ -14,8 +14,7 @@ from bluesky.run_engine import (RunEngineStateMachine,
                                 NoReplayAllowed, FailedStatus,
                                 RunEngineInterrupted,
                                 RequestStop,
-                                RequestAbort,
-                                PlanHalt)
+                                RequestAbort)
 from bluesky import Msg
 from functools import partial
 from bluesky.tests.utils import MsgCollector, DocCollector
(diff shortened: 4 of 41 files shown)


bluesky/callbacks/olog.py
```diff
@@ -158,7 +158,7 @@ def __init__(self, logbook):
         from pyOlog import SimpleOlogClient
         self.client = SimpleOlogClient()
         # Check at init time we are in an IPython session.
-        from IPython import get_ipython
+        from IPython import get_ipython  # noqa: F401
```
mrakitin: We do this import twice. It feels it would be better to to stash `get_ipython` as an attribute of `self`, and reuse it later. Maybe too much for this PR.

bluesky/callbacks/zmq.py
```diff
@@ -146,7 +146,7 @@ def __init__(self, in_port=None, out_port=None, *, zmq=None):
                 out_port = backend.bind_to_random_port("tcp://*")
             else:
                 backend.bind("tcp://*:%d" % out_port)
-        except:
+        except BaseException:
```
mrakitin: Is the intention here to catch all exception from that [hierarchy](https://docs.python.org/3/library/exceptions.html#exception-hierarchy), including `SystemExit`, `KeyboardInterrupt` and `GeneratorExit`?

bluesky/commandline/zmq_proxy.py
```diff
@@ -58,7 +58,7 @@ def main():
         threading.Thread(target=start_dispatcher,
                          args=('localhost', out_port, args.logfile),
                          daemon=True).start()  # Set daemon to all ipython exit
-                                               # kill all threads
+        #                                        kill all threads
```
mrakitin: This does not look nice. I'd recommend to place the comment in one line after the `.start()` or move the comment above that line.

bluesky/tests/test_zmq.py
```diff
@@ -67,9 +65,9 @@ def local_cb(name, doc):
         local_accumulator.append((name, doc))
 
     # Check that numpy stuff is sanitized by putting some in the start doc.
-    md = {'stuff': {'nested': np.array([1, 2, 3])},
-          'scalar_stuff': np.float64(3),
-          'array_stuff': np.ones((3, 3))}
+    # md = {'stuff': {'nested': np.array([1, 2, 3])},
+    #       'scalar_stuff': np.float64(3),
+    #       'array_stuff': np.ones((3, 3))}
```
mrakitin: I think it was intended to be passed to the count plan below. If not, this should be removed along with the comment line above.

bluesky/tests/test_zmq.py
```diff
@@ -314,9 +308,9 @@ def local_cb(name, doc):
...
```
mrakitin: See my comment about a similar piece of code above.

fuzz/fuzz.py
```diff
@@ -337,7 +337,7 @@ def run_fuzz():
           set([msg.command for msg in message_objects]))
 
     num_message = 100
-    num_SIGINT = 10
+    # INT = 10
```
mrakitin: It's used in the commented line below. Maybe the name of the variable should not change.

bluesky/callbacks/zmq.py
dmgav: I guess this was the idea. At least this is how the code used to work until now.

bluesky/tests/test_zmq.py
dmgav: If the reviewers are sure nobody will ever need this code, it can be removed. The question is if it was disabled temporarily and not enabled back or the change was permanent. That's why I commented it.

mrakitin: I see. The comment line suggests it should be used. I wonder if @danielballan or @tacaswell have any memory of it?

bluesky/callbacks/olog.py
dmgav: I see. But this is more of a code formatting PR. I guess we shouldn't change the code if it is not absolutely necessary.

mrakitin: OK, let's keep it here, and it can be refactored/cleaned later.

bluesky/commandline/zmq_proxy.py
dmgav: Done

mrakitin: Thanks.

fuzz/fuzz.py
dmgav: Yes, I guess so. It's my mistake.

klauer:
I'd strongly recommend a separate flake8 build in the matrix, otherwise one extra line break in a source code file will fail _every build_ you have. If you haven't been through that before, take my word that it's very frustrating.

mrakitin:
That's a good point @klauer, I like your suggestion - we don't want all jobs to fail due to a code styling issue. Maybe your suggestion should also go to https://github.com/NSLS-II/scientific-python-cookiecutter.

bluesky/callbacks/zmq.py
danielballan: Yes.

mrakitin: OK, thanks. Resolving this point.

danielballan:
@dmgav Here is an example of how @klauer has implemented this idea of "one designated `flake8` CI build" in the past: https://github.com/caproto/caproto/blob/0a8b57ee65416ffb128f001d8047daa192eeab6e/.travis.yml#L50-L55

I support adding a flake8 build here. I'm 50/50 about adding one to scientific-python-cookiecutter. The cookiecutter used to have just one build to keep the story simple; now we have multiple builds; and if we add this we'll have multiple _types_ of builds. It's all justified and makes sense if you think about it, but I worry that it's hard for readers to jump from "What's CI?" to something this complex. The scientific-python-cookiecutter is not intended to be a DAMA or Bluesky Collaboration cookiecutter; it targets small- to medium-sized projects managed by scientists whose primary job activity is not programming.

mrakitin:
> I support adding a flake8 build here. I'm 50/50 about adding one to scientific-python-cookiecutter. The cookiecutter used to have just one build to keep the story simple; now we have multiple builds; and if we add this we'll have multiple _types_ of builds. It's all justified and makes sense if you think about it, but I worry that it's hard for readers to jump from "What's CI?" to something this complex. The scientific-python-cookiecutter is not intended to be a DAMA or Bluesky Collaboration cookiecutter; it targets small- to medium-sized projects managed by scientists whose primary job activity is not programming.

Got it. Maybe https://github.com/NSLS-II/scientific-python-cookiecutter/issues/86 will be an overkill then.

bluesky/tests/test_zmq.py
danielballan: Yes, it looks like the very next line below---which is also commented out, and has been since this file was first committed---was at some point intended to test that the objects in `md` serialized successfully.

I agree that the fix isn't to remove these lines; it's to use them somewhere. The easiest fix might me to  change the next _uncommented_ line below

```py
RE(count([hw.det]), local_cb)
```

to

```py
RE(count([hw.det]), local_cb, **md)
```

bluesky/tests/test_run_engine.py
```diff
@@ -167,7 +166,7 @@ def test_flyer_with_collect_asset_documents(RE):
     from ophyd.sim import det, new_trivial_flyer, trivial_flyer
     from bluesky.preprocessors import fly_during_wrapper
     assert hasattr(new_trivial_flyer, 'collect_asset_docs')
-    assert hasattr(trivial_flyer, 'collec_asset_docs') == False
+    assert hasattr(trivial_flyer, 'collec_asset_docs') is False
```
danielballan: A minor point.
```suggestion
    assert not hasattr(trivial_flyer, 'collec_asset_docs')
```

dmgav:
I am trying to implement the separate `flake8` build in Ophyd, but it doesn't seem to work. Probably I should create a PR.

dmgav:
I don't think the PR is ready: I am trying to configure it so that if `flake8` fails, the build fails as well.

danielballan:
Sounds good. I intend my approving review to mean, "I don't see any unintended changes to bluesky behavior here."

danielballan:
Ah, there is a real failure here resulting from the addition of numpy arrays in `md`, I believe.

https://travis-ci.org/github/bluesky/bluesky/jobs/663661072#L2545

danielballan:
One approach we have taken in the past is to JSON-serialize and -deserialize the documents using `event_model.sanitize_doc` before comparing them.

dmgav:
All tests except `flake8` test (the first one) succeeded, so it proves that the script works. Pushing the commit with fixed formatting.

dmgav:
It looks like all the issues raised in reviews are addressed.

PR #1331, opened 2020-03-31
Use nice error message if Python too low.

danielballan:
Copied from our cookiecutter

Tested with `python2 setup.py sdist`


setup.py
```diff
@@ -14,7 +35,7 @@
     license="BSD (3-clause)",
     url="https://github.com/NSLS-II/bluesky",
```
mrakitin: Should we have the URL updated?
```suggestion
    url="https://github.com/bluesky/bluesky",
```

setup.py
```diff
@@ -14,7 +35,7 @@
...
     packages=setuptools.find_packages(),
-    python_requires='>=3.6',
+    python_requires='>={}'.format('.'.join(str(n) for n in min_version)),
     install_requires=requirements,
     classifiers=[
         "Development Status :: 3 - Alpha",
```
mrakitin: Stable/production-ready already?

mrakitin:
```suggestion
        "Development Status :: 5 - Production/Stable",
```

PR #1334, opened 2020-04-02
List scan

dmgav:
The PR is addressing and attempts to resolve the issue https://github.com/bluesky/bluesky/issues/1189

In the new API interface for `grid_scan` and `rel_grid_scan` functions, the optional `snake_axes` argument was added. The boolean value `True` or `False` of `snake_axes` enables or disables snaking along all axes (except the slowest). Alternatively, `snake_axes` can be a list (or iterable), which specifies the axes along which snaking is enabled.

```
# Default - snaking is disabled
grid_scan([hw.det], hw.motor, 1, 2, 5, hw.motor1, 7, 2, 10, hw.motor2, 3, 5, 4)

# Snaking is explicitely disabled
grid_scan([hw.det], hw.motor, 1, 2, 5, hw.motor1, 7, 2, 10, hw.motor2, 3, 5, 4, snake_axes=False)

# Snaking can also be disabled by providing empty list of motors
grid_scan([hw.det], hw.motor, 1, 2, 5, hw.motor1, 7, 2, 10, hw.motor2, 3, 5, 4, snake_axes=[])

# Snaking is enabled for all motors except the slowest hw.motor
grid_scan([hw.det], hw.motor, 1, 2, 5, hw.motor1, 7, 2, 10, hw.motor2, 3, 5, 4, snake_axes=True)

# Snaking is enabled only for hw.motor1
grid_scan([hw.det], hw.motor, 1, 2, 5, hw.motor1, 7, 2, 10, hw.motor2, 3, 5, 4, snake_axes=[hw.motor1])

# Snaking is enabled only for hw.motor1 and hw.motor2
grid_scan([hw.det], hw.motor, 1, 2, 5, hw.motor1, 7, 2, 10, hw.motor2, 3, 5, 4, snake_axes=[hw.motor1, hw.motor2])
```

The new API interface is consistent with already existing functions `list_grid_scan` and `rel_list_grid_scan` that accept `snake_axes` argument.

The implementations of `grid_scan` and `rel_grid_scan` are backward compatible. Old and new API should not be mixed. If they are mixed, then the settings in `snake_axes` override snaking settings in the `args` list. For example:

```
# Old style call that enables snaking for hw.motor1
grid_scan([hw.det], hw.motor, 1, 2, 5, hw.motor1, 7, 2, 10, True, hw.motor2, 3, 5, 4, False)

# Equivalent new style call
grid_scan([hw.det], hw.motor, 1, 2, 5, hw.motor1, 7, 2, 10, hw.motor2, 3, 5, 4, snake_axes=[hw.motor1])

# In the following call, snaking is disabled for all motors
grid_scan([hw.det], hw.motor, 1, 2, 5, hw.motor1, 7, 2, 10, True, hw.motor2, 3, 5, 4, False, snake_axes=False)

# In the following call, snaking is enabled for hw.motor2
grid_scan([hw.det], hw.motor, 1, 2, 5, hw.motor1, 7, 2, 10, True, hw.motor2, 3, 5, 4, False, snake_axes=[hw.motor2])
``` 
The description of the old style calls will be removed from documentation. It is assumed that only the new style calls are used in all new code.

## Description

The changed include adding code for processing of arguments in `grid_scan` and `rel_grid_scan`, modification of `chunk_outer_product_args` so that it is able to recognize two patterns of arguments (with/without snaking boolean flags). The tests were implemented for `grid_scan`, `rel_grid_scan`, `chunk_outer_product_args` and `outer_product` functions.

## Motivation and Context
Issue: https://github.com/bluesky/bluesky/issues/1189

## How Has This Been Tested?
The tests for `grid_scan` and `rel_grid_scan` are implemented. 
In addition, the tests for `chunk_outer_product_args` and `outer_product` were implemented.

### DIFF
--- a/bluesky/plan_patterns.py
+++ b/bluesky/plan_patterns.py
@@ -388,10 +388,19 @@ def chunk_outer_product_args(args):
     Parameters
     ----------
     args
-        patterned like (``motor1, start1, stop1, num1,```
-                        ``motor2, start2, stop2, num2, snake2,``
-                        ``motor3, start3, stop3, num3, snake3,`` ...
-                        ``motorN, startN, stopN, numN, snakeN``)
+        Two patterns are supported:
+
+        Pattern 1: (``motor1, start1, stop1, num1,```
+                    ``motor2, start2, stop2, num2,``
+                    ``motor3, start3, stop3, num3,`` ...
+                    ``motorN, startN, stopN, numN``)
+
+        Pattern 2: (``motor1, start1, stop1, num1,```
+                    ``motor2, start2, stop2, num2, snake2,``
+                    ``motor3, start3, stop3, num3, snake3,`` ...
+                    ``motorN, startN, stopN, numN, snakeN``)
+
+        In Pattern 2, all elements 'snakeX' must be boolean.
 
         The first motor is the "slowest", the outer loop. For all motors
         except the first motor, there is a "snake" argument: a boolean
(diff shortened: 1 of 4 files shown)


dmgav:
@danielballan I just noticed your comment https://github.com/bluesky/bluesky/issues/1189#issuecomment-606628088. The arguments are parsed by `chunk_outer_product_args` function, which used to validate the argument list simply by checking the number of arguments. I extended the functionality to the new pattern that doesn't contain `snakeX` boolean values. In some cases this is not sufficient (say 24 arguments can be split both in the 6 groups of 4 and in one group of 4 and 4 groups of 5. I assumed that `snakeX` values are always boolean and use this assumption to distinguish between patterns. In principel it is possible to do additional validation by checking if motors are `Movable`, but it will limit the potential use of `chunk_outer_product_args` only to the cases when the motors are movable.

danielballan:
It's safer to assume that the `motorX` values are always `Movable` than that the `snakeX` values are of type `bool`. User code that passes `0` and `1` or other truth-y values in would break here. I can't think of an instance where a non-`Movable` object could be used as a motor.

That said, it's likely that every _actual_ instance of this in the wild uses `True`/`False` for `snakeX`, not `0`/`1` or anything else, so it may not be worth spending more effort on this.

Thanks for the detailed description; this looks on the mark to me. Will do a detailed line review soon.

> If they are mixed, then the settings in snake_axes override snaking settings in the args list.

Would it be better to fail with a `ValueError` in this case? When the user asks us to do something impossible (snake but also do not snake) I prefer to fail loudly, lest the user wonder "Why aren't my `snakeX` values being respected?"

dmgav:
I actually like the idea of raising the ValueError in case the new and old styles are mixed. May be it's also a good idea to add argument validation in `chunk_outer_product_args` the and check if motors are Movable. This is not a big change.

This looks like a big PR, but most of the new code is for testing.

dmgav:
The changes: 
-  The argument list patterns are identified by looking for the motors instead of boolean values. The restrictions on `snakeX` to be boolean is removed: now 0 and 1 can be used as well. This is relevant only to deprecated API, so there is no need for restrictions.
- Now `ValueError` is raised if the deprecated and new API is mixed, i.e. if argument list contains `snakeX` entries and `snake_axes` is passed and has value different from `None`.

dmgav:
The description of the PR was modified to reflect the changes.

dmgav:
Thanks @danielballan . I thought that the list of attributes needs to be extended but I was not sure which ones to add.

danielballan:
https://github.com/bluesky/bluesky/pull/1334/commits/f69709890020241a9d7c500e175908cb87e7b07f introduced a slew of CI failure due to a missing comma. This should not have been merged. Maybe the button was green because Travis-CI didn't kick off. Fix proposed in https://github.com/bluesky/bluesky/pull/1336.

tacaswell:
Sorry about that :sheep:

PR #1336, opened 2020-04-09
Fix critical typo in Movable.

danielballan:
This fixes a critical typo in f69709890020241a9d7c500e175908cb87e7b07f
that was merged in https://github.com/bluesky/bluesky/pull/1334. The
typo failed many tests, so I do not think any additional tests are
needed. I think #1334 was merged by mistake.


danielballan:
Self-merging (with the @tacaswell approval above) because until this is fixed all PR builds will fail.

PR #1338, opened 2020-04-20
Include configuration in flyscan EventDescriptor

### LINKED ISSUE #1337: An EventDescriptor generated by a fly scan should have a `'configuration'` key.
## Expected Behavior

An EventDescriptor generated by a fly scan should have a `'configuration'` key.

## Current Behavior

It does not.

https://github.com/bluesky/bluesky/blob/7bb4129926ca0ccca1840cbc4c86397f9ad43416/bluesky/bundlers.py#L595-L604

The documents are still _valid_ because configuration is not required by the [schema](https://github.com/bluesky/event-model/blob/d5643e795cd66da7ee5110daa1200cb67c4ec54a/event_model/schemas/event_descriptor.json#L137-L142).

## Possible Solution

Fix the `_collect` method linked above to include configuration. The missing code can be found in the `_save` method on that same class. It should be factored into a new shared method, or perhaps duplicated if too many details vary to make a shared method elegant/practical.

```py
from ophyd.sim import TrivialFlyer
from bluesky import RunEngine
from bluesky.plans import fly
RE = RunEngine()
flyer = TrivialFlyer()
from pprint import pprint
def callback(name, doc):
    print(f'{name} document:')
    pprint(doc)
RE(fly([flyer]), callback)
```

Output excerpt:
```
# missing 'configuration'
descriptor document:
{'data_keys': {},
 'hints': {},
 'name': 'stream_name',
 'object_keys': {'trivial_flyer': []},
 'run_start': '2d67a036-3691-4e21-9ad8-50fca439ada0',
 'time': 1587140357.5311675,
 'uid': 'c0716490-30ea-4e9e-b4c3-6429863c40b3'}
```

jklynch:
## Description
This PR modifies`RunBundler.collect` to call `read_configuration`, if it exists, on the `collect` message object and the results are included in *all* descriptors.

## Motivation and Context
This PR resolves #1337.

## How Has This Been Tested?
`test_flyer_descriptor()` in `test_run_engine.py` checks for the presence of key `configuration`.

### DIFF
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -396,8 +396,6 @@ class RunBundler:
                 config[name]["data"] = self._config_values_cache[obj]
                 config[name]["timestamps"] = self._config_ts_cache[obj]
                 config[name]["data_keys"] = self._config_desc_cache[obj]
-                if hasattr(obj, "hints"):
-                    hints[name] = obj.hints
             descriptor_uid = new_uid()
             doc = dict(
                 run_start=self._run_start_uid,
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -396,6 +396,8 @@ class RunBundler:
                 config[name]["data"] = self._config_values_cache[obj]
                 config[name]["timestamps"] = self._config_ts_cache[obj]
                 config[name]["data_keys"] = self._config_desc_cache[obj]
+                if hasattr(obj, "hints"):
+                    hints[name] = obj.hints
             descriptor_uid = new_uid()
             doc = dict(
                 run_start=self._run_start_uid,
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -594,7 +594,6 @@ class RunBundler:
                 descriptor_uid = new_uid()
                 stream_object_keys = {collect_obj.name: list(stream_data_keys)}
                 hints = {}
-                # is this if block an error?
                 if hasattr(collect_obj, "hints"):
                     hints.update({collect_obj.name: collect_obj.hints})
                 for obj_read in objs_read:
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -596,9 +596,6 @@ class RunBundler:
                 hints = {}
                 if hasattr(collect_obj, "hints"):
                     hints.update({collect_obj.name: collect_obj.hints})
-                for obj_read in objs_read:
-                    if hasattr(obj_read, "hints"):
-                        hints[obj_read.name] = obj_read.hints
                 doc = dict(
                     run_start=self._run_start_uid,
                     time=ttime.time(),
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -614,19 +614,20 @@ class RunBundler:
                                  extra={'doc_name': 'descriptor',
                                         'run_uid': self._run_start_uid,
                                         'data_keys': stream_data_keys.keys()})
-                self._descriptors[stream_name] = (objs_read, doc)
+                self._descriptors[stream_name] = (frz_stream_data_keys, doc)
                 self._sequence_counters[stream_name] = count(1)
             else:
                 objs_read, doc = self._descriptors[stream_name]
-            if d_objs != objs_read:
-                raise RuntimeError(
-                    "Mismatched objects read, "
-                    "expected {!s}, "
-                    "got {!s}".format(d_objs, objs_read)
-                )
+                # we only need to make this check when the else block is executed
+                if frz_stream_data_keys != objs_read:
+                    raise RuntimeError(
+                        "Mismatched objects read, "
+                        "expected {!s}, "
+                        "got {!s}".format(frz_stream_data_keys, objs_read)
+                    )
 
             descriptor_uid = doc["uid"]
-            local_descriptors[objs_read] = (stream_name, descriptor_uid)
+            local_descriptors[frz_stream_data_keys] = (stream_name, descriptor_uid)
 
             bulk_data[descriptor_uid] = []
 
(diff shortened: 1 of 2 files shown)


bluesky/bundlers.py
```diff
@@ -555,63 +555,74 @@ async def collect(self, msg):
...
-        local_descriptors = {}  # hashed on obj_read, not (name, objs_read)
-        for stream_name, data_keys in named_data_keys.items():
-            desc_key = stream_name
-            d_objs = frozenset(data_keys)
-            if desc_key not in self._descriptors:
+        local_descriptors = {}  # hashed on objs_read, not (name, objs_read)
+        # collect_obj.describe_collect() returns a dictionary like this:
+        #     {name_for_desc1: data_keys_for_desc1,
+        #      name_for_desc2: data_keys_for_desc2, ...}
+        for stream_name, stream_data_keys in collect_obj.describe_collect().items():
```
danielballan, line 591: I am glad we both appreciate the value of well-chosen local variable names for code readability.

bluesky/bundlers.py
```diff
@@ -555,63 +555,74 @@ async def collect(self, msg):
...
                 objs_read = d_objs
-                # We don't not have an Event Descriptor for this set.
+                # We do not have an Event Descriptor for this set.
                 descriptor_uid = new_uid()
-                object_keys = {obj.name: list(data_keys)}
+                stream_object_keys = {collect_obj.name: list(stream_data_keys)}
                 hints = {}
-                if hasattr(obj, "hints"):
-                    hints.update({obj.name: obj.hints})
+                # is this if block an error?
```
danielballan, line 599: The `hints` attribute is a relatively new thing, so we must treat it as optional to support old code. I think I would be in favor of keeping it optional; I don't think a minimal bluesky-compatible object should need to implement it.

bluesky/bundlers.py
```diff
@@ -555,63 +555,74 @@ async def collect(self, msg):
...
+                if hasattr(collect_obj, "hints"):
+                    hints.update({collect_obj.name: collect_obj.hints})
+                for obj_read in objs_read:
```
danielballan, line 602: I think this loop is misconceived and should be removed. The `objs_read` is a `frozenset` of _strings_ represent data keys. The strings will never have a `.name` attribute or a `.hints` attribute. It's worth reviewing `_save` and `_monitor` to see if the same mistake was made there as well.

bluesky/bundlers.py
jklynch: It was very difficult to compare the save and collect methods because they used minimal, similar variable names, but I would have changed these anyway.

bluesky/bundlers.py
jklynch: It looks like the save method has
```python
for obj in objs_read:
    ...
    if hasattr(obj, "hints"):
        hints[name] = obj.hints
```
but monitor does not.

jklynch: removing
```python
    if hasattr(obj, "hints"):
        hints[name] = obj.hints
```
causes this [test](https://github.com/bluesky/bluesky/blob/master/bluesky/tests/test_run_engine.py#L1390) to fail. Is the test correct?

bluesky/bundlers.py
```diff
@@ -555,63 +553,70 @@ async def collect(self, msg):
...
-            d_objs = frozenset(data_keys)
-            if desc_key not in self._descriptors:
+        local_descriptors = {}  # hashed on objs_read, not (name, objs_read)
+        # collect_obj.describe_collect() returns a dictionary like this:
+        #     {name_for_desc1: data_keys_for_desc1,
+        #      name_for_desc2: data_keys_for_desc2, ...}
+        for stream_name, stream_data_keys in collect_obj.describe_collect().items():
+            d_objs = frozenset(stream_data_keys)
+            if stream_name not in self._descriptors:
                 objs_read = d_objs
```
danielballan, line 592: I think this is the source of the confusion. Elsewhere (e.g. in the `save` method) `objs_read` is a `frozenset` of live objects (e.g. Devices). This is a `frozenset` of strings. We should use a different name for it.

bluesky/bundlers.py
danielballan: Looks like `save` should be left as is. In that case, `self._obj_read` is a `frozenset` of actual objects. See https://github.com/bluesky/bluesky/pull/1338/files#r412169748

bluesky/bundlers.py
jklynch: Is the `frozenset` necessary? Is there a reason not to use `stream_data_keys`?

danielballan: Great question. We make a `frozenset` because are around to use it as a key in a dict. That means it had to be hashable, and mutable objects are not hashable—hence, make a copy in an immutable data structure and proceed with that. Does that answer your question?

jklynch: Thanks. I tend to use a tuple for no good reason in these cases.

jklynch:
I have added what I hope is a good test of the contents of descriptor["configuration"].

bluesky/bundlers.py
danielballan: Yeah, I would normally reach for a tuple first as well, just because it's a more familiar data structure. But we want to make sure that `{'a', 'b'}` == `{'b', 'a'}`, which of course isn't the case for a tuple. We could create a tuple and sort the elements first, but that's easy to mess up if you forget to sort somewhere.

danielballan:
Note that the CI failures are due to an unrelated issue https://github.com/bluesky/bluesky/issues/1339.

danielballan:
Re-running CI now that https://github.com/bluesky/bluesky/issues/1339 has been resolved.

PR #1345, opened 2020-05-01
verify nd step for scan_nd back compatibility

lsammut:
verify the signature of the per_step function meets requirements for back compatibility from 1.5.5 to 1.6.0.

in discussions with @tacaswell 

still to be tested.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -939,7 +939,27 @@ def scan_nd(detectors, cycler, *, per_step=None, md=None):
 
             return True
 
+        def _verify_nd_step(sig):
+            if len(sig.parameters) < 3:
+                return False
+            for name, (p_name, p) in zip_longest(['detectors', 'step', 'pos_cache'], sig.parameters.items()):
+                # this is one of the first 3 positional arguements, check that the name matches
+                if name is not None:
+                    if name != p_name:
+                        return False
+                # if there are any extra arguments, check that they have a default
+                else:
+                    if p.kind is p.VAR_KEYWORD or p.kind is p.VAR_POSITIONAL:
+                        continue
+                    if p.default is p.empty:
+                        return False
+
+            return True
+
         if sig == inspect.signature(bps.one_nd_step):
+            pass
+        elif _verify_nd_step(sig):
+            # check other signature for back-compatibility
             pass
         elif _verify_1d_step(sig):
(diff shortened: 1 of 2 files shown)


bluesky/plans.py
```diff
@@ -941,6 +958,9 @@ def _verify_1d_step(sig):
 
         if sig == inspect.signature(bps.one_nd_step):
```
tacaswell, line 959: We can probably remove this check as `_verify_nd_step` should handle this case?

tacaswell:
I think the failures due to the tests being strict about the message in the exception:

https://github.com/bluesky/bluesky/blob/e5de942d14e723e843291a1a6cb989c28dd9b9fc/bluesky/tests/test_plans.py#L254-L262

That pattern match will need to be updated to new exception text.

lsammut:
> I think the failures due to the tests being strict about the message in the exception:
> 
> https://github.com/bluesky/bluesky/blob/e5de942d14e723e843291a1a6cb989c28dd9b9fc/bluesky/tests/test_plans.py#L254-L262
> 
> That pattern match will need to be updated to new exception text.

Hi @tacaswell, 
yes I noticed this too. If the new message is ok I'll update the test accordingly.

danielballan:
The new error message looks great—go for it! :-)

Thanks for fixing this, @lsammut.

lsammut:
@tacaswell @danielballan 
I've added some nd_per_step specific tests and updated the pattern match. The tests shouldn't fail anymore.

tacaswell:
pushing "re-run" on the tests made them pass, going to look into why they are flaky (also flaky on the master cron jobs).

bluesky/tests/test_plans.py
```diff
@@ -242,6 +246,15 @@ def wrong_names(a, b, c, take_reading=None):
...
+
+    def per_nd_step_bad_pos(detectors, step, pos_cache, *, extra_no_dflt):
+        "no body"
+
+    def all_wrong(a, b, c=None, *args, d=None, g, **kwargs):
+        "no body"
+
     return pytest.mark.parametrize(
         "per_step",
         [too_few, too_many, extra_required_kwarg, wrong_names, per_step_only_args],
```
klauer, line 260: Consider using `pytest.param(callable, id='...')` here

Issue #1353, opened 2020-05-20
RunEngine raises RunEngineInterrupted exception when paused, not helpful in GUI applications


RussBerg:
I have the RunEngine embedded in a Qt application, when the user pauses the RunEngine it raises the RunEngineInterrupted  exception letting the user know what their options are (resume, stop etc), while helpful for a console application it is much less so in a GUI application because it looks like an error and adds no value as the widgets on screen indicate valid next steps etc.

## Current Behavior
user clicks stop during RunEngine execution and an exception is raised.

## Possible Solution
allow the parent class of the RunEngine to set a flag which skips the raise of RunEngineInterrupted

## Your Environment
Windows 10, Qt application
![image](https://user-images.githubusercontent.com/4698891/82491678-3141d980-9aa2-11ea-8229-af3beeaa28da.png)

RussBerg:
issue was resolved by wrapping RE calls to pause/stop/resume in try/except blocks on application side requiring no changes to base code

PR #1354, opened 2020-05-20
added gui_mode attribute to solve #1353

RussBerg:
added gui_mode attribute, False by default which will still raise RunEngineInterrupted and if set to True  for GUI applications it will skip the raise

## Description
added:
 - self.gui_mode boolean attribute
 - set_gui_mode() function to set the attribute
 - if statement now check self.gui_mode before raise of RunEngineInterrupted

## Motivation and Context
see issue #1353 

## How Has This Been Tested?
I tested this on my own Qt GUI application (pyStxm) and it does not affect any other aspects of RunEngine that I can see

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -342,6 +342,7 @@ class RunEngine:
         self.waiting_hook = None
         self.record_interruptions = False
         self.pause_msg = PAUSE_MSG
+        self.gui_mode = False
 
         if during_task is None:
             during_task = DefaultDuringTask()
@@ -419,6 +420,17 @@ class RunEngine:
         self.unsubscribe_lossless = self.dispatcher.unsubscribe
         self._subscribe_lossless = self.dispatcher.subscribe
         self._unsubscribe_lossless = self.dispatcher.unsubscribe
+
+    def set_gui_mode(self, val):
+        '''
+        Allow the RunEngine to know if it is running on the console (default) or as part of a GUI, if GUI then we want to
+        suppress the RunInterruptedException that is called when the RunEngine is paused because GUI applications have
+        other visual widgets that indicate next steps where as on the console the user must be explicitly told what those
+        options are.
+        :param val:
+        :return:
+        '''
+        self.gui_mode = val
 
     @property
     def commands(self):
@@ -795,7 +807,8 @@ class RunEngine:
         self._resume_task(init_func=_build_task)
 
         if self._interrupted:
-            raise RunEngineInterrupted(self.pause_msg) from None
+            if(not self.gui_mode):
+                raise RunEngineInterrupted(self.pause_msg) from None
 
         return tuple(self._run_start_uids)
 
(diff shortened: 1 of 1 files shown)


tacaswell:
Thanks for the issue and the PR!

I have a generic concern about flags like this that change the internal behavior.  One is ok, two gets hard to manage, 3 or more condense and produce new states of ~matter~ code complexity ;)

Is possible to wrap the call to `RE` in a `try: ... except`?

If we are going to add this flag, I would rather call it "raise_on_interrupt` rather than `gui_mode` to be clearer about what the scope if it is.  If it is called `gui_mode` and later on someone wants to add something else that makes sense for their GUI but is harmful to yours we will end up in a pickle.

I may be more supportive of pulling a base class out of the current `RunEngine` and then re-subclasing to add it the CLI specific functionality.  In addition to embedding in a GUI we are also starting to think about what it means to run a RE in a server (ex https://github.com/bluesky/bluesky-queueserver).

bluesky/run_engine.py
```diff
@@ -795,7 +807,8 @@ def set_blocking_event(future):
         self._resume_task(init_func=_build_task)
 
         if self._interrupted:
-            raise RunEngineInterrupted(self.pause_msg) from None
+            if(not self.gui_mode):
```
mrakitin, line 810:
```suggestion
            if not self.gui_mode:
```

bluesky/run_engine.py
```diff
@@ -832,7 +845,8 @@ def resume(self):
                 obj.resume()
         self._resume_task()
         if self._interrupted:
-            raise RunEngineInterrupted(self.pause_msg) from None
+            if(not self.gui_mode):
```
mrakitin, line 848:
```suggestion
            if not self.gui_mode:
```

bluesky/run_engine.py
```diff
@@ -420,6 +421,17 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
...
 
+    def set_gui_mode(self, val):
+        '''
+        Allow the RunEngine to know if it is running on the console (default) or as part of a GUI, if GUI then we want to
+        suppress the RunInterruptedException that is called when the RunEngine is paused because GUI applications have
+        other visual widgets that indicate next steps where as on the console the user must be explicitly told what those
+        options are.
+        :param val:
+        :return:
+        '''
```
mrakitin, line 432: The docstring may need some rework to match the style of the rest of the docstrings.

RussBerg: ok I feel like a dork (appropriate emoji here), not sure why I didn't think of just wrapping the pause/resume/stop calls in a try except block, can you spell tunnel vision?, anyway that solved it fine so I will nuke this PR, thanks @tacaswell for pointing that out. I am very interested in the queue-server so I will look at that tomorrow

RussBerg:
problem was resolved on the application side which didnt require changes to RunEngine

tacaswell:
> can you spell tunnel vision?,

This is why we program with friends :)  I assure you I have done the same.

RussBerg:
Thanks Tom :)

PR #1367, opened 2020-06-18
FIX: allow ':' to be used in keynames and still format LiveTable

tacaswell:
## Description

Internally hash the keys we use to format the LiveTable rows.  This ensure that the value of the data keys can not corrupt the formatting.  Opted to go with a hash (because this is completely internal) rather than trying to sanitize the input as it seemed simpler.

## Motivation and Context

fixes a bug that @stashlukj  ran into when trying to use ``:' ` as signal names.

## How Has This Been Tested?

Added another test of LiveTable.  Re-arranged the tests a bit.

### DIFF
--- a/bluesky/tests/test_callbacks.py
+++ b/bluesky/tests/test_callbacks.py
@@ -5,6 +5,7 @@ from bluesky.object_plans import AbsScanPlan
 from bluesky.preprocessors import run_wrapper, subs_wrapper
 from bluesky.plan_stubs import pause
 import bluesky.plans as bp
+import bluesky.preprocessors as bpp
 from bluesky.callbacks import CallbackCounter, LiveTable, LiveFit, CallbackBase
 from bluesky.callbacks.core import make_callback_safe, make_class_safe
 from bluesky.callbacks.mpl_plotting import (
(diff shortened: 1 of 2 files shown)


tacaswell:
The blank space in the test is actually 😻 that seems to not existing in the font the diffs are rendered in.

klauer:
Python already has unique hashes for strings - how about `str(hash(key))`? Though I suppose that wouldn't be consistent between processes, so perhaps hashlib is better. 🤷‍♂️

tacaswell:
I just did not think of that, using `str(hash(key))` is definitely better.

tacaswell:
Had to use `f'h{str(hash(k))}'` as the key because other was the format string ended up being '{3405904703012941009:d}` which tells python to look for the too-many-ith positional argument to format.

bluesky/tests/test_callbacks.py
```diff
@@ -266,6 +270,30 @@ def test_table_external(RE, hw, db):
 +------------+--------------+----------------+----------------+"""
 
 
+def test_evil_table_names(RE):
+    from ophyd import Signal
+    sigs = [
+        Signal(value=0, name="a:b"),
+        Signal(value=0, name="a,b"),
+        Signal(value=0, name="a'b"),
+        Signal(value=0, name="🐍"),
```
klauer, line 279: 🐍

tacaswell: :snake:

tacaswell:
It is failing on the docs build due to warnings from PIMS

tacaswell:
power-cycled in the hope that the docs are fixed on master.

klauer:
Bump @tacaswell @danielballan, can we get this one in? @ZLLentz found that we're running into this with our default config...

ZLLentz:
New users of `bluesky` who want to scan simple `EpicsSignal` instances run into this because the default `name`, if not provided, is just the pvname again, which always has `:` in it at SLAC.

danielballan:
Yes. We're doing another round of patch releases this week.

Re-ran the two failed builds...

PR #1368, opened 2020-06-19
FIX/API: do not create a DefaultDuringTask on import

tacaswell:
Doing the init at import time is breaking xi-cam (which is a Qt
application and does many of its imports in a background thread).

This was here for back-compatibility to replace a function by the same
name, but because we went with a `block` method rather than a
`__call__` method if anyone was relying on this they are already
broken so remove with no attempt to shim it back.

### DIFF
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -1507,9 +1507,6 @@ class DefaultDuringTask(DuringTask):
                 blocking_event.wait()
 
 
-default_during_task = DefaultDuringTask()
-
-
 def _rearrange_into_parallel_dicts(readings):
     data = {}
     timestamps = {}


dmgav:
@tacaswell Do you remember why exactly we decided create `default_during_task`? Just for compatibility with older versions, which had the function `default_during_task()` defined? As far as I remember, there were some issues with teleporter being initialized too late, but I don't remember the details and all the examples seem to run well, i.e. `RE.__init__` seems to take care of the initialization. I think that if the only issue is back-compatibility, then `default_during_task` can be removed.

tacaswell:
re-reviewing https://github.com/bluesky/bluesky/pull/1300 I think it was just meant to be for back compatibility (we solved the init-early-enough problem in `RE.__init__`), however the object with that name is no longer callable so we still broke any code that was relying on it.

dmgav:
Yes, I guess the conclusion was that nobody is using custom `during_task` functions, so we could safely replace it by an object.

danielballan:
@ronpandolfi When you get a chance, please confirm that this fully resolves your issue before we merge.

dmgav:
@danielballan Should we merge it?

PR #1369, opened 2020-06-22
RunBundler.collect() expects the wrong structure from read_configuration

jklynch:
## Description
RunBundler.collect() incorrectly assumed the dictionary returned by a flyer's read_configuration() method would be keyed by component names. The structure actually returned by read_configuration() is the same as returned by ReadableDevice.read_configuration() and is keyed on field names.

## Motivation and Context
This closes #1366.

## How Has This Been Tested?
The existing flyer test has been corrected. A new flyer test has been added.

PR #1370, opened 2020-06-24
Backport #1369 to v1.5.x

jklynch:
Backport of #1369.


jklynch:
CI failed on python 3.6 with what looks like a server issue. (?)

danielballan:
Agreed. Restarted that build.

jklynch:
@danielballan I could not restart the build, but I have been able to do that in the past maybe just for other repositories. Do you know why?

danielballan:
Double check that you are still signed into Travis. When this happens to me, it’s because my session with Travis has expired.

PR #1378, opened 2020-07-17
ENH: allow "movable" objects that match bluesky interface with instance attributes

### LINKED ISSUE #1377: check of isinstance(obj, Movable) only accepts class attributes, rejecting potentially valid objects
I'm trying to get yaqc-bluesky settable objects to work in a `grid_scan`, but bluesky is claiming that my arguments are invalid. In particular, my objects are failing to pass `isinstance(obj, Movable)` [1].

The `Movable` class is an abstract base class [2] which, according to the comment, should turn the `isinstance` check into, essentially, a check against the list of expected attributes. Not actually understanding the ABC magic, all I can say is that I can pass `hasattr` without passing `isinstance`:

```python
import bluesky
from bluesky.utils import Movable


EXPECTED_ATTRS = (
    'name',
    'parent',
    'read',
    'describe',
    'read_configuration',
    'describe_configuration',
    'set',
    'stop',
)


class Device(object):

    def __init__(self):
        for attr in EXPECTED_ATTRS:
            setattr(self, attr, None)


device = Device()
print("hasattr", all(hasattr(device, attr) for attr in EXPECTED_ATTRS))  # True
print("isinstance", isinstance(device, Movable))  # False
print("subclasshook", Movable.__subclasshook__(device))  # True
```

Even more strangely, when I play with your example notebook [3] your own ophyd motor objects also fail this `isinstance` check and, indeed, are rejected within `grid_scan` (Hello Bluesky.ipynb). Is it possible that this plan is just broken?

Any guidance would be appreciated---thanks! :smile: 

[1] https://github.com/bluesky/bluesky/blob/master/bluesky/plan_patterns.py#L431
[2] https://github.com/bluesky/bluesky/blob/master/bluesky/utils.py#L1519-L1548
[3] https://try.nsls2.bnl.gov/

untzag:
This PR changes the way that bluesky checks if an argument is "movable". Before: `isinstance(obj, Movable)`, now: `is_movable(obj)`.

Current behavior requires all of the expected attributes to be implemented as class attributes. With this PR, bluesky now checks instance attributes. For example, a class that defines `name` as an instance attribute at init time would have failed in old state but now passes. All objects that passed before should still pass.

closes #1377 

As an example of where current state gets weird, try running the following cell in a new notebook on https://try.nsls2.bnl.gov/

```python
from bluesky.utils import Movable
from bluesky.plans import grid_scan
from bluesky import RunEngine

%run scripts/beamline_configuration.py
motor = motor_ph
det = ph

RE(grid_scan([det], motor, 0, 1, 11))  # fails
```

### DIFF
--- a/bluesky/tests/test_utils.py
+++ b/bluesky/tests/test_utils.py
@@ -3,7 +3,7 @@ import pytest
 from functools import reduce
 import operator
 
-from bluesky.utils import ensure_generator, Msg, merge_cycler, Movable
+from bluesky.utils import ensure_generator, Msg, merge_cycler, is_movable
 from cycler import cycler
 
 
@@ -102,11 +102,11 @@ def test_cycler_merge_mixed(hw, children):
     assert mcyc.by_key() == cyc.by_key()
 
 
-def test_Movable(hw):
+def test_is_movable(hw):
 
     obj_list = [(10, False), (1.05, False), ("some_string", False),
                 (hw.det, False), (hw.motor, True)]
     for obj, result in obj_list:
-        assert isinstance(obj, Movable) == result, \
+        assert is_movable(obj) == result, \
             f"The object {obj} is incorrectly recognized "\
             f"as {'' if result else 'not '}movable"
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -1565,12 +1565,12 @@ def is_movable(obj):
 class Movable(metaclass=abc.ABCMeta):
     """
     Abstract base class for objects that satisfy the bluesky 'movable' interface.
-    
+
     Examples
     --------
-    
+
     .. code-block:: python
-    
+
         m = hw.motor
         # We need to detect if 'm' is a motor
         if isinstance(m, Movable):
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -1516,25 +1516,51 @@ def _rearrange_into_parallel_dicts(readings):
     return data, timestamps
 
 
+def is_movable(obj):
+    """Check if object satisfies bluesky 'movable' interface.
+
+    Parameters
+    ----------
+    obj : Object
+        Object to test.
+
+    Returns
+    -------
+    boolean
+        True if movable, False otherwise.
+    """
+    EXPECTED_ATTRS = (
+        'name',
+        'parent',
+        'read',
+        'describe',
+        'read_configuration',
+        'describe_configuration',
+        'set',
+        'stop',
+    )
+    return all(hasattr(obj, attr) for attr in EXPECTED_ATTRS)
+
+
 class Movable(metaclass=abc.ABCMeta):
     """
     Abstract base class for objects that satisfy the bluesky 'movable' interface.
-
     Examples
     --------
-
     .. code-block:: python
-
         m = hw.motor
         # We need to detect if 'm' is a motor
         if isinstance(m, Movable):
             print(f"The object {m.name} is a motor")
-
     """
     @classmethod
     def __subclasshook__(cls, C):
         # If the following condition is True, the object C is recognized
         # to have Movable interface (e.g. a motor)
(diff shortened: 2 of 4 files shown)


klauer:
What level of checking is "sufficient"?
- has attribute that matches
- is callable/not
- has compatible call signature
- etc

untzag:
@klauer being that I'm a total newbie I was trying to be minimally invasive

I like the idea of verifying callable signatures to the extent possible

ksunden:
Is `Movable` considered part of the "Public API"? if so, probably should keep the class and warn (probably in the subclasshook) rather than deleting entirely.

tacaswell:
Please leave `Moveable` is place but deprecate it.

I am :+1: to moving to `hasattr` rather than this ABC implementation, I think this was an early design experiment that we should back away from slowly....

tacaswell:
Anyone can dismiss my review.

prjemian:
I agree with hasattr and will learn from this.

untzag:
@tacaswell thanks for your review

I've added back the `Movable` class and now warn within `__subclasshook__` regarding future removal. I'd like input from maintainers on what type of warning it should be:

- true `DeprecationWarning` (invisible by default)
- base warning (seems like you do this in a few other places to indicate deprecation, e.g. `utils.install_qt_kicker`)
- `VisibleDeprecationWarning` (new warning class for bluesky? common convention to do)

danielballan:
> I think this was an early design experiment that we should back away from slowly....

Well, it was a recent addition. I suggested this approach based on a James Powell post advocating for this over `hasattr` checks because the semantics are clearer. But an `is_movable` function is nearly as good in that regard, and if there is a need for the added flexibly (adding attributes dynamically) I see no reason to block it. So, I’m convinced.

I haven’t encountered VisibleDeprecationWarning yet. Nice idea!

dmgav:
The `Movable` object is a recent addition. Does anyone actually using it?

bluesky/tests/test_utils.py
```diff
@@ -102,11 +102,11 @@ def test_cycler_merge_mixed(hw, children):
     assert mcyc.by_key() == cyc.by_key()
 
 
-def test_Movable(hw):
+def test_movable(hw):
```
dmgav, line 105:
```suggestion
def test_is_movable(hw):
```

dmgav: Just a minor change: we are testing the function `is_movable`.

bluesky/utils.py
```diff
@@ -1516,25 +1516,39 @@ def _rearrange_into_parallel_dicts(readings):
     return data, timestamps
 
 
+def is_movable(obj):
```
dmgav, line 1519: I think it's better to include a docstring as for `Movable` class.

untzag: thanks for catching that @dmgav

untzag:
(just squashed locally)

untzag:
@mrakitin @dmgav @tacaswell gentle bump---any chance for this to get merged in the near future?

ksunden:
@untzag, looks like you need to rebase/fix the merge conflict...

bluesky/utils.py
```diff
@@ -1536,25 +1536,51 @@ def _rearrange_into_parallel_dicts(readings):
...
+
 class Movable(metaclass=abc.ABCMeta):
     """
     Abstract base class for objects that satisfy the bluesky 'movable' interface.
-
     Examples
     --------
-
     .. code-block:: python
-
```
mrakitin, line 1547: Looks like this change may break proper rendering by Sphinx. Can it be reverted (and the removed lines above)? Thanks.

dmgav: I pushed a commit with inserted strings. I guess once CI passes we can merge it.

mrakitin:
Thanks @dmgav!

PR #1381, opened 2020-07-28
FIX: add draw_idle to LiveGrid and LiveScatter

tacaswell:
Previously we were relying on the qt kicker to trigger the re-draws,
move the logic into the callbacks instead.

This is both more robust (it is still safe for something external to
check the stale status and trigger e re-draw) and more-correct (the
callback object knows when it is back to a consistent state and should
be re-drawn).

## Description

Issue of LiveGrid not updating reported via informal channels from CSX

## How Has This Been Tested?

Tested manually.

```python
from bluesky import RunEngine
import bluesky.plans as bp
from ophyd.sim import *
plt.ion()

RE = RunEngine()
bec = BestEffortCallback()
RE.subscribe(bec)

motor.delay = 1
motor2.delay = .1

RE(bp.rel_grid_scan([det], motor, -5, 5, 25, motor2, -4, 4, 16))
```

### DIFF
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -24,6 +24,7 @@ def initialize_qt_teleporter():
     ------
     RuntimeError
         If called from any thread but the main thread
+
     """
     if threading.current_thread() is not threading.main_thread():
         raise RuntimeError(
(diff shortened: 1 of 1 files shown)


tacaswell:
Fixed the example.  We can probably wrap / mock the canvas to make sure `draw_idle` is called, but there are some things that are best tested by humans using it day-to-day...

mrakitin:
We would like to have it deployed to the NSLS-II experimental floor shortly. Merging.

mrakitin:
Thanks @tacaswell!

PR #1384, opened 2020-08-12
Give the RunEngine thread a useful name.

danielballan:
Now `threading.current_thread()` will show a Thread with the name
bluesky-run-engine instead of something like Thread-1.


danielballan:
Writing a code snippet in a conversation with @thomascobb prompted this.

PR #1389, opened 2020-08-24
WIP: Add flush and flush upon gc/exit.

### LINKED ISSUE #1388: PersistentDict doesn't support __setitem__ on nested dictionaries
Reported by @ambarb during CSX deployment.

## Expected Behavior
```
RE.md['sample'] = {}
RE.md['sample']['foo'] = 'bar'
RE.md
<PersistentDict {'sample': {'foo': 'bar'}}>
```

## Current Behavior

Fails quietly, without error messages:

```
RE.md['sample'] = {}
RE.md['sample']['foo'] = 'bar'
RE.md
<PersistentDict {'sample': {}}>
```

## Possible Solution

Needs further investigation. Something appears to go wrong in the base class `__setitem__` implementation.

See above. Reproducible with any instance of PersistentDict, RE.md not needed.

## Context

Just trying to set up the dictionary and change values after initial setup.

## Your Environment

CSX, but should be reproducible with any instance of PersistentDict.

danielballan:
**WIP -- does not quite work yet**

Fixes #1388.

_"Those who do not study history are doomed to repeat it."_

We had an identical issue on https://github.com/Nikea/historydict/pull/27 which, funnily enough, was _also_ reported by @ambarb and _also_ fixed by me, back in 2016.

### DIFF
--- a/bluesky/tests/test_persistent_dict.py
+++ b/bluesky/tests/test_persistent_dict.py
@@ -1,5 +1,6 @@
 import collections.abc
 import gc
+import pytest
 
 import numpy
 from numpy.testing import assert_array_equal
@@ -23,6 +24,12 @@ def test_persistent_dict(tmp_path):
     expected = dict(d)
     actual.reload()  # Force load changes from disk
     recursive_assert_equal(actual, expected)
+
+    # Test element deletion
+    del d['b']
+    with pytest.raises(KeyError):
+        d['b']
+    assert 'b' not in d
 
     # Smoke test the accessor and the __repr__.
     assert d.directory == tmp_path
--- a/bluesky/tests/test_persistent_dict.py
+++ b/bluesky/tests/test_persistent_dict.py
@@ -19,8 +19,13 @@ def test_persistent_dict(tmp_path):
     recursive_assert_equal(actual, expected)
 
     # Update a value and check again.
-    actual['a'] = 2
-    expected = dict(actual)
+    d['a'] = 2
+    expected = dict(d)
+
+    # Force sync changes from d to actual
+    d.flush()
+    actual.reload()
+
     recursive_assert_equal(actual, expected)
 
     # Smoke test the accessor and the __repr__.
--- a/bluesky/tests/test_persistent_dict.py
+++ b/bluesky/tests/test_persistent_dict.py
@@ -19,8 +19,8 @@ def test_persistent_dict(tmp_path):
     recursive_assert_equal(actual, expected)
 
     # Update a value and check again.
-    d['a'] = 2
-    expected = dict(d)
+    actual['a'] = 2
+    expected = dict(actual)
     recursive_assert_equal(actual, expected)
 
     # Smoke test the accessor and the __repr__.
--- a/bluesky/tests/test_persistent_dict.py
+++ b/bluesky/tests/test_persistent_dict.py
@@ -21,11 +21,7 @@ def test_persistent_dict(tmp_path):
     # Update a value and check again.
     d['a'] = 2
     expected = dict(d)
-
-    # Force sync changes from d to actual
-    d.flush()
-    actual.reload()
-
+    actual.reload()  # Force load changes from disk
     recursive_assert_equal(actual, expected)
 
     # Smoke test the accessor and the __repr__.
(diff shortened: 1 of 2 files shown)


MikeHart85:
I think this should do it. Somewhat similar to our previous adventures with cleanup and finalizers in ophyd.

Note that even avoiding self reference in the finalizer, and even using `del` and `gc.collect`, I wasn't always able to force an implicit finalization before actual Python shutdown. This means `flush()` should be called explicitly if immediate sync to disk is required.

For example, the following sequence called the finalizer as expected (before shutdown):

```python
>>> from bluesky.utils import PersistentDict
>>> d = PersistentDict('/tmp/foobar123')
>>> d['a'] = 1
>>> del d
(finalized here)
>>>
```

But just one extra access prevented this from working:

```python
>>> from bluesky.utils import PersistentDict
>>> d = PersistentDict('/tmp/foobar123')
>>> d['a'] = 1
>>> d
<PersistentDict {'a': 1}>
>>> del d
(not finalized)
>>> import gc
>>> gc.collect()
(still not finalized)
>>> exit
(finalized here)
```

MikeHart85:
Hm, thinking about it some more, is the change to the unit test I made correct?

Or was that kind of syncing between two `PersistentDict`s using the same directory actually expected? If so, the problem is that the cache prevents loading from disk on every access. So even if we flush `d`, `actual` won't see the change because it only looks at its own `_cache`.

MikeHart85:
An alternate solution, adding a `reload` method to mirror `flush` and allow manual syncing.

It's not great, since you have to know to call them, so we lose the intuitive updates between them. Not sure there's a better way though, short of wrapping every single value to catch all updates to mutables.

bluesky/tests/test_persistent_dict.py
```diff
@@ -20,13 +21,35 @@ def test_persistent_dict(tmp_path):
     # Update a value and check again.
     d['a'] = 2
     expected = dict(d)
+
+    # Force sync changes from d to actual
+    d.flush()
```
danielballan, line 26: This is technically not needed because we are setting a top level key, right?

MikeHart85: You're right, just the reload is needed.

MikeHart85:
Grumble. Thinking about deletions, there's actually a few things that can still go wrong. E.g.:

```python
>>> d = PersistentDict('/tmp/foobar')
>>> d['a'] = 1
>>> d['b'] = 2
>>> d
<PersistentDict {'a': 1, 'b': 2}>
>>> del d['b']
>>> d
<PersistentDict {'a': 1}>
>>> !ls /tmp/foobar/
a
# Looking good... but...
>>> d._cache
{'a': 1, 'b': 2}
>>> d['b']
2
# Sadface
```

(`b` also reappears on disk once `d` is finalized)

Easy enough to fix, by overriding `__delitem__`, but similar issues are going to pop up with all the other methods in `zict.Func`. Boils down to some of the methods relying on _cache, and others going straight for filesystem. 

I think fixing this would require reimplementing pretty much all of `zict.Func`. Might as well skip it and go for `ZictBase`? Is that even worth it? Maybe roll our own?

danielballan:
Maybe roll our own and/or vendor whatever parts of zict are still applicable, if any.

danielballan:
I think zict's emphasis is its ability to make two-tier mappings, where some items are backed by "fast" storage and some by "slow" storage. This is how dask "spills" to disk transparently. We're not really using that aspect, so we may be better of still ideas and patterns from zict than importing it directly.

MikeHart85:
In the current state, this should now fully cover the reported issue, as well all *common* usage patterns I can think of. 

You can still encounter cache sync issues if you start doing unusual things... but it should be possible to work around them by using `flush` and/or `reload`.

Since we're trying to push this out quickly, do you think it would be reasonable to leave the rest as is for this PR?

tacaswell:
I think that with the remote work that is coming we should radically re-consider how we manage this sort of meta-data.

danielballan:
Agreed! It's possible that the built-in bluesky solution for this should just an in-memory dict, and we should make these fancy syncing dicts a deployment concern.

PR #1392, opened 2020-08-27
mark stop as optional for settable hardware

untzag:
Currently bluesky requires objects to implement `stop` to pass the `is_moveable` check [1]. However if you look at RunEngine behavior it's clear that you expect devices that cannot be stopped to _not_ implement a stop method [2]. This PR attempts to clean up this discrepancy by making `stop` optional explicitly.

[1] https://github.com/bluesky/bluesky/blob/master/bluesky/utils.py#L1608
[2] https://github.com/bluesky/bluesky/blob/master/bluesky/run_engine.py#L1219-L1230

### DIFF
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -1605,7 +1605,6 @@ def is_movable(obj):
         'read_configuration',
         'describe_configuration',
         'set',
-        'stop',
     )
     return all(hasattr(obj, attr) for attr in EXPECTED_ATTRS)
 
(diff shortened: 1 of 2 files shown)


untzag:
Now that I think about it, should we be documenting `success` as an arg to stop? :thinking: What does `success` even do?

danielballan:
We should be documenting that as a required argument to `stop`. It signals whether we are stopping in the course of some planned procedure or because of some error/failure condition. It allows the device to respond differently to these two cases—perhaps more defensively to the latter, to “make itself safe”. I am not sure there exists any example where this is acted on, in practice. If I were designing it today I would consider passing an Exception object rather than a boolean, so as to give more detail about the nature of the failure.

untzag:
thanks for the context @danielballan :smile:

PR #1395, opened 2020-08-28
Fix/bec use teleporter

aegger13:
Allow for QtAwareCallback to get use_teleporter arg from kwargs.  There was no way to pass this in BestEffortCallback
## Description

get use_teleporter arg from kwargs. 
## Motivation and Context

This could be a local environment issue, but if use_teleporter is None, we get a segfault trying to check if 'qt' is in mpl backend.  When we know we're going to use the teleporter, it'd be nice to be able to pass this to BEC kwargs to QtAwareCallback

## How Has This Been Tested?

from ophyd.sim import motor1, det1
from bluesky import RunEngine
from bluesky.callbacks.mpl_plotting import initialize_qt_teleporter
from Bluesky.plans import scan
RE = RunEngine({})
initialize_qt_teleporter()
bec = BestEffortCallback(). # Change BestEffortCallback(use_teleporter=True)
RE.subscribe(bec)
RE(bp.scan([det1], motor1, -10, 10, 21))
This caused segfault when use_teleporter is None.  The usual tk not finding main loop on main thread (again, this seems to be an environment issue for our NFS machines)
Now this produces live plot no problem.

We are using the latest pcds_conda environment https://github.com/pcdshub/pcds-envs

All changes should be benign, everything functions the same way if you don't pass in the use_teleporter arg.  I've also made sure the plotting utilities (LivePlot, etc...) still function as they should.

### DIFF
--- a/bluesky/callbacks/mpl_plotting.py
+++ b/bluesky/callbacks/mpl_plotting.py
@@ -58,7 +58,8 @@ def _get_teleporter():
 
 
 class QtAwareCallback(CallbackBase):
-    def __init__(self, *args, use_teleporter=None, **kwargs):
+    def __init__(self, *args, **kwargs):
+        use_teleporter = kwargs.get('use_teleporter', None)
         if use_teleporter is None:
             import matplotlib
             use_teleporter = 'qt' in matplotlib.get_backend().lower()
@@ -66,7 +67,7 @@ class QtAwareCallback(CallbackBase):
             self.__teleporter = _get_teleporter()
         else:
             self.__teleporter = None
-        super().__init__(*args, **kwargs)
+        super().__init__()
 
     def __call__(self, name, doc, *, escape=False):
         if not escape and self.__teleporter is not None:
@@ -117,7 +118,7 @@ class LivePlot(QtAwareCallback):
     """
     def __init__(self, y, x=None, *, legend_keys=None, xlim=None, ylim=None,
                  ax=None, fig=None, epoch='run', **kwargs):
-        super().__init__(use_teleporter=kwargs.pop('use_teleporter', None))
+        super().__init__(**kwargs)
         self.__setup_lock = threading.Lock()
         self.__setup_event = threading.Event()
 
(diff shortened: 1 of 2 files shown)


bluesky/callbacks/mpl_plotting.py
```diff
@@ -117,7 +118,7 @@ class LivePlot(QtAwareCallback):
     """
     def __init__(self, y, x=None, *, legend_keys=None, xlim=None, ylim=None,
                  ax=None, fig=None, epoch='run', **kwargs):
-        super().__init__(use_teleporter=kwargs.pop('use_teleporter', None))
+        super().__init__(**kwargs)
```
tacaswell, line 121: The kwargs here are aimed at being passed through to the line plot (~L178),

bluesky/callbacks/mpl_plotting.py
```diff
@@ -58,15 +58,16 @@ class Teleporter(QtCore.QObject):
...
+        use_teleporter = kwargs.get('use_teleporter', None)
         if use_teleporter is None:
             import matplotlib
             use_teleporter = 'qt' in matplotlib.get_backend().lower()
         if use_teleporter:
             self.__teleporter = _get_teleporter()
         else:
             self.__teleporter = None
-        super().__init__(*args, **kwargs)
+        super().__init__()
```
tacaswell, line 70: Any stray args/kwargs should be passed on to the super, not dropped on the floor here.

danielballan:
Thanks for the contribution, @aegger13. I agree with @tacaswell’s assessment that this would break things. Can you elaborate more about the problem you are seeing? Could it be that you are using matplotlib’s tk backend, whereas these objects are designed to integrate with Qt?

ZLLentz:
The goal of this PR to was to fix a problem we had at SLAC/LCLS where any `LivePlot` we created (on some machines...) caused a segfault that implied an issue with the teleporter, despite configuring matplotlib and setting up the teleporter.

We later realized that the thing we got wrong was the order: if we set the matplotlib backend first, then create the teleporter, then create the BEC, everything is fine.

I think we should close this PR, but consider updating the documentation for https://blueskyproject.io/bluesky/callbacks.html. That page is fairly out of date and still references the qt kicker.

danielballan:
Thanks, @ZLLentz. That's been somewhere on my to-do list for awhile, but I opened https://github.com/bluesky/bluesky/issues/1403 to track it.

tacaswell:
Some of that issue is also due to bugs in Matplotlib (that are now fixed) with how the automatic backend selection works...

ZLLentz:
Thanks for the info @tacaswell, we're not using the latest matplotlib but I do not remember why. I'll look to updating our install.

PR #1397, opened 2020-08-29
DOC: fix scan_nd docstring

untzag:
Current docstring led me down the wrong path as someone with no prior familiarity with the cycler package.


danielballan:
Awesome. Thank you!

danielballan:
And sorry for leading you down the wrong path—the original, erroneous docstring was almost definitely my doing. 🐑

PR #1400, opened 2020-08-30
DOC/MNT: add configure count time wrapper / decorator to docs

tacaswell:
## Description

configure_count_time_wrapper and configure_count_time_decorator
have existed from 2017, but were not included in the built docs.

Also change to use rd internally.

## Motivation and Context

This is something that is frequently asked for and we already have the machinery extracted from the (now removed) spec API.  in ~2017 we generalized it but have not been using it.

## How Has This Been Tested?

The wrapper already has tests in the test suite.

### DIFF
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -9,7 +9,7 @@ from .utils import (normalize_subs_input, root_ancestor,
                     short_uid as _short_uid, make_decorator,
                     RunEngineControlException, merge_axis)
 from functools import wraps
-from .plan_stubs import (open_run, close_run, mv, pause, trigger_and_read)
+from .plan_stubs import (open_run, close_run, mv, pause, trigger_and_read, rd)
 
 
 def plan_mutator(plan, msg_proc):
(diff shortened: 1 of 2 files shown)


tacaswell:
I am not sure that this is actually the best way to do this as it assumes that there is a single signal that we can set to adjust the count time, it may be better to have a protocal of checking for am `obj.adjust_my_count_time_plan` on the objects something like

```python
def adjust_count_time(self, target: float) -> PlanFunction:
     current_time = yield from bps.rd(self.count_time)
     yield from bps.mv(self.count_time, target)
     return partial(self.adjust_count_time, target=current_time)  # return how to "undo" the change
```

Could also see a case for returning `(old_time, new_time)` as a tuple and assuming that we can use the same plan method to re-set it to the old one?

prjemian:
Could it be a dict instead, since some detectors have multiple attributes to be configured?  Keys would be as in `stage_sigs` (possibly referring to some.deeper.subdevice.with.count_time in the Device).

danielballan:
@prjemian: Abstract away from count time, exposure time to "time for measurement". Make it the responsibility of the ophyd Device to handle this rather than pushing it into plans.

@danielballan: Maybe should think of ophyd interface "conventions" for specialized cases like this, a tier of standard methods below "required" and even below the universal "optional" ones like `stage`.

@ksunden: Yaq has a similar notion, "traits": https://yaq.fyi/traits/ 

@prjemian: Implement as mix-in class.

PR #1406, opened 2020-09-22
ENH: add message to run a co-routine

tacaswell:
## Description

Adds a default Msg to run a (arbitrry) user supplied co-routine.

Leaving as draft until we think this is actually a good idea.

## Motivation and Context

We have talked about this on-and-off at it turned out to be super simple to implement...

## How Has This Been Tested?

Has a new test.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -418,7 +418,9 @@ class RunEngine:
             'wait_for': self._wait_for,
             'input': self._input,
             'install_suspender': self._install_suspender,
-            'remove_suspender': self._remove_suspender, }
+            'remove_suspender': self._remove_suspender,
+            'run_coroutine': self._run_coro,
+        }
 
         # public dispatcher for callbacks
         # The Dispatcher's public methods are exposed through the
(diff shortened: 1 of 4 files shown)


danielballan:
This runs any awaitable on an asyncio event loop. In the far future it might support a different kind of loop as well or instead (such as trio). I suggest the name `'await'`.

bluesky/run_engine.py
```diff
@@ -2263,6 +2265,16 @@ async def _input(self, msg):
...
 
+    async def _run_coro(self, msg):
+        """
+        Run a user-supplied co-routine in the RE's event loop
+
+            Msg('run_coroutine', coro, a, b=3)
+
+        """
+        _, coro, args, kwargs, _ = msg
+        return await coro(*args, **kwargs)
```
tacaswell, line 2276: Do we want to offer `self` to the co-routine?

danielballan:
On community call, @ZLLentz remembers having some use cases for this.

danielballan:
@tacaswell Can you recall and list some use cases here to help us judge whether this hook is useful or maybe too large of an escape hatch?

tacaswell:
The main things I had in mind were the ability to run extra I/O (e.g. post to slack / query a database for metada) in a way that will _not_ be executed in simulation mode.

PR #1407, opened 2020-09-24
log when emitting resource and datum documents

### LINKED ISSUE #1404: Bundler does not log when emitting Resource of Datum
When all other document types except 'resource' and 'datum' are emitted, DEBUG log messages are emitted as well with some details useful for debugging (but intentionally _not_ the whole document). The pattern should be extended to these document types as well. This was likely an oversight during the logging overhaul.

jklynch:
## Description
Log a debug-level message when resource and datum documents are emitted.

Documentation has been expanded and variables renamed for clarity in the affected method `save`.

## Motivation and Context
Closes #1404.

## How Has This Been Tested?
All tests pass. No new tests added.

### DIFF
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -338,6 +338,12 @@ class RunBundler:
     async def save(self, msg):
         """Save the event that is currently being bundled
 
+        Create and emit an Event document containing the data read from devices
+        in self._objs_read. Emit any Resource and Datum documents cached by
+        those devices before emitting the Event document. If this is the first
+        Event of its stream then create and emit the Event Descriptor document
+        before emitting Resource, Datum, and Event documents.
+
         Expected message object is::
 
             Msg('save')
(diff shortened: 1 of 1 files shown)


bluesky/bundlers.py
```diff
@@ -408,39 +415,60 @@ async def save(self, msg):
...
+                resource_or_datum_doc["run_start"] = self._run_start_uid
+
+            doc_logger.debug(
+                "[%s] document emitted %r",
+                resource_or_datum_name,
+                resource_or_datum_doc,
+                extra={
+                    "doc_name": resource_or_datum_name,
+                    "run_uid": self._run_start_uid,
+                    "doc": resource_or_datum_doc
```
danielballan, line 448: When we last revamped logging here, there was lengthy discussion about how much to include. Considerations are readability of logs and also performance. (Making a large dict into a string for writing can impose noticeable overhead.) We decided to include the document name, the current Run uid, and any particular info especially likely to be useful for debugging, such as the data keys in an event or descriptor. In this case, I think we are justified in emitting the entire document for Resource and Datum because:

1. These document types never grow very large.
2. All of the information in the document is likely useful for debugging. No field stands out as less important than the others in this context.

danielballan (approved):

dmgav (approved):

mrakitin (approved):

PR #1408, opened 2020-09-25
Declare matplotlib as optional dependency

mrakitin:
Matplotlib is used by `BestEffortCallback` and other vis-callbacks. We need it as a dependency here. Relevant discussion started in https://github.com/bluesky/bluesky-queueserver/pull/35/files#r489571206 and then in the DAMA teams.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 cycler
 event-model>=1.14.0
 historydict
+matplotlib
 msgpack
 msgpack-numpy
 numpy


tacaswell:
I am :-1:  on this.  We intentionally did not add this as a core dependency as it is _only_ used in some of the viz callbacks (and we check if it is already imported to get the qt shims).  If we are going to add something like this is should be in `install_extras`.

tacaswell (changes requested):
This was an intentional omission.

mrakitin:
I am OK to continue with `install_extras`. Can we reopen?

danielballan:
Sure. If we are adding extras there are a couple others we might want to declare, like pyzmq.

PR #1412, opened 2020-10-15
check limits should call objects' check_value() method

### LINKED ISSUE #1411: check_limits() should call objects' check_value() method
The `bluesky.simulators.check_limits()` needs to be revised.

## Expected Behavior

The `bluesky.simulators.check_limits()` acts on `set` messages in a plan, checking that the position will be within the limits allowed by the object.

## Current Behavior

The code now gets the `.limits` tuple from the object and does a comparison of position within those limits.  It raises an exception if there is a problem.  https://github.com/bluesky/bluesky/blob/09bd73bd05ac353ae9b2cc4ea4f015c8b934e658/bluesky/simulators.py#L83-L115

This is a problem for many objects, as noted by this GitHub search: https://github.com/search?p=1&q=org%3Abluesky+check_limits&type=Issues

Notably, the [*hklpy/diffract/Diffractometer*](https://github.com/bluesky/hklpy/pull/45) object has two types of positioner, pseudo and real.  The real positioners derive their limits from real positioners, such as EPICS motors.  The pseudo-positioners are reciprocal-space terms.  For these terms, the limits are computed dynamically, based on current values of all the real positioners.  Thus, the simple code implemented in  `bluesky.simulators.check_limits()` [fails](https://github.com/bluesky/hklpy/issues/13) for bluesky plans that position `Diffractometer` pseudo-axes.

## Possible Solution

The `check_limits` code should call each objects `check_value(pos)` method.  Objects with special needs should override the base class `check_value()` method and raise exceptions if there is a problem.

The new code will be much simpler:

```
def check_limits(plan):
    """
    Check that a bluesky plan will not move devices outside of their limits.

    Parameters
    ----------
    plan : iterable
        Must yield `Msg` objects
    """
    for msg in plan:
        if msg.command == 'set':
            msg.obj.check_value(msg.args[0])
```

see various GH issues as noted: https://github.com/search?p=1&q=org%3Abluesky+check_limits&type=Issues

## Context

The impact of this change is important.  One of the selling points of the Bluesky framework is the ability to inspect a plan before executing it.  Some of the [issues noted above](https://github.com/search?p=1&q=org%3Abluesky+check_limits&type=Issues) propose adding a `check_limits()` call as a routine practice.

## Your Environment

```
In [25]: RE.md["versions"]                                                                                                                                    
Out[25]: 
{'apstools': '1.3.7+43.g59c4610',
 'bluesky': '1.6.5',
 'databroker': '1.0.6',
 'epics': '3.4.2',
 'event_model': '1.15.2',
 'matplotlib': '3.2.2',
 'numpy': '1.19.1',
 'ophyd': '1.5.2',
 'pyRestTable': '2020.0.2',
 'spec2nexus': '2021.1.7'}
```

prjemian:
Fixes #1411 which provides details.

## Description
Revise to call each object's `check_value(pos)` method.  Removed local check of `msg.obj.limits`.

Note that `check_limits()` cannot know if same limits are a problem.  Also it cannot know if limits are exceeded if the object does not raise.

## Motivation and Context
see #1411

## How Has This Been Tested?
Tested this code with the `hklpy.diffract.Diffractometer` class.  Other projects may also need to be tested with this code.

### DIFF
--- a/bluesky/simulators.py
+++ b/bluesky/simulators.py
@@ -88,28 +88,7 @@ def check_limits(plan):
     ----------
     plan : iterable
         Must yield `Msg` objects
-
-    Raises
-    ------
-    LimitsExceeded
     """
-    ignore = []
     for msg in plan:
         if msg.command == 'set':
-            if msg.obj in ignore:
-                continue  # we have already warned about this device
-            try:
-                low, high = msg.obj.limits
-            except AttributeError:
-                warn("Limits of {} are unknown and can't be checked.".format(msg.obj.name))
-                ignore.append(msg.obj)
-                continue
-            if low == high:
-                warn("Limits are not set on {}".format(msg.obj.name))
-                ignore.append(msg.obj)
-                continue
-            pos, = msg.args
-            if not (low < pos < high):
-                raise LimitsExceeded("This plan would put {} at {} "
-                                     "which is outside of its limits, {}."
-                                     "".format(msg.obj.name, pos, (low, high)))
+            msg.obj.check_value(msg.args[0])
--- a/bluesky/simulators.py
+++ b/bluesky/simulators.py
@@ -88,9 +88,9 @@ def check_limits(plan):
     ignore = []
     for msg in plan:
         if msg.command == 'set' and msg.obj not in ignore:
-            if hasattr(msg.obj, "limits"):
+            if hasattr(msg.obj, "check_value"):
                 msg.obj.check_value(msg.args[0])
             else:
-                warn("Limits of {} are unknown and can't be checked."
-                     .format(msg.obj.name))
+                warn(f"{msg.obj.name} has no check_value() method"
+                     f" to check if {msg.args[0]} is within its limits.")
                 ignore.append(msg.obj)
--- a/bluesky/simulators.py
+++ b/bluesky/simulators.py
@@ -76,10 +76,6 @@ def summarize_plan(plan):
 print_summary = summarize_plan  # back-compat
 
 
-class LimitsExceeded(Exception):
-    ...
-
-
 def check_limits(plan):
     """
     Check that a plan will not move devices outside of their limits.
--- a/bluesky/simulators.py
+++ b/bluesky/simulators.py
@@ -85,12 +85,12 @@ def check_limits(plan):
     plan : iterable
         Must yield `Msg` objects
     """
-    ignore = [] 
+    ignore = []
     for msg in plan:
         if msg.command == 'set' and msg.obj not in ignore:
             if hasattr(msg.obj, "limits"):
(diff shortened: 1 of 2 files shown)


bluesky/simulators.py
```diff
@@ -88,28 +84,13 @@ def check_limits(plan):
...
-            try:
-                low, high = msg.obj.limits
-            except AttributeError:
-                warn("Limits of {} are unknown and can't be checked.".format(msg.obj.name))
-                ignore.append(msg.obj)
-                continue
-            if low == high:
-                warn("Limits are not set on {}".format(msg.obj.name))
+        if msg.command == 'set' and msg.obj not in ignore:
+            if hasattr(msg.obj, "limits"):
```
tacaswell, line 91: should his be checking for `"check_value"` instead of `"limits"`

prjemian: I agree.  This assumes nothing else.

tacaswell:
Do we want to keep both letting objects that have these brains use them and falling back to naive limit checking if not?

prjemian:
I don't think we should.

For example, one of the unit tests I removed takes an object which has no `.limits` attribute, adds it, then tests with it.  If the object has no implementation of `.check_value()`, why would adding any attribute unexpected by that object have any particular effect?  After some experimentation, I found that the `check_value()` support is rather strong, even the `Diffractometer` object worked by default when a proper position tuple is passed.

We are emphasizing the responsibility of any movable class to override `check_value()` as needed and raise an appropriate exception for cases where the proposed limit would fail.

tacaswell:
fair enough.  The prototype for `check_value` is on `OphydObject` (which means it is definitely defined on _every_ `Device` subclass) where as `obj.limits` is first defined on `PostionerBase`.

prjemian:
What about Signal?  Certainly that is an object for the `set` Msg?

tacaswell:
`Signal` also inherits from `OphydObj` (but not from `PositionerBase`) but does include a `limits`.

bluesky/simulators.py
prjemian: Can we use an f-string in the `warn()` here?

prjemian: The line length is 79.  f-string would save 8 characters.

tacaswell:
I remember there being some controversy about how exactly do to limits / value checking early on but none of the details (other than @klauer and I did not agree on some things).

prjemian:
Understood.  I was expecting to find a boolean method indicating if a position was `in_limits()`.  The name `check_value()` was not intuitive but I learned to work with it rather than fight it.  I like the idea of raising appropriate exceptions instead of returning simple `False`.

tacaswell (approved):

PR #1417, opened 2020-11-04
Mention check_value() in bluesky interface docs

danielballan:
## Description

In #1412 we effectivley made ``check_value(...)`` an optional part of
the "bluesky interface" that movable devices should implement. This adds
it to the documentation on Movables and makes a couple other refinements
in the process.


prjemian (approved):
LGTM!  Thanks.

danielballan:
Travis-CI is continually backed up and I don't want to hold the release on it. Verified locally that the docs build with warnings and are formatted as expected.

image

PR #1420, opened 2020-11-06
Add github action for autorelease via CI

ksunden:
## Description
Adds a github action which builds and uploads sdist and wheel to pypi.

With this applied, a release can be initiated using the github release in the web ui (or api, eg the `gh` cli).

I do _not_ think pushing a tag alone triggers this, but have not actually tested that.

This does require someone with proper permissions adding Secrets via github settings for PYPI_USERNAME and PYPI_PASSWORD (I use a pypi token with limited scope for such things).

Since bluesky is pure python, universal wheel (well, py3 only, but noarch, no abi specified) should work, and matches what is uploaded currently (though this could be modified should that change).

## Motivation and Context
This follows on from some conversations in the slack with @danielballan with an eye towards making releasing more automated.

That said, this does a real minimum of `setup.py sdist bdist_wheel` and `twine upload`.

## How Has This Been Tested?
I used this with another python repository.
That said, I do not actually know if more is done for bluesky releases. If it is, it can be added. In particular, the repo I used this with did not use versioneer (and I have limited experience with it, though it looks like it pulls the version from the tag itself, which should work, I expect)

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,31 @@
+# This workflows will upload a Python Package using flit when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install wheel twine setuptools
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+      run: |
+        python setup.py sdist bdist_wheel
+        twine upload
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -24,7 +24,9 @@ jobs:
         pip install wheel twine setuptools
     - name: Build and publish
       env:
-        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_USERNAME: __token__
+        # The PYPI_PASSWORD must be a pypi token with the "pypi-" prefix with sufficient permissions to upload this package
+        # https://pypi.org/help/#apitoken
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
       run: |
         python setup.py sdist bdist_wheel


ksunden:
If desired, I can add this to other repos as well, it is just copying... there are also ways of setting up org wide, but I haven't looked too closely at those (And that involves making a new repo)

danielballan:
@ksunden This somehow slipped past our notice until now. Apologies for the very slow follow-up. I look forward to digging into this in the new year.

danielballan:
Back from the holidays, finally looking at this closely.

I like that this places the publication step in a workflow by itself, separate from tests and documentation building/publishing. In this way, publication is not contingent on other steps. I wouldn't want a random flaky test failure or ill-timed dependency problem to interfere with publication, for example, as might be the case if they were placed as different steps in a shared workflow. (Any real test failures should have been caught _before_ we go to the point of pushing the tag.)

I also like that it runs contingent on making a GH release (including the "prerelease" option). We (most especially I) have been lazy/inconsistent about bothering to do this, and this will ensure that we don't forget that step.

I think we should strongly consider using [the official PyPA GitHub Action](https://github.com/marketplace/actions/pypi-publish) to handle this because it will abstract away any future changes to best practices around the usage of twine etc. This isn't exactly a _fast_-moving space, but there are a lot of improvements being discussed, and it would be convenient if the upgrade path for tracking evolving best practices were as simple as updating the version the action that we use. Here is an example from our friend @thomascobb: https://github.com/dls-controls/scanspec/blob/8aed8ebdfeb75062ff8cb7d5e36dd98e1e3c3da1/.github/workflows/build.yml#L89-L97

Finally, I would prefer to use a deploy token rather than a username/password. Contributors come and go (and sometimes update their credentials) but the token is scoped to the _project_.

ksunden:
My personal preference for these things when they are really just a few commands is to have it be flat and readable, so I can see exactly whats going on, I find that easier to debug, and/or change (e.g. some of my projects use flit, which has its own `publish` command built in).

All that said, if you prefer using that, I'm happy to rework the PR.

As for the token, there is no real difference to the action, you just use the secrets menu to set `PYPI_USERNAME=__token__` and `PYPI_PASSWORD=the actual token`.

In this case, the username is not at all "secret", in reality, but it is the most flexible route.

thomascobb:
Personally I prefer to only release from a tag when the tests pass. Since I moved to asyncio for most of my projects I don't seem to have problems with flaky test anymore...

Also, what benefits do you get from doing a GitHub release on a Python project? I've used them when I need to add built artifacts to a release, but never on a Python projects as the artifacts are on PyPI...

danielballan:
We get flaky tests around some (deprecated but still tested) SQLite code in databroker, various tests that involve networking without asyncio in caproto, and very occasionally the zmq + asyncio tests here in bluesky. I could accept the argument that we should just _fix_ that somehow rather than let it distort how we run our tests.

About GH releases, I think others in my group have argued for it just for the sake of completeness. It also figured prominently in the GH UI now, and people might look to it to check the latest release. I find the necessity a little irritating because, as you say, we already have tags and the releases don’t add anything in this situation except another chore, another step in the process.

@ksunden Let’s switch to the __token__ approach and then I’ll be happy to merge this. I’ll upload a token to the secrets. We can always change course as needed as we see how this feels to use and receive any broader feedback.

thomascobb:
Looks like you can automatically make a release with an action:
https://github.com/actions/create-release

I'll try that out...

prjemian:
FWIW: there is a semantic distinction between tags and releases that has
strong value to the user.

Here's how I see it.

* Tags represent checkpoints
* Releases (fewer number than tags) mark a notable tag
* Milestones associate with specific goals, such as releases

ksunden:
closing/reopening to rerun tests, test failure is unrelated, surely (as I didn't change any library code) but I like green checkmarks.

danielballan:
I have generated a PyPI token and added it as a repository secret. We'll find out if this works when we do our next release. :-D

Thanks again for pushing on this, @ksunden.

PR #1423, opened 2020-11-30
Progress bars in Jupyter Notebook

callumforrester:
# WIP: Progress bars in Jupyter Notebook

## Description

Adds a progress manager that is compatible with Jupyter notebooks and Jupyterlab.
WIP solution as discussed with @danielballan 

Summary of changes:
* Added a NotebookProgressBar class that manages tqdm.notebook's progress bar widgets, aims for broadly the same behaviour as ProgressBar with ASCII bars. 
* Modified ProgressBarManager so that it can use either ProgressBar (default) or NotebookProgressBar

## Motivation and Context

The standard tqdm ASCII progress bars do not work well in Jupyter; they are not cleared up properly and they repeat every time they change, e.g.
```
30%    [---     ]
60%    [-----   ]
100%   [--------]
``` 
tqdm provides a solution with its notebook module, which uses Jupyter widgets rather than ASCII output. But it requires a custom wrapper class in Bluesky because it's behaviour/interface is different.

## How Has This Been Tested?

Testing has been limited. As discussed with @danielballan this is an ad-hoc solution to build on.

I have run a number of plans in a Jupyter notebook with these bars including:
* scan
* mv
* mvr
* adaptive_scan
* grid_scan
* count
* abs_set

Environment:
Python 3.8
Bluesky 1.6.7
Jupuyterlab 1.2.6

## Screenshots:
image


danielballan:
This is great! Thanks for your patience with the slow turnaround @callumforrester. We're normally better about promptly reviewing contributions. The unplanned migration from Travis-CI to GitHub Actions, due to a sudden change in Travis-CI pricing, has put most everything else on hold.

Power-cycling this to re-run it against the new GitHub Actions CI.

danielballan:
Some of the CI failures are due to the use of `dataclasses`, which was introduced in Python 3.7. Per [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), we can drop support for Python 3.6 in major or minor releases as of November 2020, so I think the right way forward here is to drop the 3.6 build. (If we need to sustain support for 3.6 for some special reason, we could require the backport of the `dataclasses` module https://pypi.org/project/dataclasses/.)

This test failure (from the 3.7 build) seems more important:

```
_______________________________ test_mv_progress _______________________________

RE = <bluesky.run_engine.RunEngine object at 0x7f7978201190>
hw = namespace(ab_det=ABDetector(prefix='', name='det', read_attrs=['a', 'b'], configuration_attrs=[]), bool_sig=Signal(nam...name='signal', value=0, timestamp=1608069126.6235502), trivial_flyer=<ophyd.sim.TrivialFlyer object at 0x7f79782f9450>)

    def test_mv_progress(RE, hw):
        motor1 = hw.motor1
        motor2 = hw.motor2
    
>       assert RE.waiting_hook.delay_draw == 0.2
E       AttributeError: 'ProgressBarManager' object has no attribute 'delay_draw'

bluesky/tests/test_progress_bar.py:90: AttributeError
```

The `delay_draw` feature is designed to avoid showing short-lived progress bars for actions that complete very quickly.

callumforrester:
No problem Dan, I've had another look at this and removed the dataclass since it didn't serve much of a purpose.
Regarding the delay time, I re-added it to NotebookProgressBar but not ProgressBarManager. ProgressBarManager now takes a factory function that generates appropriate progress bars. My view was that specific parameters should be included there. For example:

```
def default_notebook_progress_bar(status_objs_or_none) -> ProgressBar:
    return NotebookProgressBar(status_objs_or_none, delay_draw=0.325)

my_manager = ProgressBarManager(default_notebook_progress_bar)
```

That way you don't have to pass delay_draw to some future ProgressBar implementation where it doesn't make sense.
I also added a helper method to make this easier:

```
my_manager = ProgressBarManager.for_notebook(delay_draw=0.325)
```

I'm happy to change it, let me know your thoughts.

bluesky/utils.py
```diff
@@ -14,12 +16,16 @@
 import abc
 from collections.abc import Iterable
 import numpy as np
+from IPython.core.display import display, HTML
```
danielballan, line 19: This module needs to be importable without IPython installed. (IPython is an optional requirement.) We could move this to a separate module or we could just hide these imports inside the functions methods where they are used. I'd be inclined toward the latter in this case, but either is fine by me.

Ditto for the other new imports below.

bluesky/utils.py
```diff
@@ -1149,7 +1155,28 @@ def apply_to_dict_recursively(d, f):
...
+            time_elapsed: float = None,
+            time_remaining: float = None,
+    ):
+        ...
+
+    def clear(self):
+        ...
+
+
+class AsciiProgressBar(ProgressBar):
```
danielballan, line 1179: This is definitely a better name, but we should also retain `ProgressBar` as a backward-compatible alias to `AsciiProgressBar`, at least for awhile. A lot of code will need to be updated. Maybe the ABC could be called `ProgressBarBase` or something.

bluesky/utils.py
```diff
@@ -1258,19 +1285,231 @@ def clear(self):
...
+        Manages creation and tearing down of progress bars.
+        :param pbar_factory: A function that creates a progress bar given an optional list of status objects
+        :return:
+        """
+
+        self.pbar_factory = pbar_factory
         self.pbar = None
 
+    @classmethod
+    def for_notebook(cls, delay_draw: float = 0.2) -> 'ProgressBarManager':
```
danielballan, line 1490: Should the `delay_draw` parameter be removed?

bluesky/utils.py
```diff
@@ -1258,19 +1285,231 @@ def clear(self):
...
+            self,
+            pos: int,
+            label: str = "",
+            meta: str = "",
+            color: str = "#97d4e8",
+            total: float = 1.0,
+            value: Optional[float] = None,
+    ) -> None:
+        """
+        Draws the progress bar or a message if there is
```
danielballan, line 1389: We use [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) style rather than sphinx. I don't mind pushing a commit to do the reformatting myself if you'd like.

danielballan (commented):
I like your approach. I left a couple minor comments.

bluesky/utils.py
danielballan: Actually, to the extent that this is _sometimes_ handled by beamline scientists, would `TextProgressBar` or `TerminalProgressBar` be approximately as descriptive and a little more accessible?

bluesky/utils.py
callumforrester: Whoops!

bluesky/utils.py
callumforrester: I think TerminalProgressBar is more descriptive, renamed. 
How do scientists use this class? My assumption was that they would only touch ProgressBarManager

bluesky/utils.py
callumforrester: Hmm, possibly. The purpose of adding a factory function was to make it easier to change in the future. The purpose of this method is to quickly create a suitable manager in your startup script for a notebook.

The constructor currently retains the default behaviour for a similar use case in a terminal. If you call `ProgressBarManager()` the object does exactly what it did before. I could also add a `.for_terminal(delay_draw=0.2)` for consistency.

Aside: Does removing the delay_draw from the constructor also have significant backwards-compatibility implications? If so I could use another ABC with an overridable method instead of passing in a function. I'm not sure I like that though because it is verbose and tightly-coupled.

I welcome your thoughts.

bluesky/utils.py
callumforrester: I usually favour many small modules so I'd prefer move it if that's okay. 
Do you think there is scope for creating a notebook utility module and moving NotebookProgressBar into it?

danielballan: Yes. This `utils.py` module has grown accidentally huge. Might be a good moment to create a `bluesky.utils` sub-package and move this file into it in a backward-compatible fashion so we can carve it up at our leisure, as in:

```py
# bluesky/utils/__init__.py
from ._legacy_utils import *
```

Then perhaps this could go into `bluesky.utils.jupyter`.

bluesky/utils.py
danielballan: Oh, when I last looked at it, the `delay_draw` parameter was accepted but then ignore (not passed through. Now there ia `partial` that passes it through. I am in favor of keeping it here so that it can be used.

danielballan:
@callumforrester I believe this is close. Can you spare some time to pick this back up and get it across the finish line?

bluesky/utils.py
callumforrester: I've had a go, let me know what you think.
A wildcard import like you suggest did not work because some modules import private functions from utils.
Instead I put the legacy code into utils/__init__.py

danielballan:
When I try this on our simulated hardware I get flaky output: two gray bars blink on and off in the widget area. Can you reproduce that issue @callumforrester? What ophyd objects have you been using to interactively test this?

```py
from  bluesky import RunEngine
from ophyd.sim import det, motor
motor.delay = 0.5
RE = RunEngine()
from bluesky.utils.jupyter import *
from bluesky.utils.jupyter import pbar_manager_for_notebook
RE.waiting_hook = pbar_manager_for_notebook()
RE(scan([det], motor, -10, 10, 10))
```

danielballan:
If that works correctly for you let me know and I'll report versions and provide a screencap.

callumforrester:
Seeing the same behaviour with your code. Although when I run it with the `TerminalProgressBar` I get 
```
motor [In progress. No progress bar available.]                                
motor [In progress. No progress bar available.]
```
Is it the same for you? Do we not expect to get a true progress bar from the simulated hardware? I have been using simulated EPICS motor records (so Ophyd thinks they are real `EpicsMotor`s).
If the TerminalProgressBar's behaviour is correct then this is a bug with the Jupyter version. It should display an empty progress bar with `[In progress. No progress bar available.]` next to it.

callumforrester:
I think it works with `EpicsMotor`, is the behaviour in the screenshot below correct? I am moving two motors with `mvr()`, the top one is still moving and the bottom one has completed motion.
image

danielballan:
I observe the "flashing" behavior with the old terminal-based progress bar manager in IPython as well. So this issue is either will the simulated hardware or the progress bars feature generally.

I don't see any reason to hold this PR up over that. I'll open a separate issue. Thanks @callumforrester!

danielballan:
xref https://github.com/bluesky/bluesky/issues/1441

PR #1429, opened 2020-12-15
Ensure that multi-step bash does not fail silently.

danielballan:
Follow-up to #1427


danielballan:
Test failure is another random failure from the delightful deprecated databroker sqlite backend.

PR #1430, opened 2020-12-15
Use databroker.temp instead of sqlite databroker.

danielballan:
In the test suite, we currently reach into databroker's test fixtures and make a databroker using its deprecated sqlite backend. In this PR, we use `databroker.temp()` instead, which currently generates a msgpack-backed catalog.

### DIFF
--- a/bluesky/tests/conftest.py
+++ b/bluesky/tests/conftest.py
@@ -58,9 +58,8 @@ class NumpySeqHandler:
 def db(request):
     """Return a data broker
     """
-    from databroker.tests.utils import build_sqlite_backed_broker
-    db = build_sqlite_backed_broker(request)
-    db.reg.register_handler('NPY_SEQ', NumpySeqHandler)
+    from databroker import temp
+    db = temp()
     return db
 
 
--- a/bluesky/tests/test_callbacks.py
+++ b/bluesky/tests/test_callbacks.py
@@ -22,7 +22,6 @@ from event_model import compose_run, DocumentNames
 import pytest
 import numpy as np
 import matplotlib.pyplot as plt
-from sqlite3 import InterfaceError
 from io import StringIO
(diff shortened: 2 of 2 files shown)


danielballan:
This seems good in general (avoid using another library's deprecated object; avoid reaching into another library's test fixtures) but the immediate motivate is that the sqlite backend frequently generates cursor `TimeoutErrors` in CI. See for example #1429.

danielballan:
Power-cycled to re-run CI

danielballan:
I'd love to get this in but haven't had time to investigate whether any of the failing tests are pertinent to the change.

cjtitus:
I was able to replicate the failed test on my system and investigate. It seems that what is happening is that `test_broker_base` is failing because the [`BrokerCallbackBase`](https://github.com/bluesky/bluesky/blob/452cf4cab041688186ba00048f88abc491068468/bluesky/callbacks/broker.py#L8) object is trying to call `db.fill_events`, which has been deprecated for the v1 databroker (but probably not for the sql databroker I guess). Therefore, you get a `NotImplementedError`.

Changing the db fixture from the SQL databroker to the `temp()` databroker has correctly exposed that `test_broker_base` is failing. Indeed, the `BrokerCallbackBase` has probably been broken for quite some time if you use a v1 databroker.

I think that the failing test should be fixed in a different PR, and that this test fix should just be merged. The failing test is an indication of how over-due this PR really is...

cjtitus:
Maybe `test_broker_base` could be marked as xfail for `NotImplementedError` so that it doesn't confuse people while a fix for that is discussed. I haven't though enough about Databroker to know what the solution should be.

danielballan:
Fix pushed. IF CI is happy, merge at will whoever sees this first. :-)

danielballan:
Thanks for pushing on this @cjtitus!

PR #1434, opened 2021-02-01
Fix persistent dict

danielballan:
## Description
Critical bug fix

## Motivation and Context
The `PersistentDict.update(..)` method writes updates to disk but does not update the in-memory cache. When the `PersistentDict` is finalized, the copy of disk is overwritten by the copy in the cache, so the update never takes effect at any point.

This bug occurred because methods of base class `zict.Func` were being called without updating our intermediate layer, `self._cache`. It seems to be the best way to avoid this kind of bug is to use composition instead of inheritance. `PersistentDict` now inherits from `collections.abc.MutableMapping`, and we can ensure that all updates flow through our code.

## How Has This Been Tested?

Unit tests added for each of the mutating methods (update, pop, popitem, setdefault, clear)

### DIFF
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -801,8 +801,9 @@ class PersistentDict(collections.abc.MutableMapping):
         yield from self._cache
 
     def popitem(self):
-        key, _value = self._cache.popitem()
+        key, value = self._cache.popitem()
         del self._func[key]
+        return key, value
 
     @staticmethod
     def _dump(obj):
--- a/bluesky/utils.py
+++ b/bluesky/utils.py
@@ -1,3 +1,4 @@
+import collections.abc
 from collections import namedtuple
 import asyncio
 import os
@@ -742,7 +743,7 @@ def all_safe_rewind(devices):
     return True
 
 
-class PersistentDict(zict.Func):
+class PersistentDict(collections.abc.MutableMapping):
     """
     A MutableMapping which syncs it contents to disk.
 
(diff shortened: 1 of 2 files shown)


danielballan:
P.S. Thanks @EliotGann for the report.

danielballan:
The test failure is unrelated to the PR. It's the sqlite databroker backend, which we are attempting to remove from the tests in #1430.

mrakitin (approved):

bluesky/utils.py
```diff
@@ -742,7 +743,7 @@ def all_safe_rewind(devices):
     return True
 
 
-class PersistentDict(zict.Func):
+class PersistentDict(collections.abc.MutableMapping):
```
MikeHart85, line 746: I like this much better. It feels more canonical if that makes sense.

The only thing that I could find, which worked before, but will no longer work as a result of this, is:

```python
p = PersistentDict(...)
p.close()
```

Not sure we ever used that though, or would even want it to work. It would close the `self._file`, putting the PersistentDict into an awkward state.

tacaswell (changes requested):
Agree with @MikeHart85 about the return on popitem.

Anyone can clear this review.

MikeHart85:
Made a quick edit to add the requested change. Feel free to revert, edit or change if needed. Looks good to merge to me otherwise.

danielballan:
Thanks, all. Good catch, @MikeHart85.

PR #1439, opened 2021-02-18
FIX: logging for CLI version ZMQ Proxy

dmgav:
Minor changes that fix the issue of missing logging output for CLI ZMQ proxy. Addresses the issue https://github.com/bluesky/bluesky/issues/1438.

Logging output functionality may be manually verified by starting ZMQ proxy in a terminal as
```
bluesky-0MQ-proxy 5577 5578 -v
```
or
```
bluesky-0MQ-proxy 5577 5578 -vvv
```
and running the following code in IPython session another terminal:
```
from bluesky import RunEngine
from ophyd.sim import det
publisher = Publisher('localhost:5577') 

RE = RunEngine()
RE.subscribe(publisher)  
RE(count([det])) 
```
The logging output in the terminal running proxy should display the contents of 'start' and 'stop' documents if the proxy was started with `-v` and all documents ('start', 'stop', 'descriptor' and 'event') if the proxy was started with `-vvv`.

If the proxy is started as
```
bluesky-0MQ-proxy 5577 5578
```
then no logging output will be printed.

Instead of sending to stdout, logging data could be saved to a file on disk:
```
bluesky-0MQ-proxy 5577 5578 -v --logfile log_file.txt
```

### DIFF
--- a/bluesky/commandline/zmq_proxy.py
+++ b/bluesky/commandline/zmq_proxy.py
@@ -5,7 +5,6 @@ import threading
 from bluesky.callbacks.zmq import Proxy, RemoteDispatcher
 from bluesky.log import set_handler
 
-
 logger = logging.getLogger('bluesky')
 
 
@@ -51,10 +50,13 @@ def main():
     args = parser.parse_args()
     in_port = args.in_port[0]
     out_port = args.out_port[0]
+
     if args.verbose:
-        logger.setLevel('INFO')
-        if args.verbose > 2:
-            logger.setLevel('DEBUG')
+        from bluesky.log import config_bluesky_logging
+        if args.verbose <= 2:  # called with -v or -vv
+            config_bluesky_logging(level='INFO')
+        else:  # called with -vvv
+            config_bluesky_logging(level='DEBUG')
         # Set daemon to kill all threads upon IPython exit
         threading.Thread(target=start_dispatcher,
                          args=('localhost', out_port, args.logfile),
--- a/bluesky/commandline/zmq_proxy.py
+++ b/bluesky/commandline/zmq_proxy.py
@@ -46,20 +46,22 @@ def main():
                         help=("Show 'start' and 'stop' documents. "
                               "(Use -vvv to show all documents.)"))
     parser.add_argument('--logfile', type=str,
-                        help="Write logfile")
+                        help="Redirect logging output to a file on disk.")
     args = parser.parse_args()
     in_port = args.in_port[0]
     out_port = args.out_port[0]
 
     if args.verbose:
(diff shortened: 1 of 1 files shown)


danielballan (approved):

jklynch (approved):
I think this looks good.

bluesky/commandline/zmq_proxy.py
```diff
@@ -47,17 +46,22 @@ def main():
...
     args = parser.parse_args()
     in_port = args.in_port[0]
     out_port = args.out_port[0]
+
     if args.verbose:
-        logger.setLevel('INFO')
-        if args.verbose > 2:
-            logger.setLevel('DEBUG')
+        from bluesky.log import config_bluesky_logging
+        # "INFO" if called with '-v' or '-vv', "DEBUG" if called with '-vvv'
```
mrakitin, line 56: Would it be enough to have 2 `v`s - `-v` for `INFO` and `-vv` for `DEBUG? I guess `WARNING` will be output by default.

Ref for myself: https://docs.python.org/3/library/logging.html#levels

dmgav: @mrakitin This is how it worked before it was broken.

mrakitin: OK, should we revisit it?

dmgav: I think verbosity settings it should be consistent with the rest of our projects. I would vote for `-vvv` for `DEBUG` mode, since it will print a lot of data. Attn: @danielballan

danielballan: >  I would vote for -vvv for DEBUG mode, since it will print a lot of data.

This seems right to me too. I agree that it’s a bit funny that -v and -vv behave the same for now, both here and in caproto CLIs, but that gives us room to add a “medium” level of verbosity as -vv in the future if we want. Looking to programs like ssh and ansible, three or more v’s seems to be where they cross over into extremely high verbosity, a mode too verbose for normal operation, imposing possible performance overhead and generally too much logging to be worth capturing unless you are trying to debug a specific issue.

dmgav: Also the assignment of mode here is artificial. We are not truly select DEBUG mode. Instead the logging mode is used to enable logging of start/stop documents (`logger.info`) and the rest of the documents (`logger.debug`): https://github.com/bluesky/bluesky/blob/b3038a0f80caec4a94f8cf88392962bc44acfaba/bluesky/commandline/zmq_proxy.py#L30
It is also easier to visually locate start/stop documents the logger output in `-vvv` mode.

PR #1443, opened 2021-02-22
add hardware-interfaces page

untzag:
## Description

Add hardware interfaces page to docs.

## Motivation and Context

As discussed with @danielballan I am trying to update the docs to be more useful to small labs (and non-EPICS users in general). My next step is to flesh out `hardware.rst` to give much more concrete advice about how to implement Bluesky interfaces. In the long run I hope to reach out to some of the existing hardware support libraries and try to see if they would be willing to add Bluesky Hardware Interface compliance to their "device" classes.

I'm aware that some of these linked packages aren't polished, but I feel like if we're going to start building an ecosystem we have to start by at least advertising what we have.

### DIFF
--- a/docs/source/hardware-interfaces.rst
+++ b/docs/source/hardware-interfaces.rst
@@ -1,0 +1,34 @@
+.. _hardware_interface_packages:
+
+Hardware Interface Packages
+===========================
+
+The Bluesky library does not provide direct support for communicating with real hardware.
+Instead, we define a high-level abstraction: the :ref:`Bluesky Hardware Interface <hardware_interface>`.
+This allows different experimentalists to use different hardware control systems.
+
+The following packages provide support for real hardware communication from Bluesky:
+
+=============  ========================================================================
+Instrbuilder_  Lightweight package with a focus on SCPI_.
+Ophyd_         Primary hardware interface package for Bluesky with a focus on EPICS_.
+Ophyd-Tango_   Tango_ integration for Bluesky.
+yaqc-bluesky_  yaq_ integration for Bluesky.
+=============  ========================================================================
+
+Importantly, you may mix hardware interfaces from multiple different packages within the same RunEngine.
+Please note that the above packages are developed and maintained separately from Bluesky itself.
+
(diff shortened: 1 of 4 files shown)


untzag:
@lucask07 please let me know if you don't want Instrbuilder listed here for any reason. I know that Bluesky support is kinda broken right now, but as we discussed I think it should be pretty easy to get working---just gotta sit down and do it.

lucask07:
@untzag OK with me. Bluesky integration with instrbuilder has been demonstrated but through a fork of ophyd. As you suggested, a better approach would be to integrate the needed interface code within instrbuilder so that ophyd remains untouched.

danielballan:
Swinging by just to say I am fighting fires this week but excited to look at this in detail. Thanks for the contribution!

docs/source/hardware-interfaces.rst
```diff
@@ -0,0 +1,34 @@
...
+
+The Bluesky library does not provide direct support for communicating with real hardware.
+Instead, we define a high-level abstraction: the :ref:`Bluesky Hardware Interface <hardware_interface>`.
+This allows different experimentalists to use different hardware control systems.
+
+The following packages provide support for real hardware communication from Bluesky:
+
+=============  ========================================================================
+Instrbuilder_  Lightweight package with a focus on SCPI_.
+Ophyd_         Primary hardware interface package for Bluesky with a focus on EPICS_.
```
danielballan, line 14: I think this should go first, and the description should probably include the phrase "reference implementation".

docs/source/hardware-interfaces.rst
```diff
@@ -0,0 +1,34 @@
...
+Ophyd-Tango_   Tango_ integration for Bluesky.
```
danielballan, line 15: This is in such early stages that I think including alongside the others could be confusing. (I know I suggested including it.) It should either go in a separate table of experimental works-in-progress or have some disclaimer language like "incomplete and experimental early work".

danielballan:
Sorry for letting this languish so long, @untzag. I took another pass. Looks good!

PR #1444, opened 2021-03-17
Fixed typo in a unit test for ProgressBarManager

dmgav:
Fixed minor bug in the unit test for `ProgressBarManager`.


mrakitin (approved):
@dmgav, thanks for catching and fixing this copy-paste error.

PR #1446, opened 2021-04-16
Initial ABC for Status, Readable, Movable, Flyable

### LINKED ISSUE #1445: ABCs for Devices that implement the "Bluesky Interface"
@untzag and @ksunden have suggested adding some ABCs for the expected Bluesky Interfaces. ABCs would be a useful tool for scaffolding implementations of the Readble, Movable, and Flyable interfaces. I would want to avoid accidentally giving the impression that implementation _must_ subclass from them. A key selling point for Bluesky is that other libraries can implement its interfaces and be interoperable without a software dependency. In our documentation, we should emphasize that they are optional, to be used or referred to if they help, or ignored otherwise.

We should also be sure that these ABCs implement https://docs.python.org/3/library/abc.html#abc.ABCMeta.__subclasshook__ so that they will behave correctly if `isinstance` is used.

ksunden:
This provides an initial implementation of [PEP 544 Protocol](https://www.python.org/dev/peps/pep-0544/) Abstract Base Classes for some core interfaces in the bluesky ecosystem.

## Description
The use of protocols allows for both static type checking using mypy (or similar) as well as runtime duck-type checking via isinstance (without requiring that the object being checked _actually_ inherits from the ABC)

Still to do:
- [x] Documentation
- [x] Unit tests
- [x] Confirm action on `configure` which currently active implementations do not actually provide (ophyd flyer, yaqc_bluesky)
- [x] Confirm action for `Status.finished_cb` which is listed as deprecated elsewhere

## Motivation and Context
Closes #1445

This will provide an easy `isinstance` check that still allows duck typed implementations to work as expected, just ensuring that all required methods/properties are present

## How Has This Been Tested?
I intend to add actual unit tests before merging, but initial ad-hoc tests with ophyd simulated hardware to show it works:
(Note, no modifications to ophyd were needed, and none should be except perhaps regarding `configure` on flyers, which is currently not defined, and thus left out of the ABC)

```python
>>> from bluesky import abc
>>> from ophyd import sim
>>> isinstance(sim.motor1, abc.Readable)
True
>>> isinstance(sim.motor1, abc.Movable)
True
>>> isinstance(sim.motor1, abc.Flyable)
False
>>> isinstance(sim.det1, abc.Readable)
True
>>> isinstance(sim.det1, abc.Movable)
False
>>> isinstance(sim.det1, abc.Flyable)
False
>>> isinstance(sim.flyer1, abc.Readable)
False
>>> isinstance(sim.flyer1, abc.Movable)
False
>>> isinstance(sim.flyer1, abc.Flyable)
True
```


ksunden:
This implementation should be able to replace the `is_movable` check and `Movable` ABC that was defined in `utils/__init__.py`

I have not gone looking for other places where it could be useful internally, but it will be useful for at least one `isinstance(obj, OphydDevice)` in bluesky-queueserver.

ksunden:
The signature of `set` is a bit more trouble yet...

From a quick perusal of ophyd (plus my knowledge of how we use it in yaqc_bluesky), it looks like the proper signature for the protocol should actually be `def set(self, value: Any) -> Status:...`

Such a signature would be compatible with adding any kwargs (with defaults), but not if more than one _required_ positional args.

As it stands, defining with `*args, **kwargs` and then wanting an implementation that does anything else (even just `val, *args, **kwargs`) fails mypy.

Are there any cases where `set` does _not_ take a single positional arg? 

for reference, heres the grep for "def set\(" on ophyd:

ophyd/positioner.py
46:    def set(self, new_position: Any, *,

ophyd/sim.py
184:    def set(self, value, *, timestamp=None, force=False):
287:    def set(self, *args, **kwargs):
328:    def set(self, value, *, timestamp=None, force=False):
422:    def set(self, value, *args, **kwargs) -> StatusBase:

ophyd/pseudopos.py
907:    def set(self, position, **kwargs):

ophyd/signal.py
272:    def set(self, value, *, timeout=None, settle_time=None):
508:    def set(self, value, *, timestamp=None, force=False):
1322:    def set(self, *args, **kwargs):
1754:    def set(self, value, *, timeout=DEFAULT_WRITE_TIMEOUT, settle_time=None):

ophyd/areadetector/plugins.py
679:    def set(self, region):
```

Most of those are of the style of "single positional arg" though the name is not always the same.
Those that are just *args, **kwargs are simple passthroughs with e.g. deprecation warnings

Note that unless the signature is defined as positional _only_, implementations that rename that argument will fail mypy as well...
(And positional _only_ is only available in 3.8+ for python syntax)

**Comment by ksunden** [comment]
I have moved to `add_callback` method as opposed to `finished_cb` attribute as that appears to be the actual current preferred/required modulo deprecation period way of implementing the callbacks.

**Comment by ksunden** [comment]
Re set signature:

if we really do want to be very laissez-faire about the signature of that, there is a way to make mypy happy with that, its a little funky, but seems to work:

```python
class Movable(Readable, Protocol):
    @property
    def set(self) -> Callable[..., Status]:
        ....
```

This will make implementations which define _any_ function that returns a status pass the type checking. 
It needs to be a property instead of just a callable attribute because attributes must be read/write and functions are read only.

ksunden:
Okay, as far as I'm concerned, this is ready for review.

My biggest question is in regards to the signature of `set` (and consequently `check_value`, as they share the same inputs) as detailed above.

Other than that, I'm pretty happy with the implementation as it stands, please do look at the type hints, see if anything stands out etc.

So far I have the runtime tests in, which _only_ look at whether attributes exist, not their signatures/types. I added (and skipped) tests using mypy to check, because ophyd does not mark its type hints as usable by mypy.

I added py.typed to this repo because it is needed for mypy and to use these abcs effectively.

bluesky/abc.py
```diff
@@ -0,0 +1,264 @@
...
+    from typing_extensions import Protocol, runtime_checkable
+
+from typing import Dict, Any, Optional, Callable, Generator, List, Union, Tuple, NoReturn
+
+
+Configuration = Dict[str, Dict[str, Any]]
+
+
+@runtime_checkable
+class Status(Protocol):
```
thomascobb, line 13: I think there is also `watch(watcher)`: https://nsls-ii.github.io/ophyd/status.html#ophyd.status.DeviceStatus.watch
At least I needed it here: https://github.com/thomascobb/bluefly/blob/master/bluefly/core.py#L122

bluesky/abc.py
```diff
@@ -0,0 +1,264 @@
...
+        """
+        ...
+
+    def describe_configuration(self) -> Configuration:
+        """Same API as ``describe``, but corresponding to the keys in ``read_configuration``."""
+        ...
+
+
+@runtime_checkable
+class Configurable(Protocol):
```
thomascobb, line 121: I've used `configure`: https://github.com/thomascobb/bluefly/blob/0ee7ea837d5f74e3cdc69e2315e9beed28dd7766/bluefly/core.py#L411
but I thought it belonged with `read_configuration` and `describe_configuration`. I could be wrong though...

thomascobb (commented):
Liking the type hints! I didn't realize that half those methods were optional (like Configurable, Stoppable, etc)...

bluesky/abc.py
ksunden: this is considered optional:

per @danielballan:

> In this case the docs are just wrong. The watch method is optional. If not implemented, there will be no progress bar, but it will otherwise operate fine.
https://github.com/bluesky/bluesky/blob/566dce9e7ada95e2346c9aa61206c6a2f0a189bf/bluesky/utils/__init__.py#L1221

We could make a `Watchable` abc if we want (also the signature of the watcher function is a bit chaotic to represent because there are a lot of arguments that need to be accepted, but not all given)

bluesky/abc.py
ksunden: I was wondering about that. from what I can tell read/describe configuration are required for _all_ valid devices, but `configure` is only required for devices which can _update_ their configuration (static config is available on all devices)

tacaswell:
I meant the ability to run a (sub) plan with a restricted set of messages, not a second implementation.

thomascobb:
There is also an optional `collect_asset_docs` but documentation on it is hard to find. I added it here, but it might be wrong:
https://github.com/thomascobb/bluefly/blob/0ee7ea837d5f74e3cdc69e2315e9beed28dd7766/bluefly/detector.py#L127

thomascobb:
A couple of weeks I had a call with @danielballan, @tacaswell and @callumforrester about extensions to the `Device` interface to allow asyncio and an expanded Flyable interface. I've already written a wrapper to an `asyncio.Task` that conforms to `Status`, so I would like any method that requires control system access to return a `Status`. I would also like to create a `validate()` method that mirrors `configure()` but provide validation possibly while the device is in use, so that users can validate their scans before queuing them up.

During the call we discussed turning `stage` and `unstage` into plans, but I've done some thinking since then and I have some ideas that would fit better with a Status object. I can discuss these further in a future call.

I propose:
```python
@runtime_checkable
class Configurable(Protocol):
    def configure(self, conf: Configuration) -> Status:
        """The same as before, but remove the return value and allow it to take time"""
    def validate(self, conf: Configuration) -> Configuration:
        """Like validate, but without modifying the Device. Returns a validated Configuration, with error status for each field to allow reporting to a GUI"""      

@runtime_checkable
class Pausable(Protocol):
    def pause(self) -> Status:
          """As before, but can take time"""
    def resume(self) -> Status:
          """As before, but can take time"""

@runtime_checkable
class Stageable(Protocol):
    def stage(self) -> Status:
        """As before, but can take time. How would dropping the original return value affect Bluesky?"""
    def unstage(self) -> Status:
        """As before, but can take time. How would dropping the original return value affect Bluesky?"""

@runtime_checkable
class Stoppable(Protocol):
    def stop(self, sucess=True) -> Status:
        """As before, but can take time"""

# As returned from compose_datum()
# https://github.com/bluesky/event-model/blob/d373989fee22b30050ce39643ecd85e255128910/event_model/__init__.py#L1573
Datum = Dict[str, Any]

@runtime_checkable
class Flyable(Protocol):
    # All the other methods
    ...

    def collect_asset_docs(self) -> Generator[Tuple[str, Datum], None, None]
        """I'm a bit vague on what the docs should say here"""
```

thomascobb:
So a flyscan would call:
```python
stage()
configure(scan_spec)
kickoff()
collect()
# While scan in progress keep collecting...
collect()
complete()
# Maybe go back up to configure() to put some more data in the same file
unstage()
```

mrakitin:
For the flyer, the sequence would be ..., kickoff, complete (the check returning the status object to monitor when the movement is done), and then collect (to emit a chunk of documents).

bluesky/abc.py
danielballan: @tacaswell and I now consider `configure` to be a misfeature. It is similar to `set` (or a bunch of `set`s) with the side-effect of notifying the RunEngine, "Any Event Descriptor documents you emitted with these configuration readers are now out of date, so if/when you emit new readings from this device, emit a fresh Event Descriptor as well."

The major problem with this is that the configuration setting could be changed _externally_. (Example: Someone runs `caput` or pokes a button in CSStudio.) In that way, the Event Descriptor would become "wrong" but the RunEngine would never know. The proper way to handle this is to let ophyd _notify_ the RunEngine that some signal that was dubbed "configuration" has changed under it, and that it should reissue an Event Descriptor.

We have known about this problem for awhile (maybe as far back as 2016) but it has never been a priority because `configure` is extremely rarely used. Let's take this moment to formally deprecate it, and drop this class entirely from the PR.

In fact, because it is so rarely used, we discussed going as far as immediately repurposing the method name along the lines proposed by @thomascobb elsewhere in the conversation on this PR.

bluesky/abc.py
```diff
@@ -0,0 +1,264 @@
...
+        ...
+
+    def clear_sub(self, function: Callable) -> Union[NoReturn, None]:
+        """Remove a subscription."""
+        ...
+
+
+@runtime_checkable
+class Checkable(Protocol):
+    def check_value(self, value: Any) -> Union[NoReturn, None]:
```
danielballan, line 253: @thomascobb @tacaswell Do you recall what we said we might do about this? Would this be superseded by `validate`?

danielballan (commented):
Thanks for your work on this @ksunden and sorry for our slow follow-up. Let's try to merge this this week.

danielballan:
@ksunden I think we can move ahead with this and iterate from there. Before we merge I want to suggest a module name change. Since this is a _protocol_ and a protocol is [conceptually not quite the same as an abc](https://glyph.twistedmatrix.com/2021/03/interfaces-and-protocols.html) it might be better to name this new module `bluesky.protocols`. It also has the advantage of being a more obvious name for non-practitioners to whom "abc" is jibberish.

dmgav:
On detailed inspection of the results of scanning the namespace that contains all simulated devices, I noticed that `motor_no_hints1` and `motor_no_hints2` of the type `SynAxisNoHints` are not recognized as Readable. Is it the desired behavior?

ksunden:
It was my understanding from reading the docs that `hints` is a required field, though it may be empty `{}`. The existance of `SynAxisNoHints` seems to refute that, but I wonder if that was an edge case someone thought of that we wish to standardize on the property at least existing (and that there is likely no actively used implementation that does not provide at least empty hints).

dmgav:
@tacaswell and @danielballan may know more about the origins of `SynAxisNoHints` and whether we want to support types that don't have hints field.

danielballan:
Good catch, @dmgav. The hints attribute is intended to be optional. There are custom devices in the wild that do not implement which we need to support.

danielballan:
Forward!

PR #1449, opened 2021-05-15
Tolerate garbage on zmq

tacaswell:
## Description

This adds a number of try-except blocks to the 0mq RemoteDispatcher to not silently fail when un-expected messages come across the socket.

## Motivation and Context

This fixes failures of the xpdan servers at XPD/PDF that we believe are caused by the network scanners pushing random byte strings into the zmq proxy.  The proxy is OK with this, but the subscribed clients are not.

## How Has This Been Tested?

This can be tested by starting up the proxy, starting a dispatcher as

```python
RemoteDispatcher('localhost:5578').start()
```

and in another terminal

```python
import zmq

url = 'tcp://localhost:5577'
context = zmq.Context()
socket = context.socket(zmq.PUB)
socket.connect(url)

socket.send(b' \x80abc ')
socket.send(b' start ')
socket.send(b' start abc')
socket.send(b'')
```

### DIFF
--- a/bluesky/callbacks/zmq.py
+++ b/bluesky/callbacks/zmq.py
@@ -6,6 +6,12 @@ import warnings
 from ..run_engine import Dispatcher, DocumentNames
 
 
+class Bluesky0MQDecodeError(Exception):
+    """Custom exception class for things that go wrong reading message from wire."""
+
+    ...
+
+
 class Publisher:
     """
     A callback that publishes documents to a 0MQ proxy.
@@ -217,7 +223,7 @@ class RemoteDispatcher(Dispatcher):
     """
     def __init__(self, address, *, prefix=b'',
                  loop=None, zmq=None, zmq_asyncio=None,
-                 deserializer=pickle.loads):
+                 deserializer=pickle.loads, strict=False):
         if isinstance(prefix, str):
             raise ValueError("prefix must be bytes, not string")
         if b' ' in prefix:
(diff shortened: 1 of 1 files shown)


bluesky/callbacks/zmq.py
```diff
@@ -243,34 +263,69 @@ def __init__(self, address, *, prefix=b'',
...
-            name = name.decode()
+            try:
+                prefix, name, doc = message.split(b' ', 2)
+            except ValueError as e:
+                if self._strict:
+                    raise Bluesky0MQDecodeError from e
+                else:
+                    print(f"The message {message} could not be split into "
+                          "three parts by b' '.  Dropping message on floor "
+                          "an continuing"
```
mrakitin, line 281:
```suggestion
                          "and continuing"
```

bluesky/callbacks/zmq.py
```diff
@@ -243,34 +263,69 @@ def __init__(self, address, *, prefix=b'',
...
+                          f"\n\n{e}")
+                    continue
+
+            try:
+                name = name.decode()
+            except UnicodeDecodeError as e:
+                if self._strict:
+                    raise Bluesky0MQDecodeError from e
+                else:
+                    print(f"the name {name} can not be decoded as utf-8. "
```
mrakitin, line 291:
```suggestion
                    print(f"The name {name} can not be decoded as utf-8. "
```

mrakitin (commented):
The changes look good to me, and thanks for documenting the purposes of the 3 parts.

I have a concern about the port numbers used in the examples: `127.0.0.1:5567`. In different places (such as https://github.com/NSLS-II-SST/sst-workers/blob/c786d1d40779f49a5649297b1345799fc233e5a2/suitcase_worker.py#L13). Also, the example for the `RemoteDispatcher` has the wrong port `5567`, while it should be `5568`.

Can those be changed to be `5577` for the `Publisher` and `5578` for the `RemoteDispatcher` while you are there, @tacaswell?

danielballan:
This is close. Can you address the last review comment, @tacaswell?

tacaswell:
@mrakitin I'm not sure which example you are referring to.

mrakitin:
https://github.com/NSLS-II-SST/sst-workers/blob/c786d1d40779f49a5649297b1345799fc233e5a2/suitcase_worker.py#L13 is the example used on the floor.

The examples in docstrings of https://github.com/bluesky/bluesky/blob/master/bluesky/callbacks/zmq.py refer to another set of ports:
https://github.com/bluesky/bluesky/blob/db8276d3e01e763ff8afd791eafcd8ded8fdb686/bluesky/callbacks/zmq.py#L36
and
https://github.com/bluesky/bluesky/blob/db8276d3e01e763ff8afd791eafcd8ded8fdb686/bluesky/callbacks/zmq.py#L111

Not sure what we want to treat as a default port, and why the usecase at SST uses a different port.

danielballan:
By convention, we use 5567 and 5568 for publish and subscribe, respectively. These ports aren’t in the Python code, only in the docs, so they are more conventions than defaults.

Not sure where SST got their number from—could  just a (valid but “unconventional”) typo.

I can’t find the example either—can you link to it?

mrakitin:
By example I meant the code in the docstrings linked above in my previous comment.

Probably SST got their preferred values. We can keep 5567 and 5568 here without changes. I wonder what ports are used for zmq at IOS?

mrakitin:
I guess the ports 5577/5578 came from the ansible playbook we used to deploy 0MQ proxies - https://github.com/NSLS-II/playbooks/blob/fe8dac9be419c26ec18f2e5176f91c5f8ef72b57/roles/bluesky-0MQ-proxy/files/run-default-bluesky-0MQ-proxy.sh#L31 (private NSLS-II repo). Maybe making it consistent across the deployed configuration and the inline examples in the docstrings (see https://github.com/bluesky/bluesky/pull/1449#issuecomment-852355021) would make sense.

danielballan:
The convention is 5567/5568. SST is the outlier and could be updated to match this convention to avoid potential confusion. AFAIK the examples in this repository are all internally consistent and consistent with the convention, so there is nothing left to do on this PR.

danielballan:
Follow-up on SST is here: https://github.com/NSLS-II/playbooks/issues/460 (private NSLS-II repo)

PR #1464, opened 2021-08-05
ENH: RunEngine `__call__` returns plan value

### LINKED ISSUE #1457: Proposal: Save or return the plan's return value after RunEngine.__call__
## Current Behavior

The RunEngine returns the run UID, or list of run UIDs for the plan that was run.
`uids = RE(plan(*plan_args))`
The actual return value of the plan is jettisoned into space without being saved, stored, cached, or returned in any way (as far as I can tell).

## Reasons for Change

Every plan can return values, that can be consumed by _other_ plans. This is crucial in adaptive plans, such as the simple examples in the [bluesky tutorial](https://blueskyproject.io/bluesky/tutorial.html#adaptive-logic-in-a-plan).

More complex plans that do peak finding, signal/motor optimization, and sample alignment may have multiple layers of adaptive plans chained together via return values. As an example, my sample alignment plan calls several edge finding plans, each of which calls a signal maximization plan.

There is currently no way to run a plan and get the return value. So I can't test any of the nested plans independently (i.e, just run the signal maximization plan, check the return value to make sure it is correct, then test the edge finding plan, check that return value, etc etc). It would be both convenient, and arguably more consistent, if the RunEngine returned the top-level plan's return value. This would allow for better debugging and interactive use, for example a user could run `max_signal_position = RE(find_maximum(dets, x_motor))` and note the best motor position for use in other plans, etc.

## Possible Solutions

The most straightforward solution would be to change `RunEngine.__call__` so that the top level plan's return value is returned by the RunEngine. Unfortunately this would be a fundamental API change, breaking any scripts that actually did rely on getting the UIDs of the plans that ran.

An alternative would be to store the return value as an attribute of the RunEngine, so that it could be accessed when needed, such as `last_return_val = RE.ret`

Or it could be saved in the run metadata.

cjtitus:
`RunEngine` now accepts a flag at initialization that will cause `__call__` to return either a uid list (the default) or pass through the return value of the plan it run

## Description

`RunEngine.__init__` has received a new argument that lets the user choose have `__call__` return either the UID list (current behaviour) or the plan return value.

`__call__`, `_run`, `resume`, and `_resume_task` have been modified to grab the plan's return value and pass it on if requested. If a plan fails to return anything (due to an error), the sentinel value of object() is returned. (to distinguish from plans that simply return None)
## Motivation and Context

This enhancement makes it vastly easier to test plans that have a return value. This is especially critical for testing sub-plans independently from the top-level plans that consume their values.

It also opens up the ability for plans to return information that could be useful to interactive users

## How Has This Been Tested?

All tests are currently passing on my local machine

- I need to decide on return values for halt(), stop(), and abort(). Should they just return object() (assuming the flag is set)?
- I also need to add tests to Bluesky that check whether or not the correct values are being returned in various circumstances.
- I would like to discuss whether or not we should add a warning about this flag. @tacaswell suggests that we warn users who do not explicitly set the flag

closes #1457


bluesky/run_engine.py
```diff
@@ -215,6 +215,12 @@ def block(ev: Threading.Event) -> None:
            - Matplotlib is imported and using a nbagg or ipympl backend (
              wait on the event and poll to push updates to the browser)
 
+    call_returns_uid: bool, optional, default True
```
tacaswell, line 218: Maybe do this as 

```
call_return_type : str, optional {'uids', 'plan_return'}
```

which is a bit more "self documenting" and does not paint us into a corner if in the future we decide we want a third different return type!

bluesky/run_engine.py
```diff
@@ -880,8 +897,10 @@ def _resume_task(self, *, init_func=None):
                 init_func()
 
             if self._task_fut is None or self._task_fut.done():
-                return
-
+                try:
+                    return self._task_fut.result()
+                except:
+                    return object()
```
tacaswell, line 903: you will want to create this once at the top of this file like

```
NO_PLAN_RETURN = object()
```

so that later you can do

```
ret = RE(...) 
if ret is NO_PLAN_RETURN:
    ...
else:
   ...
```

bluesky/run_engine.py
```diff
@@ -1251,6 +1274,7 @@ async def _run(self):
         self._reason = ''
         # sentinel to decide if need to add to the response stack or not
         sentinel = object()
+        plan_result = sentinel
```
tacaswell, line 1277: probably use a different sentinel here?

tacaswell:
Can you also add a few tests that this works as expected?

This is looking good!

bluesky/run_engine.py
cjtitus: What scope would this be in? In this example would you have run `from bluesky import NO_PLAN_RETURN`?

tacaswell:
The test failure looks transient and un-elated.

bluesky/run_engine.py
cjtitus: Yes, I will use `NO_PLAN_RETURN` instead

bluesky/run_engine.py
tacaswell: That is one option.

I could also see doing this as a instance variable on the RE (so it can have a nice repr and "know" who its RE is etc) so you could do

```
ret = RE(...)
if ret is not RE.NO_PLAN_RETURN:
 ....
```

tacaswell:
`stop`, `abort`, and `halt` should probably all play by the same rules as `__call__` and `resume` in terms of what they return.  Good catch thinking of those!

bluesky/run_engine.py
cjtitus: What would the default behavior be if the user passed something different from `'uids'` or `'plan_return'`? I can think of two options -- either we check `call_return_type in ['uids', 'plan_return']` at the top of the initialization and throw an error if something different is passed in, or we do
```
if call_return_type == 'plan_return':
    return plan_result
else:
    return tuple(self._run_start_uids)
```
and just make returning a uid list the default for any string value besides `'plan_return'`

bluesky/run_engine.py
cjtitus: Should I shove the definition of `NO_PLAN_RETURN` into `bluesky/utils/__init__.py`?

cjtitus:
There are two things to do with testing. I need to 
- Test that return values work
- Test that switching to return values hasn't broken any of the normal tests?

I'm probably going to add a `RE_plan_return()` fixture to conftest.py. The question is -- do I want to run this version of the RE through some of the normal tests? If so, how many, and how do I accomplish this (parametric testing?) In theory, I have written the code so that the only path that isn't being tested by the "normal" tests is
```        
if self._call_return_type == 'plan_return':
    return plan_return
else:
    return tuple(self._run_start_uids)`
```
And all other behavior is identical. So if you believe that which branch you take of this if-statement won't affect anything in the current tests, then I just need to add tests to make sure that the return values are correct. Otherwise, we need to discuss how many/which tests should be run on the plan_return version of the RunEngine.

tacaswell:
You can parameterize the RE fixture to run every test 2x.  That will help find which tests it breaks :)

cjtitus:
Predictably, `test_force_stop_exit_status(bail_func, status, RE)` fails when you don't return the uids list. This test pauses the run engine, calls either resume, stop, halt, or abort, and checks that the returned UID is still correct.

I can modify the test so that it only runs on the "uids" fixture, and not the "plan_return" fixture for now.

cjtitus:
Ahh shoot, didn't test all the things, only ran test_run_engine.py, so I'll test everything and fix.

cjtitus:
Ok, so every test that fails, fails because the test tried to do 
```
d = DocCollector()
rs, = RE(some_test_plan(), d.insert)
```
and then usually some assert with `d.descriptor[rs]`

For all of these cases, I think you can get rs the way that `test_exceptions_exit_status` does it, via `rs = d.start[0]['uid']`. Does this make sense as a change to make to all of these tests that were getting the uids from the RunEngine?

tacaswell:
If I recall correctly, part of what we were testing with `DocCollector` is that the documents that came out were the right documents!  I would mark those with a `pytest.skipif` for now and then we can think about the best way to deal with this going forward (which might be to re-write the plans in those tests to return a list of uids?)

cjtitus:
I think this could be accomplished by just changing `@run_decorator` to return a list rather than a string, but perhaps a change for another day...

tacaswell:
Changing `@run_decorator` might be too big of a change to consider.

cjtitus:
`@run_decorator` really just needs to wrap its _current_ return value in a list, and then most built-in plans will then always return exactly what they would have returned before I made this PR.

Right now, I can't imagine that many people are capturing the return value from `@run_decorator` are they? If you're typing `RE(scan([dets], motor, ...))`, the `@run_decorator` is what's returning a value, but it's currently just discarded. Maybe I'll propose that as a separate enhancement (one day) and people can debate there.

tacaswell:
I know we have some cases like this on the floor:

```python

def my_plan(...):
    uid = yield from bp.scan(...)
    if uid is not None:
        hdr = db[uid]

```

which would be broken by changing the return type of `@run_decorator` to be a list from a string.

bluesky/run_engine.py
```diff
@@ -354,14 +360,15 @@ def setup_run_permit():
...
+        self.NO_PLAN_RETURN = object()
 
         if during_task is None:
             during_task = DefaultDuringTask()
         self._during_task = during_task
 
         # The RunEngine keeps track of a *lot* of state.
         # All flags and caches are defined here with a comment. Good luck.
-
+        self._call_return_type = call_return_type  # should __call__ return UID or plan value
```
tacaswell, line 371:
```suggestion
        self._call_return_type = call_return_type  # should __call__ return UIDs or plan value
```

bluesky/run_engine.py
```diff
@@ -929,6 +949,16 @@ def _resume_task(self, *, init_func=None):
                     if (exc is not None
                             and not isinstance(exc, _RunEnginePanic)):
                         raise exc
+                    # Only try to get a result if there wasn't an error,
+                    # (other than a cancelled error)
+                    if exc is None:
```
tacaswell, line 954: I don't think we need this conditional as the `if exc is not None` above will bail out of the function in all cases that would not go through this branch?

bluesky/tests/test_run_engine.py
```diff
@@ -1412,15 +1415,23 @@ def test_force_stop_exit_status(bail_func, status, RE):
...
 
-    assert len(d.start) == 1
-    assert d.start[0]['uid'] == rs
-    assert len(d.stop) == 1
-    assert d.stop[rs]['exit_status'] == status
+    if RE._call_return_type == 'plan_return':
+        if bail_func == "resume":
+            assert rs != RE.NO_PLAN_RETURN
+        else:
+            assert rs == RE.NO_PLAN_RETURN
```
tacaswell, line 1428:
```suggestion
            assert rs is RE.NO_PLAN_RETURN
```

bluesky/tests/test_run_engine.py
```diff
@@ -1412,15 +1415,23 @@ def test_force_stop_exit_status(bail_func, status, RE):
...
-    rs, = getattr(RE, bail_func)()
+    rs = getattr(RE, bail_func)()
 
-    assert len(d.start) == 1
-    assert d.start[0]['uid'] == rs
-    assert len(d.stop) == 1
-    assert d.stop[rs]['exit_status'] == status
+    if RE._call_return_type == 'plan_return':
+        if bail_func == "resume":
+            assert rs != RE.NO_PLAN_RETURN
```
tacaswell, line 1426:
```suggestion
            assert rs is not RE.NO_PLAN_RETURN
```

tacaswell:
I think this is technically sound (but I left a few line comments).  I think this is ready for wider review, in particular the choice of names!

We should also decide how heavily we want to advertise this in the docs.  It might be worth "soft launching" this feature with just the docstring, seeing how it goes in practice, and then highlighting it in the tutorials / narrative?

untzag:
Disclaimer: I truly have no context of the broader picture here. I have never actually used an adaptive plan and don't know how return values work in constructing such things.

I don't like the idea of a RE with highly dynamic method signatures. This PR introduces complex new behavior that every developer building on top of Bluesky must account for. I would prefer to see the RE behavior formalized and "locked down" (including type hinting eventually)... that's what I think the ecosystem needs if Bluesky is going to become a shared technology across a wide variety of contexts.

I prefer your other idea of adding an attribute to the RE like `last_plan_returned`. Developers can simply ignore that attribute.

Zooming out, do we _need_ to change the RE at all? Since plans are just generators, you can just pass them as objects and wrap them... I tried this on my machine with copy-pasted class from Stack Overflow [1] and it works perfectly as far as I can tell. Here's the script I ran for full context:

```python
import bluesky
from bluesky import plans
import yaqc_bluesky

class Generator:
    def __init__(self, gen):
        self.gen = gen

    def __iter__(self):
        self.value = yield from self.gen

RE = bluesky.RunEngine()
sensor = yaqc_bluesky.Device(39900)
my_plan = Generator(plans.count([sensor], 5))
out = RE(my_plan)
print(out)
print(my_plan.value)
```

A little bit of extra work, true... but totally doable. If you're primarily thinking about testing, you could probably use some magic with fixtures and singletons to automatically wrap your plans. There are probably other ideas out there that are even more clever---there's some awesome patterns around generator usage and I'm just a beginner with that kind of syntax.

[1] https://stackoverflow.com/a/34073559

untzag (approved):
Approving because I don't see anything wrong with the implementation. If y'all want this behavior go for it :smile: 

I think holding off on documenting the behavior is fine, but I would raise an issue at least so that documentation actually happens in the future.

bluesky/run_engine.py
cjtitus: A non-none exec could, at this point, still be a _RunEnginePanic, which would not get raised by
` if (exec is not None and not isinstance(exec, _RunEnginePanic)):`
If exec is a runtime panic, bad things happen when trying to get `self._task_fut.result()`

There may be more elegant combinations of conditionals and try/except statements, but this one at least passes all the tests, and leaving out that conditional causes it to fail on the stopped/aborted tests.

cjtitus:
@untzag I should say that the plans don't necessarily have to be adaptive like the AI in "bluesky-adaptive", or even like plans that make decisions mid-scan for return values to be helpful. For instance, sometimes it is very useful to scan a motor, monitor a detector, and find the motor position that yields the maximum detector value. It is easier to have the plan itself deal with the details of how to find the maximum, rather than having a separate function that has to extract the data from a databroker and do fitting. Therefore, a plan should be able to return a value somehow, otherwise you have no way to actually use the fit value that the plan found.

@tacaswell and I discussed various solutions to this problem, and having a `last_plan_return` attribute of the RunEngine was one idea, but I believe our conclusion was that "less state is better". If we get push-back on this, and people think that it would be better to introduce a new attribute, all of the code that I have just written will still be useful, as it is needed to consistently deal with interrupts. But instead of returning the value, we would stash it in `last_plan_return`. 

I do think that there is an overall consistency to having the RunEngine act like a `yield from` statement. With this PR, calling
`ret = RE(plan())` acts conceptually like `ret = yield from plan()`, in that you get the value of the last return statement.

We also aren't sure how much the returned UIDs are used "in the wild". If you use them anywhere, that would certainly be helpful to know.

klauer:
I haven't been following this too closely, so keeping that in mind, throwing in my 2c:

What if I want the run starts UIDs _and_ the return value?
Could instead there just be a custom return type that is easily accepted by type checkers and yet also largely backward-compatible - without opting in to get return values?
e.g.,
```python
@dataclass
class PlanRunResult:
    run_start_uids: typing.Tuple[str, ...]
    result: typing.Any

    # behave like the run start UID tuple
    def __iter__(self):
        yield from self.run_start_uids

    # omitting __len__, __repr__, etc. that would make this compliant/useful
```

The `uids is not None` case above seems like it should just be changed to be backward-incompatible. If there are no UIDs to return, it should just be an empty tuple and then a falsey-check would be sufficient.

tacaswell:
I have come to the conclusion that returning the list of uids rather than the return of the plan was a bad idea as
it is dropping a communication channel on the floor (which we are super careful to never do inside plans).
it also leads to terrible things like setting up "Peaks" on the bec, aliasing `peaks = bec.peak_stats` into the user namespace and then we end up with users doing
```
> RE(some_plan)
> peaks.cntr   # MAGIC!
```

it would be much better if we could do

```
> peaks = RE(some_plan)
```

this comes up a lot on beamlines that do "rock the motor go to the peak" alignment

Accepting that we need to provide a way to get the the return value out, the options I thought of were:
 1. embrace the side band! (which what @untzag  suggested)
flag-day switch of the return type
what is proposed here (which is a bit sock-puppet written by me through significant slack discussions)
(edited)
2:55
I think 1 is "too odd" and as cjtitus points out it seems nice that ret = RE(plan) and ret = yield from plan ryhme so well
2:55
2 runs the risk of major breakage on the floor and is a non-starter
2:57
hence we landed on 3 so we can test this out, see how it feels, etc and then long-term move to it being the default (some chain like "can opt in", "can opt in and we warn if you dont", "have to opt-out", "we warn if you opt out", "removed"

ZLLentz:
First, I agree that the uids are not the optimal return type. I don't use them at all.

Everything in this pull requests seems good as implemented, and I would not object to a merge, but I'd like to echo Ken's suggestion as a better path forward. What if somebody wants additional information from the RE return later, in a way we can't envision yet?  In general I like to discourage multiple return types on principle because it makes working with these methods confusing.

Side-thoughts: most plans don't have an obvious value to return, unlike something like getting the peak. What should the return type of `bp.scan` be, for example? If I switch to the return value mode, and I see something like a `4` return from an arbitrary scan, how should I interpret that? Or am I expecting to modify my plans to return a well-keyed dict or a plan-specific data type? Or I suppose most of these returning `None` is fine.

In cases like this, a standard dataclass with uid, plan succeeded/failed, inital call args, etc. would be the most useful thing to see represented in `IPython`. But maybe there's a follow-up PR embedded here.

cjtitus:
@klauer that's certainly an interesting idea that shouldn't be too hard to prototype. I _was_ wondering about the case where someone wants both values, but didn't want to over-complicate things by making an object. Right now, you could check `_run_start_uids`, but this is exactly the sort of state that @tacaswell would like to eliminate.

It could be a good idea to roll the exit status into this as well, which is currently set in `RunEngine._exit_status`. This is exactly the kind of state that I was trying to avoid by returning plan values directly rather than in a `RunEngine.last_plan_return`. If we implemented a `PlanReturnResult` object, it could incorporate `run_start_uids`, `result`, and `exit_status`, and any other last-plan-state-saving variables that we might want to remove.

However, there is an elegant simplicity in simply returning the _result_, and not some complicated object. I would be open to trying to implement a return class if we decide that it is the correct decision.

untzag:
Thanks for the clarification @tacaswell, as anticipated I didn't appreciate the broader context. Knowing that the intention is to ultimately phase out the option in favor of always returning plan output makes me less afraid of long-term consequences from this change.

I still vote for a dataclass for extensibility. Perhaps rather than trying to fake backwards compatibility like @klauer suggested we could just make a really simple vanilla dataclass and move towards it while gradually phasing out existing behavior.

cjtitus:
Having just seen the comment from @ZLLentz, I think that plans returning `None` is fine, but the documentation should probably be modified to note the return value. Of course, plans currently return `None` for precisely the reason that this PR was opened -- if there is no way to get the plan return value, why return anything at all? Maybe clever people will find sensible things for some of the built-in plans to return.

Most built-in plans don't actually return `None` however, because any plan that has an inner function wrapped by @run_decorator will actually return a single uid. So `scan` will actually return the uid already, and most built-in single-run plans will do the same. This is currently not mentioned in the documentation, as it is not really relevant right now.

Expecting plans to return some complex dictionary or data structure sounds like too much. Plans that have a logical value to return should be able to return it, (and document it!) plans that don't logically return anything should not return anything.

I would be interested to see what @danielballan thinks about this, but I believe he's on vacation.

tacaswell:
> What if somebody wants additional information from the RE return later, in a way we can't envision yet?

That is part of why I think we should get the RE out of the way and return what ever the plan returns (like dask.delayed or asyncio tasks).

If we are going to with a rich object, maybe we want to track how futures / asyscio tasks work and return a rich object that has methods to get the result, the exception, the RE it came from, ect.  This would map `RE.__call__` <-> `loop.create_task()` which is not a terrible analogy.  I have a another half thought of adding a message for "please run this (sub) plan in the background" think the people who want to oscillate a sample underneath data collection or run "cheat" flyscans where you take data as fast as you can in the main plan while the background plan runs you through some trajectory (which you can totally do now, but you end up mixing the trajectory logic with the "please take data logic" in a way that is un pleasing (to me)) 

----

I am still in favor of doing the "opt-in -> opt-out -> remove" path as no matter what we land on as the _right_ thing to return is, I am pretty confident that the tuple of uids is the _wrong_ thing and we are still small enough that we can get away with a hard (but gentle) break on this.

cjtitus:
One additional thing I will note is that not all plans actually have a UID. A UID is a property of a _run_, which is really a specific type of plan (one wrapped by `@run_decorator`). If you just call `RE(mv(motor1, value))`, there's no UID created, so it just returns `()` currently. In some sense you can't get the RE out of the way if you ever want to return a UID list, because the RE is what's tracking the UIDs from the plans during `_open_run`. Otherwise, you'd need plans to start keeping track of UIDs themselves.

Returning a UIDs list is one of those things that seems really nice (why wouldn't you want to know, for sure, what the UID of your data is?), but nobody really uses in practice, and adds some complication (the need to save internal state in the RunEngine), and isn't general (not all plans generate a UID).

I will have to look again at the results objects that asyncio returns. Making it analogous to asyncio tasks sounds nice, but not that many people use asyncio, so it may not be a helpful analogy for many. Is there a reason we would want to have people call a result() method rather than access a result attribute? I guess in the case where there is no return, (like in an aborted plan) you could raise an exception, but I'm not convinced this is the best idea, or useful (I'm also not convinced it's _not_ the best idea, and it _could_ be useful).

cjtitus:
I am going to prototype the `dataclass` implementation of a simple return object that has the `run_start_uids`, `result`, and `exit_status` and see where that gets me.

ZLLentz (approved):
Pre-emptively adding my approval, I do like this update and have been thinking about what return values I want to include in my own plans.

prjemian:
My 2c.  Never use the uids returned from `RE(plan())`.  I like @klauer's data class idea but wonder if a dict might be the more flexible return type, given that our ideas on this may evolve further in the coming years of use.  What's the downside of such a dict?

cjtitus:
@prjemian I believe it's less about downsides and more about the _upsides_ of having a static, type-checkable dataclass, rather than a dictionary whose keys could be anything that we dream up. I don't think it would be harder to add a data class attribute than add a dictionary key. I wasn't envisioning any ability to add keys at runtime, so either way the RunEngine code needs to be modified to change the return values. I think @klauer and @untzag seem gung-ho about type-checking, so maybe they can chime in if I am incorrect.

klauer:
Well, ignoring TypedDict from 3.8+: mutable mappings are much more free-form and (in my opinion) clunkier to deal with than a good-old attribute-having object such as a dataclass instance.  In either case, a new attribute to the return type would be trivial.

cjtitus:
Is there any reason not to be maximalist about what goes in the RunEngineReturn dataclass? So far we have discussed:
* uids list
* plan result
* exit status
But one could easily imagine also returning
* _exception
* _interrupted
* _reason
Which just about rounds out the list of per-call RE state. Presumably these are all currently private because it is inadvisable to reach into the guts of the RunEngine and try to get information about state that gets cleared every call. But if we shove them into a return dataclass, then you have a safe way to access all possible information about the plan that was just run!

cjtitus:
Does Bluesky use dataclasses anywhere? I'm failing on Python 3.6 tests because dataclasses wasn't introduced until 3.7. However, there is a backport for 3.6. Is there some compatibility trick that we already use that I can replicate?

untzag:
@cjtitus if it helps, this syntax can be used for version-specific dependencies 

https://github.com/bluesky/bluesky/blob/master/requirements.txt#L10

klauer:
> Is there any reason not to be maximalist about what goes in the RunEngineReturn dataclass?

Good point - not that I can see. If we all agree this return type is a good thing (which it seems we mostly do), I'd say the more information, the better.

tacaswell:
By NEP29 we really should drop py36, but have been holding onto it out of concern for any facilities who have been slow to upgrade.

cjtitus:
I have added `_interrupted`, `_reason`, and `_exception` into the `RunEngineResult`. 

As an aside, it would really be nice for someone to fix the stochastic failure of test_table_external, or else skip it or mark it as an expected failure.

danielballan:
I was on vacation all last week and missed the action so far. I like the way this discussion has evolved. I will do a detailed implementation review in the next week or two.

bluesky/run_engine.py
```diff
@@ -42,6 +44,16 @@ class _RunEnginePanic(Exception):
...
 
 
+@dataclass
+class RunEngineResult():
+    run_start_uids: typing.Tuple[str]
+    plan_result: typing.Any
+    exit_status: str
+    interrupted: bool
+    reason: str
+    exception: typing.Any  # could be exception or None
```
ksunden, line 54:
```suggestion
    exception: typing.Optional[Exception]
```

bluesky/run_engine.py
```diff
@@ -42,6 +44,16 @@ class _RunEnginePanic(Exception):
     ...
 
 
+@dataclass
+class RunEngineResult():
+    run_start_uids: typing.Tuple[str]
```
ksunden, line 49:
```suggestion
    run_start_uids: typing.Tuple[str, ...]
```

This change allows variable length tuples by default the existing code would have only type checked for 1-tuples

cjtitus: Oh great, thanks for catching. I'm a bit new at type hints, and I guess they aren't actually checked by anything in the tests.

cjtitus:
I changed the `call_return_type` flag from a string to a boolean, which seems much simpler. It is now called `call_returns_result`, and the code and documentation have been changed to reflect this minor refactor. I added back in one more test that I forgot to parametrize, so I don't think any tests are skipped anymore.

I updated documentation, and added an entry for The `RunEngineResult` on the RunEngine API Documentation page, so that people can figure out what the heck you get from `__call__` now.

bluesky/run_engine.py
```diff
@@ -215,6 +244,13 @@ def block(ev: Threading.Event) -> None:
            - Matplotlib is imported and using a nbagg or ipympl backend (
              wait on the event and poll to push updates to the browser)
 
+    call_returns_result : bool, default True
+        A flag that controls the return value of __call__.
+        If ``True``, the ``RunEngine`` will return a :class:``RunEngineResult`` object that
+        contains information about the plan that was run. If ``False``, the
+        ``RunEngine`` will return a list of uids.
```
danielballan, line 251:
```suggestion
        ``RunEngine`` will return a tuple of uids.
```

bluesky/run_engine.py
```diff
@@ -724,6 +774,9 @@ def __call__(self, *args, **metadata_kw):
         -------
         uids : list
```
danielballan, line 775:
```suggestion
        uids : tuple
```

bluesky/run_engine.py
```diff
@@ -1106,6 +1183,14 @@ def abort(self, reason=''):
         """
         Stop a running or paused plan and mark it as aborted.
 
+        Returns
+        -------
+        uids : list
```
danielballan, line 1188:
```suggestion
        uids : tuple
```

bluesky/run_engine.py
```diff
@@ -1132,13 +1217,25 @@ async def _abort_coro(self, reason):
...
+        else:
+            return tuple(self._run_start_uids)
 
     def stop(self):
         """
         Stop a running or paused plan, but mark it as successful (not aborted).
 
+        Returns
+        -------
+        uids : list
```
danielballan, line 1233:
```suggestion
        uids : tuple
```

bluesky/run_engine.py
```diff
@@ -1161,12 +1258,25 @@ async def _stop_coro(self):
...
+        else:
+            return tuple(self._run_start_uids)
 
     def halt(self):
         '''
         Stop the running plan and do not allow the plan a chance to clean up.
 
+        Returns
+        -------
+        uids : list
```
danielballan, line 1274:
```suggestion
        uids : tuple
```

danielballan (approved):
I only found a minor docstring issue. Looks good.

danielballan:
I will merge this on Friday if there is no objections raised.

untzag:
Personally I really like this implementation, thanks @cjtitus :smile:  

Consider recording your intent to change default behavior in the docstring and/or sphinx. Personally I might even add an invisible warning right now for those with `call_returns_result=False`.

cjtitus:
I'll sneak a notification about the intended default behavior change into the RunEngine docstring, and leave it up to @danielballan when/how they want to add warnings and when they want to switch the default.

cjtitus:
Ok, hopefully that's the last change (just documentation!) before merge.

cjtitus:
Somehow a test that has never failed before, and is totally unrelated, failed on only the 3.6 build. (test_descriptor_layout_from_monitor in test_new_examples.py). No idea what that's about. Hopefully it can just be ignored.

danielballan:
Thanks @cjtitus .

I think we should move _real_ slow on this. One release with no warnings at all while we kick the tires, then (invisible) `PendingDeprecationWarning`, then (visible) `UserWarning`, then flip the default.

tacaswell:
Congrats @cjtitus !

PR #1466, opened 2021-08-16
[BUG] protocol should not require parent to be @property

untzag:
## Description

The current implementation does not allow the `parent` attribute to be set, which means that a simple device that specifies `self.parent = None` fails to pass. I don't understand why `parent` should be implemented as an `@property` rather than a simple attribute. It's likely I'm missing details, review from @ksunden will be very important here.

While we're at it, why do we use `Any` here. Couldn't we say `Optional[Readable]`?

The hints dictionary is also currently implemented as an `@property`, we might want to change that as well.

## Motivation and Context

I ran into this within https://github.com/bluesky/yaqc-bluesky/pull/59

## How Has This Been Tested?

Tested locally, seems to behave the way I want.

### DIFF
--- a/bluesky/protocols.py
+++ b/bluesky/protocols.py
@@ -28,11 +28,7 @@ class Status(Protocol):
 @runtime_checkable
 class Readable(Protocol):
     name: str
-
-    @property
-    def parent(self) -> Optional[Any]:
-        """``None``, or a reference to a parent device."""
-        ...
+    parent: Optional[Any]
 
     def trigger(self) -> Status:
         """Return a ``Status`` that is marked done when the device is done triggering.
@@ -117,11 +113,7 @@ class Movable(Readable, Protocol):
 @runtime_checkable
 class Flyable(Protocol):
     name: str
-
-    @property
-    def parent(self) -> Optional[Any]:
-        """``None``, or a reference to a parent device."""
-        ...
+    parent: Optional[Any]
 
     def kickoff(self) -> Status:
         """Begin acculumating data.


ksunden:
The big reason for things to be properties rather than attributes is that properties are not writable.

Essentially the contract is that the user cannot change the value of parent (or hints for that matter).

As for the Optional[Any], My original plan was to do `Optional[Readable]`, however e.g.  Flyable devices are not necessarily Readable (though many _are_)  Any was kind of a stand in... could probably be replaced with Optional[Union[Readable, Flyable]] (All Movables are Readable by inheritance/definition)

klauer:
Might [`typing.Final`](https://docs.python.org/3.8/library/typing.html#typing.Final) be appropriate here? (https://www.python.org/dev/peps/pep-0591/) It's more for "do not override in subclass", but aside from `@property` I'm not sure there's a better option.

untzag:
OK, I somewhat expected this answer. I understand that changing `parent` is an ugly thing to do. But from a pragmatic perspective we already have implementations that do self-assign `parent`. `yaqc_bluesky.Base` does this, `ophyd.Component` does this. Is the  small amount of  extra hostility towards existing implementations worth this small amount of extra assurance in object behavior?

Actually, why _can't_ parent be dynamic? What breaks?

On the other hand, why is `name` OK to change, should that all so be an `@property` without a setter?

To be explicit, I'm fine with this behavior. It's not hard to implement at all. But it was surprising to me so I think it warrants some discussion.

I like the idea of `Optional[Union[Readable, Flyable]]`!

I don't see how `typing.Final` helps us here.

ksunden:
I don't really think `Final` has any applicability at the protocol level

After thinking a bit more, and running some tests, I remembered that instance attributes are difficult for static type checking in general... Actually the protocol defined as a `@property` doesn't say that the thing must _not_ be writable, but rather that it is okay for it to be read only, that the only guarantee is that it is readable.

All of this only matters for static type checking, of course, as runtime checking purely checks `hasattr`: writability and signature/implementation do not matter at all for those runtime `isinstance` checks


This change would make it so that classes which implement things as properties are required to have a setter, which is not what we want

ksunden:
```python
from typing import Protocol

class A(Protocol):
 
    @property
    def test(self) -> int:
        ...

class B:
    test = 0

class C:
   def __init__(self):
       self.test = 1
c: A = B()from typing import Protocol

class A(Protocol):
 
    @property
    def test(self) -> int:
        ...

class B:
    test = 0

class C:
   def __init__(self):
       self.test = 1
c: A = B()
d: A = C()

d: A = C()

```
This file type checks fine, the fact of it being a property in the protocol _only_ increases flexibility for implementations (except, perhaps, as you seem to be running into, if you are trying to directly subclass the Protocol class, which I'm not sure is at all necessary, I actually think the right level of integration is tests which run mypy...

ksunden:
To put a bit more context, the last line of this fails:

```python
from typing import Protocol

class A(Protocol):
 
    @property
    def test(self) -> int:
        ...

class B:
    test = 0

class C:
   def __init__(self):
       self.test = 1

class D(Protocol):
    test: int

class E:
    @property
    def test(self) -> int:
        return 4


c: A = B()
d: A = C()

e: D = B()
f: D = C()
g: D = E()
```

The error message is:

```
mypy_test.py:30: error: Incompatible types in assignment (expression has type "E", variable has type "D")
mypy_test.py:30: note: Protocol member D.test expected settable variable, got read-only attribute
Found 1 error in 1 file (checked 1 source file)
```

untzag:
Thanks @ksunden, I learned something today!

Issue #1459, opened 2021-07-16
test test_table_external is stochastically failing


untzag:
`test_callbacks/test_table_external` [1] seems to be failing randomly. I can't track down exactly why---the fixtures or whatever that are creating the environment for that test are opaque to me.

[1] https://github.com/bluesky/bluesky/blob/master/bluesky/tests/test_callbacks.py#L150-L152

cjtitus:
I don't have the power to close this, but #1430 should actually solve this problem completely, so this issue is now fixed! We can run our tests in peace (and with the knowledge that failures are actual failures)!

mrakitin:
Closed via #1430.

PR #1467, opened 2021-08-16
BUG: done and success should not be class attributes of Status protocol

untzag:
## Description

Working further, and having been schooled by @ksunden in #1466, I'm now on the exact opposite side of the problem. The current implementation of `Status` does not work with our existing classes because we _do_ use `@property`. Error given is "expected settable variable, got read-only attribute".

## Motivation and Context

Current behavior does not pass existing Status objects.

## How Has This Been Tested?

Passes on yaqc_bluesky after this change.

### DIFF
--- a/bluesky/protocols.py
+++ b/bluesky/protocols.py
@@ -11,8 +11,6 @@ Configuration = Dict[str, Dict[str, Any]]
 
 @runtime_checkable
 class Status(Protocol):
-    done: bool
-    success: bool
 
     def add_callback(self, callback: Callable[["Status"], NoReturn]) -> NoReturn:
         """Add a callback function to be called upon completion.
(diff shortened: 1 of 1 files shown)


ksunden (approved):
LGTM, I agree this should be the correct protocol definition

untzag:
@ksunden shall we also change `name` to an `@property` here? It's going to be a problem in the future...

untzag:
I believe failure is unrelated, see #1459

ksunden:
@untzag may as well

untzag:
I believe nobody actually cares about protocols right now except myself and @ksunden, so I'm merging

danielballan:
Thanks for the initiative. I was on vacation all last week, just catching up.

I agree with the changes to Status. The name of a Device or Flyable is runtime settable. Whether they _should_ be is arguable, but at the time we believed they should be settable and I think imposing them to be read-only would break things.

Is there a way for Protocol to express “This may be read-only or settable”?

ksunden:
That is actually how these type check, the assertion is only that it is readable, but an attribute satisfies that much.

That said, the type checker would balk at something typed as `Readable` where you attempted to do `x.name = "new_name"`, I think still (though if it were more narrowly typed to something that had a simple attr or a setter for the property, it would not error)

PR #1470, opened 2021-09-03
Fix list_grid_scan md for plan pattern args

ksunden:
Update function always returns None, thus the expected dictionary was not actually included

## Description

Rework plan pattern args to not use `update`, as that does not return the dictionary it is updating.

## How Has This Been Tested?

Ran a list_grid_scan, ensured that the plan_pattern_args were not null.


danielballan:
Thank you, @ksunden!

PR #1472, opened 2021-09-09
Fix descriptors for flyers that do not implement read_configuration

tacaswell:
## Description

Fix descriptors for flyers that do not implement read_configuration.

## Motivation and Context

Previously if `read_configuration` was not defined we would not put an entry in the configuration key in the descriptor.  While the RE would run OK, this would case issues on data access with newer versions of databroker. 

I do not think we want to make a distinction between in the documents between "the device reported configuration and it was empty" from "the device did not report any configuration".

## How Has This Been Tested?

Added new test

### DIFF
--- a/bluesky/tests/test_flyer.py
+++ b/bluesky/tests/test_flyer.py
@@ -7,6 +7,8 @@ from bluesky import Msg
 from bluesky.plans import fly, count
 from bluesky.run_engine import IllegalMessageSequence
 from bluesky.tests import requires_ophyd
+from bluesky.tests.utils import DocCollector
+
 from ophyd import Component as Cpt, Device
 from ophyd.sim import NullStatus, TrivialFlyer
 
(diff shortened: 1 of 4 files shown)


tacaswell:
This was found by @MikeHart85 @mrakitin and @JunAishima at AMX today.

bluesky/tests/test_flyer.py
```diff
@@ -370,3 +371,50 @@ def read_configuration(self):
...
+        [
+            Msg("open_run"),
+            Msg("kickoff", flyer, group="foo"),
+            Msg("wait", group="foo"),
+            Msg("complete", flyer, group="bar"),
+            Msg("wait", group="bar"),
+            Msg("collect", flyer),
+            Msg("close_run"),
+        ],
+        d.insert,
```
mrakitin, line 413: Why not use `bp.fly` here?

mrakitin (approved):
Looks good! I left an optional suggestion to reuse the existing plan, but it's good to go as is.

bluesky/tests/test_flyer.py
tacaswell: With snark: because this was a few lines above, easy to copy-paste, and I did not remember the name of `bp.fly` ;)

More seriously: The important thing from the point of view of the RE is the sequence of messages so by passing it just the sequence it needs here it is very clear exactly what sequence of messages the test thinks it is testing and insulates this test from any possible changes to bp.fly.

PR #1473, opened 2021-09-14
DOC: release notes for v1.8.0

mrakitin:
These are proposed release notes for the upcoming release. The version number is up for discussion. Should we bump the minor number and release it as v1.8.0 due to the plan return value now?


tacaswell:
This should definitely be 1.8.0

docs/source/api_changes.rst
```diff
@@ -2,18 +2,53 @@
  Release History
 =================
 
+v1.7.1 (2021-09-01)
```
tacaswell, line 5:
```suggestion
v1.8.0 (2021-09-14)
```

tacaswell (approved):

PR #1474, opened 2021-09-21
FIX: catch last py310 issue

tacaswell:
I do not understand why this is not failing on CI as we are passing a dictionary instead of an event loop!

## Description

We need to unpack the loop kwargs dictionary.

## Motivation and Context

This fails many many tests on py310, but I am very confused why this is not failing on older verisons of Python as well.

## How Has This Been Tested?

Run locally.


tacaswell:
🤦🏻 sorry branch fail, but it works...

mrakitin (approved):

PR #1475, opened 2021-09-21
Read-only property `RunEngine.deferred_pause_requested`

dmgav:
Changes:

- Implement read-only property `RunEngine.deferred_pause_requested` which returns the value of private variable `RunEngine._deferred_pause_requested`.

- Raise `TransitionError` exception if `RunEngine.request_pause()` is called when Run Engine in a state different from `idle`. The exception is raised if the parameter `defer` is `True` or `False`.

- Updated `can_pause` named checker of `RunEngineStateMachine` so that it returns `True` if transition to `pausing` state is allowed. Obviously, the name checker was not updated when additional transitional states were added. The name checker was not anywhere in the code, so the change is unlikely to affect operation of the Run Engine.

## Motivation and Context

The `RunEngine.deferred_pause_requested` property may be useful for Queue Server, where the state of Run Engine is monitored from a separate thread.

## How Has This Been Tested?
The unit tests are implemented for each new feature.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -736,6 +736,11 @@ class RunEngine:
 
     async def _request_pause_coro(self, defer=False):
         # We are pausing. Cancel any deferred pause previously requested.
+        if not self.state.can_pause:
+            raise TransitionError(
+                f"Run Engine is in '{self.state}' state and can not be paused."
+            )
+
         if defer:
             self._deferred_pause_requested = True
             print("Deferred pause acknowledged. Continuing to checkpoint.")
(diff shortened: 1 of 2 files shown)


requirements-dev.txt
```diff
@@ -28,7 +28,7 @@ requests
 scikit-image
 scipy
 sphinx
-sphinx_rtd_theme
+sphinx_rtd_theme==0.5.2
```
danielballan, line 31: Please open an issue reminding us to follow up on this.

danielballan (approved):

tacaswell (approved):

requirements-dev.txt
dmgav: I am afraid `sphinx_rtd_theme` needs to be pinned in all projects. The new version v1.0.0 does not work for some reason. So this is a general issue.

PR #1478, opened 2021-09-29
adding derivative calculation to the compute function

gfreychet:
Adding a derivative calculation in the compute function. This will be used at SMI when we want to go to the inflection point after a scan (which is commonly done when scanning sample height).

## Description

The change consist in using numy.diff function to calculate the derivative and calculate the corresponding min, max, cen and com

## Motivation and Context

When we scan the sample height during alignment routine at SMI when we need to go to the inflection point of the scan

## How Has This Been Tested?

It has not been tested. It should not affect the previous calculation just add more 'parameters/options' to compute

### DIFF
--- a/bluesky/callbacks/fitting.py
+++ b/bluesky/callbacks/fitting.py
@@ -250,6 +250,7 @@ class PeakStats(CollectThenCompute):
                 y.append(_y)
         x = np.array(x)
         y = np.array(y)
+
         if not len(x):
             # nothing to do
             return
@@ -276,8 +277,8 @@ class PeakStats(CollectThenCompute):
         crossings = np.where(np.diff((y > mid).astype(np.int)))[0]
         _cen_list = []
         for cr in crossings.ravel():
-            _x = x[cr:cr+2]
-            _y = y[cr:cr+2] - mid
+            _x = x[cr:cr + 2]
+            _y = y[cr:cr + 2] - mid
 
             dx = np.diff(_x)[0]
             dy = np.diff(_y)[0]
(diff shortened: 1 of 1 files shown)


gfreychet:
@tacaswell This is an attempt of adding derivative calculation in the compute function

bluesky/callbacks/fitting.py
```diff
@@ -290,5 +291,33 @@ def compute(self):
...
                 self.fwhm = np.abs(self.crossings[-1] - self.crossings[0],
                                    dtype=float)
+
+        # Calculate the derivative of the data
+        x_der = x[1:]
+        y_der = np.diff(y)
+        self.x_der_data = x_der
+        self.y_der_data = y_der
+        self.max_der = x[np.argmax(y_der)], self.y_data[np.argmax(y_der)],
+        self.min_der = x[np.argmin(y_der)], self.y_data[np.argmin(y_der)],
```
mrakitin, line 301: Argmin and argnax are calculated twice each, that is not efficient. I think it'd make sense to stash the values into temp vars and reuse.

mrakitin (commented):
Looks like the `..._der` values could be combined into a separate attribute such as `self.der.<quantity>`, so that they are not mixed with the non-derived values. What do you think? [`namedtuple`](https://docs.python.org/3/library/collections.html#collections.namedtuple) can be a good candidate for the `der` class.

bluesky/callbacks/fitting.py
gfreychet: 2 questions:
-Should the derivative calculation be optional? I.e. only when asked should we spend time computing it, using a flag?
-I am defining the der namedtuple dic in the __init__ instance. Is this the correct way of doing so or should it be only defined in compute?

mrakitin: I like the optional approach as it won't break the existing `PeakStats` content.

For the second question, defining it in `__init__` makes sense, and then populate it in the `compute` method.

bluesky/callbacks/fitting.py
```diff
@@ -290,5 +294,35 @@ def compute(self):
...
+        # Calculate the derivative of the data
+        x_der = x[1:]
+        y_der = np.diff(y)
+        self.der.x = x_der
+        self.der.y = y_der
+        max_der = np.argmax(y_der)
+        min_der = np.argmin(y_der)
+        self.der.max = x[max_der], self.y_data[max_der],
+        self.der.min = x[min_der], self.y_data[min_der],
+        self.der.com, = np.interp(center_of_mass(y), np.arange(len(x_der)), x_der)
```
mrakitin, line 307: Don't you want `center_of_mass(y_der)` here?

mrakitin (commented):
Looks good. I have a single question to address.

Also, a simple test would be great to have, to test the new functionality when it's enabled.

bluesky/callbacks/fitting.py
gfreychet: You are right.

I tried to incorporate a der flag in compute with a default value being false, so it will compute the derivative only if specify to be True. What do you think?

How can I test this at the beamline (not a big expert with git). Should I create a specific environement at the beamline and install git+my forked bluesky?

mrakitin (commented):
I think it'll be better to add a `compute_derivative` (or similarly named) attribute via `__init__` as a keyword argument. That way `BestEffortCallback` (`bec`) can work without modifications.

bluesky/callbacks/fitting.py
mrakitin: Answered via Slack about testing.

mrakitin:
@gfreychet, I am done with my updates to your PR branch. I added a kwarg to enable derivatives calculation via `bec`, so that could be enabled for the beamlines that need it. Do you mind testing this code at the beamline (the way you did it yesterday)?

gfreychet:
After testing the code at SMI beamline (cloning the branch with the help of @mrakitin), we were able to calculate on an old scan the derivative and the corresponding CEN, COM, FWHM, ...

dmgav:
If it is important to include this feature in the next release, we could merge this PR and create an issue to add tests and revise the code before the following release.

dmgav:
@mrakitin Why did you make `calc_derivative_and_stats` **required** kwarg?

mrakitin:
See https://github.com/bluesky/bluesky/pull/1478#discussion_r723736105.

PR #1484, opened 2021-10-11
TST: Fixed RE conftest parametrization for call_returns_plan_result

cjtitus:
## Description

conftest.py now correctly produces RE fixtures with call_returns_plan_result = True and False. Before, it was always essentially true. Also fixes one test that failed with correct parameterization

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -295,6 +295,11 @@ class RunEngine:
     ignore_callback_exceptions
         Boolean, False by default.
 
+    call_returns_result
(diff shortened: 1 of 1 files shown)


mrakitin:
Ah, did not realize it's that test 😆. Thanks, @cjtitus!

cjtitus:
Added a public property `call_returns_result` and changed all the existing tests that were calling the "private" underlying `RE._call_returns_result` flag to call `RE.call_returns_result`. This was maybe the first time that I have ever used sed and had it save me time.

tacaswell:
Thank you @cjtitus !

PR #1487, opened 2021-11-09
FIX: Also clear the describe cache on configure

tacaswell:
## Description

Nuke more caches on configure

## Motivation and Context

This is to fix an issue at FXI where they were changing the number of
images (and hence the shape) on an AD.  Through enough stage/unstaging the
Resource documents had correct information, but because we were caching the
result of `describe` the first time we saw the object all of the streams were
erroneously reporting that they had the same shape which was eventually causing
tiled to error out.

For this to work it makes the assumption that any PV values that might sneak
into the descriptor (rather than the configuration) would also be configuration
values (if they are not, the configure message will fail on the call to the
configure ophyd method).

This is also slightly over-kill for many cases where nothing in describe
changes, but we have no other channel to clear these caches.

On a bit more consideration, I think this has been a bug from the very beginning as we did not consider that the _shape_ could change over the course of a run (and that you would get that by putting each version of the shape in its own stream).

## How Has This Been Tested?

Hot-patched at a beamline! 

https://github.com/NSLS-II-FXI/profile_collection/blob/e34bf4b79588efe972ccabe0e6faab3df28fbcd5/startup/01-patch_bundler.py#L1-L34


ZLLentz (approved):
- Assigned as reviewer during the bluesky pr dive call
- Code looks self-explanatory and good
- Ran the unit tests for all of our libraries against master + this PR, all the bluesky-related tests pass, I think we should merge this

bluesky/bundlers.py
```diff
@@ -777,4 +782,12 @@ async def configure(self, msg):
             obj_set, _ = self._descriptors[name]
             if obj in obj_set:
                 del self._descriptors[name]
-        await self._cache_read_config(obj)
+
+        # nuke the caches, let read re-fill them.  Saves us work
+        # for multiple
+        if obj in self._describe_cache:
+            # assume the caches are coherent!
```
coretl, line 789: The only time I can think of when the caches are not coherent is for the flyer case, and then `obj` will be in the `config_*_cache` but not in the `describe_cache`. I guess we don't need to clear these caches for the flyer case?

coretl: Also, we didn't document `configure` in the protocols because we thought we were going to deprecate it, does this mean we need to keep it? Or should we think of a different way of invalidating caches? Or is this a very special case that is unlikely to reoccur?

stan-dot: @coretl what is the deprecation status for `configure` in the protocols?

the function is there in the `bundlers.py` in the latest main at  the moment

PR #1499, opened 2022-01-14
Run decorator returns the plan result, not uid

cjtitus:
When writing adaptive plans, it's very useful to return a result from one plan, and use that result to decide whether, or how, to run another plan.

Unfortunately, if you are orchestrating a series of "full" plans, rather than stub plans, the run_decorator will gobble up your plan's intended return value and substitute the run uid. Not cool! My plans want to return things for a good reason! All that run_decorator's uid return does, as far as I can tell, is stop decorated plans from returning useful results.

## Motivation and Context

An example would be an alignment plan, where you want to align one motor first, and then align a second after repositioning.

```python
def align_z(det, motor):
    @stage_decorator([det])
    @run_decorator()
    def inner_align_z():
        # some fancy alignment scan and logic
        # that provides the correct value of the motor, z
        return z
    return (yield from inner_align_z())

def alignment_plan():
    z = yield from align_z(det, zmotor)
    yield from abs_set(zmotor, z + 2)
    x = yield from align_x(det, xmotor)
```

In this example, it looks like align_z will return a motor value, `z`. `inner_align_z()` returns `z`, and `align_z` returns the result of `inner_align_z()`. But because `inner_align_z` has been wrapped by the run_decorator, `z` will be hijacked and turn into a uid, which is confusing, and not what I intended. Now, you could get around this by making `align_z` a stub plan, but what if you want to be able to run it by itself? How would you test `align_z` by itself if it were a stub plan?

## How Has This Been Tested?

Because the plan result wasn't previously returned, only breaks one test, `test_uid_passthrough` in test_supplemental_data.py, which for some reason tests this exact case, internally to a plan. 

## What might this actually break?
If people have been writing plans that rely on other plans returning UID, this will break those plans. I can't imagine why you would need a UID in the middle of a plan, since as far as I know, this would only be useful to interact with the DataBroker, which you probably shouldn't do _while running a plan_.

However, people do lots of things, usually because they found that they work, so I'd love to know if this is something that a lot of people are doing. If it is, I can find another workaround, I suppose. I would love an explanation of why it is desirable that a plan returns a uid, and what this could be used for.

### DIFF
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -314,7 +314,7 @@ def run_wrapper(plan, *, md=None):
     md : dict, optional
         metadata to be passed into the 'open_run' message
     """
-    rs_uid = yield from open_run(md)
+    yield from open_run(md)
 
     def except_plan(e):
         if isinstance(e, RunEngineControlException):
@@ -322,10 +322,9 @@ def run_wrapper(plan, *, md=None):
         else:
             yield from close_run(exit_status='fail', reason=str(e))
 
-    yield from contingency_wrapper(plan,
-                                   except_plan=except_plan,
-                                   else_plan=close_run)
-    return rs_uid
+    return (yield from contingency_wrapper(plan,
+                                           except_plan=except_plan,
+                                           else_plan=close_run))
 
 
 def subs_wrapper(plan, subs):
--- a/bluesky/tests/test_supplemental_data.py
+++ b/bluesky/tests/test_supplemental_data.py
@@ -141,6 +141,7 @@ def test_pickle():
     assert D2.flyers == D.flyers
 
 
+"""
 def test_uid_passthrough(RE, hw):
     # Test that none of the preprocessors in SupplementalData break plans
     # returning uids (a bug that was caught on the floor).
@@ -168,3 +169,4 @@ def test_uid_passthrough(RE, hw):
         assert isinstance(uid, str)
 
     RE(mycount2())
+"""


jklynch:
This feature seems like a good one to me, but changing the behavior of @run_decorator has the potential to affect a lot of code. How about adding a new decorator and leaving @run_decorator unchanged?

cjtitus:
I just wonder, was anybody actually using the return value from @run_decorator? As the RunEngine used to _discard_ the return value from all plans, the @run_decorator return could only have been used internally to a plan. Were people building plans that ran sub-plans and then collected their UIDs? What would that possibly have been useful for, unless you built a plan that accessed the databroker (which I believe is very inadvisable)?

I could of course just copy the run_decorator code with a different return behavior, but I think that having two nearly-identical run_decorators is probably a bad idea long-term.

tacaswell:
> As the RunEngine used to discard the return value from all plans, the @run_decorator return could only have been used internally to a plan. 

Yes, some of the plans that take background or calibration data capture this and then write it into the start document of further runs.  We have started to recommend against this, but it is a reasonable thing to do.

>  unless you built a plan that accessed the databroker

There are a lot of plans on the floor at NSLS-II that access databroker (but to be fair most of them do `db[-1]`).

>  I think that it was a mistake to have the run_decorater drop your plan's return value and substitute the UID, and a further mistake to have the RunEngine itself drop any plan's return value, and also substitute the UID. 

The causality is the other way around, the choice to have the RE return the uids was made first (so we did not have many plans with meaningful returns).  For internal usage collecting uids makes sense (record of what you did!).  

----

I think the right path here is to add a flag to `run_decorator` (and wrapper) that is `pass_plan_return=False` so you can opt-in to getting the return value, but still get something sensible if you are lazy and do now want to manage your own returns.

cjtitus:
Makes sense. Actually, in the time since opening this PR, I have realized that I also get UIDs for calibration in this way... how embarrassing! I actually also agree that this is something that I shouldn't have done (especially the way that I do it, which is particularly awful, as it involves updating a persistent global dictionary) but how to create links between plans in a less hack-y way is probably a subject for a longer discussion. I wish I had a better way to do it.

I wonder if there is a compromise that, like the RunEngine return, lets us have our cake and eat it too. For the RunEngine return, we eventually realized that we could just create a class that returns both the plan result AND the UIDs, along with any errors.

Do you think a similar thing would work here? pass_plan_return=True could, at minimum, return a tuple consisting of the UID + the return value. Or a dictionary, or an object. Maybe it could have success messages! I sure know that I have some adaptive plans that fail sometimes, and should be able to report their failure, because there is sometimes a very obvious way to re-try.

tacaswell:
> Do you think a similar thing would work here? pass_plan_return=True could, at minimum, return a tuple consisting of the UID + the return value. 

I concern with that is it makes understanding the return value hard as the code will say

```python
     return 'bob'
```

and what you get back is `(uid, 'bob')` which is very non-obvious.  With the RE case we are explitily a running-stuff-for-you service (think analogy to `subprocess.run` so returning a structured object with both the return of what ever the user wanted run along with some meta-data about how the running of the thing went feel natural.

Another thought that might work here is to also add a `pass_uid` flag (and at this point a new decorator may be a better choice that just does both of these things by default) so

```
@run_decorator(return_plan_return=True, pass_uid=True)
def my_plan(blah, blarg, *, start_uid):
    ...
```

and then the run decorator will not expose `start_uid` out and will source it itself?  That way if you want both, you can now do what ever structuring you need to.

cjtitus:
> > Do you think a similar thing would work here? pass_plan_return=True could, at minimum, return a tuple consisting of the UID + the return value.
> 
> I concern with that is it makes understanding the return value hard as the code will say
> 
> ```python
>      return 'bob'
> ```
> 
> and what you get back is `(uid, 'bob')` which is very non-obvious.

Of course currently, the code says
```python
    return 'bob'
````
and what you get back is `uid`, which is even less obvious.

I am becoming persuaded that maybe a custom decorator that replaces the run_decorator functionality may be the way to go. I can prototype this in my own code and see what makes sense. It could be
`@run_return_decorator` and it would return a tuple, and if you use `@run_return_decorator` you should just know that it will do this.

tacaswell:
> and what you get back is uid, which is even less obvious.

Fair, but making a poor choice one way does not justify making another poor choice to compensate ;) 

If we are starting from scratch, I think a decorator that both returns what ever the wrapped function returns _and_ passes the start uid (or full start document could see a case either way) into the wrapped plan.  It is still a _bit_ odd but not unheard of for decorators to add / remove / modify input to the wrapped function.

cjtitus:
I'm not as concerned with passing the start UID into the plan, as getting both the UID and return value out of the plan.

If you have to write your plan to accept a UID, just because you intend to have the plan return it in a tuple with the return value, I think it's just simpler to have a decorator do that tuple-ization for you, no? That's what decorators are for, after all.

Wouldn't it be the case that anybody using the new decorator would know what they are signing up for?

jklynch:
From my armchair, I am concerned that having the run_decorator insert a parameter into a function (plan) call could tie that function (plan) to the decorator. What if one wants to call the function (plan) without that decorator?

cjtitus:
I am thinking now about a run_decorator alternative that leaves the plan return value completely alone, paired with a RunEngine command to get the most recent run start UID, or the list of run start UIDs.

stan-dot:
@cjtitus what are your latest thoughts on this?

Is this still an ongoing problem?

> Unfortunately, if you are orchestrating a series of "full" plans, rather than stub plans, the run_decorator will gobble up your plan's intended return value and substitute the run uid. Not cool! My plans want to return things for a good reason! All that run_decorator's uid return does, as far as I can tell, is stop decorated plans from returning useful results.

cjtitus:
@stan-dot I still believe this would have been a better way to structure the run_decorator from the beginning, but it's something I've been able to work around when necessary. Trying to actually change this would break too much for too many people. As I mentioned, if I really need this, I'll probably just re-implement run_decorator.

Issue #1501, opened 2022-01-26
Questions about protocols


thomascobb:
#1445 implemented some protocols, and I'm in the process of proposing some asyncio compatible versions of them. In doing so I'm trying to understand what is there at present and have some questions:

1. I don't think `Status.add_callback` should never return, should it just return None? https://github.com/bluesky/bluesky/blob/master/bluesky/protocols.py#L14
2. When might `pause` and `resume` never return? https://github.com/bluesky/bluesky/blob/master/bluesky/protocols.py#L195
3. Could we tie down the definition of Configuration down a bit more using a TypedDict? https://github.com/bluesky/bluesky/blob/master/bluesky/protocols.py#L9 Although I was struggling to find where all the optional things that could go in a Configuration were defined...
4. I was thinking about putting `name` and `parent` in a `Named` protocol, and `read_configuration` and `describe_configuration` in a `Configurable` protocol and inheriting from them in `Readable` and `Flyable`, would this work?
5. I was wondering if we needed the inheritance at all, does a `Settable` also have to be `Readable`? Or should the RE do two isinstance checks if it needs something that is both?
6. `runtime_checkable` can't tell the difference between `async def read()` and `def read()`, do we want to require the runengine to check using `inspect.iscoroutinefunction` everything it's going to run or should we use `async def read_async()` instead?
7. I can't actually find anywhere that the RE does `isinstance(Readable)` or similar, did I miss something or is there a plan to use these in the future? 
8. Could we make `trigger` (and other `Status` returning functions) return an `Optional[Status]` rather than a Status that is marked as done immediately?

I've discussed some of these with @tacaswell and @danielballan but was interested to see @ksunden and @untzag 's opinions too

ksunden:
1/2) I think I got confused about `NoReturn` when I was doing this... I now realize that is intended for "always Raises/gets in infinite loop" rather than "has no _explicit_ return", which is what I was trying to indicate, and should just be indicated by `None`

3) I'm open to options, though much of this is quite loosely defined, and I went pretty generic to have _something_ as the return value which current implementations passed.

4) No real argument here, I _almost_ did at least the `Configurable` one at the time, but decided what I had was good enough and could be iterated on.

5) I think the idea is that anything that can be set can also be read, I was mostly going off of prior docs for that being a required relationship.

6) No strong opinion, I think that is more on what makes sense for RE

[...]

PR #1502, opened 2022-01-27
Add async capabilities to protocols

thomascobb:
## Description
An attempt at the tweaks outlined in #1501 to provoke discussion

## Motivation and Context
So that asyncio can be used in `ophyd.v2` and in preparation for https://github.com/bluesky/ophyd/issues/1013

## How Has This Been Tested?
Not tested yet, will be tested as part of `ophyd.v2` when changes are agreed, but here as a draft for discussion


bluesky/protocols.py
```diff
@@ -1,17 +1,39 @@
...
+    # TODO: what about status/severity?
+
+
+class EventDescriptorMandatory(TypedDict):
+    source: str
+    dtype: Literal["string", "number", "array", "boolean", "integer"]
+    shape: List[int]
+
+
+class EventDescriptor(EventDescriptorMandatory, total=False):
```
thomascobb, line 21: As per https://blueskyproject.io/bluesky/hardware.html#bluesky.protocols.Readable.describe

Are there more things that we can add here?

bluesky/protocols.py
```diff
@@ -31,12 +53,39 @@ def success(self) -> bool:
...
+        ...
+
+    def describe_configuration(self) -> SyncOrAsync[Dict[str, EventDescriptor]]:
+        """Same API as ``describe``, but corresponding to the keys in
+        ``read_configuration``."""
+        ...
+
+
+@runtime_checkable
+class Readable(Named, Protocol):
```
thomascobb, line 88: This separates Configurable and Readable. This is because simple objects (like Signals) don't have any configuration, so this saves a bit of code. I also plan to put `configure()` and `validate()` in `Configurable` at a later date

bluesky/protocols.py
```diff
@@ -31,12 +53,39 @@ def success(self) -> bool:
 
 
 @runtime_checkable
-class Readable(Protocol):
+class Named(Protocol):
+    @property
+    def name(self) -> str:
+        ...
+
     @property
```
thomascobb, line 61: What is parent actually used for?

bluesky/protocols.py
```diff
@@ -31,12 +53,39 @@ def success(self) -> bool:
 
 
 @runtime_checkable
-class Readable(Protocol):
+class Named(Protocol):
+    @property
+    def name(self) -> str:
```
thomascobb, line 58: In our split of Device and Ability, do the following need a name, and if so what should they be:
- A Device that represents the areaDetector driver and plugins of a Pilatus detector
- An Ability that lets the Pilatus be software triggered
- An Ability that lets the Pilatus be flyscanned
- A Signal for the max intensity of a frame as calculated by the NDStats plugin

bluesky/protocols.py
```diff
@@ -151,26 +171,20 @@ def collect(self) -> Generator[Dict[str, Any], None, None]:
         """
```
thomascobb, line 171: I guess collect() output is https://blueskyproject.io/event-model/data-model.html#event-document but without `seq_num`, `uid` and `descriptor`?

bluesky/protocols.py
```diff
@@ -151,26 +171,20 @@ def collect(self) -> Generator[Dict[str, Any], None, None]:
...
         ...
 
-    def describe_collect(self) -> Dict[str, Configuration]:
+    # TODO: is this actually right? Do we need collect_asset_docs too?
```
thomascobb, line 174: I guess `collect_asset_docs` outputs resource and datum documents? I found this:
https://blueskyproject.io/event-model/data-model.html?highlight=path_semantics#resource-document

but no docs on how to write collect_asset_docs anywhere...

bluesky/protocols.py
```diff
@@ -151,26 +171,20 @@ def collect(self) -> Generator[Dict[str, Any], None, None]:
...
-
-    def describe_configuration(self) -> Configuration:
-        """Same as for a Readable device."""
-        ...
-
 
 @runtime_checkable
 class Stageable(Protocol):
+    # TODO: we were going to extend these to be able to return plans, what
+    # signature should they have?
```
thomascobb, line 187: @tacaswell we discussed returning plans but I'm not sure what the type signature should be

bluesky/protocols.py
```diff
@@ -192,18 +206,20 @@ def unstage(self) -> List[Any]:
 
 @runtime_checkable
 class Pausable(Protocol):
-    def pause(self) -> Union[NoReturn, None]:
+    # TODO: Should these be Status instead?
```
thomascobb, line 209: We could make this return an `Optional[Status]` I guess? Then maybe `trigger()` should do the same?

bluesky/protocols.py
```diff
@@ -227,14 +243,62 @@ def subscribe(self, function: Callable) -> Union[NoReturn, None]:
...
+
+    def close(self):
+        """Closes connection"""
+        ...
+
+
+@runtime_checkable
+class Monitorable(Protocol):
+    """Async compatible version of Subscribable with tweaked interface"""
+    # TODO: monitor() returns a Subscription, do we like the naming?
```
thomascobb, line 268: I had a name collision here, so I renamed the async version of Subsribable "Monitorable" after the bluesky verb... Do we like the name? And the interface?

bluesky/protocols.py
```diff
@@ -227,14 +243,62 @@ def subscribe(self, function: Callable) -> Union[NoReturn, None]:
...
+        ...
+
+
 @runtime_checkable
 class Checkable(Protocol):
-    def check_value(self, value: Any) -> Union[NoReturn, None]:
+    # TODO: should this be allowed to do I/O to do this check? If so,
+    # then this should only be used after stage() as that might change
+    # some params. If not, then it will need some cached limits to check
+    # against which still probably come from stage.
```
thomascobb, line 300: This is a tricky problem that bit us with pymalcolm, we change settings as part of setting up a device for a scan, but want to check if a scan config is valid before that happens. I thought we might be able to use saved settings to validate against, but if we have a `stage` plan then maybe we could run that and parse the `set` messages to use those values?

thomascobb:
@callumforrester any comments?

bluesky/protocols.py
```diff
@@ -1,17 +1,39 @@
...
 except ImportError:
-    from typing_extensions import Protocol, runtime_checkable
+    from typing_extensions import Literal, Protocol, TypedDict, runtime_checkable
 
-from typing import Dict, Any, Optional, Callable, Generator, List, Union, NoReturn
+from typing import Any, Awaitable, Callable, Dict, Generator, List, Optional, TypeVar, Union
 
 
-Configuration = Dict[str, Dict[str, Any]]
+class EventData(TypedDict):
```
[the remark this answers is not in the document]
callumforrester: Not sure, but for this and other event schemas, might they be better residing in https://github.com/bluesky/event-model?
@danielballan/@tacaswell, any comments.

bluesky/protocols.py
callumforrester: I'd be a little bit wary of that. While it is useful to return `Optional` and `SyncOrAsync` for backwards-compatibility, do we risk making our interface too ambiguous if we do it too often? What are other people's thoughts?

bluesky/protocols.py
callumforrester: Not sure what the best way forward is here but if we're going to have separate sync and async protocols for something I think there should be a deprecation procedure or at least some `TODO` about future reconciliation.

callumforrester (commented):
Have left a few, @tacaswell and @danielballan feel free to correct me on any of them

bluesky/protocols.py
```diff
@@ -1,17 +1,39 @@
...
+    upper_ctrl_limit: float
+    # TODO: add other limits here?
+
+
+T = TypeVar("T")
+SyncOrAsync = Union[T, Awaitable[T]]
 
 
 @runtime_checkable
 class Status(Protocol):
```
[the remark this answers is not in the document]
thomascobb: I don't think so, the existing Status doesn't have one, and it's an implementation detail in `ophyd.v2` that I'm using one to wrap an awaitable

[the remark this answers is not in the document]
callumforrester: If you're just working with protocols then you can't consider `Status` to be `Awaitable`, is that the behaviour we want? Can see a few pros and cons either way?
https://mypy-play.net/?mypy=latest&python=3.10&gist=90fc21fde784e7da73de27b751983ad0

bluesky/protocols.py
ksunden: I think I remember a question arising in Queueserver on whether we wanted bare Signals to show up as proper Readables or not, and the answer at that time being "no"

Opinions may have changed since then/queueserver can choose to further filter I suppose.

thomascobb: Ophyd Signals at the moment conforms to Readable, although I don't know if the current behaviour of returning identical dictionaries for `read()` and `read_configuration()` is desirable:
https://github.com/bluesky/ophyd/blob/master/ophyd/signal.py#L381-L444

Maybe @dmgav could answer how Queueserver treats current Ophyd Signals?

bluesky/protocols.py
thomascobb: On reflection, I think anything that does I/O should return `Status` (or `Optional[Status]` for backwards compatibility) if it doesn't return a value, and `SyncOrAsync` if it does. That would make `pause` and `resume` return `Optional[Status]` and I would suggest `trigger` return `Optional[Status]` to make it consistent. This means that if inheriting from the `Readable` protocol, `trigger` doesn't need defining for `Signal`s

callumforrester:
@thomascobb Food for thought: Have we considered maintain two sets of protocols rather than one set with doubled-up capabilities?

For example:
```python
@runtime_checkable
class Movable(Protocol):
    def set(self, value: Any) -> Status:
        """Return a ``Status`` that is marked done when the device is done moving."""
        ...

@runtime_checkable
class AsyncMovable(Protocol):
    async def set(self, value: Any) -> AsyncStatus:
        """Return ``AsyncStatus`` that is marked done when the device is done moving."""
        ...
```

It allows for a clean break and doesn't change where the differentiating logic in the RE has to go. I can imagine there are disadvantages though, bloat for one. It might help to resolve some of the backwards-compatibility issues discussed here though.

thomascobb:
> @thomascobb Food for thought: Have we considered maintaining two sets of protocols rather than one set with doubled-up capabilities?

I started there, but it turns out runtime_checkable can't tell the difference between those two protocols, which means an additional `inspect` call on each method. At that point you might as well call it an see whether it returns an `Awaitable` or not. That's what led me to try and combine the two.

bluesky/protocols.py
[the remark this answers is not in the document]
thomascobb: This depends if we are talking protocol or concrete classes:
https://mypy-play.net/?mypy=latest&python=3.10&gist=8211643e2fd3aebcfa42a7f23e7a4a0e
If I declare that `ophyd.v2.SignalRW.set()` returns a `ConcreteStatus` then we can await it no problem, but if I write a generic function that takes a `Movable` then we can't. I haven't thought far enough ahead to think whether that will be a problem though...

bluesky/protocols.py
thomascobb: Probably best to make this `SyncOrAsync[None]` but put in the docs that no signals should be written to as this can be called during a scan

bluesky/protocols.py
thomascobb: Case 2 and 3 are the same device, but 2 abilities with the same name

thomascobb: Case 4: Add `__getattr__` on `Readable` that wraps signals in readable interface, and top level Ability knows about hints etc. of its children

bluesky/protocols.py
thomascobb: Decided that bare Signals don't need a read() method, but we should be able to get a Readable ability on any Signal via its top level Ability. E.g. "pilatus_NDStats_intensity" is produced from the pilatus Ability with a couple of special method calls to drill down

bluesky/protocols.py
thomascobb: Walks up the tree to stop duplicate stages

bluesky/protocols.py
thomascobb: decided split was fine

bluesky/protocols.py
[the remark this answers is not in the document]
thomascobb: Call it a Reading

bluesky/protocols.py
```diff
@@ -1,17 +1,39 @@
...
 
-from typing import Dict, Any, Optional, Callable, Generator, List, Union, NoReturn
+from typing import Any, Awaitable, Callable, Dict, Generator, List, Optional, TypeVar, Union
 
 
-Configuration = Dict[str, Dict[str, Any]]
+class EventData(TypedDict):
+    value: Any
+    timestamp: float
+    # TODO: what about status/severity?
```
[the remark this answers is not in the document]
thomascobb: yes

bluesky/protocols.py
thomascobb: yes

untzag:
@thomascobb really basic question here---sorry I'm slow :smile: 

The protocols already support callbacks for subscription. Why do we need to change the protocols for ophyd v2? Couldn't asyncio be an implementation detail?

thomascobb:
> @thomascobb really basic question here---sorry I'm slow smile
> 
> The protocols already support callbacks for subscription. Why do we need to change the protocols for ophyd v2? Couldn't asyncio be an implementation detail?

This is driven by 2 things:
1. For asyncio, we would like callables to optionally be coroutines
2. It would appear that `subscribe` doesn't actually return what the run engine wants. At the moment it has to call `read()` each time the callback fires: https://github.com/bluesky/bluesky/blob/dc8b36fb1030a9116b6fe908ccdd32b58e254804/bluesky/bundlers.py#L273-L285

If we narrowed the protocol be: 
```python
Callback = Callable[[Dict[str, EventData]], SyncOrAsync[None]]

@runtime_checkable
class Subscribable(Protocol):
    def subscribe(self, function: Callback) -> None:
        ...
    def clear_sub(self, function: Callback) -> None:
        ...
```
Then:
1. Would probably be OK with current ophyd as the RE doesn't do sync callbacks
2. We could check in the RE what was passed to the callback function, and fallback to the current behaviour of calling `read()` if it didn't contain a `Dict[str, EventData]`

untzag:
Thanks @thomascobb 

> It would appear that subscribe doesn't actually return what the run engine wants. At the moment it has to call read() each time the callback fires.

Wow, I missed this subtlety I see what you mean that needs to change.

>  For asyncio, we would like callables to optionally be coroutines.

As in passing awaitable callbacks into my subscribe method? Can you say way? You can write a synchronous function which launches coroutine(s), isn't that enough?

thomascobb:
> As in passing awaitable callbacks into my subscribe method? Can you say way? You can write a synchronous function which launches coroutine(s), isn't that enough?

That's a good point, I was thinking about this like `read()`, where the protocol is implemented by ophyd, so needs to be async compatible, so we can handle the complexity of the return value in the RE. For `subscribe()`, the callback is implemented only in the RE, so it makes sense to put the complexity there, and make the `subscribe` function as simple as possible, i.e. sync only.

I'll make some edits to the PR and resolve some of these threads...

bluesky/protocols.py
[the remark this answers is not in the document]
thomascobb: I think now we decided raw signals don't need to be Settable, we don't need `AsyncStatus` to have an `__await__` method anymore

bluesky/protocols.py
```diff
@@ -1,17 +1,54 @@
...
+
+class Reading(ReadingMandatory, total=False):
+    severity: Severity
+    message: str
+
+
+class DescriptorMandatory(TypedDict):
+    source: str
+    dtype: Literal["string", "number", "array", "boolean", "integer"]
+    shape: List[int]
```
thomascobb, line 36: What happens if the shape of an array changes mid-way through a scan?

bluesky/bundlers.py
```diff
@@ -196,18 +222,34 @@ async def read(self, msg, reading):
...
 
-    def _cache_config(self, obj):
+    async def _cache_describe(self, obj):
+        "Read the object's describe and cache it."
+        obj = check_supports(obj, Readable)
+        self._describe_cache[obj] = await maybe_await(obj.describe())
+
+    async def _cache_describe_config(self, obj):
+        "Read the object's describe_configuration and cache it."
+        if isinstance(obj, Configurable):
```
coretl, line 238: `read_configuration()` and `describe_configuration()` will only be called if they both exist, as `Configurable` has both methods. This will get worse if we add `configure()` to `Configurable`. We will need an Ability name for `validate()` too

bluesky/protocols.py
```diff
@@ -143,34 +216,34 @@ def complete(self) -> Status:
...
         ...
 
-    # Remaining same as Readable
-    def read_configuration(self) -> Configuration:
-        """Same as for a Readable device."""
-        ...
+@runtime_checkable
+class Stageable(Protocol):
+    @property
+    def parent(self) -> Optional[Any]:
```
coretl, line 238: Does `parent` belong to Stageable?

bluesky/protocols.py
```diff
@@ -192,18 +265,18 @@ def unstage(self) -> List[Any]:
 
 @runtime_checkable
 class Pausable(Protocol):
-    def pause(self) -> Union[NoReturn, None]:
+    def pause(self) -> SyncOrAsync[None]:
```
coretl, line 268: Easier for the RE to make these `SyncOrAsync` rather than `Status`. Is there a downside to it?

bluesky/run_engine.py
```diff
@@ -1818,9 +1820,10 @@ async def _read(self, msg):
 
             Msg('read', obj)
         """
-        obj = msg.obj
+        obj = check_supports(msg.obj, Readable)
         # actually _read_ the object
-        ret = obj.read(*msg.args, **msg.kwargs)
+        # TODO: deprecation warning for args and kwargs?
```
coretl, line 1825: We pass `*args, **kwargs` through in a few places to the Ability, should we?

bluesky/run_engine.py
```diff
@@ -2316,7 +2326,7 @@ async def _stage(self, msg):
         """
         _, obj, args, kwargs, _ = msg
         # If an object has no 'stage' method, assume there is nothing to do.
-        if not hasattr(obj, 'stage'):
+        if not isinstance(obj, Stageable):
```
coretl, line 2329: `Stageable` means "has `stage` *and* `unstage`". Is this what we want?

bluesky/run_engine.py
```diff
@@ -2348,7 +2358,8 @@ async def _stop(self, msg):
 
             Msg('stop', obj)
         """
-        return msg.obj.stop()  # nominally, this returns None
+        obj = check_supports(msg.obj, Stoppable)
+        return await maybe_await(obj.stop())  # nominally, this returns None
```
coretl, line 2362: Easier to `await` then do the dance with an `Event` and a `Status` here, as with `pause` and `resume`

bluesky/run_engine.py
ksunden: See #1507

bluesky/bundlers.py
```diff
@@ -271,9 +317,22 @@ async def monitor(self, msg):
         seq_num_counter = count(1)
 
         def emit_event(*args, **kwargs):
-            # Ignore the inputs. Use this call as a signal to call read on the
-            # object, a crude way to be sure we get all the info we need.
-            data, timestamps = _rearrange_into_parallel_dicts(obj.read())
+            def is_reading_dict(d) -> bool:
```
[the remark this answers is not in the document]
coretl: Do you think the check is sufficient to check it's what bluesky wants? I tried reading the ophyd code to work out what it currently passes to the callback, but got lost in a sea of `*args` and `**kwargs`...

bluesky/run_engine.py
```diff
@@ -2213,8 +2222,9 @@ async def _resume(self, msg):
             await current_run.restore_monitors()
         # Notify Devices of the resume in case they want to clean up.
         for obj in self._objs_seen:
-            if hasattr(obj, 'resume'):
-                obj.resume()
+            if isinstance(obj, Pausable):
+                # TODO: state lock this?
```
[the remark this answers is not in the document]
coretl: @tacaswell @danielballan do we need any locking around these extra awaits for `pause`, `resume` and `stop`?

tacaswell:
A suggested name for this file: `duckery.py`  🦆 🦆

coretl:
> A suggested name for this file: `duckery.py`

I don't think `typing.Protocol` was the best choice of name, I would have gone with `typing.Trait` or `typing.Interface`, although maybe they should have tried `typing.Duck`? With that in mind we could go with:

1. `protocols.py`
2. `duckery.py`
3. `traits.py`
4. `interfaces.py`

@danielballan, @tacaswell, @untzag, @ksunden what do you reckon?

bluesky/bundlers.py
```diff
@@ -660,8 +691,39 @@ async def collect(self, msg):
...
+        for name, doc in maybe_collect_asset_docs(collect_obj):
+            # Add a 'run_start' field to the resource document on its way out.
+            if name == "resource":
+                doc["run_start"] = self._run_start_uid
+            await self.emit(DocumentNames(name), doc)
+
+        # Populate descriptors and local descriptors from describe_collect
+        # if not already called
+        if collect_obj not in self._local_descriptors:
+            await self._cache_describe_collect(collect_obj)
```
coretl, line 725: This means `describe_collect` will only be called once per run for `Flyable`, like `describe` is for `Readable`. Is this correct?

bluesky/bundlers.py
```diff
@@ -679,13 +742,7 @@ async def collect(self, msg):
             objs_read = frozenset(ev["data"])
             stream_name, descriptor_uid = local_descriptors[objs_read]
             seq_num = next(self._sequence_counters[stream_name])
-
             event_uid = new_uid()
-
-            reading = ev["data"]
-            for key in ev["data"]:
-                reading[key] = reading[key]
```
coretl, line 687: This seems like a relic from when numpy arrays were turned into lists:
https://github.com/bluesky/bluesky/blame/9ed5d21acc7509bd61a663ea06c6a0ede58a8312/bluesky/run_engine.py#L1687

I deleted it

bluesky/protocols.py
```diff
@@ -244,18 +325,32 @@ def check_value(self, value: Any) -> Union[NoReturn, None]:
...
 
+class Hints(TypedDict, total=False):
+    fields: List[str]
+    dimensions: List[Tuple[List[str], Literal["primary"]]]
+    gridding: Literal["rectilinear", "rectilinear_nonsequential"]
+
+
 @runtime_checkable
-class Hinted(Protocol):
+class HasHints(HasName, Protocol):
```
coretl, line 341: Following @danielballan's suggestion of `Has*`, we now have `HasName`, `HasParent` and `HasHints`

coretl:
Remaining `hasattr`s and `getattr`s:
- `read_attrs`
- `rewindable`
- `RealPosition`
- `dotted_name`
- `component_names`
Are any of these important enough to go in the protocols?

untzag:
I have a fairly strong preference to follow the Python language standards where possible. Python chose the word "Protocol", I think we should be consistent.

coretl:
> I think we should escalate the concept of components to protocol in a future PR.

At the moment I don't have the `Component` class in `ophyd.v2`, whether I'll need to introduce it in the future is a different matter. Apart from the use case of creating Signals (which I think I've covered), where else are they used?

untzag:
I think nested hardware interfaces are common to many different control systems. Queueserver recently added support for dot syntax to address components. As more control systems become supported it would be useful to standardize this syntax because right now it's just a convention based on how ophyd did it (as far as I know).

bluesky/bundlers.py
coretl: Resolved as we ditched `configure()` in favour of `set()` on the relevant `Signal`s, and it is reasonable to expect both `read_configuration()` and `describe_configuration()` to be implemented

bluesky/bundlers.py
coretl: Note that it used to be called every time we called `collect()`

coretl: [dallan (NSLS-II)](https://app.slack.com/team/U754JQZAM)  [1:33 PM](https://nikea.slack.com/archives/D01A8S4QTU0/p1648557235141449)
No, that should be called once (like describe()). I suspect that is a hold-over from when collect() was only ever called once, before support for partial, incremental collect was added.

bluesky/utils/__init__.py
```diff
@@ -1681,6 +1683,7 @@ def _rearrange_into_parallel_dicts(readings):
     return data, timestamps
 
 
+# TODO: this doesn't match the Movable protocol, should it?
```
coretl, line 1686: Do we need to make this match, or keep it?

coretl:
- Add `get_setpoint` and `get_readback` and call it `class SomethingMovable(Movable)`

bluesky/protocols.py
coretl: defer this

bluesky/simulators.py
```diff
@@ -87,10 +88,11 @@ def check_limits(plan):
     """
     ignore = []
     for msg in plan:
-        if msg.command == 'set' and msg.obj not in ignore:
-            if hasattr(msg.obj, "check_value"):
-                msg.obj.check_value(msg.args[0])
+        obj = msg.obj
+        if msg.command == 'set' and obj not in ignore:
+            if isinstance(obj, Checkable):
+                obj.check_value(msg.args[0])
```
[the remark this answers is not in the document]
coretl: Keep the top level API sync, and call into run-engine event loop if needed in some way

coretl:
I think we should try and close this out sooner rather than later, so have raised https://github.com/bluesky/bluesky/issues/1517 to cover issues that don't have an immediately obvious solution, or need more discussion. They should all be additions rather than changes to this PR, so this is probably a good section to merge on its own.

tacaswell:
rewindable (which is currently implemented as a `Signal` (😱 )) should be an attribute an anything triggerable or flyable.

It is a mechanism that a device can use to inform trigger and read that going back and naively re-triggering the device when recovering from an interruption (either due to manual intervention by the user or from suspenders).

The work also needs to be done for `bluesky.plans.fly` to actually use this value.

I think that changing this to be a normal attribute is on the table.  Trying to re-construct what we were thinking I suspect we made it a signal so that we could put the value in `read_configuration` but I am not sure if I would still agree with that position.

bluesky/protocols.py
```diff
@@ -244,18 +326,32 @@ def check_value(self, value: Any) -> Union[NoReturn, None]:
...
         a suggestion to enable automated tools to provide helpful information
         with minimal guidance from the user. See :ref:`hints`.
         """
         ...
+
+
+def check_supports(obj, protocol: Type[T]) -> T:
+    assert isinstance(obj, protocol), \
+        "%s does not implement all %s methods" % (obj, protocol.__name__)
+    return obj
```
untzag, line 357: Why is this here? Is it because we don't like the way `isinstance` reads? Or is it for a prettier exception? This just jumps out at me as "why"?

bluesky/utils/__init__.py
```diff
@@ -1739,3 +1746,28 @@ def __subclasshook__(cls, C):
...
+    # TODO: deprecation warning about *args and **kwargs?
+    if isinstance(obj, WritesExternalAssets):
+        yield from obj.collect_asset_docs(*args, **kwargs)
+
+
+async def maybe_await(ret: SyncOrAsync[T]) -> T:
+    if inspect.isawaitable(ret):
+        return await ret
+    else:
+        return ret
```
untzag, line 1773: love this little pattern :smile:

bluesky/protocols.py
coretl: Good question. I've now added a docstring that hopefully explains it

    This exists so that multiple protocol checks can be run in a mypy 
    compatible way, e.g.::

        triggerable = check_supports(obj, Triggerable)
        triggerable.trigger()
        readable = check_supports(obj, Readable)
        readable.read()

The problem is that mypy won't allow multiple type narrowing with isinstance in a single function, so you couldn't do:

```python
assert isinstance(obj, Triggerable)
obj.trigger()
assert isinstance(obj, Readable)
obj.read()
```

The call to read() will fail type checking

untzag: Wow, I never knew that.

bluesky/bundlers.py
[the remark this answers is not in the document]
coretl: Make emit_event take a single positional arg called `readings`, assert no kwargs if it is used

bluesky/run_engine.py
coretl: warn about any usage of *args and **kwargs with message about opening an issue

bluesky/run_engine.py
```diff
@@ -2069,16 +2076,18 @@ async def _set(self, msg):
...
         where arguments are passed through to `obj.set(*args, **kwargs)`.
         """
+        obj = check_supports(msg.obj, Movable)
         kwargs = dict(msg.kwargs)
         group = kwargs.pop('group', None)
-        self._movable_objs_touched.add(msg.obj)
-        ret = msg.obj.set(*msg.args, **kwargs)
+        self._movable_objs_touched.add(obj)
+        # TODO: deprecate args and kwargs?
+        ret = obj.set(*msg.args, **kwargs)
```
coretl, line 2084: Should we change the signature of protocols.Set to take *args, **kwargs?

bluesky/run_engine.py
```diff
@@ -353,7 +354,8 @@ def __init__(self, md=None, *, loop=None, preprocessors=None,
                  scan_id_source=default_scan_id_source,
                  during_task=None, call_returns_result=False):
         if loop is None:
-            loop = get_bluesky_event_loop()
+            loop = asyncio.new_event_loop()
+        set_bluesky_event_loop(loop)
```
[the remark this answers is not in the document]
coretl: https://github.com/bluesky/bluesky/pull/1502/commits/ffcc7a7c326b721a0c77b2288f5bcfc428d1566c

bluesky/run_engine.py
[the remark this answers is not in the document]
coretl: This is clarified in https://github.com/bluesky/bluesky/pull/1518

bluesky/run_engine.py
coretl: https://github.com/bluesky/bluesky/pull/1502/commits/c78ba233cb2c1101f2f97d1dcd64716db8a29742

bluesky/protocols.py
coretl: It shouldn't do. https://github.com/bluesky/event-model/pull/222 defines "bulk" events where the event is fixed size, but the number emitted is not, and there is a single pointer to a slice of an external file.

bluesky/bundlers.py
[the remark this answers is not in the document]
coretl: https://github.com/bluesky/bluesky/pull/1502/commits/ddc10c61cdf18c94f8d14dc978e642924ac04f74

bluesky/run_engine.py
coretl: Decided to add timeout, and support *args and **kwargs

coretl:
I have squashed this and rebased on master in #1522

PR #1505, opened 2022-02-04
WIP: Experiment with dropping too-old events

tacaswell:
Hot fix for SST

### DIFF
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -270,7 +270,10 @@ class RunBundler:
                                 'data_keys': data_keys.keys()})
         seq_num_counter = count(1)
 
-        def emit_event(*args, **kwargs):
+        def emit_event(*args, timestamp, **kwargs):
+            if ttime.time() - timestamp > .1:
+                print("STALE CALLBACK")
+                return
             # Ignore the inputs. Use this call as a signal to call read on the
             # object, a crude way to be sure we get all the info we need.
             data, timestamps = _rearrange_into_parallel_dicts(obj.read())


tacaswell:
Apparently this discovered our tests are not a faithful reproduction of EpicsSginals, but other than this is hot-deployed at SST and solves the problem well enough for now (they actually are using a slightly modified version of this that removes the print and only triggers the timout logic on a object with a particular name.

tacaswell:
On a bit more consideration and typing at @danielballan in slack I think this PR should be adapted to

 - [ ] make the 'too old" time be configurable via kwargs in the `Msg` so that this can be selectively applied (some monitors are fast and likely to be problems, but dropping them is also OK which may be used in tandem with a monitor than is slow and catching everyone is critical)
 - [ ] add the ability to thread through a 1 in N logic (either only emitting every Nth or maybe probabilistically keeping emitting 90% of the time) . 
 - [ ] sort out what happens when both are used (no callback older than drop dead time should go out, but then the question is the 1 in N of all events or of those young enough to be considered)
 - [ ] decide if `timestamp` will be a required kwarg or not

I do not forsee having time to push on this in the next few weeks, I would be very happy if someone picked this up an ran with it.

tacaswell:
On a bit more consideration, the knobs should be:

 - max age to be emitted (so once something is too stale, always drop on floor)
 - minimum gap between monitors (if we just emitted something, drop all new updates on the floor for some time)

prjemian:
This PR is missing a clear statement of the original problem.  Then, it might become apparent why dropping of any event is deemed necessary.  Is this a fix for a general problem or specifically at one NSLS-II beam line (SST)?

tacaswell:
The issue here is if you:

- use monitoring to get an event stream based on camonitor
- the ca events come in at a higher rate than we can emit the event model Events

Then the queue in ophyd gets backed up (because we have no meaningful back pressure, because we grab from the ca stack promptly we are defeating the CA back pressure). Because that queue ensues that they are processed in the order they are received, this means that processing things like "DMOV" also get delayed which causes other issues.

We conceived the monitoring in bluesky for "slow" things, but it gets used for "fast" things.  I don't see a way to predict if something is going to be fast or slow apriopi so we can not protectively reject even setting up the bad cases so I think the only option we have is to implement something like this to ensure reliable operation.

stan-dot:
closed in favour of https://github.com/bluesky/bluesky/issues/1783

PR #1507, opened 2022-02-28
ENH: don't call stage unless stageable

untzag:
Some of the device objects we are passing into bluesky happen to have an attribute `stage` that has nothing to do with Bluesky's concept of `Stageable`. Since we already have protocols, we can use them to make the run engine smart enough to not attempt calling `stage` on our objects.

There are other places in the run engine where protocols could be used, but I'm going to ignore those in favor of solving the problem in front of us now.

As always, feel free to reject or propose alternative solutions. If the answer is "please don't ever pass anything with attribute `stage` this is expected behavior" we can work with that.

## Description

Replace `hasattr(obj, 'stage')` with `isinstance(obj, bs_protocols.Stageable)`.

## Motivation and Context

See https://gitlab.com/yaq/yaqd-attune/-/issues/6

## How Has This Been Tested?

Tests written.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -14,6 +14,7 @@ import weakref
 from dataclasses import dataclass
 import typing
 from .bundlers import RunBundler
+from . import protocols as bs_protocols
 
 import concurrent
 
@@ -2316,7 +2317,7 @@ class RunEngine:
         """
         _, obj, args, kwargs, _ = msg
         # If an object has no 'stage' method, assume there is nothing to do.
-        if not hasattr(obj, 'stage'):
+        if not isinstance(obj, bs_protocols.Stageable):
             return []
         result = obj.stage()
         self._staged.add(obj)  # add first in case of failure below
@@ -2332,7 +2333,7 @@ class RunEngine:
         """
         _, obj, args, kwargs, _ = msg
         # If an object has no 'unstage' method, assume there is nothing to do.
-        if not hasattr(obj, 'unstage'):
+        if not isinstance(obj, bs_protocols.Stageable):
             return []
         result = obj.unstage()
         # use `discard()` to ignore objects that are not in the staged set.
--- a/bluesky/tests/test_run_engine.py
+++ b/bluesky/tests/test_run_engine.py
@@ -339,6 +339,26 @@ def test_stage_and_unstage_are_optional_methods(RE):
     RE([Msg('stage', dummy), Msg('unstage', dummy)])
 
 
+def test_need_both_stage_and_unstage_for_stagable(RE):
+    class StageOnly(object):
+
+        def stage(self):
+            raise Exception
+
+    class UnstageOnly(object):
+
+        def unstage(self):
+            raise Exception
+
+    stage_only = StageOnly()
+    unstage_only = UnstageOnly()
+    RE([Msg("stage", stage_only),
+        Msg("stage", unstage_only),
+        Msg("unstage", stage_only),
+        Msg("unstage", unstage_only)]
+        )
+
+
 def test_pause_resume_devices(RE):
     paused = {}
     resumed = {}
--- a/bluesky/tests/test_run_engine.py
+++ b/bluesky/tests/test_run_engine.py
@@ -356,7 +356,7 @@ def test_need_both_stage_and_unstage_for_stagable(RE):
         Msg("stage", unstage_only),
         Msg("unstage", stage_only),
         Msg("unstage", unstage_only)]
-        )
+       )
 
 
 def test_pause_resume_devices(RE):


tacaswell:
I'm slightly against this (I like duck typing 🦆 ) but not enough to try and block it (but I also probably will not merge, I'll leave that to someone else).

This needs some loud / through documentation as this can be a breaking change (for things that _do_ have a stage method but _do not_ register with the protocol).  It might be worth going through a release were we warn loudly that either will in the future not call this stage _or_ have found a `stage` but are not going to call it.  This would be an absolutely terrible change to debug if you were not expecting it (things silently _not_ happening is hard to find in complex code!).

untzag:
Yes, I see that this could be a breaking change for someone. The protocol for Stageable is pretty minimal, I suspect (but am not sure) that those who are breaking it should already be seeing errors logged in the run engine.

untzag:
I'm not sure if we're talking past each other or not, so I just want to make it super clear how the `isinstance` check works. The following code passes:

```python
from bluesky.protocols import Stageable

class Test(object):
    def stage(self):
        return []
    def unstage(self):
        return []

t = Test()
assert isinstance(t, Stageable)
```

This code doesn't require any registration step, and I wouldn't call it strict. Protocols are for duck typing [1]. The only change here is that the check is slightly more strict because it actually requires that BOTH methods are defined by the object.

[1] https://www.python.org/dev/peps/pep-0544/

prjemian:
@untzag:  I appreciate being educated.  Thanks!

tacaswell:
> The following code passes:

🐑 whelp, OK then.  I retract my complaints (and need to go read up on protocols!).

ZLLentz:
So, in summary, these protocols allow us to implement duck-typing with more granularity than a `hasattr` check, e.g. "has the attr and the attr is a callable that can be called with no args"

I think this is a good idea and is worth merging if you need to use the `stage` attribute for something else (though, I'd caution against colliding with the `bluesky` verbs whenever possible).

Side note: I really hope there isn't anyone out there who has intentionally defined a `stage` with no `unstage` who would be made sad by the apparent API break in this change, though I doubt this was ever intended behavior.

ksunden:
Protocols allow _static_ type checkers (e.g. mypy) to do rich things about ensuring return types and signatures.

However, at runtime, it is simply `hasattr` checks.

The reason we went with protocols over traditional ABCs is that protocols have no registration step.

In this instance, yes, it requires _both_ `stage` and `unstage` to be defined (and when used they will be called), but either (or both) can be No-Ops. One can check whether their particular device would pass such using mypy on something like:

```python
from bluesky.protocols import Stageable
x: Stagable = MyStageableDevice()
```

More info about protocols in [PEP 544](https://www.python.org/dev/peps/pep-0544/#runtime-checkable-decorator-and-narrowing-types-by-isinstance) (section on runtime_checkable linked directly)

untzag:
Yeah, it's worth noting that the `isinstance` check is actually weaker than mypy. So the following also passes:

```python
from bluesky.protocols import Stageable

class Test(object):
    def stage(self):
        pass
    def unstage(self):
        pass

t = Test()
assert isinstance(t, Stageable)
```

This cursed thing also passes:
```python
from bluesky.protocols import Stageable

def f():
    pass

class Test(object):
    stage = f
    unstage = f

t = Test()
assert isinstance(t, Stageable)
```

This does NOT pass, however:
``` Python
from bluesky.protocols import Stageable

class Test(object):
    stage = None
    unstage = None

t = Test()
assert isinstance(t, Stageable)  # assertion error
```

ksunden:
err, I'm actually surprised by that last one, as this one passes the isinstance check, but does not provide callables:

```python
from bluesky.protocols import Stageable

class Test(object):
    stage = "hi"
    unstage = "hi"

t = Test()
assert isinstance(t, Stageable)
```

Looking into the source of the `Stageable.__instancecheck__`, there is special handling for things expected to be methods set as None:

```python
    def __instancecheck__(cls, instance):
        # We need this method for situations where attributes are
        # assigned in __init__.
        if (
            getattr(cls, '_is_protocol', False) and
            not getattr(cls, '_is_runtime_protocol', False) and
            not _allow_reckless_class_checks(depth=2)
        ):
            raise TypeError("Instance and class checks can only be used with"
                            " @runtime_checkable protocols")

        if ((not getattr(cls, '_is_protocol', False) or
                _is_callable_members_only(cls)) and
                issubclass(instance.__class__, cls)):
            return True
        if cls._is_protocol:
            if all(hasattr(instance, attr) and
                    # All *methods* can be blocked by setting them to None.
                    (not callable(getattr(cls, attr, None)) or
                     getattr(instance, attr) is not None)
                    for attr in _get_protocol_attrs(cls)):
                return True
        return super().__instancecheck__(instance)
```

Note comment in that last block there

untzag:
Thanks @ksunden, I learned something that's not a detail I had recognized. Also I think we're way in the weeds :seedling: at this point.

danielballan:
I agree that this is a nice improvement. Thanks @untzag.

PR #1512, opened 2022-04-06
Future support

tacaswell:
This is the start of work to support future versions of Python + QT + databroker

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
         ophyd-version: ["master", "latest"]
       fail-fast: false
     steps:
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
         ophyd-version: ["master", "latest"]
       fail-fast: false
     steps:
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -43,6 +43,14 @@ jobs:
         python -m pip install git+https://github.com/bluesky/ophyd.git@master
         python -m pip list
 
+    - name: Install pyqt6 on some versions
+      if: ${{ matrix.python-version == '3.10' }}
+      shell: bash -l {0}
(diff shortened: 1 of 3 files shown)


mrakitin (commented):
Looks good to me. Just a quick question: do we install pyqt6 anywhere to test both versions?

tacaswell:
we can pip install pyqt6, do you think it is worth an extra CI job?

Testing both in one process is between a very bad idea (because it might crash) and impossible (because it will crash).

mrakitin:
I think we can have an extra job in CI for that case (for Python 3.7 only probably as https://pypi.org/project/PyQt6/6.3.0/#files has the wheels only for that version). Do we want to maintain a separate `requirements-dev.txt` with PyQt6? Right now we have PyQt5 listed there.

tacaswell:
I added pyqt6 to py310 and if pyqt6 is installed mpl 3.5 and up will pick it automatically.

bluesky/utils/__init__.py
```diff
@@ -1551,7 +1551,16 @@ def block(self, blocking_event):
...
             if 'qt' in backend:
 
-                from matplotlib.backends.qt_compat import QtCore, QtWidgets
+                from matplotlib.backends.qt_compat import QtCore, QtWidgets, QT_API
+                import functools
+
+                @functools.lru_cache(None)
+                def _enum(name):
+                    # foo.bar.Enum.Entry (PyQt6) <=> foo.bar.Entry (non-PyQt6).
+                    return operator.attrgetter(
```
klauer, line 1560: This is a really fancy but hard to understand function, even though I can figure it out.
Worthy of at least a docstring or a potential refactor.

tacaswell: I added a docstring.

.github/workflows/testing.yml
```diff
@@ -43,6 +43,14 @@ jobs:
         python -m pip install git+https://github.com/bluesky/ophyd.git@master
         python -m pip list
 
+    - name: Install pyqt6 on some versions
+      if: ${{ matrix.python-version == '3.10' }}
+      shell: bash -l {0}
+      run: |
+        set -vxeuo pipefail
+        python -m pip install pyqt6
```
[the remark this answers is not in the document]
tacaswell: https://peps.python.org/pep-0425/ is the reference for version tags.

The `abi3` tag indicates that it only uses the restricted stable CPython API.  I am not fully up on the details, but I would not be surprised that the interface is narrower than needed for other projects.  See https://docs.python.org/3/c-api/stable.html for details.

bluesky/utils/__init__.py
klauer: I'll say you did! 👏

mrakitin:
@klauer, are you OK with this update? Let's have it in if there are no objections.

mrakitin:
Thanks, Ken! I'll go ahead and will merge it then.

Thanks for the future support fix, @tacaswell!

Issue #1136, opened 2019-01-07
Watch memory usage during a 500,000-point scan.


danielballan:
User report:

> [At CMS we] started a continuous exposures (one after the other in the speed of 2-3s/image) for Pilatus2M. The exposure lasted about 8 days and generated about 500,000 images. After this long exposures, the system became very slow and showed a high resource usage of bluesky (76% memory). It remained slow even after killing the ipython process by command 'kill -9'. The workstation has to be restarted.

This could be a caproto issue, but I figure we'll start with bluesky + ophyd.sim and start making this more realistic until we can reproduce the leak.

Reported as ticket 2901 in NSLS-II Trac.

tacaswell:
I suspect that the slowness was stuff got pushed to swap and it would have recovered eventually without a restart.

Do we have anymore information about what the plan they were running was?  How much RAM did the machine have (75% of 2GB is much less worrying thas 75% of 256GB)?

tacaswell:
Do we know what callbacks or monitors they had running during this?

danielballan:
No, I posted all the information I have. I think it's likely this is a local configuration issue, but there's no harm in doing some basic checks.

RuipengLi:
It ran with 32G memory on xf11bm-ws1.

danielballan:
Here's one example of a place in the callbacks where we cache N items for an N-point scan. I don't think this itself explains 75% usage of 32G, but we might follow the pattern elsewhere. I'll try to avoid this as I write our "next generation" callbacks.

https://github.com/NSLS-II/bluesky/blob/7eea5f30d5e3a4c386585e23d17048b16cdac776/bluesky/callbacks/core.py#L343

mrakitin:
Should be resolved via https://github.com/bluesky/bluesky/pull/1513.

PR #1513, opened 2022-04-06
Fix prompt task cleanup

### LINKED ISSUE #1415: How to recover memory consumed during a run?
While testing proposed ophyd updates to ensure uninterrupted multi-run sessions, the tests were [interrupted](https://github.com/prjemian/ipython-aps/issues/3#issuecomment-719935364) and the ipython session was _killed_ when _Out of memory_ (as reported by the system `dmesg` command).  (Details in issue: https://github.com/prjemian/ipython-aps/issues/3#issuecomment-719935364)

The workstation involved has 16 GB RAM of which 6 GB were available to the ipython process.

## Expected Behavior
It is expected to make measurements that last for days, including possibly thousands of bluesky runs (run = start document emitted from RunEngine).  These measurements should not be interrupted by _Out of memory_ problems due to multiple runs of the same subplan.

## Current Behavior
In the next figure, a plot of memory consumed for each of the first 137 iterations of `bp.scan([noisy_det], motor, -2.1, 2.1, 23)` (using `noisy_det` and `motor` from `ophyd.sim`) is shown for a bare `RE` configuration (no callbacks or added metadata) as shown in the code: https://github.com/prjemian/ipython-aps/issues/3#issuecomment-720045276

Clipboard01

Observations:

1. [rss memory](https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_info) is consumed and not recovered when running a bluesky plan
1. Change in rss memory is comparable (statistically identical) between compared versions of bluesky & ophyd.
1. `bp.count` consumes less rss memory per run than `bp.scan`
1. Amount of memory consumed is not uniform, has wide variation

## Possible Solution
not a clue

How to free up memory consumed by previous runs *while* the RunEngine is operating?

Create a new, clean environment for testing:

    conda create -n bstest bluesky ophyd psutil -c conda-forge

Includes:

* bluesky            conda-forge/noarch::bluesky-1.6.6-py_0
* ophyd              conda-forge/noarch::ophyd-1.5.4-py_0
* psutil             conda-forge/linux-64::psutil-5.7.3-py37hb5d75c8_0

This is the test code `issue3_test.py`:

```python
#!/usr/bin/env python

"""
issue3_test.py

https://github.com/prjemian/ipython-aps/issues/3#issuecomment-720042520
"""

import os
import psutil
import sys
import time

from bluesky import RunEngine
from bluesky import plans as bp
from ophyd.sim import motor, noisy_det

RE = RunEngine({})
pid = os.getpid()
process = psutil.Process(pid)


def rss_mem():
    """return memory used by this process"""
    return process.memory_info()


def example(iters=1):
    mem_start = rss_mem().rss
    for _i in range(iters):
        t0 = time.time()
        mem0 = rss_mem().rss
        # yield from bp.count([noisy_det])
        yield from bp.scan([noisy_det], motor, -2.1, 2.1, 23)
        mem = rss_mem().rss
        print(
            f"Finished #{_i+1} of {iters} iterations"
            f", dt={time.time() - t0:.3f} s"
            f", bytes={mem}, bytes_changed={mem - mem0}"
            )
    mem_end = rss_mem().rss
    print(f"changed: {mem_end - mem_start}")


uids = RE(example(25))
```

tacaswell:
This address the issue raised on slack https://nikea.slack.com/archives/C7QC41C1G/p1649160481932079 and closes #1415.

I ended up ripping out this entire cache as a) not useful b) tricky to get right

Below is the commit message:

-----

The way the bridging between ophyd objects <-> RE works for anything with a
Status object:

 - ophyd returns a status object
 - RE subscribes to the done callback on the status object
 - that callback closes over an asyncio.Event
 - RE also holds onto the asyncio.Event for use in _wait
 - because the done callback runs on a different thread (the epics thread not
   the RE thread) we use call_soon_threadsafe which schedules a sync functions
   to be called by the EventLoop as soon as possible
 - call_soon_threadsafe returns an asyncio.Handle object which _only_ lets
   you cancel or ask if the callback was canceled (these are what we
   were caching and calling "tasks").  You can not even ask if it is done or
   started, it can only cancel.

It is not clear to me if calling this a task is due to a change in CPython (we
started when asyncio was a bit provisional) or due to us not understanding
things when we wrote this code (but I would bet on door number 2).  The idea
was that if we needed to cancel things we would want to cancel these tasks,
however:

 - the Handle is created by a callback only when that status object is done
 - from inside the callback there is no way to get back to the Handle we
   are holding
 - There should be no harm in letting these run

I tried changing `status_tasks` to be a dict mapping p_event ->
Handle, however that opened up a race condition where the
callback gets run before Handle can be added to the dictionary
which lead to race-conditions.

Caching the Handles without clearing them is harmless for short plans, however
every `set`, `trigger`, and `kickoff` adds a Handle to the cache.  For
long (days to months) long plans this will eventually become a problem.

In conclusion:

 - These Handles are very limited and transient
 - Capturing them robustely is very hard
 - The callbacks should be very fast so the Handle is only useful very
   transiently
 - I do not think there is harm is always letting these callbacks run.  The
   only possible failure path I see is that it is possible we may clobber
   an exception with a failure.  If that is the case, we should work that
   out mare generally as that was already a risk but with an added side of a
  race condition.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -445,7 +445,6 @@ class RunEngine:
         self._reason = ''  # reason for abort
         self._task = None  # asyncio.Task associated with call to self._run
         self._task_fut = None  # future proxy to the task above
-        self._status_tasks = deque()  # from self._status_object_completed
         self._pardon_failures = None  # will hold an asyncio.Event
         self._plan = None  # the plan instance from __call__
         self._command_registry = {
@@ -654,7 +653,6 @@ class RunEngine:
         self._reason = ''
         self._task = None
         self._task_fut = None
-        self._status_tasks.clear()
         self._pardon_failures = asyncio.Event(**self._loop_for_kwargs)
         self._plan = None
         self._interrupted = False
@@ -1257,8 +1255,7 @@ class RunEngine:
                 self._exception = RequestAbort()
         else:
             self._task.cancel()
-        for task in self._status_tasks:
-            task.cancel()
+
         if self._call_returns_result:
             plan_return = self.NO_PLAN_RETURN
             run_engine_result = self._create_result(plan_return)
@@ -1416,8 +1413,7 @@ class RunEngine:
                     if not self.resumable:
                         self._run_permit.set()
                         stashed_exception = FailedPause()
-                        for task in self._status_tasks:
-                            task.cancel()
+
                         self._state = 'aborting'
                         continue
                 # currently only using 'suspending' to get us into the
@@ -1976,10 +1972,9 @@ class RunEngine:
                 obj,
                 ret.success,
             )
-            task = self._loop.call_soon_threadsafe(
+            self._loop.call_soon_threadsafe(
                 self._status_object_completed, ret, p_event, pardon_failures
             )
-            self._status_tasks.append(task)
 
         try:
             ret.add_callback(done_callback)
@@ -2030,10 +2025,9 @@ class RunEngine:
                 obj,
                 ret.success,
             )
-            task = self._loop.call_soon_threadsafe(
+            self._loop.call_soon_threadsafe(
                 self._status_object_completed, ret, p_event, pardon_failures
             )
-            self._status_tasks.append(task)
 
         try:
             ret.add_callback(done_callback)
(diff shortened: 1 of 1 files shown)


untzag:
ping @ddkohler

ksunden (approved):

ksunden:
This resolves a problem we have been seeing in the wright group where long acquisitions have a high memory burden causing the entire system to slow down and eventually break.

untzag:
closes #1136

tacaswell:
I rebased to get rid of the merge commit (mostly aesthetic).  The failures with the dev version of ophyd are happening on the default branch as well so we can ignore them here.

@untzag or @ksunden should merge this when CI passes on the "latest" ophyd runs.

untzag:
this feels like a big thing I'm excited to have this released :tada:

PR #1518, opened 2022-05-13
MNT: clarify how suspenders get evaluated

tacaswell:
## Description

While looking at the #1501 and friends we had questions about if we needed to protect possible I/O with locks.  This refactors that code so that we do not need a lock.

## Motivation and Context

Suspenders work by injecting (a bunch of ) plans onto the plan stack (why not just one you ask....that is a very good question...~for a later date~ the last commit in this PR).

Previously the sync function (called from the Suspenders classes from a background thread (in the case of ophyd the ophyd trampoline thread)) would create a task that would then muck about to add the plans to the plan stack, do any rewinding, set up the wait etc.  If we were to put any awaits in this coroutine than there is a chance that the `_run` task would get time again and we would end up in a possibly inconsistent state!

This refactor moves all of the work to modify the plan stack into a coroutine which is added to the plan stack in the prior co-routine and then does it the set up.  It will be safe to run any I/O related async calls in this new co-routine as it is being awaited in the `_run` loop!

## How Has This Been Tested?

Because there is an extra message some of the tests did have to be modified.  The test suite caught many bugs in the early drafts of this PR.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -1152,7 +1152,9 @@ class RunEngine:
                 print(f"Justification for this suspension:\n{justification}")
 
             # add starting the suspender logic to the stack
-            self._plan_stack.append(single_gen(Msg('_start_suspender', None, pre_plan, post_plan, justification, fut)))
+            self._plan_stack.append(
+                single_gen(Msg('_start_suspender', None, pre_plan, post_plan, justification, fut))
+            )
             self._response_stack.append(None)
 
             # The event loop is still running. The pre_plan will be processed,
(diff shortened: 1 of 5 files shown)


bluesky/run_engine.py
```diff
@@ -464,7 +464,8 @@ def setup_run_permit():
             'clear_checkpoint': self._clear_checkpoint,
             'rewindable': self._rewindable,
             'pause': self._pause,
-            'resume': self._resume,
+            '_resume_from_suspender': self._resume,
```
coretl, line 467: I understand `self._resume` is triggered via a message from the suspender, but `self.resume` is called from an iPython terminal by the user. They seem to do very different things however, is there a case for making `self.resume` push a plan onto the stack in the same way as this PR does?

Would avoid this nastiness: https://github.com/coretl/bluesky/blob/2fce3f153073e73a916c92374d3452a62c22b426/bluesky/run_engine.py#L938-L939

tacaswell: yes, I think there is a case for that.

coretl:
This seems sensible to me, although I can't say I understand the finer details of the code. Will you merge this then I merge master into #1502?

PR #1520, opened 2022-05-17
ENH: use larger default length for string values

tacaswell:
We use a default precision of 3, but when we create the string format string
that turns in the length which is not great!

## Description

While working on something else I noticed than string values were truncated to length 3 which was...less than useful.  The issue it that we over-pattern-matched and used the same value for "precision" for "length" when formatting strings 🤦🏻 (I think I wrote this originally.....) 

## Motivation and Context

A three character wide string is not useful!

## How Has This Been Tested?

Needs tests!

------

If someone wants to take this over, please do!

### DIFF
--- a/bluesky/callbacks/core.py
+++ b/bluesky/callbacks/core.py
@@ -281,7 +281,7 @@ class LiveTable(CallbackBase):
     def __init__(self, fields, *, stream_name='primary',
                  print_header_interval=50,
                  min_width=12, default_prec=3, extra_pad=1, separator_lines=True,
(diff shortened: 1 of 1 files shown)


tacaswell:
There is something wrong with the headings.

stan-dot:
@tacaswell  is it about those headings? 
(bluesky.callbacks.core) line 371
headings = [k if k != self.ev_time_key else "time" for k in self._format_info]
self._header = "|" + self._main_fmnt.format(*headings) + "|"
        
should it be a separate issue?

Issue #1521, opened 2022-05-25
refactor RunEngine.resume to match suspenders code


tacaswell:
> I understand `self._resume` is triggered via a message from the suspender, but `self.resume` is called from an iPython terminal by the user. They seem to do very different things however, is there a case for making `self.resume` push a plan onto the stack in the same way as this PR does?
>
>Would avoid this nastiness: https://github.com/coretl/bluesky/blob/2fce3f153073e73a916c92374d3452a62c22b426/bluesky/run_engine.py#L938-L939

_Originally posted by @coretl in https://github.com/bluesky/bluesky/pull/1518#discussion_r872380120_

PR #1522, opened 2022-05-26
Add async capabilities to protocols

coretl:
To better document how Bluesky interfaces with hardware and make the asyncio compatible.
Supersedes #1502 

## Description
- Add async capabilities to protocols
- Make bundler and run engine use them
- Cache describe_* for both read and subscribe
- PendingDeprecationWarning for args and kwargs passed to collect_asset_docs, read, kickoff, complete or trigger
- Allow subscriptions to callback with Readings, rather than having to call read to get them
- Add helpers for calling in bluesky event loop
- Remove Movable ABC

## Motivation and Context
So I can write Ophyd.v2

## How Has This Been Tested?
Unit tests pass


bluesky/run_engine.py
```diff
@@ -933,8 +935,9 @@ def resume(self):
         self._response_stack.append(None)
         # Notify Devices of the resume in case they want to clean up.
         for obj in self._objs_seen:
-            if hasattr(obj, 'resume'):
-                obj.resume()
+            if isinstance(obj, Pausable):
+                fut = asyncio.run_coroutine_threadsafe(maybe_await(obj.resume()), self._loop)
```
coretl, line 939: I think this is safe, but it is messy. Maybe #1521 will solve it?

bluesky/run_engine.py
```diff
@@ -1180,12 +1183,12 @@ async def _start_suspender(self, msg):
             )
         # During suspend, all motors should be stopped. Call stop() on
         # every object we ever set().
-        self._stop_movable_objects(success=True)
+        await self._stop_movable_objects(success=True)
```
coretl, line 1186: These awaits are in a plan, so I think they are safe

bluesky/run_engine.py
coretl: @tacaswell this is the area of code that most needs reviewing

untzag (changes requested):
The signature for `Movable` breaks ophyd and yaqc-bluesky. The following class should pass mypy:

```python
from ophyd.status import Status
class Device:
    def set(self, value: float) -> Status:
         ...
```

If you just go back to the old signature it should work for anything you could imagine...

Note that I'm talking about static type checking, not runtime checking! We have some tests for this:

- https://github.com/bluesky/yaqc-bluesky/blob/master/bluesky_protocols.py
- https://github.com/bluesky/ophyd/blob/master/ophyd/tests/test_bluesky_protocols.py

In my opinion these tests should pass before merge.

ksunden:
I tend to agree with @untzag, though the `float` annotation does not capture the fullness of what can be passed there... the point is that to my understanding, timeout is not a required kwarg, it is an implementation detail and to type check actual implementations, all that is required is a single positional argument. 

(I personally tend towards devices where that _is_ a float, but various devices use either arrays, str, or dict for various reasons, and not sure we can narrow that type)

I will note that something specified as a single positional argument does type check against an implementation which allows variadic args and kwargs just fine, but that is an implementation detail of the particular library.

coretl:
I didn't know there were those tests, I've put the original definition back as adding anything to the type signature would require all implementers to add it to the implementation

coretl:
BTW the docs build error should go away once #1524 is merged

PR #1525, opened 2022-06-18
FIX: make trigger and read careful

tacaswell:
## Description

Previously we would get confusing errors when:

  - event had been opened
  - read or describe raise any errors once
  - a wrapper (such as the baseline pre-processor) tries to take a reading as
  part of plan clean up

The error would be that you could not open an event with an event open which
would effectively drop the original error (the one you actually need to debug
the issue!) on the floor.

This could also be fixed by improving the multi-exception handling in the RE,
but this is a simpler fix.

## Motivation and Context

These errors are among the most confusing.  I helped @maffettone debug one of these today and finally got my act together to write this patch.

The fix is to make sure that `trigger_and_read` never leaves an event "open" when it returns control to its caller.

## How Has This Been Tested?

Added a test.

### DIFF
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -887,6 +887,7 @@ def trigger_and_read(devices, name='primary'):
     msg : Msg
         messages to 'trigger', 'wait' and 'read'
     """
+    from .preprocessors import contingency_wrapper
     # If devices is empty, don't emit 'create'/'save' messages.
     if not devices:
         yield from null()
@@ -904,13 +905,29 @@ def trigger_and_read(devices, name='primary'):
         if not no_wait:
             yield from wait(group=grp)
         yield from create(name)
-        ret = {}  # collect and return readings to give plan access to them
-        for obj in devices:
-            reading = (yield from read(obj))
-            if reading is not None:
-                ret.update(reading)
-        yield from save()
+
+        def read_plan():
+            ret = {}  # collect and return readings to give plan access to them
+            for obj in devices:
+                reading = (yield from read(obj))
+                if reading is not None:
+                    ret.update(reading)
+            return ret
+
+        def happy_path():
+            yield from save()
+
+        def sad_path(exp):
+            yield from drop()
+            raise exp
+
+        ret = yield from contingency_wrapper(
+            read_plan(),
+            except_plan=sad_path,
+            else_plan=happy_path
+            )
         return ret
+
     from .preprocessors import rewindable_wrapper
     return (yield from rewindable_wrapper(inner_trigger_and_read(),
                                           rewindable))
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -914,17 +914,17 @@ def trigger_and_read(devices, name='primary'):
                     ret.update(reading)
             return ret
 
-        def happy_path():
+        def standard_path():
             yield from save()
 
-        def sad_path(exp):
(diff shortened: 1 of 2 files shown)


maffettone:
Overall this makes sense, and tests well assuming that the only source of failures would be in the `describe()` method. 

On the surface this will also work if the error is generally raised in the `readI()` method, whether in the sub call to describe or from somewhere else. 

Are there common errors in `trigger()` that we should also be concerned about?

bluesky/tests/test_plans.py
```diff
@@ -657,3 +657,19 @@ def test_grid_scans_failing(RE, hw, plan):
...
+
+def test_describe_failure(RE):
+    ophyd = pytest.importorskip("ophyd")
+
+    class Aardvark(Exception):
+        ...
+
+    class BadSignal(ophyd.Signal):
+        def describe(self):
+            raise Aardvark("Look, and aardvark!")
```
maffettone, line 670:
```suggestion
            raise Aardvark("Look, an aardvark!")
```

bluesky/tests/test_plans.py
```diff
@@ -657,3 +657,19 @@ def test_grid_scans_failing(RE, hw, plan):
...
+
+    bad_signal = BadSignal(value=5, name="Arty")
+
+    with pytest.raises(Aardvark, match="Look, and aardvark!"):
```
maffettone, line 674:
```suggestion
    with pytest.raises(Aardvark, match="Look, an aardvark!"):
```

bluesky/tests/test_plans.py
```diff
@@ -657,3 +657,19 @@ def test_grid_scans_failing(RE, hw, plan):
                     hw.motor1, 4, 5, 6,
                     hw.motor2, 7, 8, 9)
             RE(plan([hw.det], *args, snake_axes=snake_axes))
+
+
+def test_describe_failure(RE):
+    ophyd = pytest.importorskip("ophyd")
+
+    class Aardvark(Exception):
```
maffettone, line 665:
```suggestion
    class Aardvark(Exception):
    # Unique exception to test behavior of describe.
```

bluesky/plan_stubs.py
```diff
@@ -904,13 +905,29 @@ def inner_trigger_and_read():
...
+            yield from save()
+
+        def sad_path(exp):
+            yield from drop()
+            raise exp
+
+        ret = yield from contingency_wrapper(
+            read_plan(),
+            except_plan=sad_path,
+            else_plan=happy_path
```
maffettone, line 927:
```suggestion
            except_plan=exception_path,
            else_plan=standard_path
```

bluesky/plan_stubs.py
```diff
@@ -904,13 +905,29 @@ def inner_trigger_and_read():
...
+
+        def read_plan():
+            ret = {}  # collect and return readings to give plan access to them
+            for obj in devices:
+                reading = (yield from read(obj))
+                if reading is not None:
+                    ret.update(reading)
+            return ret
+
+        def happy_path():
```
maffettone, line 917:
```suggestion
        def standard_path():
```

maffettone (commented):
Small considerations on naming things and spelling. 

Otherwise, if we think these tests are sufficient (no need to test read or trigger raising exceptions), this looks good to me.

danielballan:
Oh, this is great! This has been masking useful errors from the start.

PR #1526, opened 2022-06-18
ENH: allow the contingency wrapper to eat exceptions

tacaswell:
## Description

Add an option to contingency_wrapper to not automatically re-raise if the except plan returns a value rather than raising its own exception.

We have to do this with a flag like this for back-compatibility reasons

## Motivation and Context

We (I) made a design mistake with `contingency_wrapper` because there is no way to actually suppress the exception.  The reason we wrote this wrapper (and finalize wrapper) is to keep users from having to understand the extra set of constraints around the cancel and generator exit exceptions so this omission defeats the point.

## How Has This Been Tested?

Needs tests!

### DIFF
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -533,7 +533,8 @@ def contingency_wrapper(plan, *,
                         except_plan=None,
                         else_plan=None,
                         final_plan=None,
-                        pause_for_debug=False):
+                        pause_for_debug=False,
+                        auto_raise=True):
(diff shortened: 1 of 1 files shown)


DominicOram:
This would be useful for us at DLS, can I pick it up and write tests etc?

tacaswell:
@DominicOram yes please.

DominicOram:
Continued in https://github.com/bluesky/bluesky/pull/1599

PR #1527, opened 2022-06-18
MNT: clear status tasks while running

tacaswell:
## Description

These were previously cleared at the start of a `__call__`.  If all `__call__`
runs are relatively short, this is OK because the dequeue never gets very big,
however with long runs (days -> months) this build up will eventually consume
an appreciably amount of RAM and become a problem.

## Motivation and Context

We had reports of out-of-memory errors.

## How Has This Been Tested?

Needs tests!


stan-dot:
`RunEngine._status_tasks` does not seem to be present any longer as a `deque`, making the need for changing it to `set` obsolete. 

action suggestion - to close the PR

@tacaswell

PR #1530, opened 2022-07-05
DOC: Document deployment recommendations.

danielballan:
I invite other facilities to comment and, even better, to add sections on their own experience.

### DIFF
--- a/docs/source/deployment-experience.md
+++ b/docs/source/deployment-experience.md
@@ -1,0 +1,135 @@
+# Deployment Experience
+
+Several facilities deploy Bluesky across many instruments (up to 25 or so).
+In this document we summarize lessons learned from that experience. At the
+moment, only NSLS-II (Bluesky's facility of origin) is represented, but we
+hope to add comments from other facilities in the future.
+
+## NSLS-II
+
+### Python environment management
+
+We deploy conda environments, verioned by operating cycle, in a central (NFS)
+software mount. The instrument staff and users have only read-only access to
+these environments. The environments contains the union of all requirements
+across all instruments: they include core scientific Python libraries (e.g.
+scipy), Bluesky libraries, and other more instrument-specific libraries, both
+internally and externally developed.
+
+The environments read-only in order to avoid accidentally breaking changes.
+To support short-term patching and minor variations betweeen instruments,
+we add an instrument-specific directory to the `PYTHONPATH` at each
+instrument respectively, granting instrument staff read/write permission
+on that directory. This enables them to install extra packages and upgrade
+or downgrade existing ones as needed, without the risk of unintended
+consequences for other instruments.
+
+The conda environments are build via a
+[CI service](https://github.com/nsls2-conda-envs),
+hosted on [Zenodo](https://zenodo.org/record/6555136), and
+deployed via conda-pack. Using conda-pack speeds up installation and
+improves reliability.
+
+We recommend:
+
+* Create centrally-managed read only conda enviornments.
+* Use `PYTHONPATH` to "overlay" custom additions or version changes.
+
+### Database Management
+
+NSLS-II and some other facilities we know of currently use a separate
+single-node MongoDB instance per instrument, but _we do not recommend this_. We
+are moving toward a central MongoDB cluster shared by all instrumenents. This
+reduces maintainence burden (e.g. less work to update MongoDB version) and
+improves reliability (avoid single-node deployments which are single points
+of failure).
+
+Each instrument will have its own MongoDB database on the cluster.
+
+We recommend:
+
+* Use a central cluster, shared by all instruments.
+* Use authentication between the MongoDB client and server.
+
+### Message Bus
+
+We use a message to "publish" Bluesky documents from the RunEngine so
+that they are available for consumers in other processes or on
+other machines on the network.
+
+To start, we made good use of 0MQ as a simple message bus. However, we
+are moving to use Kafka, a much more fully-featured and robust solution.
+The Bluesky document publishers and consumers and fairly agnostic to
+the transport technology used, so switching is easy.
+
+Finally, we current subscribe the database-insertion code (`db.insert`)
+_directly_ to the RunEngine. We have begun to move that step downstream
+of a message bus, i.e. switching from:
+
+```
+RunEngine --> MongoDB
+```
+
+to
+
+```
+RunEngine --> Kafka --> MongoDB
+```
+
+We find that this improves performance because Kafka can accept data buffer it
+faster than MongoDB can insert it. We have yet to commit to this architecture
+at scale.
+
+We recommend:
+* Start with 0MQ to play because it is easy to set up.
+* Move to Kafka early, once you are relying on these systems.
+
+### Managing facility-specific code
+
+Where possible, we encourage facilities to push embellishments to Bluesky
+libraires so the benefits and maintenace can be shared. But each facility also
+accumulates utilies, bluesky plans, ophyd Devices, and other code that is
+specific to them. Some of this work may eventually be pushed upstream but is
+deployed locally at first to meet local deadlines. Other work may not be
+useful at other facilties and therefore not appropriate to contribute upstream.
+
+At NSLS-II, we have a catch-all library [nslsii](https://github.com/NSLS-II/nslsii).
+APS and LCLS have similar packages.
+
+We recommend:
+* Make a facility-specific package.
+* Make it public.
+
+### Managing instrument configuration
+
+Each instrument needs Python code to:
+
+* Define instrument-specific ophyd Device classes
+* Instantiate ophyd Devices
+* Define instrument-specific plans
+* Subscribe instrument-specific convenience
+* Customize the user experience
+
+Setting up an environment for collection is too complex for a declarative configuration
+file (like a YAML file) and requires executing Python code.
+
+The initial target of Bluesky development was a command-line tool. We chose the
+IPython application, we used the "profiles" feature of IPython to run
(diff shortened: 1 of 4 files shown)


docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
...
+The conda environments are build via a
+[CI service](https://github.com/nsls2-conda-envs),
+hosted on [Zenodo](https://zenodo.org/record/6555136), and
+deployed via conda-pack. Using conda-pack speeds up installation and
+improves reliability.
+
+We recommend:
+
+* Create centrally-managed read only conda enviornments.
+* Use `PYTHONPATH` to "overlay" custom additions or version changes.
```
ZLLentz, line 36: It's comforting to see how many deployment similarities NSLS-II has with LCLS. This section here that I'm attaching this comment to is exactly what we do at LCLS- centrally managed conda + PYTHONPATH for custom edits as needed for an experiment when it can't wait for a formal release cycle.

ZLLentz: Our NFS has been historically sometimes slow, so in addition we've been using conda-pack to deploy copies of the central environments locally to SSDs on operator consoles.

ZLLentz (approved):
I agree with all the recommendations here- I'd be happy to add an LCLS section that goes over some of the minor differences in our deployment.

docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Database Management
+
+NSLS-II and some other facilities we know of currently use a separate
+single-node MongoDB instance per instrument, but _we do not recommend this_. We
+are moving toward a central MongoDB cluster shared by all instrumenents. This
```
awalter-bnl, line 42: 'instrumenents' should be 'instruments'

awalter-bnl (commented):
Just a few typos, nothing substantive to the texts. Hope this helps.

docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
...
+
+The initial target of Bluesky development was a command-line tool. We chose the
+IPython application, we used the "profiles" feature of IPython to run
+instrument-specific code and populate the user namespace at IPython start-up
+with variables representing ophyd Device, bluesky plans, and so on.
+We currently have at least one GitHub repository per instrument. These are public
+under separate GitHub organization. The repositories are all named `profile_collection`.
+
+Over time, these large IPython profiles have become unwieldy and difficult to
+maintain. We recommend moving at least class and plan definitions to a normal
```
cjtitus, line 124: This has resulted in conflicts with the centralized python environment, because instrument packages cannot be straightforwardly installed in development mode without clever hacks. More thought is needed about how this should work, from a beamline developer's perspective.

tacaswell: `python setup.py develop` or `pip install -ve .` does two things is the setuptools world:

1. run `python setup.py dist_info` so that the right meta-data is generated
2. adds a `.pth` file where ever you are installing to add the path to where the source is to the PYTHONPATH at start up

For security reasons (I think), `.pth` files are only used when they are in the "standard" search path not in a folder added by `PYTHONPATH`, those paths are all read only so the overlay trick does not work here.  However, you can still do these two things manually or run `pip install ....` everytime you make a change (which if you wrap it up in a helper script that then restarts bsui might not be that bad).

docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
...
+Several facilities deploy Bluesky across many instruments (up to 25 or so).
+In this document we summarize lessons learned from that experience. At the
+moment, only NSLS-II (Bluesky's facility of origin) is represented, but we
+hope to add comments from other facilities in the future.
+
+## NSLS-II
+
+### Python environment management
+
+We deploy conda environments, verioned by operating cycle, in a central (NFS)
```
prjemian, line 12: typo: `verioned` -> `versioned`

docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
...
+software mount. The instrument staff and users have only read-only access to
+these environments. The environments contains the union of all requirements
+across all instruments: they include core scientific Python libraries (e.g.
+scipy), Bluesky libraries, and other more instrument-specific libraries, both
+internally and externally developed.
+
+The environments read-only in order to avoid accidentally breaking changes.
```
prjemian, line 19: `environments read-only` -> `environments are read-only`

docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
+# Deployment Experience
+
+Several facilities deploy Bluesky across many instruments (up to 25 or so).
+In this document we summarize lessons learned from that experience. At the
+moment, only NSLS-II (Bluesky's facility of origin) is represented, but we
+hope to add comments from other facilities in the future.
+
+## NSLS-II
```
prjemian, line 8: APS is a little different.  We have a central NFS share for a common set of Bluesky packages, allowing any user to use the software right away.  (We have other Python installations provided this same way.)  The shared installation is updated in the APS downtime between operating cycles (3x/year).

For production use, we recommend a custom conda environment for each instrument so they have full control of the software suite.  The environment is saved in version control with the other configuration (devices, plans, utils, ...).  We recommend GitHub.  If there are privacy concerns, then APS has an internal GitLab server to be used.

Lately, `conda` is taking very long times to solve the package requirements.  We will start to use `micromamba` to create environments and see how that is received.

We tend to avoid relying on PYTHONPATH since many beam lines use Python in other ways.  PYTHONPATH has proven to generate conflicts.  Instead, we rely on activating an environment to get all the paths correct.  For custom user libraries, we show users how to use `sys.path.append()`.

docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
...
+improves reliability (avoid single-node deployments which are single points
+of failure).
+
+Each instrument will have its own MongoDB database on the cluster.
+
+We recommend:
+
+* Use a central cluster, shared by all instruments.
+* Use authentication between the MongoDB client and server.
+
```
prjemian, line 53: APS (currently) has switched to a set of MongoDB VMs, where several instruments will share a single VM.  The VMs are managed by the APS IT group.  The set of VMs is not clustered, since this would trigger an enormous cost increase.  We'll see how that goes.  Recommendations on how to cluster without large cost increase are welcome.

docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
...
+We use a message to "publish" Bluesky documents from the RunEngine so
+that they are available for consumers in other processes or on
+other machines on the network.
+
+To start, we made good use of 0MQ as a simple message bus. However, we
+are moving to use Kafka, a much more fully-featured and robust solution.
+The Bluesky document publishers and consumers and fairly agnostic to
+the transport technology used, so switching is easy.
+
+Finally, we current subscribe the database-insertion code (`db.insert`)
```
prjemian, line 65: `current` -> `currently`

docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
...
+RunEngine --> MongoDB
+```
+
+to
+
+```
+RunEngine --> Kafka --> MongoDB
+```
+
+We find that this improves performance because Kafka can accept data buffer it
```
prjemian, line 79: `accept data buffer it` -> `accept data and buffer it`

docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
...
+faster than MongoDB can insert it. We have yet to commit to this architecture
+at scale.
+
+We recommend:
+* Start with 0MQ to play because it is easy to set up.
+* Move to Kafka early, once you are relying on these systems.
```
prjemian, line 85: If starting with Kafka early, why bother with 0MQ?  What's the advantage?  What's the penalty to switch from 0MQ to Kafka later?

docs/source/deployment-experience.md
prjemian: Or, use a starter shell script to include the instrument package as the interactive session is started.  Or advise how to use `sys.path.append()` to identify additional packages to be found.

cjtitus: My point is that this recommendation to move to packages is made, but from my perspective the details haven't really been worked out in terms of actually making it easier to maintain and deploy in concert with the central conda env, which is also being recommended. I don't think I'd be able to convince someone who already had code in an IPython profile that they should switch to packaged code, as their life will get harder and more mysterious, and doubly so if they don't already understand the intricacies of packaging. The fact that we are discussing `python setup.py dist_info` and the locations where `.pth` files will and will not be read is a good indicator that more work is required to make this a turnkey solution for most people.

docs/source/deployment-experience.md
```diff
@@ -0,0 +1,135 @@
...
+organize, maintain, reuse, and test.
+
+### Deployment tools
+
+For deploying services (MongoDB, Kafka, Tiled) and configuration, we use
+Ansible and have embraced "configuration as code" as much as we can.
+
+We recommend:
+* Automate all deployment.
+* Reduce the variation between instruments as much as is technically feasible.
```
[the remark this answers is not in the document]
tacaswell: This is referring to the software environment not the hardware.  We used to do per-beamline environments but now we just install everyones extra packages (including beamline specific) into one environment that everyone gets.  We were worried about conflicting dependency requirements but in 7 years it has never happened and the simplicity of a unified environment is well worth the trouble if it ever comes up.

stan-dot:
@danielballan should this go into the [cookbook ](https://github.com/bluesky/bluesky-cookbook) instead?

PR #1533, opened 2022-07-19
Bad comparison created too many axes for plots

maffettone:
## Description

Change the logic to create enough axes but not too many in the grid. `>=` instead of `>`. 

## Motivation and Context

The logic here produces a 3x3 grid for 6 plots instead of a 2x3 grid. This is exceptionally problematic for these intervals, because there will be no x-axis labels with the shared axes, since the entire bottom row will have `visible = False`.
Adjusting the logic will create enough plots. 

## How Has This Been Tested?

I am guilty of not using the `tests/interactive/best_effor_cb.py` because of errors with old versions of databroker. 
I did run this sanity check to make sure the logic would always produce enough axes, and that removing a full row would create an invalid figure with not enough axes (therefore is necessary). 
```python
import numpy as np
for l in range(5, 20):
    nrows = ncols = int(np.ceil(np.sqrt(l)))
    while (nrows-1) * ncols >= l:
        nrows -= 1
    assert(nrows*ncols >=l)
    assert(nrows*ncols - ncols < l)
```

## Screenshots
Before
Before

After
After

### DIFF
--- a/bluesky/callbacks/best_effort.py
+++ b/bluesky/callbacks/best_effort.py
@@ -245,7 +245,7 @@ class BestEffortCallback(QtAwareCallback):
                     layout = (len(columns), 1)
                 else:
                     nrows = ncols = int(np.ceil(np.sqrt(len(columns))))
-                    while (nrows - 1) * ncols > len(columns):
+                    while (nrows - 1) * ncols >= len(columns):
                         nrows -= 1
                     layout = (nrows, ncols)
                 if ndims == 1:


tacaswell (approved):
modulo CI passing.

maffettone:
Looks like the GHA for 3.6/3.7 are grabbing ophyd from git, but the current ophyd has reqs of 3.8

tacaswell:
The issues is that ohpyd dropped support for 3.6 and 3.7, merging over the failures.

PR #1534, opened 2022-07-20
[BLD]: follow NEP29 version support schedule

maffettone:
Increasing the minimum version of bluesky to python 3.8 inline with NEP29.

## Description

Remove 3.6/3.7 from testing. 
Set min version to 3.8 in setup.py

## Motivation and Context

NEP29

## How Has This Been Tested?

Passes unit tests.


maffettone:
The only other place where outdated python pops up is in the Azure pipelines?  Are those no longer in use, and should they be removed?

tacaswell (approved):

tacaswell:
Please hold of merging this until we get one more release out.

PR #1536, opened 2022-08-08
Add Locatable

### LINKED ISSUE #1535: Need a new type of `Movable` that lets you see where it actually is
Name:
- `Locatable`
- ...
Methods:
- `get_setpoint() -> SyncOrAsync[Any]` and `get_position() -> SyncOrAsync[Any]`
- `get_setpoint() -> SyncOrAsync[Any]` and `get_readback() -> SyncOrAsync[Any]`
- ...

coretl:
## Description
Adds Locatable protocol

## Motivation and Context
Fixes #1535 

## How Has This Been Tested?
It hasn't, see comments

### DIFF
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -309,6 +309,9 @@ def rd(obj, *, default_value=0):
     This is a helper plan to get the scalar value out of a Device
     (such as an EpicsMotor or a single EpicsSignal).
 
+    For devices that implement the Locatable protocol, the location is canonical
+    and is returned without parsing the read keys.
+
     For devices that have more than one read key the following rules are used:
 
     - if exactly 1 field is hinted that value is used
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -13,7 +13,7 @@ try:
 except ImportError:
     from toolz import partition
 
-from .protocols import Triggerable
+from .protocols import Locatable, Triggerable
 from .utils import (
     get_hinted_fields,
     merge_cycler,
@@ -342,6 +342,15 @@ def rd(obj, *, default_value=0):
         The "single" value of the device
 
     """
+    # Location is canonical if it exists
+    if isinstance(obj, Locatable):
+        location = yield Msg("locate", obj)
+        if location is None:
+            # list-ify mode
+            return default_value
+        else:
+            return location["readback"]
+
     hints = get_hinted_fields(obj)
     if len(hints) > 1:
         msg = (
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -3,6 +3,8 @@ from __future__ import generator_stop
 from collections import OrderedDict, deque, ChainMap
 from collections.abc import Iterable
 import uuid
+
+from bluesky.protocols import Locatable
 from .utils import (get_hinted_fields, normalize_subs_input, root_ancestor,
                     separate_devices,
                     Msg, ensure_generator, single_gen,
@@ -974,6 +976,14 @@ def _normalize_devices(devices):
     return devices, coupled_parents
 
 
+def __get_result_of_message(msg_type: str, obj):
+    result = yield Msg(msg_type, obj)
+    if result is None:
+        # this plan may be being list-ified
+        print(f"*** all positions for {obj.name} are relative to current position ***")
+    return result
+
+
 def __read_and_stash_a_motor(obj, initial_positions, coupled_parents):
     """Internal plan for relative set and reset wrappers
 
@@ -983,37 +993,40 @@ def __read_and_stash_a_motor(obj, initial_positions, coupled_parents):
        Do not use this plan directly for any reason.
 
     """
-    # obj should have a `position` attribution
-    try:
-        cur_pos = obj.position
-    except AttributeError:
-        # ... but as a fallback we can read obj and grab the value of the
-        # first key
-        reading = yield Msg('read', obj)
+    # First check if it is a locatable, and use its location if it is
+    if isinstance(obj, Locatable):
+        location = yield from __get_result_of_message("locate", obj)
+        if location is None:
+            setpoint = 0
+        else:
+            setpoint = location["setpoint"]
+    # Otherwise it might have a `position` attribution
+    elif hasattr(obj, "position"):
+        setpoint = obj.position
+    # Otherwise fallback to read obj and grab the value of the first key
+    else:
+        reading = yield from __get_result_of_message("read", obj)
         if reading is None:
-            # this plan may be being list-ified
-            print("*** all positions for {m.name} are "
-                  "relative to current position ***".format(m=obj))
-            cur_pos = 0
+            setpoint = 0
         else:
             fields = get_hinted_fields(obj)
             if len(fields) == 1:
                 k, = fields
-                cur_pos = reading[k]['value']
+                setpoint = reading[k]['value']
             elif len(fields) == 0:
                 k = list(reading.keys())[0]
-                cur_pos = reading[k]['value']
+                setpoint = reading[k]['value']
             else:
                 raise Exception("do not yet know how to deal with "
                                 "non pseudopositioner multi-axis.  Please "
                                 "contact DAMA to justify why you need "
                                 "this.")
 
-    initial_positions[obj] = cur_pos
+    initial_positions[obj] = setpoint
 
     # if we move a pseudo positioner also stash it's children
     if obj in coupled_parents:
-        for c, p in zip(obj.pseudo_positioners, cur_pos):
+        for c, p in zip(obj.pseudo_positioners, setpoint):
             initial_positions[c] = p
 
     # if we move a pseudo single, also stash it's parent and siblings
--- a/bluesky/protocols.py
+++ b/bluesky/protocols.py
@@ -274,7 +274,7 @@ class Location(TypedDict, Generic[T]):
 
 @runtime_checkable
 class Locatable(Movable, Protocol):
-    def get_location(self) -> SyncOrAsync[Location]:
+    def locate(self) -> SyncOrAsync[Location]:
         """Return the current location of a Device.
 
         While a ``Readable`` reports many values, a ``Movable`` will have the
--- a/bluesky/protocols.py
+++ b/bluesky/protocols.py
@@ -3,7 +3,10 @@ try:
 except ImportError:
     from typing import Literal, Protocol, TypedDict, runtime_checkable
 
-from typing import Any, Awaitable, Callable, Dict, Iterator, List, Optional, Sequence, Tuple, Type, TypeVar, Union
+from typing import (
+    Any, Awaitable, Callable, Dict, Generic, Iterator, List, Optional,
+    Sequence, Tuple, Type, TypeVar, Union
+)
 
 
 # TODO: these are not placed in Events by RE yet
@@ -259,6 +262,26 @@ class Movable(Protocol):
     def set(self, value) -> Status:
         """Return a ``Status`` that is marked done when the device is done moving."""
         ...
+
+
+class Location(TypedDict, Generic[T]):
+    """A dictionary containing the location of a Device"""
+    #: Where the Device was requested to move to
+    setpoint: T
+    #: Where the Device actually is at the moment
+    readback: T
+
+
+@runtime_checkable
+class Locatable(Movable, Protocol):
+    def get_location(self) -> SyncOrAsync[Location]:
+        """Return the current location of a Device.
+
+        While a ``Readable`` reports many values, a ``Movable`` will have the
+        concept of location. This is where the Device currently is, and where it
+        was last requested to move to. This protocol formalizes how to get the
+        location from a ``Movable``.
+        """
 
 
 @runtime_checkable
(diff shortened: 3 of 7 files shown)


coretl:
I was going to change `mvr` to use this, but like `__read_and_stash_a_motor` is sync, and would need to make it either async, or invent a new message for this. I think I favour a new message, what do you think @tacaswell ?

coretl:
Also, I removed the doc reference to `Movable.position` as this replaces it

bluesky/plan_stubs.py
```diff
@@ -342,6 +342,15 @@ def rd(obj, *, default_value=0):
...
 
     """
+    # Location is canonical if it exists
+    if isinstance(obj, Locatable):
+        location = yield Msg("locate", obj)
+        if location is None:
+            # list-ify mode
+            return default_value
+        else:
+            return location["readback"]
```
coretl, line 352: Should `rd` return `setpoint` or `readback`?

bluesky/preprocessors.py
```diff
@@ -983,37 +993,40 @@ def __read_and_stash_a_motor(obj, initial_positions, coupled_parents):
...
-        reading = yield Msg('read', obj)
+    # First check if it is a locatable, and use its location if it is
+    if isinstance(obj, Locatable):
+        location = yield from __get_result_of_message("locate", obj)
+        if location is None:
+            setpoint = 0
+        else:
+            setpoint = location["setpoint"]
+    # Otherwise it might have a `position` attribution
+    elif hasattr(obj, "position"):
```
coretl, line 1004: `hasattr` rather than try/except, is this ok?

bluesky/preprocessors.py
```diff
@@ -983,37 +993,40 @@ def __read_and_stash_a_motor(obj, initial_positions, coupled_parents):
...
-    initial_positions[obj] = cur_pos
+    initial_positions[obj] = setpoint
 
     # if we move a pseudo positioner also stash it's children
     if obj in coupled_parents:
-        for c, p in zip(obj.pseudo_positioners, cur_pos):
+        for c, p in zip(obj.pseudo_positioners, setpoint):
             initial_positions[c] = p
 
     # if we move a pseudo single, also stash it's parent and siblings
```
coretl, line 1032: I didn't update the pseudo positioners, they still use position

bluesky/protocols.py
```diff
@@ -261,6 +264,26 @@ def set(self, value) -> Status:
...
+    """A dictionary containing the location of a Device"""
+    #: Where the Device was requested to move to
+    setpoint: T
+    #: Where the Device actually is at the moment
+    readback: T
+
+
+@runtime_checkable
+class Locatable(Movable, Protocol):
+    def locate(self) -> SyncOrAsync[Location]:
```
coretl, line 277: `locate` seemed to fit better with `read` and `set` than `get_location`

bluesky/run_engine.py
```diff
@@ -453,6 +453,7 @@ def setup_run_permit():
             'save': self._save,
             'drop': self._drop,
             'read': self._read,
+            'locate': self._locate,
```
coretl, line 456: I went with a new message in the end

coretl:
Unit tests failing as Ophyd seems to require a min of python3.6, looks like this is fixed in #1487

bluesky/run_engine.py
```diff
@@ -1853,6 +1854,19 @@ async def _read(self, msg):
 
         return ret
 
+    async def _locate(self, msg):
+        """
+        Locate a Movable and return its location.
+
+        Expected message object is:
+
+            Msg('locate', obj)
```
coretl, line 1863: Actually, having the ability to locate multiple objects concurrently would be useful for load/save. Would this be `Msg('locate', *objs)`? Similarity with `set`, but we don't have `Status` objects, so would have to do the asyncio gather within this function

danielballan (approved):

danielballan:
I intend to merge this on Wednesday if there are no other comments.

bluesky/plan_stubs.py
```diff
@@ -342,6 +342,15 @@ def rd(obj, *, default_value=0):
         The "single" value of the device
 
     """
+    # Location is canonical if it exists
```
ZLLentz, line 345: I think this should be added to the docstring above along with the other descriptions about how a read key is picked for the return value.

Something like, for example:
```
For devices that implement the Locatable protocol, the location is canonical and is returned without parsing the read keys.
```

bluesky/tests/test_examples.py
```diff
@@ -451,13 +451,13 @@ def test_duplicate_keys(RE, hw):
...
 
     def gen():
-        yield(Msg('open_run'))
-        yield(Msg('create', name='primary'))
-        yield(Msg('trigger', hw.det))
-        yield(Msg('trigger', hw.identical_det))
-        yield(Msg('read', hw.det))
-        yield(Msg('read', hw.identical_det))
-        yield(Msg('save'))
+        yield (Msg('open_run'))
```
ZLLentz, line 454: thank you for doing this

docs/source/hardware.rst
```diff
@@ -176,15 +176,19 @@ with the following additional methods and attributes.
...
-        A heuristic that describes the current position of a device as a
-        single scalar, as opposed to the potentially multi-valued description
-        provided by ``read()``.
+.. autoclass:: bluesky.protocols.Locatable
+    :members:
+    :show-inheritance:
 
-        Optional: bluesky itself does not use the position attribute, but other
-        parts of the ecosystem might.
-        Developers are encouraged to implement this attribute where possible.
```
ZLLentz, line 187: It might make sense to keep the `position` attribute in the documentation in some sort of deprecated sense to help people who stumble upon old devices that have not implemented `locate`.

ZLLentz (approved):
I like all the decisions made here. `locate` is an unambiguous name for the new message and method and is thematically consistent with the other bluesky verbs. Your implementation that fixes the relative set and reset wrappers is very clean.

bluesky/run_engine.py
ZLLentz: I think extending this in the same way `set` does for multiple objects makes a lot of sense. I don't fully understand the implications of doing the asyncio gather here vs elsewhere.

coretl: The `set()` method returns a `Status` object, so `Msg('set', obj)` only takes one device as an argument, and the combining of `Status` objects is done in the RE. `locate()` can be sync or async, and a `Status` object is not suitable as it is meant for long running tasks that don't have a return value. I think doing the grouping of locate objects in the message and doing the gather in this function is cleaner than pushing it up a level to the main RE loop like `set` does makes more sense here, but I'll defer to @tacaswell on this...

docs/source/hardware.rst
danielballan: Agreed.

coretl:
I have rebased on master, and applied the docs changes.

I also allowed multiple objects to be passed to `Msg('locate')`, and it would call `obj.locate()` on them all concurrently. I would welcome a review specifically of e47d78f2285d82fb0bda5217b1f74e8a63fad370 to see if it's a good idea or not.

coretl:
There are only 2 unresolved conversations left here, I'd welcome comments on them. When they're resolved I think this is ready for merge

bluesky/preprocessors.py
```diff
@@ -974,6 +976,14 @@ def _normalize_devices(devices):
     return devices, coupled_parents
 
 
+def __get_result_of_message(msg_type: str, obj):
+    result = yield Msg(msg_type, obj)
+    if result is None:
+        # this plan may be being list-ified
+        print(f"*** all positions for {obj.name} are relative to current position ***")
```
klauer, line 983: Though this was merely copied from the original code, I wonder if there's no better destination for such information. 

It'd be surprising to have this print be a side-effect of listifying a plan generator.

bluesky/run_engine.py
klauer: `Msg("locate", obj1, obj2, ...)` - a single message with multiple objects seems appropriate, especially with the ability to use `gather()` and perform the requests in parallel.

Looking back at the `read` message (which is closer to `locate` than `set` is, in my opinion), args/kwargs to that message currently get passed to `obj.read(*args, **kwargs)`. However, that usage is *maybe* deprecated according to https://github.com/bluesky/bluesky/blob/e47d78f2285d82fb0bda5217b1f74e8a63fad370/bluesky/utils/__init__.py#L1754-L1763 This makes me think that a change of convention for `locate` is probably a good thing. Of course, @tacaswell may disagree...

bluesky/plan_stubs.py
klauer: +1 for `readback`

danielballan: Agreed

bluesky/run_engine.py
danielballan: I agree.

coretl: Are we happy with `Msg('locate', obj) -> obj_location` but `Msg('locate', obj1, obj2) -> [obj1_location, obj2_location]`?

tacaswell (approved):

bluesky/run_engine.py
```diff
@@ -1849,6 +1850,27 @@ async def _read(self, msg):
...
+
+        If a single obj is passed, obj.locate() is returned. If multiple objs
+        are passed, obj.locate() is called in parallel for all objs and a list
+        of the results returned.
+
+        """
+        objs = [check_supports(obj, Locatable) for obj in (msg.obj,) + msg.args]
+        # actually _locate_ the objects
+        coros = [maybe_await(obj.locate()) for obj in objs]
+        if len(coros) == 1:
```
tacaswell, line 1869: I have a weak preference to always return a list even if it is length 1.  I think it is slightly more verbose if you always just use 1, but much nicer if you want to write generic code that is 0 or more.

bluesky/preprocessors.py
coretl: Captured in #1552

tacaswell:
Responding to https://github.com/bluesky/bluesky/pull/1536#discussion_r1012352071  I think that passing args/kwargs through to `read` was probably a mistake (and not a feature I think we actually use at NSLS-II).  Once you start to do that then the plan becomes (very silently) dependent on only hardware that will take those inputs.  If you are going to make hardware specfiic plans, better set the state of what ever you were going to pass into read via `bps.mv(..., ...)` before the read than _in_ the read.

tacaswell:
A specific example of why the locate message should always return a list is so that


```python
def some_plan(objs: List[Locatable]):
    objs = ...  # maybe passed to us?
    pos_s = yield from bps.locate(*objs)
    for p in pos_s:
       print(p)
```

It is nice to not have to special case that based on the length of `objs`.

tacaswell:
Sorry for the rapid fire short comments (talking to @coretl and @danielballan on slack).

We may want to take a queue from Matplotlib with `plt.subplots()` here and add a `squeeze=True` keyword to the message to control if we try to drop the outer dimension (https://matplotlib.org/stable/api/figure_api.html#matplotlib.figure.Figure.subplots ).  In the mpl case there can be 2 extra dimensions, but I think it still works here.

That way we can keep the niceity for one-of usage, rhyme with `read` in the one-off case, have the option for the more precise behavior, and have a back-compatible path to extending this to `read` in the future.

coretl:
> Sorry for the rapid fire short comments (talking to @coretl and @danielballan on slack).
> 
> We may want to take a queue from Matplotlib with `plt.subplots()` here and add a `squeeze=True` keyword to the message to control if we try to drop the outer dimension (https://matplotlib.org/stable/api/figure_api.html#matplotlib.figure.Figure.subplots ). In the mpl case there can be 2 extra dimensions, but I think it still works here.
> 
> That way we can keep the niceity for one-of usage, rhyme with `read` in the one-off case, have the option for the more precise behavior, and have a back-compatible path to extending this to `read` in the future.

So are you proposing:
- `Msg(`locate`, obj) -> location`
- `Msg(`locate`, obj1, obj2) -> [location1, location2]`
- `Msg(`locate`, obj, squeeze=False) -> [location]`
- `Msg(`locate`, obj1, obj2, squeeze=False) -> error`

tacaswell:
`Msg("locate", obj1, obj2, squeeze=False) -> [location1, location2]` 

The analogy in numpy.squeeze

```
In [3]: import numpy as np

In [4]: np.ones(5).squeeze()
Out[4]: array([1., 1., 1., 1., 1.])

In [5]: np.ones(5).reshape(1, 5).squeeze()
Out[5]: array([1., 1., 1., 1., 1.])

In [6]: np.ones(5).reshape(5, 1).squeeze()
Out[6]: array([1., 1., 1., 1., 1.])

In [7]: np.ones(6).reshape(2, 3).squeeze()
Out[7]:
array([[1., 1., 1.],
       [1., 1., 1.]])

In [8]: np.ones(1).squeeze()
Out[8]: array(1.)

In [9]: np.ones(1).squeeze().shape
Out[9]: ()
```

so would mean "get rid of any singleton dimensions all the way down to a scalar"

coretl:
Perfect, will make it so...

coretl:
This should be ready to go

PR #1540, opened 2022-08-22
Tidy test fixture

coretl:
## Description
Don't need to call `set_bluesky_event_loop` in test fixture anymore

## Motivation and Context
#1539 sets the bluesky event loop to the user supplied one, so the fixture no longer needs to

## How Has This Been Tested?
CI


ZLLentz (approved):
Looks good to me, the test suite passes so let's merge

PR #1543, opened 2022-09-02
Use multiplied cyclers for non-snaked axes

ksunden:
## Description

This short circuits the `snake_cyclers` method in `utils` to use the equivalent multiplied cyclers when there is no snaking.

## Motivation and Context

This is primarily a memory usage consideration.
The cycler object represented by multiplied cyclers is much smaller than the fully expanded cycler (which is equivalent) that this method currently produces.
Even for some modestly sized scans such as (11, 101, 101), the full cycler produced was eating hundreds of MB of RAM when all of the data was repeated anyway.

## How Has This Been Tested?

The existing snaking tests already do exactly what I would want to do to test this: ensure that the output is as expected.

### DIFF
--- a/bluesky/utils/__init__.py
+++ b/bluesky/utils/__init__.py
@@ -613,6 +613,8 @@ def snake_cyclers(cyclers, snake_booleans):
     """
     if len(cyclers) != len(snake_booleans):
         raise ValueError("number of cyclers does not match number of booleans")
+    if not any(snake_booleans[1:]):
+        return reduce(operator.mul, cyclers)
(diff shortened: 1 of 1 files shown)


ksunden:
At some point it may be worth putting snaking natively in cycler, then even the snaked versions would benefit from reduction of repeated values kept in memory

untzag (approved):

untzag:
This makes perfect sense to me, a very nice performance change for what the Wright Group does.

Test failures are unrelated as far as I know.

PR #1544, opened 2022-09-06
release prep v1.10.0

untzag:
release prep v1.9.1


docs/source/api_changes.rst
```diff
@@ -2,7 +2,23 @@
  Release History
 =================
 
-v1.9.0 (2022-XX-YY)
+v1.9.1 (2022-09-06)
```
tacaswell, line 5:
```suggestion
v1.10.0 (2022-09-06)
```

tacaswell:
I think the change to stop tracking the handles for tasks is a big enough change that I do not think it is safe to do in a micro/patch release.

tacaswell (approved):

PR #1562, opened 2023-01-12
Made protocol methods abstract

### LINKED ISSUE #1547: Add abstractmethod decorators to all protocols
## Expected Behavior
I like type checking protocols with mypy. For example, this gives an error with mypy to say that `Implicit` is not a `Readable` because it is missing the `read` method:
```python
from bluesky.protocols import Readable

class Implicit:
    def describe(self):
        return {}

r: Readable = Implicit()
```
However, when defining the class it would be good to declare what protocols it implements. You can do this by subclassing:
```python
from bluesky.protocols import Readable

class Explicit(Readable):
    def describe(self):
        return {}

r: Readable = Explicit()
```
Unfortunately this makes mypy happy with the assignment because we define stub methods in the protocols. The way to make mypy correctly error seems to be to add the `@abstractmethod` decorator to all the stub methods of the protocols:
```python
class Readable(Protocol):
    @abstractmethod
    def read(self) -> Dict[str, Any]:
        ...
```
As defined in the docs: https://mypy.readthedocs.io/en/stable/class_basics.html#abstract-base-classes-and-multiple-inheritance

> Note that mypy performs checking for unimplemented abstract methods even if you omit the [ABCMeta](https://docs.python.org/3/library/abc.html#abc.ABCMeta) metaclass. This can be useful if the metaclass would cause runtime metaclass conflicts.

Any objections if I decorate all protocol class methods with this?

evalott100:
Made protocol methods abstract. Added missing method definitions in protocol subclasses defined in unit tests.

## Motivation and Context
We'd like to use abstract classes for protocols as discussed here #1547 
Closes #1547 

## Testing
Tested inside `test_new_examples.py` using initialisation of classes `Jittery` (subclass of  `(Readable, Locatable)`) and ` AsyncLocatable` (subclass of `Locatable`).

### DIFF
--- a/bluesky/tests/test_new_examples.py
+++ b/bluesky/tests/test_new_examples.py
@@ -170,9 +170,14 @@ def test_mvr(RE, hw):
     assert actual[2] == Msg('wait', None)
 
 
-def test_locatable_message_multiple_objects(RE):
+def test_locatable_message_multiple_objects(RE, hw):
     class AsyncLocatable(Locatable):
         value = 1.0
+
+        def set(self, value) -> Status:
+            self.position = value + 0.1
+            # We don't actually need a motor, but use one just to give us a Status
+            return hw.motor.set(value)
 
         async def locate(self) -> Location:
             # Grab the original value
@@ -204,7 +209,7 @@ def test_locatable_message_multiple_objects(RE):
     ]
 
 
-def test_rd_locatable(RE):
+def test_rd_locatable(RE, hw):
     class Jittery(Readable, Locatable):
         def describe(self) -> Dict[str, Descriptor]:
             return dict(x=dict(source="dummy", dtype="number", shape=[]))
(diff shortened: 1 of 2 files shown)


bluesky/tests/test_new_examples.py
```diff
@@ -215,6 +220,15 @@ def read(self) -> Dict[str, Reading]:
...
             return dict(setpoint=1.0, readback=1.1)
 
+        def set(self, value) -> Status:
+            self.position = value + 0.1
+            # We don't actually need a motor, but use one just to give us a Status
+            return hw.motor.set(value)
+
+        def name(self) -> str:
+            # Abstractmethod name should be defined in the subclass.
+            return ""
```
coretl, line 230: Please can you check that the following change is sufficient to make mypy happy...
```suggestion
        name = ""
```

evalott100: With that change mypy is "happy" (has the same single error in that file for `ophyd.sim` that it did before).

klauer (commented):
I'm in favor of the changes overall 👍 

~Is it worth considering adding even a limited set of mypy checks to continuous integration to ensure there are no regressions?~ Looks like it's already in the test suite - great

untzag:
@klauer take a look at this if you're not aware

https://github.com/bluesky/bluesky/blob/master/bluesky/tests/test_protocols.py

untzag (approved):
This branch passes both yaqc-bluesky and ophyd tests.

untzag:
This is a minimal change that introduces new functionality that shouldn't collide with anything people are already doing. I'm merging in the interest of keeping things moving forward.

PR #1563, opened 2023-02-01
ENH: make stage and unstage return status objects

### LINKED ISSUE #1559: Let stage and unstage return Status
At the moment `stage()` and `unstage()` are sync and return a `List[Any]` of the devices that have been staged.

Discussions with @tacaswell and @danielballan suggest this return value is rarely used, and it would be more valuable to stage multiple devices at the same time.

Suggest we change the protocols to be:
- `def stage() -> Status`
- `def unstage() -> Status`

Need to change https://github.com/bluesky/bluesky/blob/d2b6bddd4415cdb2946fe8d552bd704bc9932c11/bluesky/run_engine.py#L2347-L2378 to look at the return values. If it is a `Status` object then track and return it like in https://github.com/bluesky/bluesky/blob/d2b6bddd4415cdb2946fe8d552bd704bc9932c11/bluesky/run_engine.py#L2130-L2160 
but if not then just return the result for backwards compatibility.

Also need to change https://github.com/bluesky/bluesky/blob/d2b6bddd4415cdb2946fe8d552bd704bc9932c11/bluesky/plan_stubs.py#L692-L729 to also call `wait()` if the result of the call is a Status message like we do in https://github.com/bluesky/bluesky/blob/d2b6bddd4415cdb2946fe8d552bd704bc9932c11/bluesky/plan_stubs.py#L913-L918

tizayi:
Let stage and unstage return Status objects, so multiple objects can be staged at once.

## Description

I have made changes to  `_stage` and `_unstage` to allow Stageables to return status objects from their stage and unstage methods. The changes mimic `_trigger`. 

I have changed the type hints on the Stageable protocol.

I have created a test for staging 2 Stageables at once.  

I have also changed the plan stubs accordingly.

## Motivation and Context

resolves #1559 

This change will allow the run engine to stage multiple devices without blocking. And these can then be awaited but a wait message.

## How Has This Been Tested?

These changes have been tested by a new test `test_stage_and_unstage_status_objects()`.
Which stages/unstages 2 dummy Stageables then awaits their completion.

The new `_stage` and `_unstage` methods use `isinstance()` to check to see if the output of an objects stage/unstage method is a status object. If it is not they will return the result preserving previous behaviour for Stageables that returns lists of staged objects.

### DIFF
--- a/bluesky/plan_stubs.py
+++ b/bluesky/plan_stubs.py
@@ -13,7 +13,7 @@ try:
 except ImportError:
     from toolz import partition
 
-from .protocols import Locatable, Triggerable
+from .protocols import Locatable, Triggerable, Status
 from .utils import (
     get_hinted_fields,
     merge_cycler,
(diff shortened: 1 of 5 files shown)


bluesky/plan_stubs.py
```diff
@@ -689,44 +689,58 @@ def configure(obj, *args, **kwargs):
     return (yield Msg('configure', obj, *args, **kwargs))
 
 
-def stage(obj):
+def stage(obj, *, group=None, wait=False):
```
coretl, line 692: @tacaswell should this default to waiting for completion?

bluesky/run_engine.py
```diff
@@ -2344,6 +2343,25 @@ async def _configure(self, msg):
             await current_run.configure(msg)
         return old, new
 
+    def _add_status_to_group(self, obj: typing.Any, status_object: Status,
```
coretl, line 2346: Please can you also make `_trigger` use this function?

bluesky/protocols.py
```diff
@@ -320,21 +320,19 @@ def describe_collect(self) -> SyncOrAsync[Dict[str, Dict[str, Descriptor]]]:
 class Stageable(Protocol):
     # TODO: we were going to extend these to be able to return plans, what
     # signature should they have?
-    def stage(self) -> List[Any]:
+    def stage(self) -> Status:
```
coretl, line 323: @tacaswell do you think we should make this `Union[Status, List[Any]]` and support both as equally valid, or have the protocols only show the new approach and put the old one in the docs as deprecated?

danielballan: My two cents: I am happy either way on this. I might lean toward annotating both as valid at this early stage.

bluesky/plan_stubs.py
danielballan: I lean "no", for consistency with `abs_set` and others. These low-level plans should be "async" (in the general sense) by default.

coretl: That's true, if we have an old-style stage then it will wait no matter what we pass here. Which begs the question, should we make the return from the `stage` message always be a `Status` object?

bluesky/plan_stubs.py
```diff
@@ -689,44 +689,58 @@ def configure(obj, *args, **kwargs):
...
     msg : Msg
-        Msg('stage', obj)
 
     See Also
     --------
     :func:`bluesky.plan_stubs.unstage`
     """
-    return (yield Msg('stage', obj))
+    ret = yield Msg('stage', obj, group=group)
+    if wait:
```
coretl, line 714: @danielballan should we only do a wait if `isinstance(ret, Status)`? Or is it harmless?

bluesky/plan_stubs.py
danielballan: Decision from call:

```py
ret = yield Msg("stage", obj, group=group)
if wait is None:
    # Wait for old-style devices (no choice!) but do not wait for new-style.
    wait = not isinstance(ret, Status)
if wait:
    yield from Msg("wait", None, group=group)
```

bluesky/protocols.py
coretl: Let's go with the `Union` then

bluesky/plan_stubs.py
```diff
@@ -689,44 +689,138 @@ def configure(obj, *args, **kwargs):
...
+    See Also
+    --------
+    :func:`bluesky.plan_stubs.stage`
+    :func:`bluesky.plan_stubs.unstage_all`
+    """
+    group = group or str(uuid.uuid4())
+    status_objects = []
+
+    for obj in args:
+        ret = yield Msg('stage', obj, group=group)
```
coretl, line 753: We need this to work with old-style too. I think we want to call `stage()` here, append to `status_objects` only if we get a `Status` back, then only call `wait` if we have any `status_objects`

bluesky/tests/test_new_examples.py
```diff
@@ -807,3 +812,62 @@ def bad_sig(detectors, mtr, step):
...
+            threading.Timer(0.5, self._callback, args=[unstaged, st]).start()
+            return st
+
+    olddummy1 = OldStyleDummy("o1")
+    olddummy2 = OldStyleDummy("o2")
+    newdummy1 = NewStyleDummy("n1")
+    newdummy2 = NewStyleDummy("n2")
+
+    def plan():
+        yield stage_all(olddummy1, olddummy2, newdummy1, newdummy2)
```
coretl, line 863: I think this should be `yield from`. But wait, how does this work at all?

bluesky/plan_stubs.py
coretl: Or maybe instead of calling `stage()` we only yield `Msg('wait')` if there are any `Status` objects in `status_objects`.

bluesky/tests/test_new_examples.py
tizayi: It wasn't working but pytest was skipping the test I made. Because it had the async keyword in front of it.

coretl (approved):
I think this is good to go.

@danielballan or @tacaswell can we merge?

mrakitin:
@awalter-bnl was interested in the progress on that topic.

coretl:
@mrakitin will review and merge tomorrow if he's happy

bluesky/tests/test_new_examples.py
```diff
@@ -807,3 +812,62 @@ def bad_sig(detectors, mtr, step):
...
+    def plan():
+        yield from stage_all(olddummy1, olddummy2, newdummy1, newdummy2)
+        assert list(staged.keys()) == ["o1", "o2", "n1", "n2"]
+
+        yield from unstage_all(olddummy1, olddummy2, newdummy1, newdummy2)
+        assert list(unstaged.keys()) == ["o1", "o2", "n1", "n2"]
+
+    start = ttime.time()
+    RE(plan())
+    stop = ttime.time()
```
mrakitin, line 871: `ttime.monotonic()` will be more stable to handle the DST-related changes (see https://docs.python.org/3/library/time.html#time.monotonic for details).
```suggestion
    start = ttime.monotonic()
    RE(plan())
    stop = ttime.monotonic()
```

bluesky/tests/test_run_engine.py
```diff
@@ -386,6 +387,64 @@ def resume(self):
...
+
+        assert "d1" in started_unstage and "d2" in started_unstage
+        assert not unstaged
+
+        yield Msg("wait", group="unstage_group")
+        assert "d1" in unstaged and "d2" in unstaged
+
+    start = ttime.time()
+    RE(my_plan())
+    stop = ttime.time()
```
mrakitin, line 443: Same `ttime.monotonic()` suggestion here.

mrakitin (approved):
Great addition! I added a couple of non-blocking suggestions.

mrakitin (approved):
Great, thanks for the follow-up updates, @tizayi!

As a side note, I find it pretty convenient to stack commits so that incremental changes can be reviewed rather than the whole PR again. I think the force-push should not be used for that reason unless a trivial typo or something similar was introduced. We don't do squashing on merge normally for this repo.

Issue #1550, opened 2022-10-06
*loop* parameter was removed from asyncio.Queue() in Py3.10


prjemian:
As part of unit testing a new f[eature](https://github.com/BCDA-APS/apstools/issues/635) in apstools, encountered this error in the Py3.10 unit testing workflow:

```py
>               RE(my_plan(msg, default, agree, bypass))

/home/runner/work/apstools/apstools/apstools/plans/tests/test_input_plan.py:38: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/runner/micromamba-root/envs/anaconda-test-env-py-3.10/lib/python3.10/site-packages/bluesky/run_engine.py:893: in __call__
    plan_return = self._resume_task(init_func=_build_task)
/home/runner/micromamba-root/envs/anaconda-test-env-py-3.10/lib/python3.10/site-packages/bluesky/run_engine.py:1032: in _resume_task
    raise exc
/home/runner/micromamba-root/envs/anaconda-test-env-py-3.10/lib/python3.10/site-packages/bluesky/run_engine.py:1663: in _run
    raise err
/home/runner/micromamba-root/envs/anaconda-test-env-py-3.10/lib/python3.10/site-packages/bluesky/run_engine.py:1498: in _run
    msg = self._plan_stack[-1].throw(stashed_exception or resp)
/home/runner/work/apstools/apstools/apstools/plans/tests/test_input_plan.py:13: in my_plan
    yield from request_input(
/home/runner/work/apstools/apstools/apstools/plans/input_plan.py:45: in request_input
    r = yield from bps.input_plan(full_text)
/home/runner/micromamba-root/envs/anaconda-test-env-py-3.10/lib/python3.10/site-packages/bluesky/plan_stubs.py:554: in input_plan
    return (yield Msg('input', prompt=prompt))
/home/runner/micromamba-root/envs/anaconda-test-env-py-3.10/lib/python3.10/site-packages/bluesky/run_engine.py:1583: in _run
    new_response = await coro(msg)
/home/runner/micromamba-root/envs/anaconda-test-env-py-3.10/lib/python3.10/site-packages/bluesky/run_engine.py:2431: in _input
    async_input = AsyncInput(self.loop)
[...]

PR #1566, opened 2023-02-09
DEP: remove loop param from AsyncInput in bluesky.utils

hyperrealist:
See #1550

## Description

Instantiation of asyncio.Queue in bluesky/utils is changed from the following aproach:
https://github.com/bluesky/bluesky/blob/a60e5c7323a61b9a988ce9f8d1b4d7f4ac495ee8/bluesky/utils/__init__.py#L1052
to the following conditional approach:
```py
        if sys.version_info < (3, 10):
            self.q = asyncio.Queue(loop=self.loop)
        else:
            self.q = asyncio.Queue()
```

## Motivation and Context

The problematic syntax used in bluesky's async input implementation has been raised as an issue here: #1550. This patch is an attempt to fix the issue.

Synchronization primitives such as `asyncio.Queue` in python 3.10+ no longer accept a *loop* parameter at instantiation. Instead, new primitives internally use a thread-safe mechanism to capture the current *loop* on await (see [discussion](https://github.com/python/cpython/issues/86558) for more details). This patch updates Queue instantiation to match the expected syntax in python 3.8, 3.9, 3.10.

## How Has This Been Tested?

A full unit test on python 3.8, 3.9, and 3.10 [passed](https://github.com/hyperrealist/bluesky/actions/runs/4133422888). Note: the unit test were performed on a separate [branch based on this topic branch](hyperrealist/bluesky/tree/unit_test) with a slightly modified workflow (modification: added a `workflow_dispatch` to [testing.yml](https://github.com/hyperrealist/bluesky/blob/c21704061bbd65ca501bcda48929c103753041c5/.github/workflows/testing.yml#L4) in order to manually trigger a unit test. Reason: github did not trigger an automatic unit test on push).

Moreover, bluesky.utils [adapts](https://github.com/bluesky/bluesky/blob/72ea2473359e3020a27f4338bad5446534dc7d02/bluesky/utils/__init__.py#L1048) async input implementation from http://stackoverflow.com/a/35514777/1221924 (which has since been updated to python 3.10 syntax). I manually tested the original example, but with the same proposed fix. All manual tests on python versions 3.6, 3.7, 3.8, 3.9, 3.10 (using separate conda environments) were successful.

### DIFF
--- a/bluesky/utils/__init__.py
+++ b/bluesky/utils/__init__.py
@@ -1049,7 +1049,10 @@ class AsyncInput:
     """
     def __init__(self, loop=None):
         self.loop = loop or asyncio.get_event_loop()
-        self.q = asyncio.Queue(loop=self.loop)
+        if sys.version_info < (3, 10):
+            self.q = asyncio.Queue(loop=self.loop)
+        else:
+            self.q = asyncio.Queue()
         self.loop.add_reader(sys.stdin, self.got_input)
 
     def got_input(self):


tacaswell (approved):

tacaswell:
Thanks @hyperrealist !  I hope we hear from you again.

Can I ask what context you are using bluesky that you came across this?

PR #1567, opened 2023-02-21
ENH: Add ignore suspenders context decorator

maffettone:
## Description
The context decorator allows using either a context manager or a decorator on a plan to temporarily remove a suspender and then add it when the plan is complete. 

## Motivation and Context
Some plans should not depend on all suspenders. (E.g. move the detector a long distance, may not care that a beamstop is closed). 

## How Has This Been Tested?
TODO: 
- [ ] Unit tests for this a context manager and decorator
- [ ] Evaluate if this is the desired approach in the context of [suspenders preprocessor](https://github.com/bluesky/bluesky/blob/da095e0ecd5792ff1695ae405906df49ed6df243/bluesky/preprocessors.py#L388-L421)

### DIFF
--- a/bluesky/suspenders.py
+++ b/bluesky/suspenders.py
@@ -1,10 +1,13 @@
 import asyncio
-from datetime import datetime, timedelta
-from abc import ABCMeta, abstractmethod, abstractproperty
 import operator
 import threading
+from abc import ABCMeta, abstractmethod, abstractproperty
+from contextlib import ContextDecorator
+from datetime import datetime, timedelta
 from functools import partial
 from warnings import warn
+
+import bluesky
 
 
 class SuspenderBase(metaclass=ABCMeta):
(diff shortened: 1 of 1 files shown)


danielballan:
This is an interesting concept.

It would be good to know more about the beamline use case before we go ahead. If the use case is to switch suspenders off while the beam is down and then back on, I suspect this construction is too verbose.

Adding it opens up a whole new branch of Ways To Do Things In Bluesky. I'm open to it, but we should be sure we really need it before we open it up.

```
RE.clear_suspenders()
```

is really enough in that case

cjtitus:
This sounds like a great decorator. I would use this to take dark counts, where we intentionally close the shutters and take measurements with no beam, which requires "suspending" our beam intensity suspender. This has tripped me up several times. I suspect many other people have the same issue.

stan-dot:
@maffettone what is the state of the need for this in 2024?

tacaswell (approved):
I think this is reasonable.

In general, going the route of adding suspenders to the plans you want to have them rather than poking holes in the ones you do not want to have them seems better (you have to know what suspenders exist to poke them out, where as you only have to know your needs to install the ones you want).  I also see the case that if for a vast majority you want the standard ones only poking them out for the special cases is way less boiler plate.

maffettone:
From a "how to design it correctly" I agree with Tom. This is more of a, "how people end up using it" regardless of best intent.

PR #1569, opened 2023-03-16
MNT: move fig_factory default resolution in BestEffortCallback

tacaswell:
## Description

Move the resolution of the default figure factory in BestEffortCallback to the `__init__` from its call site.

## Motivation and Context

General cleanup, doing this early is better than doing this late (and every time).

## How Has This Been Tested?

CI


mrakitin (approved):
I like the enhancement!

PR #1570, opened 2023-03-17
Propagate exception through failed status

RAYemelyanova:
It was noticed that ophyd v2 AsyncStatus objects do not carry their exceptions through to the run engine. This relates to the following issue: https://github.com/bluesky/ophyd/issues/1098

## Description
There are several small changes involved in this PR:
1. Status protocol now has an 'exception' property, which can return exceptions.
2. When a FailedStatus is created (in `RunEngine::_status_object_completed`) this method now checks if the status has exception as a method (this is for backwards compatibility with ophyd v1 devices). It retrieves the exception and raises a FailedStatus from it.

These have been tested against a PR in the ophyd library, found here: https://github.com/bluesky/ophyd/pull/1103

## Motivation and Context
It is annoying for users to have bluesky plans fail with ophyd v2 devices, and get no more information than a simple FailedStatus. In addition, the FailedStatus is often one of the most recent calls in traceback, so it appears first visually, and seeing something like:

```
ERROR FailedStatus: None
```

can be a bit disheartening.

The aim is, that this PR will help to propagate errors through the FailedStatus in ophyd v2 devices.
## How Has This Been Tested?
I have added a couple of tests, one of which will fail until the ophyd v2 pull request has been merged. These just check how ophyd devices propagate their errors - for the ophyd v2 case, one can easily test the implications of the changes in these PRs:

1. Clone this repository
2. install requirements, i.e. `pip install -e . / pip install -r requirements-dev.txt`
3. install ophyd, i.e. `pip install ophyd`
4. `pytest bluesky/tests/test_run_engine.py -k test_failures_propagated_through_FailedStatus_ophyd_v2`
5. Observe the output; the test will fail, and in the failure the FailedStatus will contain no more information.
6. uninstall ophyd and reinstall from the PR mentioned above by typing, `pip install git+https://github.com/bluesky/ophyd.git@propagate_exception_through_FailedStatus`
7. repeat steps 5 and 6, noticing the tests pass as the FailedStatus now recognizes the correct exception.


RAYemelyanova:
Note: I have just noticed that there are a lot more changes than I realised in this PR, due to formatting - my IDE applies black formatting. Should a separate PR be made to make bluesky conform to black standards?

bluesky/protocols.py
```diff
@@ -134,6 +137,10 @@ def add_callback(self, callback: Callable[["Status"], None]) -> None:
         """
         ...
 
+    @property
+    def exception(self) -> Optional[StatusException]:
+        ...
+
```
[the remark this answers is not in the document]
RAYemelyanova: Why should it raise? It could just be an unused parameter, no? Or would this cause an issue with flake8/isort/linting?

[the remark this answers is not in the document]
coretl: It should raise because it can't honour a timeout of anything other than 0, asyncio does not allow you to block in a sync function while still keeping the event loop alive. If the user requests anything other than 0, then they are explicitly asking for a wait with timeout which we can't do, so it should raise saying so.

bluesky/run_engine.py
```diff
@@ -2194,7 +2194,8 @@ def _status_object_completed(self, ret, p_event, pardon_failures):
             # TODO: need a better channel to move this information back
             # to the run task.
             with self._state_lock:
-                self._exception = FailedStatus(ret)
```
coretl, line 2197: A nicer version of this...
```python
try:
    raise FailedStatus(ret) from ret.exception(timeout=0)
except Exception as e:
    self._exception = e
```

RAYemelyanova: Have added this and made sure to propagate the traceback also

bluesky/run_engine.py
```diff
@@ -2194,7 +2194,11 @@ def _status_object_completed(self, ret, p_event, pardon_failures):
...
+                try:
+                    exc = ret.exception(timeout=0)
+                    raise FailedStatus(exc).with_traceback(exc.__traceback__) from exc
```
tacaswell, line 2199:
```suggestion
                    raise FailedStatus(ret).with_traceback(exc.__traceback__) from exc
```

RAYemelyanova: Sure, have now changed this.

RAYemelyanova:
Have modified a test to check for the full traceback contents of a FailedStatus.

tacaswell (approved):

PR #1574, opened 2023-04-19
Point at newly versioned event model docs

coretl:
## Description
When event model started writing versioned docs the path for intersphinx changed, this fixes it

## How Has This Been Tested?
Let's see if CI passes...


coretl:
I have no idea why the unit tests are failing, but master is also failing with the same error...

tacaswell:
I'm going to merge this over the test failures (I'm confident this will only change the docs build).

The failures look like something is going funny with the tmpdir fixture + disk on CI?

PR #1575, opened 2023-05-03
removed with_traceback call

RAYemelyanova:
## Description

Previously we had an exception being raised like so:

`raise FailedStatus(ret).with_traceback(ret.__traceback__) from exc`

It's been brought to my attention that this is redundant, so I've just removed the dunder method here.

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -2225,7 +2225,7 @@ class RunEngine:
             with self._state_lock:
                 try:
                     exc = ret.exception(timeout=0)
-                    raise FailedStatus(ret).with_traceback(exc.__traceback__) from exc
+                    raise FailedStatus(ret) from exc
                 except Exception as e:
                     self._exception = e
         p_event.set()


RAYemelyanova:
bit strange that the tests are failing.. seem unrelated to my change.

coretl:
@tacaswell @danielballan Seems to be a single failture with `test_persistent_dict` but all the other tests in that file also fail. This is also failing on master. Any ideas?

@RAYemelyanova does this also fail locally?

RAYemelyanova:
@coretl no, tests do not fail locally. At least, when I ran pytest on the tests that would be affected.

Am running all tests now, and will update if they fail locally...

coretl:
Yeah, last passing test was `zict 2.2.0`, and first failing test was `zict 3.0.0`. Changelog here: https://zict.readthedocs.io/en/latest/changelog.html

coretl (approved):

PR #1579, opened 2023-05-09
Implement compose functions from event model into the bundlers

### LINKED ISSUE #1573: Implement compose functions from event-model into the bundlers
Currently the when bundling documents the event-model schema is validated against, but the compose functions are not used to create the documents. We should get it working with the current model before trying the [rethink external assets](https://github.com/bluesky/event-model/issues/236).

evalott100:
Closes #1573 

Implements compose functions from event-model into the bluesky bundler.

# To test

~~Change `event-model` to `event-model@git+https://github.com/evalott100/event-model@switch_from_partials_to_classes` in `requirements.txt"~~

### DIFF
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -1,6 +1,5 @@
 from collections import deque
 import inspect
-from itertools import count, tee
 import time as ttime
 from typing import Any, Deque, Dict, FrozenSet, List, Tuple, Callable
 from event_model import DocumentNames, compose_run
@@ -43,7 +42,7 @@ class RunBundler:
         self._local_descriptors: Dict[Any, Dict[FrozenSet[str], Tuple[str, str, Callable, Callable]]] = dict()
         # a seq_num counter per stream
         self._sequence_counters: Dict[Any, Dict[FrozenSet[str], Tuple[str, str, Callable, Callable]]] = dict()
-        self._teed_sequence_counters = dict()  # for if we redo data-points
+        self._sequence_counters_copy = dict()  # for if we redo data-points
         self._monitor_params: Dict[Subscribable, Tuple[Callback, Dict]] = dict()  # cache of {obj: (cb, kwargs)}
         self.run_is_open = False
         self._uncollected = set()  # objects after kickoff(), before collect()
@@ -62,6 +61,7 @@ class RunBundler:
         self._run_start_uid = new_uid()
         self._interruptions_desc_uid = None  # uid for a special Event Desc.
         self._interruptions_counter = 0  # seq_num, special Event stream
+        run = compose_run(uid=self._run_start_uid, event_counters=self._sequence_counters, metadata=self._md)
 
         (
             doc,
@@ -69,7 +69,7 @@ class RunBundler:
             self._compose_resource,
             self._compose_stop,
             self._compose_stream_resource
-        ) = compose_run(uid=self._run_start_uid, event_counters=self._sequence_counters, **self._md)
+        ) = run.__dict__.values()
         await self.emit(DocumentNames.start, doc)
         doc_logger.debug("[start] document is emitted (run_uid=%r)", self._run_start_uid,
                          extra={'doc_name': 'start',
@@ -284,7 +284,10 @@ class RunBundler:
 
     async def _cache_describe_config(self, obj):
         "Read the object's describe_configuration and cache it."
-        if isinstance(obj, Configurable):
+
+        # Pylance doesn't understand this is an acceptable isinstance with the
+        # @runtime_checkable, hence the type ignore.
+        if isinstance(obj, Configurable):  # type: ignore
             conf_keys = await maybe_await(obj.describe_configuration())
         else:
             conf_keys = {}
@@ -292,7 +295,7 @@ class RunBundler:
 
     async def _cache_read_config(self, obj):
         "Read the object's configuration and cache it."
-        if isinstance(obj, Configurable):
+        if isinstance(obj, Configurable):  # type: ignore
             conf = await maybe_await(obj.read_configuration())
         else:
             conf = {}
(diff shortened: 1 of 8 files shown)


evalott100:
Tests will fail here since the `event-model` release isn't up to date with main - a new release would be necessary .

bluesky/bundlers.py
```diff
@@ -40,9 +39,10 @@ def __init__(self, md, record_interruptions, emit, emit_sync, log, *, strict_pre
         self._config_ts_cache: ObjDict[Any] = dict()  # " obj.read_configuration() timestamps
         self._descriptors = dict()  # cache of {name: (objs_frozen_set, doc)}
         # cache of {obj: {objs_frozen_set: (name, uid)}}
-        self._local_descriptors: Dict[Any, Dict[FrozenSet[str], Tuple[str, str]]] = dict()
-        self._sequence_counters = dict()  # a seq_num counter per stream
-        self._teed_sequence_counters = dict()  # for if we redo data-points
+        self._local_descriptors: Dict[Any, Dict[FrozenSet[str], Tuple[str, str, Callable, Callable]]] = dict()
```
coretl, line 42: Does this work?
```suggestion
        self._local_descriptors: Dict[Any, Dict[FrozenSet[str], ComposeDescriptorBundle] = dict()
```
Also for `self._descriptors` and others.

docs/source/api_changes.rst
```diff
@@ -2,6 +2,17 @@
  Release History
 =================
 
+Unreleased
+==========
+
+* Changed `~bluesky.bundlers` to use event-model compose functions
+  * In the run stop document, 'num_changes' will now contain descriptors even if
+  they have no events associated with them.
+  * Documents are now tested against schema when they are composed.
```
coretl, line 11:
```suggestion
  * Events produced by monitors are now checked against their corresponding Descriptor document.
```

coretl (changes requested):

bluesky/bundlers.py
evalott100: ComposeDescriptorBundle doesn't contain the two strings in the tuple for values in `_local_descriptors` (for the stream_name and descriptor uid), similarly `_descriptors` contains an `obj_set` not present in the ComposeDescriptorBundle

evalott100: Should I rework the code to obtain that stuff from the doc and get rid of the current descriptor logic?

coretl: If you think that would come out neater then yes

evalott100:
Tested to work against https://github.com/evalott100/event-model/tree/switch_from_partials_to_classes, we will need to get this merged into event model before we merge this

bluesky/bundlers.py
coretl: `stream_name` should already be in the descriptor doc anyway (it's `name`), but I'm not convinced we should add `objs_read` into it, that's a bluesky thing rather than an event model thing... We could have a tuple of `DescriptorBundle` and `objs_read`?

bluesky/bundlers.py
```diff
@@ -293,15 +285,18 @@ async def _cache_describe(self, obj):
 
     async def _cache_describe_config(self, obj):
         "Read the object's describe_configuration and cache it."
-        if isinstance(obj, Configurable):
+
+        # Pylance doesn't understand this is an acceptable isinstance with the
+        # @runtime_checkable, hence the type ignore.
+        if isinstance(obj, Configurable):  # type: ignore
```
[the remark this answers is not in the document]
evalott100: It would require using a 
``` python
# pyright: ignore [X]
```

where `X` is one of [these options](https://github.com/microsoft/pyright/blob/main/docs/configuration.md). I've tried searching for it but I can't find which of these this error belongs to...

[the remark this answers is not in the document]
coretl: The latest update of vscode doesn't seem to error on these lines for me even without the ignore

coretl (commented):
This looks reasonable to me, although we should wait for event model to be released before merging

tacaswell:
This is a regression in mpl 3.7.2!  We should ban that version in requirments-dev.  I'll both fix this upstream and open a PR to bluesky to switch to a better way of doing the layout.

evalott100:
> This is a regression in mpl 3.7.2! We should ban that version in requirments-dev. I'll both fix this upstream and open a PR to bluesky to switch to a better way of doing the layout.

Great, thank you! Feel free to add me as a reviewer.

coretl (approved):
I'm happy with this, @tacaswell are we ok to merge?

bluesky/tests/test_run_engine.py
```diff
@@ -1742,8 +1743,8 @@ def plan():
             yield Msg('read', det)
             yield Msg('save')
         finally:
-            yield Msg('declare_stream', None, det2,  name='primary')
-            yield Msg('create', name='primary')
+            yield Msg('declare_stream', None, det2,  name='primary2')
+            yield Msg('create', name="primary2")
```
coretl, line 1747: Why primary2?

danielballan:
@coretl Merge at will once these last two comments are resolved.

bluesky/tests/test_run_engine.py
evalott100: @tacaswell This should indeed be a different name, the code fails here

https://github.com/bluesky/bluesky/blob/febaa1096b20ce4c883f719f72a08f4c07aba1ac/bluesky/bundlers.py#L150-L156

event-model already has a descriptor with name `"primary"` and so it rightly throws an error if another is made. 

This doesn't fail in current bluesky since the doc is made without this check:

https://github.com/bluesky/bluesky/blob/90bdda70877e4f825d877f83449e09e0e9eed89b/bluesky/bundlers.py#L157-L166

My most recent commit changes `"primary"` to `"det1"`, and `"primary2"` to `"det2"` to avoid confusion.

PR #1581, opened 2023-05-16
MNT/API: validate that stream names must be strings

tacaswell:
We make that assumption downstream, enforce this at the RE level.

## Motivation and Context

@evalott100 found a code path where we were handling the case of the bundle name being `None`.  This will ensure that case is never hit.

## How Has This Been Tested?

Added tests.


danielballan:
Should this be moved into event-model?

tacaswell:
I am 👍🏻 on moving this to event model.

PR #1583, opened 2023-05-19
Remove useless instructions about install_kicker

graeme-winter:
## Description

Removing instruction which when performed says it is not necessary any more

## Motivation and Context

Housekeeping

## How Has This Been Tested?

It's text


mrakitin (approved):

tacaswell:
test failure is a timing issue on CI.

PR #1584, opened 2023-05-19
DOC: Tidy up tutorial around gridscans

DominicOram:
## Description
Next section of the tutorial uses 3 motors and fails, so we should import the third here. In running a session based on the tutorial users were confused about why they saw no plot from the 3D grid so added an explination.

## Motivation and Context
Help avoid confusion


tacaswell (approved):

tacaswell:
Thank you.

The failure looks like a timing issue on CI.

In the future can you please make PRs from your fork not by adding branches to the main repo?

Issue #1572, opened 2023-04-05
Need typing-extensions in <3.11


coretl:
## Report
```
$git clone https://github.com/bluesky/bluesky.git
$cd bluesky
$python3
>>> import bluesky
Traceback (most recent call last):
  File “<stdin>“, line 1, in <module>
  File “/home/dell/bluesky_project/bluesky/bluesky/__init__.py”, line 1, in <module>
    from .utils import Msg  # noqa: F401
  File “/home/dell/bluesky_project/bluesky/bluesky/utils/__init__.py”, line 32, in <module>
    from bluesky.protocols import (
  File “/home/dell/bluesky_project/bluesky/bluesky/protocols.py”, line 281, in <module>
    class Location(TypedDict, Generic[T]):
  File “/home/dell/anaconda3/envs/py3.9/lib/python3.9/typing.py”, line 1838, in __new__
    raise TypeError(‘cannot inherit from both a TypedDict type ’
TypeError: cannot inherit from both a TypedDict type and a non-TypedDict base class
```

## Analysis
https://stackoverflow.com/questions/71944154/variadic-generic-type-for-pythons-typeddict

## Idea for fix
This:
https://github.com/bluesky/bluesky/blob/37eb3e21cf136db7ddf1f6a063f373f06c229578/requirements.txt#L10
Should be <3.11 instead

This:
https://github.com/bluesky/bluesky/blob/37eb3e21cf136db7ddf1f6a063f373f06c229578/bluesky/protocols.py#L1-L4
Should import from `typing_extensions` if `sys.version_info < (3, 11)`

coretl:
This wasn't picked up by the tests because one of the dev-requirements has a dependency on typing_extensions... When we skeletonify this module it will run an import test that should catch this...

PR #1585, opened 2023-05-23
Use Python version check rather than import error check to import Protocol

callumforrester:
Use a softer import mode for protocols

## Motivation and Context
The `Protocol` class is available in `typing_extensions` until Python3.8, when it becomes part of `typing`.
Currently, we try to import from `typing_extensions` and, if there is an error, assume we are in Python3.8 or higher and import from `typing`. This presents a problem if you are using Python3.8+ and still using `typing_extensions` for another purpose, then the import from `typing_extensions` will succeed. This can be a problem if you wish to subclass one of the protocols. This PR enforces always using `typing` if the Python version is greater than 3.8.

## How Has This Been Tested?
Checked protocol still imports.
CI passes

### DIFF
--- a/bluesky/protocols.py
+++ b/bluesky/protocols.py
@@ -1,6 +1,8 @@
-try:
+import sys
+
+if sys.version_info < (3, 8):
     from typing_extensions import Literal, Protocol, TypedDict, runtime_checkable
-except ImportError:
+else:
     from typing import Literal, Protocol, TypedDict, runtime_checkable
 
 from asyncio import CancelledError
--- a/bluesky/protocols.py
+++ b/bluesky/protocols.py
@@ -1,18 +1,25 @@
-import sys
-
-if sys.version_info < (3, 8):
-    from typing_extensions import Literal, Protocol, runtime_checkable
-else:
-    from typing import Literal, Protocol, runtime_checkable
-
+from abc import abstractmethod
 from asyncio import CancelledError
 from typing import (
-    Any, Awaitable, Callable, Dict, Generic, Iterator, List, Optional,
-    Sequence, Tuple, Type, TypeVar, Union
+    Any,
+    Awaitable,
+    Callable,
+    Dict,
+    Generic,
+    Iterator,
+    List,
+    Literal,
+    Optional,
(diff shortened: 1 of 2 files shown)


tacaswell:
Python 3.8 is our current minimum so we can just drop the conditionals all together.

bluesky/protocols.py
```diff
@@ -1,15 +1,25 @@
-try:
-    from typing_extensions import Literal, Protocol, TypedDict, runtime_checkable
-except ImportError:
-    from typing import Literal, Protocol, TypedDict, runtime_checkable
```
tacaswell, line 4: How did this even work?!

tacaswell (dismissed):

bluesky/protocols.py
ksunden: https://github.com/bluesky/bluesky/blob/1b0e3bfc623405ef7ab422bc0a9e3036d57a5702/requirements.txt#L10

because of conditional dependencies... guaranteed to have typing_extensions for 3.8, but all available in 3.8+, perfectly fine to use the `typing_extensions` if exists, though

tacaswell: I must have typoed something when I was testing it earlier as I thought `from typing import TypedDict` failed 🐑 .

That leads to a different question as to why we are pulling `TypedDict` from `typeing_extensions` rather from `typing`.

callumforrester: The one from `typing` does seem to fail. I did a bit of investigation but can't figure out why. @coretl any ideas?

coretl: Cos of https://github.com/bluesky/bluesky/issues/1572

tacaswell:
I think we need to make the dependency on `typeing_extensions` non-version gated then?

coretl:
#1572 shows that we can use `typing.TypedDict` from python 3.11 upwards, but we need `typing_extensions.TypedDict` for <3.11. I don't mind either way whether we version gate so `typing_extensions` is not needed on 3.11. @tacaswell which do you prefer?

tacaswell (approved):

PR #1589, opened 2023-06-09
Tweak Status.exception() protocol

coretl:
## Description

Make it return `Optional[BaseException]` like `Future.exception()`

## Motivation and Context

So it's easier to make `ophyd.v2.core.AsyncStatus.exception` (as this is what `Task.exception()` returns)


coretl:
@tacaswell mind if I merge this?

PR #1590, opened 2023-06-19
Rethinking external assets

### LINKED ISSUE #1558: Add StreamDatum and StreamResource support
This is a reasonably large chunk of work to add support for what is produced in https://github.com/bluesky/event-model/issues/219 and https://github.com/bluesky/event-model/issues/236.

While we're at it, we can also tidy up the matching of data keys in `collect()` to descriptor ids:
https://github.com/bluesky/bluesky/blob/master/bluesky/bundlers.py#L710-L727

The proposed protocols are:
```python
# Like PartialEvent, but arrays
class PartialEventPage(TypedDict):
   data: Dict[str, Array]
   timestamps: Dict[str, Array]
   time: Array[float]

# These are actually tuples of name and dict, but for brevity...
Asset = Resource | Datum | StreamResource | StreamDatum

# If it writes external data, step or flyscan, it's one of these
class WritesExternalAssets(Protocol):
    def collect_asset_docs(self) -> Iterator[Asset]: ...

# If it can be step scanned and makes data, it's one of these, note that read might be empty if it only writes external data
class Readable(Protocol):
    def describe(self) -> Dict[str, DataKey]: ...    
    def read(self) -> Dict[str, Reading]: ...

# If it can be fly scanned and makes data, it's one of these
class Collectable(Protocol):
    # Support the old way of nesting with stream name, or the new way of getting stream name from collect message
    def describe_collect(self) -> Dict[str, DataKey] | Dict[str, Dict[str, DataKey]]: ...

# If it makes embedded fly scan data as single events, it's one of these
class EventCollectable(Collectable):
    def collect(self) -> Iterator[PartialEvent]: ...

# If it makes embedded fly scan data as event pages, it's one of these
class EventPageCollectable(Collectable):
    def collect_pages(self) -> Iterator[PartialEventPage]: ...
```
The changes are:
- Rename `Descriptor` to `DataKey`
- Add `StreamResource` and `StreamDatum` to `Asset`s that can be emitted from `collect_asset_docs`
- Split off `describe_collect` into `Collectable`
- Split off `collect` into `EventCollectable`
- Split off `collect_pages` into `EventPageCollectable`
- Add optional `name` argument to `collect` message, and use it instead of the stream name we used to get from `describe_collect`. Error if both are supplied and they do not match.
- Add optional `collect` argument to `declare_stream` message, and if set call `describe_collect` instead of `describe` on the Devices (also fix the docstring of `bps.declare_stream` while we're at it)
- Change anything that produces bulk events to produce event pages

The sequence for a step scan:
- `stage()`
- `describe_configuration()`, `read_configuration()`, `describe()` (either on first step, or in advance if pre-declared
- for each step:
  - `trigger()` if it exists
  - `read()`
  - `collect_asset_docs()` if it exists
- `unstage()`

The sequence for a fly scan:
- `stage()`
- `describe_configuration()`, `read_configuration()`, `describe_collect()` (either on first collect, or in advance if pre-declared
- `kickoff()`
- status = `complete()`
- while not status.done:
  - `collect()` if it exists
  - `collect_pages()` if it exists
  - `collect_asset_docts()` if it exists
- `unstage()`

This means that if we have a Device that supports step and flyscanning, the setup that is different goes in `trigger` and `kickoff`

evalott100:
Closes #1558


evalott100:
Tests pass locally using the correct branch of event model in `requirements.txt`.

bluesky/bundlers.py
```diff
@@ -620,18 +600,84 @@ async def complete(self, msg):
...
+            assert isinstance(should_be_a_dict_to_data_key, dict),\
+                "describe_collect() returned a Dict[str, SomeTypeNotADict], this would only be acceptable "\
+                "if a name argument was included in the collect Msg"
+
+            if should_be_a_dict_to_data_key:
+                should_be_a_data_key = should_be_a_dict_to_data_key[next(iter(should_be_a_dict_to_data_key))]
+                assert "source" in should_be_a_data_key and isinstance(should_be_a_data_key["source"], str),\
+                    "describe_collect() returned a Dict[str, Dict[str, SomeTypeNotADataKey]]"
+
+            return describe_collect_dict.items()
```
coretl, line 661: This logic didn't work for my use case of `Dict[str, DataKey]` with a single key and the message stream name given, it fired the first assert statement on line 622. This logic worked for me:

```suggestion
    def _maybe_format_datakeys_with_stream_name(
        self,
        describe_collect_dict: Union[Dict[str, DataKey], Dict[str, Dict[str, DataKey]]],
        message_stream_name: Optional[str] = None
    ) -> List[Tuple[str, Dict[str, DataKey]]]:
        """
        Check if the dictionary returned by describe collect is a dict
            `{str: DataKey}` or a `{str: {str: DataKey}}`.

        If a `message_stream_name` is passed then return a singleton list of the form of
            `{message_stream_name: describe_collect_dict}.items()`.
        If the `message_stream_name` is None then return the `describe_collect_dict.items()`.
        """
        def has_str_source(d: dict):
            return isinstance(d, dict) and isinstance(d.get("source", None), str)

        if describe_collect_dict:
            first_value = list(describe_collect_dict.values())[0]
            if has_str_source(first_value):
                # We have Dict[str, DataKey], so return just this
                # If stream name not given then default to "primary"
                return [(message_stream_name or "primary", describe_collect_dict)]
            elif all(has_str_source(v) for v in first_value.values()):
                # We have Dict[str, Dict[str, DataKey]] so return its items
                if message_stream_name and list(describe_collect_dict) != [message_stream_name]:
                    # The collect contained a name and describe_collect returned a Dict[str, Dict[str, DataKey]],
                    # this is only acceptable if the only key in the parent dict is message_stream_name
                    raise RuntimeError(
                        f"Expected a single stream {message_stream_name!r}, got {describe_collect_dict}"
                    )
                return list(describe_collect_dict.items())
            else:
                raise RuntimeError(
                    f"Invalid describe_collect return: {describe_collect_dict} when collect "
                    f"was called on {message_stream_name}"
                )
        else:
            # Empty dict, could be either but we don't care
            return []
```

bluesky/bundlers.py
```diff
@@ -176,12 +168,15 @@ async def _prepare_stream(self, desc_key, objs_read):
...
-        return descriptor_doc
+        self._descriptor_objs[desc_key] = objs_read
+        return self._descriptors[desc_key].descriptor_doc, self._descriptors[desc_key].compose_event, objs_read
 
-    async def _ensure_cached(self, obj):
+    async def _ensure_cached(self, obj, collect=False):
         if obj not in self._describe_cache:
-            await self._cache_describe(obj)
+            if collect:
+                await self._cache_describe_collect(obj)
```
coretl, line 177: We need to think about this one a bit more, `_cache_describe_collect` so `_ensure_cached` called twice will do describe collect twice

coretl: Also, `_prepare_stream` is called which takes data out of `_describe_cache` so I don't know how that works...

bluesky/protocols.py
```diff
@@ -191,7 +122,8 @@ def parent(self) -> Optional[Any]:
 class WritesExternalAssets(Protocol):
     @abstractmethod
     def collect_asset_docs(self) -> Iterator[Asset]:
```
coretl, line 124: Make this `SyncOrAsync` too

coretl: Actually, this doesn't quite work. `SyncOrAsync = Union[T, Awaitable[T]]`, but an `AsyncIterator` is not an `Awaitable[Iterator]`. I think we instead need:
```python
SyncOrAsyncIterator = Union[Iterator[T], AsyncIterator[T]]
...
    def collect_asset_docs(self) -> SyncOrAsyncIterator[Asset]:
```
The code that calls `collect_asset_docs` and friends will then need something different to `for doc in obj.collect_asset_docs()`. Something like:
```python
async def iterate_maybe_async(iterator: SyncOrAsyncIterator[T]) -> AsyncIterator[T]:
    if inspect.isasyncgen(ret):
        async for v in ret:
            yield v
    else:
        for v in ret:
            yield v
...
    async for doc in iterate_maybe_async(obj.collect_asset_docs()):
         ...
```
Similar for `collect` and `collect_pages`.

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -0,0 +1,433 @@
...
+    return iter([("resource", resource)])
+
+
+def collect_asset_docs_Datum(self) -> Iterator[Asset]:
+    datum = Datum(
+        datum_id="some_resource_name/123123",
+        resource="uid_of_that_resource",
+        datum_kwargs={"some_argument": 1234},
+    )
+    return iter([("datum", datum)])
```
coretl, line 48: ditto...

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -0,0 +1,433 @@
...
+        lambda *args: collector.append(args)
+       )
+    assert len(collector) == 5
+    assert collector[2][0] == asset_type
+
+
+def describe_with_name(self) -> SyncOrAsync[Dict[str, DataKey]]:
+    data_keys = {
+        str(x): DataKey(shape=[1], source="a", dtype="string")
+        for x in ["x", "y", "z"]
```
coretl, line 109: Please could we change these to be some more typical strings, like "det1" for datakeys, and "stream1" for stream names?

bluesky/bundlers.py
```diff
@@ -595,68 +626,293 @@ async def complete(self, msg):
...
+            elif name == DocumentNames.stream_datum.value:
+                if not stream_resource_data_keys:
+                    raise RuntimeError(
+                        "Received `stream_datum` without recieving any `stream_resource`"
+                    )
+                if not stream_datum_data_key_counter:
+                    # Create a counter of data_keys for `stream_datum` `indices` to be validated against
+                    counter_list = list(range(len(stream_resource_data_keys)))
+                    counter_list.reverse()
+                    stream_datum_data_key_counter = cycle(counter_list)
```
coretl, line 784: I'm not sure what this counter is doing. I was intending that the only counter we use is `self._sequence_counter`. When we pack external assets then we should check that all StreamDatums have the same indices_difference, then assign `StreamRange(self._sequence_counter[stream_name], self._sequence_counter[stream_name] + indices_difference)` to each of them, finally doing `self._sequence_counter[stream_name] += indices_difference`. (It might be more complicated than that because we should support producing Events at the same time, but you should still be able to use the same sequence_counters`). 

I might have misunderstoon the purpose of this counter though...

evalott100: It is to validate that the indices passed in are of equal width for all the data_keys:

collect is called then we have four `stream_resource`
`stream_resource_1 + stream_resource_2 + stream_resource_3 + stream_resource_4`

then the asset docs iterator will spit out `stream_datum` for each stream resource:
```
   stream_datum_1_1 + stream_datum_2_1 + stream_datum_3_1 + stream_datum_4_1
 + stream_datum_1_2 + stream_datum_2_2 + stream_datum_3_2 + stream_datum_4_2
 + ... 
 + stream_datum_1_n + stream_datum_2_n + stream_datum_3_n + stream_datum_4_n
```


`stream_datum_m_l` is the `l`-th stream_datum for the `m`-th data_key. This counter checks to see how many contiguous stream datum we have left for each data_key before we receive a new stream_datum for the first data key: We want to check the indices difference is the same for `stream_datum_m_l` and `stream_datum_(m-1)_l`, but then allow the next stream datum `stream_datum_1_(l+1)` to have a different indices difference.

coretl: Discussed in person. We can simplify because we decided `l` will be precisely one for every call to `collect_asset_docs`, so we can check the lengths are equal for all stream_datums and assign them seq_nums accordingly

evalott100: Never mind, this was for the old plan.

bluesky/bundlers.py
```diff
@@ -597,68 +631,274 @@ async def complete(self, msg):
...
+        stream_datum_previous_indices_difference = 0
+        number_of_stream_datum_receieved = 0
+
+        for name, doc in asset_docs:
+            if name == DocumentNames.resource.value:
+                doc["run_start"] = self._run_start_uid
+            elif name == DocumentNames.stream_resource.value:
+                if number_of_stream_datum_receieved:
+                    raise RuntimeError("Received a `stream_resource` document after `stream_datum` documents.")
+
```
coretl, line 760: This is valid for the case of a flyer that wraps 2 detectors

coretl: Instead we should cache in self the stream_resource by uid, then check when we get a stream_datum that it refers to a stream_resource that exists

bluesky/bundlers.py
```diff
@@ -597,68 +631,274 @@ async def complete(self, msg):
...
+                    name,
+                    doc,
+                    extra={
+                        "doc_name": name,
+                        "run_uid": self._run_start_uid,
+                        "doc": doc
+                    }
+                )
+
+        return stream_datum_previous_indices_difference
```
coretl, line 804: Before we return we should check that we have received a StreamDatum for every `STREAM:` data_key in the descriptor

bluesky/bundlers.py
```diff
@@ -597,68 +631,274 @@ async def complete(self, msg):
...
-            Msg('collect', flyer_object, stream=True, return_payload=False)
+            Msg('collect', flyer_object, stream=True, return_payload=False, name='stream_name')
         """
-        collect_obj = check_supports(msg.obj, Flyable)
+
+        collect_obj = check_supports(msg.obj, Collectable)
+        assert not (isinstance(collect_obj, EventCollectable) and isinstance(collect_obj, EventPageCollectable)), \
+            "collect() was called for a device which is both EventCollectable and EventPageCollectable. "\
+            "If you want to have an EventCollectable device format only some events as event_pages "\
+            "then use `stream=False`."
```
coretl, line 901: Should warn instead of assert, then prefer pages

bluesky/tests/test_external_assets_and_paging.py
coretl, line 1: I think the set of tests looks about right, but I'd like to see the results tested. Doesn't have to be exhaustive, but something that checks we get an Event with 3 keys in it, then StreamResource and StreamDatum for the other 2 detectors. I did something similar here:
https://github.com/bluesky/ophyd-async/blob/c33b9e7e5f889233a05a327612d8e900e53d05aa/tests/epics/areadetector/test_hdf_streamer_det.py#L68C16-L68C16

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -0,0 +1,433 @@
...
+def collect_asset_docs_Datum(self) -> Iterator[Asset]:
+    datum = Datum(
+        datum_id="some_resource_name/123123",
+        resource="uid_of_that_resource",
+        datum_kwargs={"some_argument": 1234},
+    )
+    yield "datum", datum
+
+
+def collect_asset_docs_StreamDatum(self) -> Iterator[Asset]:
```
coretl, line 51: Please can we also add a function that gives 2 StreamResources and StreamDatums. This is what a flyer will do, proxying Stream documents from 2 detectors. Or even a single detector making 2 datasets

PR #1591, opened 2023-06-21
FIX: do not leak public token maps

tacaswell:
We were not removing the old public tokens on subscribe so the internal state would grow without bounds.

## Description

@dmgav Noticed that the internal mapping between the public and private subscription tokens was never reduced in size

## Motivation and Context

We should not have unbounded caches!

## How Has This Been Tested?

Added a test.


untzag (approved):
Gosh seems like a great fix to me, thanks @dmgav @tacaswell

dmgav (approved):

PR #1592, opened 2023-06-22
DOC clarify how to exit from a plan

prjemian:
Sometimes during execution of a plan by the RunEngine, it is deemed essential that the Bluesky session must be terminated immediately in an orderly manner.  Describe these steps clearly.

## Description
Documentation was added to provide a plan that will stop the session both immediately and orderly, under the control of the RunEngine.

## Motivation and Context
APS beam lines (tomo, USAXS, depo lab) have requested instructions how to make an emergency stop of the entire session in response to some monitored conditions.  Previously, the `clear_checkpoint()` plan was added in response.  

The documentation of how to finish the task after calling `yield from clear_checkpoint()` was vague.

## How Has This Been Tested?

```py
"""
demo_orderly_exit.py

Demonstrate how to STOP and exit Python from a Bluesky suspender.

1. Set BIT_PV set to 0 before starting this code.
2. Start this code.
3. From a separate EPICS client, change BIT_PV to 1.  
   This will activate the suspender.  Subsequently, the Python session will exit
   with an exception trace.
   After the trace, the ``atexit`` handler is called.  That's where to put any
   final instructions.
"""

from bluesky import RunEngine
from bluesky import plan_stubs as bps
from bluesky.suspenders import *
from ophyd import EpicsSignalRO
import atexit

BIT_PV = "gp:gp:bit1"  # Some bo PV
RE = RunEngine()


def pre_response(*args, **kwargs):
    print(f"pre_response(): {args=}  {kwargs=}")
    print("request clear_checkpoint()")
    yield from bps.clear_checkpoint()
    print(f"pre pause(): {RE.state=}")
    yield from bps.pause()
    print(f"post pause(): {RE.state=}")


def exit_notice(*args, **kwargs):
    print("Python session has ended.  Ignore the errors.  EXPLAIN further...")


bit = EpicsSignalRO(BIT_PV, name="bit")
bit.wait_for_connection()
suspender = SuspendBoolHigh(
    bit, pre_plan=pre_response, tripped_message="This suspender will not resume."
)
RE.install_suspender(suspender)
atexit.register(exit_notice)


def loop_forever():
    "a silly plan"
    while True:
        print(f"Hi")
        yield from bps.sleep(1)


# Execute the plan.
RE(loop_forever())
```

### DIFF
--- a/docs/source/state-machine.rst
+++ b/docs/source/state-machine.rst
@@ -332,6 +332,19 @@ abort instead.
     actions that it is not safe to replay: staging a device, adding a
     monitor, or adding a subscription.
 
+.. caution::
+
+    This plan will exit the entire Python session immediately:
+
+    .. code-block:: python
+
+        import bluesky.plan_stubs as bps
+
+        def exit_python_session_now():
+            """Exit this Python session immediately."""
+            yield from bps.clear_checkpoint()  # this plan will not resume
+            yield from bps.pause()  # trigger the Python session to exit
+
 .. _planned_pauses:
 
 Planned Pauses
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -732,7 +732,8 @@ class RunEngine:
         Command the Run Engine to pause.
 
         This function is called by 'pause' Messages. It can also be called
-        by other threads. It cannot be called on the main thread during a run,
+        by other threads. It cannot be called on the main thread during a run
+        (in a plan, use ``yield from bps.pause()``),
         but it is called by SIGINT (i.e., Ctrl+C).
 
         If there current run has no checkpoint (via the 'clear_checkpoint'


prjemian:
I suggest the documentation provide this example (since the existing docs led me astray, with a call to `RE.pause()` from the plan).

untzag:
Thanks so much for providing this example @prjemian, this is EXACTLY the functionality I was searching for and I was just about to bother everyone on slack about it!

untzag:
I agree you should add a more complete example about integrating these messages with suspenders. I don't think the average reader will be able to figure it out on their own.

docs/source/state-machine.rst
```diff
@@ -332,6 +332,19 @@ abort instead.
     actions that it is not safe to replay: staging a device, adding a
     monitor, or adding a subscription.
 
+.. caution::
+
+    This plan will exit the entire Python session immediately:
```
tacaswell, line 337: This puts the RE in a paused state, "exit the entire Python session" to me reads as `sys.exit(0)` which is not what this is doing....

tacaswell:
I think this suggests we need to add `stop` and `abort` messages to the RE such that you can reliably stop any further messages from being processed.

Another alternative would be to add an `Exception` sub-class like `class ExitRunEngine(Exception):...` which we could catch in the RE and not re-raise out of the RE but would cause all of the plans to be (semi-gracefully) exited.

tacaswell:
This should also be distinguished from returning early which will exit the _local_ plan, but if it is part of a bigger plan the outer plan will continue on.

tacaswell:
(sorry for closing, fat fingers on my keyboard)

docs/source/state-machine.rst
prjemian: That's a more precise statement.  Yet isn't the immediate _consequence_ that the Python process is terminated?

tacaswell: I would expect that this would give you the IPython prompt back?

tacaswell:
Yes, but I am not 100% sure that is the best place for this content to go.

stan-dot:
@coretl , @callumforrester what are the plan - exiting requirements that we know to be at DLS use cases?

PR #1593, opened 2023-07-05
FIX: make sure declared but empty streams report 0 length

tacaswell:
## Description

If the stream is declared (via pre-declare) but never populated report that it has 0 events in the stop documents (rather than just memory-holing it and not including an entry).

## Motivation and Context

If there is a descriptor doc it should show up in the list of streams in the stop document.

## How Has This Been Tested?

Added test (and manually tested).

### DIFF
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -109,7 +109,7 @@ class RunBundler:
         num_events = {}
         for bundle_name, counter in self._sequence_counters.items():
             if bundle_name is None:
-                # rare but possible via Msg('create', name='primary')
+                # rare but possible via Msg('create', name=None)
                 continue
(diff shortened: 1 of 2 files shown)


coretl:
Can this be closed when #1579 is merged? It has the same effects on the tests:
https://github.com/bluesky/bluesky/pull/1579/files#diff-a1803e5fd63150c94a4de264f4373c56e398446dfc495d9b3d27bac36dc88afcL1595

tacaswell:
Yes, I think that this will be made redunedent by #1579

stan-dot:
#1579  can we close this, @tacaswell and @coretl ?

tacaswell:
Yes.  Thank you for doing the leg work on this @stan-dot

PR #1599, opened 2023-07-25
ENH: Allow the contingency wrapper to eat exceptions

DominicOram:
## Description
Add an option to contingency_wrapper to not automatically re-raise if the except plan returns a value rather than raising its own exception.

We have to do this with a flag like this for back-compatibility reasons

## Motivation and Context
We made a design mistake with contingency_wrapper because there is no way to actually suppress the exception. The reason we wrote this wrapper (and finalize wrapper) is to keep users from having to understand the extra set of constraints around the cancel and generator exit exceptions so this omission defeats the point.

## How Has This Been Tested?
Added unit tests that should cover all functionality. The first unit test added ensures backwards compatibility and also passes on `master`

## Note
Built on original PR https://github.com/bluesky/bluesky/pull/1526

### DIFF
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -535,7 +535,8 @@ def contingency_wrapper(plan, *,
                         except_plan=None,
                         else_plan=None,
                         final_plan=None,
-                        pause_for_debug=False):
+                        pause_for_debug=False,
+                        auto_raise=True):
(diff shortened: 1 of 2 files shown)


untzag (approved):
I didn't know I wanted this, but now I need it :heart:

tacaswell:
Thank you for finishing this.

PR #1601, opened 2023-08-01
Add child protocol

RAYemelyanova:
It has become useful in ophyd.v2 to add a `children` property to devices.

To remain consistent with the `parent` property, it was suggested that adding a protocol like `HasChildren`, analogous to `HasParent`, would be useful

## Description
A PR has been started about this for ophyd.v2. See here: https://github.com/bluesky/ophyd/pull/1138

## How Has This Been Tested?
I am unsure how protocols are tested, please advise and I will add tests.

### DIFF
--- a/bluesky/protocols.py
+++ b/bluesky/protocols.py
@@ -3,18 +3,29 @@ try:
 except ImportError:
     from typing import Literal, Protocol, TypedDict, runtime_checkable
 
+from abc import abstractmethod
 from asyncio import CancelledError
 from typing import (
-    Any, Awaitable, Callable, Dict, Generic, Iterator, List, Optional,
-    Sequence, Tuple, Type, TypeVar, Union
+    Any,
+    Awaitable,
+    Callable,
+    Dict,
+    Generic,
+    Iterator,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
 )
-
-from abc import abstractmethod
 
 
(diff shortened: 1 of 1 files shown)


RAYemelyanova:
not sure why lint is failing, seems to be complaining about something my changes have not altered.

untzag:
@RAYemelyanova consider adding some tests

bluesky/tests/test_protocols.py

Also can this be an `@property` so that instance attributes are valid?

untzag:
For what it's worth this isn't the implementation I happened to go with for `yaqc-bluesky`, I just have a list of child objects. I don't know how important the concept of a device name is to bluesky, so I don't really know why we want tuples of (name, device).

That said I'm all for standards and my devices do have names so I'd be happy to change yaqc-bluesky to match this protocol once merged.

We could look at e.g. instrbuilder to see about other implementations.

stan-dot:
@coretl this would rather belong in https://github.com/bluesky/ophyd-async , right?

I am not sure in which situation we'd like to know about a device that it `HasParent`. This risks more coupling between the devices

coretl:
Here is the current `ophyd-async` implementation:
https://github.com/bluesky/ophyd-async/blob/8273a2e0980fc586db38abb15fc6d354e0d40e68/src/ophyd_async/core/_device.py#L139-L142

 The name part of the iterator was needed so that you can use this method to:
- connect child devices, reporting which named child failed to connect
- name the child device with something that included the attribute name

stan-dot:
@coretl I do not understand this comment

PR #1605, opened 2023-08-23
Add metadata to descriptor document

RAYemelyanova:
Relates to https://github.com/bluesky/event-model/pull/286

Currently the DAQ-Core team in diamond is needing access to custom field setting on descriptor documents. This is because, for example, tomography scans should output to several streams, and we want our tools to read descriptor documents and be aware of the nexus application definitions that the data from those streams should be written in.

### DIFF
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -127,7 +127,24 @@ class RunBundler:
         self.run_is_open = False
         return doc["run_start"]
 
-    async def _prepare_stream(self, desc_key, objs_read):
+    async def _prepare_stream(self, desc_key, objs_read, md=None):
+        """Generate and emit a new descriptor document for a stream.
+
+        Parameters
+        ----------
+        desc_key : string
+            The name of the stream
+        objs_read : Iterable of objects
+            Readable objects which will be used in the stream, which will be
+            collected into the descriptor document.
+        md: optional dict of metadata
+            Dictionary of metadata to be passed to the descriptor document.
+
+        Returns
+        -------
+        descriptor_doc : dict
+            The descriptor document for the stream
+        """
         # We do not have an Event Descriptor for this set
         # so one must be created.
         data_keys = {}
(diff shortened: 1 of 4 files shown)


danielballan:
This one will need some careful thought. Getting toward end-of-day in US/Eastern, but I'll leave this tab open to follow up. Might need discussion at the next meeting.

RAYemelyanova:
linting failing due to https://github.com/bluesky/bluesky/issues/1606, which is currently being addressed in another PR.

bluesky/bundlers.py
```diff
@@ -416,6 +434,7 @@ async def save(self, msg):
         Expected message object is::
 
             Msg('save')
+            Msg('save', {"key": "value"})
```
DiamondJoseph, line 437: This should be Msg('save', key=value), else {"key": "value"} is going to be obj (the only positional argument) not kwargs.

danielballan:
That makes sense.

In early days (~2015) we intentionally left this open at the schema level in case a use case arose, but made it inaccessible from plans until we found a use case. In 8 years, this is the first use case. Maybe it's time to open the doors. There are two things we should think through:

1. Is this definitely superior to every other way handling this use case that we can think of?
2. Are downstream tools (visualizaiton and stuff) definitely not accidentally going to break when keys are added? They shouldn't but we should verify to avoid surprises.

DiamondJoseph:
Decided to place this metadata on StartDocument, with some mapping object. Assumption is that enough information available at run start to resolve onto any eventual stream creations (even if creating streams proceedurally).

danielballan:
Elaborating on some reasoning, in case we have a reason to revisit this in the future...

* The process/service consuming the documents will need to hold on to some state either way (an open file to write into). Starting that state with RunStart rather than EventDescriptor does not change much.
* In the broader context of scientific computing, attaching metadata to columns of a table is a somewhat niche thing to do, and expanding our use of that feels less "standard" than simply putting more things int our top-level JSON bundle.
* We typically know what we need to know at RunStart time, or at least we can describe the range of things we _might_ need to know. If we run into a more adaptive/dynamic use case, that would be the time to revisit this.

PR #1607, opened 2023-08-24
Simple fixes to placate flake8

RAYemelyanova:
Fixes https://github.com/bluesky/bluesky/issues/1606


danielballan:
To elaborate, the reasoning is that if future changes make more fine-grained exceptions, as long as they inherit from the accepted one, that should not be treated as a backward-incompatible change by the test suite.

danielballan (approved):

danielballan:
Thanks for taking care of this upkeep @RAYemelyanova.

PR #1610, opened 2023-08-30
Add type annotations to plans and plan stubs, update docstrings

callumforrester:
## Description
Add type annotations to plans and plan stubs. IDEs such as vscode should now provide hints when using them. See screenshot below. Fix and update out-of-date docstrings on plans and plan stubs.

## Motivation and Context
Makes development with plans easier with auto-fill etc. Also allows downstream projects to make use of static type checking utilities such as mypy.

## How Has This Been Tested?
Docstrings and type annotations should not semantically change the code. All tests passing supports this.

## Screenshots (if appropriate):
image
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ dev = [
     "mypy",
     "myst-parser",
     "networkx",
-    "numpydoc",
+    "numpydoc>=1.8.0rc1",
     "ophyd",
     "packaging",
     "pandas",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -180,6 +180,10 @@ lint.select = [
     "I",  # isort - https://docs.astral.sh/ruff/rules/#isort-i
     "UP", # pyupgrade - https://docs.astral.sh/ruff/rules/#pyupgrade-up
 ]
+lint.ignore = [
+    "C408",  # dict(...) is fine to use
+    "C409",  # tuple(...) is fine to use
+]
 exclude = [
     "docs/source/conf.py",
     "docs/source/examples",
--- a/src/bluesky/plan_stubs.py
+++ b/src/bluesky/plan_stubs.py
@@ -32,6 +32,7 @@ except ImportError:
     from toolz import partition
 
 from .protocols import (
+    check_supports,
     Configurable,
     Flyable,
     Locatable,
@@ -640,7 +641,12 @@ def sleep(time: float) -> MsgGenerator:
 
 
 @plan
-def wait(group: Optional[Hashable] = None, *, timeout: Optional[float] = None):
+def wait(
+    group: Optional[Hashable] = None,
+    *,
+    timeout: Optional[float] = None,
+    move_on: bool = False,
+):
     """
     Wait for all statuses in a group to report being finished.
 
@@ -986,9 +992,10 @@ def complete_all(*args, group=None, wait=False, **kwargs):
 @plan
 def collect(
     obj: Flyable,
-    *,
+    *args,
     stream: bool = False,
     return_payload: bool = True,
+    name: Optional[str] = None
 ) -> MsgGenerator[List[PartialEvent]]:
     """
     Collect data cached by one or more fly-scanning devices and emit documents.
@@ -1019,7 +1026,7 @@ def collect(
     :func:`bluesky.plan_stubs.complete`
     :func:`bluesky.plan_stubs.wait`
     """
-    return (yield Msg("collect", obj, stream=stream, return_payload=return_payload))
+    return (yield Msg("collect", obj, *args, stream=stream, return_payload=return_payload, name=name))
 
 
 @plan
--- a/src/bluesky/plan_stubs.py
+++ b/src/bluesky/plan_stubs.py
@@ -58,7 +58,7 @@ from .utils import (
 from .utils import (
     short_uid as _short_uid,
     MsgGenerator,
-    ScalarOrIterable,
+    ScalarOrIterableFloat,
 )
 
 from event_model import ComposeEvent
@@ -1771,7 +1771,7 @@ def one_nd_step(
 def repeat(
     plan: Callable[[], MsgGenerator],
     num: int = 1,
-    delay: Optional[ScalarOrIterable] = None,
+    delay: Optional[ScalarOrIterableFloat] = None,
 ) -> MsgGenerator[Any]:
     """
     Repeat a plan num times with delay and checkpoint between each repeat.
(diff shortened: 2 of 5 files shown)


bluesky/plans.py
```diff
@@ -16,13 +18,41 @@
...
+        Mapping[Movable, Any],
+        Mapping[Movable, Any],
+        Optional[bps.TakeReading],
+    ],
+    MsgGenerator,
+]
+
+
+def count(
+    detectors: List[Readable],
```
dmgav, line 49: This should probably be `Union[List[Readable], Tuple[Readable]]` for generality.

bluesky/plans.py
```diff
@@ -16,13 +18,41 @@
...
+    num: Optional[int] = 1,
```
dmgav, line 50: The type should probably be simply `int`. `Optional[int]` assumes that the parameter accepts `None`, which is not true here.

bluesky/plans.py
```diff
@@ -1102,24 +1307,29 @@ def scan(detectors, *args, num=None, per_step=None, md=None):
...
     full_cycler = plan_patterns.inner_product(num=num, args=args)
 
-    return (yield from scan_nd(detectors, full_cycler,
-                               per_step=per_step, md=_md))
+    return (yield from scan_nd(detectors, full_cycler, per_step=per_step, md=_md))
 
 
-def grid_scan(detectors, *args, snake_axes=None, per_step=None, md=None):
+def grid_scan(
+    detectors,
```
dmgav, line 1327: Missing type.

bluesky/plans.py
callumforrester: Count does accept `num=None`, see line 84 (on this branch)
```python
    if num is None:
        num_intervals = None
```

bluesky/plans.py
callumforrester: Good point, I think `Sequence` works as a supertype of both. Would `Sequence[Readable]` work with Queue Server validation? It's much more concise.

bluesky/plans.py
callumforrester: Good spot! Will replace based on outcome of https://github.com/bluesky/bluesky/pull/1610/files/7ebbfcf3408f15d11ece2f847a70b198103aef3d#r1310969275

bluesky/plans.py
dmgav: I made some changes to the Queue Server. Using `Iterable` is good now.

bluesky/utils/__init__.py
```diff
@@ -61,6 +61,19 @@ def __repr__(self):
...
+P = TypeVar("P")
+
+#: Object usually returned from plan functions that is fed to the RunEngine
+MsgGenerator = Generator[Msg, Any, Optional[P]]
+
+#: Metadata passed from a plan to the RunEngine for embedding in a start document
+CustomPlanMetadata = Dict[str, Any]
+
+#: Scalar or iterable of values, one to be applied to each point in a scan
+ScalarOrIterable = Union[float, Iterable[float]]
```
dmgav, line 74: Should it be `ScalarOrIterableFloat` or `ScalarOrIterableNumber` for clarity?

bluesky/plans.py
```diff
@@ -133,55 +172,66 @@ def list_scan(detectors, *args, per_step=None, md=None):
...
 
-    return (yield from scan_nd(detectors, full_cycler, per_step=per_step,
-                               md=_md))
+    return (yield from scan_nd(detectors, full_cycler, per_step=per_step, md=_md))
 
 
-def rel_list_scan(detectors, *args, per_step=None, md=None):
+def rel_list_scan(
+    detectors: List[Readable],
+    *args: Any,
```
dmgav, line 231: Here is another issue with the Queue Server, since it will not attempt to convert strings to objects (motors in this particular example) if parameter is annotated with a type not associated with Ophyd devices (such as `Readable`, `Movable` etc). So in this case `*args` should have no annotation (by default the queue server tries to convert all strings to objects) or the type should be something like `Union[Movable, Any]`. It should also work with mypy.

bluesky/plans.py
callumforrester: It doesn't support `Sequence` then?

dmgav: It does.

callumforrester:
@dmgav as an aside in terms of types the queueserver supports, I've been using the types supported by pydantic as a guide: https://docs.pydantic.dev/1.10/usage/types/, it provides a useful rule of thumb.

dmgav:
Unfortunately, Pydantic functionality is insufficient for proper type checking. The problem with `Iterable`, for example, is that if the parameter type is `Iterable[str]`, Pydantic is absolutely happy with simply a string value. It also accepts a number represented as a string for parameters instead of `float` or `int`. Pydatic v1 is even less strict. The problem is not to make the queue server 'accept' the parameter values, but to detect mismatch of types and decide which values has to be converted to objects. There is some post-processing for this.

bluesky/plans.py
danielballan: I'm generally in favor of "Accept lazy iterables (generators) everywhere!" but I think that it's pretty reasonable to expect a `Sequence` here.

callumforrester: I think make it `Sequence` here to reflect what it will accept and make a separate issue for making all plans accept lazy iterables. https://github.com/bluesky/bluesky/issues/1613

bluesky/plans.py
callumforrester: Interesting, for blueapi we went the opposite way. Only attempts to convert strings if they are annotated with protocol types or things that follow protocol types.

callumforrester:
@danielballan Could make a custom type but I wouldn't like to hide away the `Optional`, it seems perfectly reasonable to write a plan that takes a group that cannot be none, so it would just be `Optional[Group]` for the current plans rather than `Optional[Hashable]`. If that sounds useful I'm happy to add it in.

As an aside, I note that `wait` takes `None` as the default for `group`. Do you know why that is? It seems like a perfect candidate for requiring a value because it's just a noop with `None`, right?

bluesky/plans.py
dmgav: This is what the Queue Server does. But it also attempts to convert all strings it finds in values of parameters that are not annotated. This may lead to errors, since any string matching a plan or a device name will be converted to object, but this never happened in practice. Unfortunately, our local users do not want to learn about type annotations, which makes this feature very useful.

dmgav: So the parameter annotated with `typing.Any` is considered having type and the Queue Server does not try to convert strings to objects.

tacaswell:
I did the rebase and pinned numpydoc forward the their rc that support `returns` section in generator docstrings.

callumforrester:
@dmgav I agree, let's close this, it's quite out of date and also superseded by https://github.com/bluesky/bluesky/pull/1712

PR #1615, opened 2023-09-13
Added run_start_uid for stream_resource

### LINKED ISSUE #1614: Bundler doesn't pack the run start uid into a stream resource
In the code here

https://github.com/bluesky/bluesky/blob/8deba753a63de84a76b1a9aceb9edf3ae3268e1c/bluesky/bundlers.py#L466-L467

We add the run start uid into the resource document. We should do this if it's a stream resource as well (this is already assumed in event-model, the bundler just hasn't implemented it).

Additionally, the bundler should be made to check against `DocumentNames` when checking document type

https://github.com/bluesky/bluesky/blob/8deba753a63de84a76b1a9aceb9edf3ae3268e1c/bluesky/bundlers.py#L672

should become `== DocumentNames.resource`

evalott100:
Closes #1614

### DIFF
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -463,7 +463,9 @@ class RunBundler:
         for resource_or_datum_name, resource_or_datum_doc in self._asset_docs_cache:
             # Add a 'run_start' field to resource documents on their way out
             # since this field could not have been set correctly before this point.
-            if resource_or_datum_name == "resource":
+            if resource_or_datum_name in (
+                DocumentNames.resource, DocumentNames.stream_resource
+            ):
                 resource_or_datum_doc["run_start"] = self._run_start_uid
 
             doc_logger.debug(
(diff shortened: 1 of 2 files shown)


bluesky/bundlers.py
```diff
@@ -463,7 +463,9 @@ async def save(self, msg):
         for resource_or_datum_name, resource_or_datum_doc in self._asset_docs_cache:
             # Add a 'run_start' field to resource documents on their way out
             # since this field could not have been set correctly before this point.
-            if resource_or_datum_name == "resource":
+            if resource_or_datum_name in [
+                DocumentNames.resource or resource_or_datum_name == DocumentNames.stream_resource
+            ]:
```
DiamondJoseph, line 468:
```suggestion
            if resource_or_datum_name in (
                DocumentNames.resource, DocumentNames.stream_resource
            ):
```

bluesky/bundlers.py
```diff
@@ -669,7 +671,9 @@ async def collect(self, msg):
         # Resource and Datum documents
         for name, doc in maybe_collect_asset_docs(msg, collect_obj):
             # Add a 'run_start' field to the resource document on its way out.
-            if name == "resource":
+            if name in [
+                DocumentNames.resource or name == DocumentNames.stream_resource
+            ]:
```
DiamondJoseph, line 676:
```suggestion
            if name in (
                DocumentNames.resource, DocumentNames.stream_resource
            ):
```

DiamondJoseph (commented):
Looks like you accidentally munged your two approaches together

DiamondJoseph (approved):
LGTM

PR #1619, opened 2023-10-09
Add new flyable plan

rosesyrett:
## Description

I am currently making some changes to ophyd-async to allow an experiment around the middle of November to take place using it. This requires an updated flyable plan.

## How Has This Been Tested?

Not tested yet

@coretl I think it's worth explaining how the flyable plan added here is different from the existing one, and why we don't just modify the existing one instead.

### DIFF
--- a/bluesky/plans.py
+++ b/bluesky/plans.py
@@ -16,7 +16,7 @@ except ImportError:
 from . import plan_patterns
 
 from . import utils
-from .utils import Msg, get_hinted_fields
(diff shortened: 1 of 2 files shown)


coretl:
This is for a single Flyable that is collected while it is completing. I think we will convert this to draft while we discuss it

stan-dot:
@coretl is this relevant anymore? I assume this is in the context of the linkam plan?

coretl:
Superceded by #1641

PR #1620, opened 2023-10-10
Wait for async staging in stage_wrapper

coretl:
## Description

Use stage_all and unstage_all to wait for staging to complete when stage returns a Status rather than blocking

## Motivation and Context

When we added async stage we didn't update all the plan preprocessors to `Msg('wait', group=g)` after a series of`Msg(`stage`, group=g)`. This is the cheapest change, although we will have to also fix `lazy_stage_wrapper` in another PR, and there may be other places emitting `Msg('stage')` directly rather than using `bps.stage()`. 

Is this correct? Or should we make the run engine wait for stage/unstage always unless it gets a group in the stage message?

## How Has This Been Tested?
Locally, no additional tests written yet

### DIFF
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -919,8 +919,7 @@ def lazily_stage_wrapper(plan):
             return None, None
 
     def unstage_all():
-        for device in reversed(devices_staged):
-            yield Msg('unstage', device)
+        yield from unstage_all(*reversed(devices_staged))
 
     return (yield from finalize_wrapper(plan_mutator(plan, inner),
                                         unstage_all()))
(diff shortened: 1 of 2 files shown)


coretl:
Need to add this for other `Msg("stage")` and `unstage` instances

rosesyrett:
The only other instance of Msg("stage") or Msg("unstage") in preprocessors says it's a hack so, I kind of don't want to touch it. I feel we need a separate PR to avoid scope creep to fix that issue.

coretl:
> I am surprised that `stage_all` and `unstage_all` do not return the union of the returns of the things that they staged.
> 
> Did we decide to abandon the returns from stage/unstage?

I thought that was the decision? I think as we can't return a result from a `Status` we decided to deprecate the return value. As `stage_all` and `unstage_all` are new functions that are written primarily for `Status` returning functions, we didn't bother including the optional return types for old-style stage functions.

Is that ok?

tacaswell (approved):

PR #1623, opened 2023-10-20
Remove group from groups after Msg('wait_for'), not before, in RunEngine._wait

rosesyrett:
RunEngine._wait now waits until wait_for message has been produced before removing futures from groups/status objects

### DIFF
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -46,6 +46,10 @@ class _RunEnginePanic(Exception):
     ...
 
 
+class WaitForTimeoutError(TimeoutError):
+    ...
+
+
 @dataclass
 class RunEngineResult():
     """
@@ -1738,7 +1742,7 @@ class RunEngine:
         futs = [asyncio.ensure_future(f()) for f in futs]
         _, pending = await asyncio.wait(futs, **self._loop_for_kwargs, **msg.kwargs)
         if pending:
-            raise TimeoutError("Plan failed to complete in the specified time")
+            raise WaitForTimeoutError("Plan failed to complete in the specified time")
 
     async def _open_run(self, msg):
         """Instruct the RunEngine to start a new "run"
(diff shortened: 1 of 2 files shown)


DiamondJoseph (approved):

tacaswell:
I think that if the wait for raises we will not clear the state?

rosesyrett:
I would like to add a test to this; however, that is currently non-trivial, because it would require adding a dependency on ophyd-async, because the best way of testing this is doing something along the lines of:

```python
def test_wait_propagates_error():
    from ophyd_async.core import Device

    class DummyDevice(Device):
        async def set(self, value:int):
            await asyncio.sleep(0.01)
            raise RuntimeError()
    dummy = DummyDevice("somename")
    RE(abs_set(dummy, 1))
```

This way, when abs set immediately sets it, there is no exception, so that line passes. But when we get into the run_engine._wait, the error is caught and reported on (instead of silently failing, which is what we found was happening...)

So the question is, do we want to rely on ophyd-async?

coretl:
How about we make it so that we pop unless we get a `TimeoutError`?

tacaswell:
There is a difference if the wait times out or if the Status object times out and fail itself.

coretl:
Ok, how about this:

- We make a new `class WaitForTimeoutError(TimeoutError):`
- In: https://github.com/bluesky/bluesky/blob/44579d5201280bcfee20574a72fa063bf7d0160a/bluesky/run_engine.py#L1741C25-L1741C25 we raise this
- In https://github.com/bluesky/bluesky/blob/44579d5201280bcfee20574a72fa063bf7d0160a/bluesky/run_engine.py#L2205 the only time we don't pop is on `WaitForTimeoutError`

tacaswell:
I miss-remembered how the raising for failed status worked https://github.com/bluesky/bluesky/blob/44579d5201280bcfee20574a72fa063bf7d0160a/bluesky/run_engine.py#L2214-L2235

so the sad-path on the status object failing itself is the happy-path here so that is not a problem.

However, a second thing I thought of here is `CancelError` s being through in by the event loop (for example if the user pauses while waiting).  

Can we go with the `TimeoutError` subclass and then if we catch that put the futures back?  You could argue that on pause is some other part of the code's job to clean up all of the stale status objects, but I think it is clearer if `_wait` cleans up after itself and "re-arms" the group on a plan-side timeout.

coretl:
> Can we go with the TimeoutError subclass and then if we catch that put the futures back? 

Sounds good, I'll update the PR

tacaswell (approved):

Issue #1625, opened 2023-10-27
Preprocessors behave unexpectedly with async logic/ophyd-async devices


rosesyrett:
As part of Diamond's I22-Tomo project, I have been trying to debug an issue we have seen with blueapi, where if we try to run a bp.count plan via blueapi with a ophyd-async device the run engine seems to hang on a 'wait' message. I have now discovered that the reason why it hangs is a fundamental problem with the way the run engine processes sequences/generators of messages, which is only cropping up now because we have async stage/trigger methods on detectors that don't block when they get called by the run engine.

## Minimal reproducible example

It is quite hard to write a minimal example of this issue that works, so I am going to preface this by saying, this example will only work if run inside Diamond's network, specifically to access the d11 detector for our test beamline, p38. Regardless the main points can still be demonstrated here:

```python
from bluesky.run_engine import RunEngine
from ophyd_async.core import DeviceCollector, StaticDirectoryProvider
from dodal.devices.areadetector.adaravis import SumHDFAravisDetector
from bluesky.plans as bp

RE = RunEngine()

with DeviceCollector():
    d11_det = SumHDFAravisDetector("BL38P-DI-DCAM-03", directory_provider=StaticDirectoryProvider("/some/path/a/hdf/plugin/can/write/to", "tmp"))
    
def with_yield_from():
    yield from bp.count([d11_det], num=1)

def with_loop():
    plan = bp.count([d11_det], num=1)
    for msg in plan:
        yield msg

RE(with_yield_from()) # will pass
RE(with_loop()) # will hang forever
```

listing both of these plans (i.e. `list(with_yield_from())`, `list(with_loop())`  ) reveals that the messages produced by both are exactly the same. However, running the plan which loops through the messages in an existing plan before yielding them seems to fail.

I know why this is happening on the ophyd-async side, but not why it is happening on the bluesky side. On the ophyd-async front, a `SumHDFAravisDetector` inherits it's stage and trigger methods from it's parent class (a `StandardDetector`) and look like this:

```python
    @AsyncStatus.wrap
    async def stage(self) -> None:
        """Disarm the detector, stop filewriting, and open file for writing."""
        await self.check_config_sigs()
        await asyncio.gather(self.writer.close(), self.controller.disarm())
        self._describe = await self.writer.open()

    @AsyncStatus.wrap
    async def trigger(self) -> None:
        """Arm the detector and wait for it to finish."""
        indices_written = await self.writer.get_indices_written()
[...]

PR #1627, opened 2023-10-30
Fix/msg mutator

### LINKED ISSUE #1626: no stop document ^C after  a previous ^C and typing `RE.stop()` or `RE.abort()`
If `^C` is issued by the user after `RE.stop()` or `RE.abort()`, bluesky does not respond.  This may happen because the user made a mistake and should have typed `RE.hault()` or simply thought they could stop every thing from moving for convenience (probably a habit of users of another common DAQ software).

Whether the users should or should not do this for convenience is probably not a topic for discussion here.  What is important to note is that pressing `^C` many times seems to break the `RE` and the terminal freezes.  Once this happens, one needs to exit the session in order to kill the process, and as a consequence, the stop document is never written to the database.

## Expected Behavior
stop document gets written to the database after pressing `Enter` for `RE.stop()` ,  `RE.abort()` or `RE.hault()`?  I don't see anything else in the stop document that happens after this, but an expert should confirm this (or others with more advanced usage that is beyond out-of-the-box bluesky)

## Current Behavior
stop document gets written to the database after as soon as possible (not waiting on motors to return to the position dictated by underlying plan code for exceptions since this doesn't seem to be a critical part of the stop document

## Possible Solution
see expected behavior, but any solution anyone one dreams of is ok by me.

see above

## Context
see above

## Your Environment
`{'bluesky': '1.8.3.post1+g34931bb9', 'ophyd': '1.6.4'}` on RHEL8

tacaswell:
## Description

Correctly handle the exception communication channel in `msg_mutator`.

Fixes code from 3dd2ef03d73fc9b614abff1724a55a06ba74374b (2016). 

## Motivation and Context

closes #1626 and #1625 

## How Has This Been Tested?

Two new tests added:

 - one "pure" test that checks mechanical changes to a sequence of messages
 - one "integration" test that verifies that for `rel_scan` the close run message (which creates the stop document) will come out before the motors are returned to their initial position.

There are incidental changes make pytest not fail on me for pytest reasons locally.

### DIFF
--- a/bluesky/preprocessors.py
+++ b/bluesky/preprocessors.py
@@ -214,7 +214,7 @@ def plan_mutator(plan, msg_proc):
 
 def msg_mutator(plan, msg_proc):
     """
-    A simple preprocessor that mutates or deletes single messages in a plan
+    A simple preprocessor that mutates or deletes single messages in a plan.
 
     To *insert* messages, use ``plan_mutator`` instead.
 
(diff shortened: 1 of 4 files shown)


rosesyrett (approved):
Thanks :+1:

danielballan (approved):

PR #1634, opened 2023-11-14
WIP: allow plans to handle FailedPause

tacaswell:
This makes my interactive test work, still need to remove all of my debugging prints.

Probably need to revie

## Motivation and Context

Following up on a discussion with a call this morning.  This makes it possible for the plans to handle the `FailedPause` exception.  If they eat the exception it effectively cancel the Pause.  Best practice should be to in the handling yield a checkpoint + pause to _actually_ pause the plan as the user requested but 🤷🏻‍♂️ .

The fun bug I found is that when pausing via message the `CancelException` gets thrown in _after_ we have finished handling the pause (slightly wondering how this ever worked) hence the new `_request_pause_coro_inner`.  

## How Has This Been Tested?

```python
from bluesky.utils import Msg, ts_msg_hook

def test():
    yield Msg('rewindable', None, False)
    yield Msg('clear_checkpoint')
    yield from bps.null()
    try:
        yield from bps.pause()
        # yield from bps.sleep(50)
    except Exception as e:
        print(f'e: {type(e)}')
        yield from bps.checkpoint()
    for j in range(10):
        yield from bps.null()

RE = RunEngine()
RE.msg_hook = ts_msg_hook
RE(test())
        


```


tacaswell:
ok, I think I have the old tests passing, but this still needs new tests.

I've also purged all of my debugging prints.

coretl:
Ok, I've tried and failed to understand this.

I want to catch a CTRL-C (twice), and get the RE to raise a `FailedPause` into the plan. Then I can checkpoint and go back round the loop on resume. I thought I could raise it from `obj.pause()` but can see no code path to do that. Also I am now doubting how the resume path will work.

Here's the code so far in case you can shed any light...

```python
def test_caught_pause_in_flyer(RE):
    from ophyd.status import Status


    class MyFlyer:
        name = "flyer"

        def kickoff(self):
            status = Status()
            status.set_finished()
            return status
        
        def complete(self):
            self.complete_status = Status()
            return self.complete_status
        
        def pause(self):
            # Let the plan handle it
            raise FailedPause()
        
        def resume(self):
            pass

        
    flyers = [MyFlyer(), MyFlyer()]
    msgs = []
    RE.msg_hook = lambda x: msgs.append((time.time(), x))

    def do_scan(start_from: int):
        # This will be prepare, but this will do for the test
        msgs.append({"prepare": start_from})
        for flyer in flyers:
            yield from bps.kickoff(flyer, group="kickoff", wait=False)
        yield from bps.wait(group="kickoff")
        for flyer in flyers:
            yield from bps.complete(flyer, group="complete", wait=False)
        yield from bps.wait(group="complete")
        # Just to check we got to the end
        yield from bps.null()


    @bpp.run_decorator()
    def pausing_plan():
        done = False
        start_from = 1
        yield Msg('rewindable', None, False)
        yield Msg('clear_checkpoint')
        while not done:
            try:
                yield from do_scan(start_from)            
            except FailedPause:
                yield from bps.checkpoint()
                # Here we would inspect the counter values, and work out where to start from
                start_from += 1
            else:
                done = True

    pid = os.getpid()

    def sim_kill():
        os.kill(pid, signal.SIGINT)

    def complete_flyers():
        for flyer in flyers:
            flyer.complete_status.set_finished()

    threading.Timer(1, sim_kill).start()  # deferred pause
    threading.Timer(1.5, sim_kill).start()  # actual pause
    threading.Timer(3, complete_flyers).start()  # actual pause

    RE(pausing_plan())
    assert msgs == <something with 2 iterations of do_scan>  # fails
```

tacaswell:
I think this is doing what I expected (I modified the code a little bit to add the includes + use `bluesky_utils.ts_msg_hook` + prints):

```
09:25:18.808105 open_run          -> None            args: (), kwargs: {}, run: None
09:25:18.808567 rewindable        -> None            args: (False,), kwargs: {}, run: None
09:25:18.808631 clear_checkpoint  -> None            args: (), kwargs: {}, run: None
At top of do_scan with 1
09:25:18.808675 kickoff           -> flyer           args: (), kwargs: {'group': 'kickoff'}, run: None
09:25:18.809166 kickoff           -> flyer           args: (), kwargs: {'group': 'kickoff'}, run: None
09:25:18.809648 wait              -> None            args: (), kwargs: {'group': 'kickoff', 'timeout': None}, run: None
09:25:18.809927 complete          -> flyer           args: (), kwargs: {'group': 'complete'}, run: None
09:25:18.810194 complete          -> flyer           args: (), kwargs: {'group': 'complete'}, run: None
09:25:18.810469 wait              -> None            args: (), kwargs: {'group': 'complete', 'timeout': None}, run: None
A 'deferred pause' has been requested. The RunEngine will pause at the next checkpoint. To pause immediately, hit Ctrl+C again in the next 10 seconds.
Deferred pause acknowledged. Continuing to checkpoint.
trying a second time
Pausing...
ate failed pause
09:25:20.309621 checkpoint        -> None            args: (), kwargs: {}, run: None
At top of do_scan with 2
09:25:20.309913 kickoff           -> flyer           args: (), kwargs: {'group': 'kickoff'}, run: None
09:25:20.311043 kickoff           -> flyer           args: (), kwargs: {'group': 'kickoff'}, run: None
09:25:20.312245 wait              -> None            args: (), kwargs: {'group': 'kickoff', 'timeout': None}, run: None
09:25:20.312915 complete          -> flyer           args: (), kwargs: {'group': 'complete'}, run: None
09:25:20.313563 complete          -> flyer           args: (), kwargs: {'group': 'complete'}, run: None
09:25:20.313894 wait              -> None            args: (), kwargs: {'group': 'complete', 'timeout': None}, run: None
09:25:21.809457 null              -> None            args: (), kwargs: {}, run: None
09:25:21.809661 close_run         -> None            args: (), kwargs: {'exit_status': None, 'reason': None}, run: None
```

<details>

```python
import time
import os
import signal
import threading

from bluesky import RunEngine
from bluesky.run_engine import FailedPause
from bluesky.utils import Msg, ts_msg_hook
import bluesky.preprocessors as bpp

from ophyd.status import Status


class MyFlyer:
    name = "flyer"

    def kickoff(self):
        status = Status()
        status.set_finished()
        return status

    def complete(self):
        self.complete_status = Status()
        return self.complete_status

    def pause(self):
        # Let the plan handle it
        raise FailedPause()

    def resume(self):
        pass


RE = RunEngine()
flyers = [MyFlyer(), MyFlyer()]
msgs = []
RE.msg_hook = lambda x: msgs.append((time.time(), x)) or ts_msg_hook(x)

def do_scan(start_from: int):
    print(f"At top of do_scan with {start_from}")
    # This will be prepare, but this will do for the test
    msgs.append({"prepare": start_from})
    for flyer in flyers:
        yield from bps.kickoff(flyer, group="kickoff", wait=False)
    yield from bps.wait(group="kickoff")
    for flyer in flyers:
        yield from bps.complete(flyer, group="complete", wait=False)
    yield from bps.wait(group="complete")
    # Just to check we got to the end
    yield from bps.null()


@bpp.run_decorator()
def pausing_plan():
    done = False
    start_from = 1
    yield Msg('rewindable', None, False)
    yield Msg('clear_checkpoint')
    while not done:
        try:
            yield from do_scan(start_from)
        except FailedPause:
            print("ate failed pause")
            yield from bps.checkpoint()
            # Here we would inspect the counter values, and work out where to start from
            start_from += 1
        else:
            done = True

pid = os.getpid()

def sim_kill():
    os.kill(pid, signal.SIGINT)

def complete_flyers():
    for flyer in flyers:
        flyer.complete_status.set_finished()

threading.Timer(1, sim_kill).start()  # deferred pause
threading.Timer(1.5, sim_kill).start()  # actual pause
threading.Timer(3, complete_flyers).start()  # actual pause

RE(pausing_plan())

```

</details>

The `FailedPause` is this case is generate by the RE when it decides that the RE state will refuse to pause (which is what the `clear_checkpoint` does...it is the difference between "there is NO history" and "there is a history, but it is empty").

There is a existing protocol objects to say "please do not replay messages", but they puts us in the "there is a history, but it is empty" state so the RE will happily pause and on RE resume will start pulling the plan again (rather than (re) processing the `msg_cache`).  The last commit on this PR adds a new protocol (only handled for `pause` not for suspenders yet) that allows objects to `RefusePause` which puts the RE in the "there is NO history" state and then will then cause the `FailedPause` to be pushed through.

```python
import time
import os
import signal
import threading

from bluesky import RunEngine
from bluesky.run_engine import NoReplayAllowed, FailedPause, RefusePause
from bluesky.utils import Msg, ts_msg_hook
import bluesky.preprocessors as bpp

from ophyd.status import Status


class MyFlyer:
    name = "flyer"

    def kickoff(self):
        status = Status()
        status.set_finished()
        return status

    def complete(self):
        self.complete_status = Status()
        return self.complete_status

    def pause(self):
        print("trying and failing to fail pause")
        raise RefusePause()

    def resume(self):
        pass


RE = RunEngine()
flyers = [MyFlyer(), MyFlyer()]
msgs = []
RE.msg_hook = lambda x: msgs.append((time.time(), x)) or ts_msg_hook(x)

def do_scan(start_from: int):
    print(f"At top of do_scan with {start_from}")
    # This will be prepare, but this will do for the test
    msgs.append({"prepare": start_from})
    for flyer in flyers:
        yield from bps.kickoff(flyer, group="kickoff", wait=False)
    yield from bps.wait(group="kickoff")
    for flyer in flyers:
        yield from bps.complete(flyer, group="complete", wait=False)
    yield from bps.wait(group="complete")
    # Just to check we got to the end
    yield from bps.null()


@bpp.run_decorator()
def pausing_plan():
    done = False
    start_from = 1
    while not done:
        try:
            yield from do_scan(start_from)
        except FailedPause:
            print("ate failed pause")
            yield from bps.checkpoint()
            # Here we would inspect the counter values, and work out where to start from
            start_from += 1
        else:
            done = True

pid = os.getpid()

def sim_kill():
    os.kill(pid, signal.SIGINT)

def complete_flyers():
    for flyer in flyers:
        flyer.complete_status.set_finished()

threading.Timer(1, sim_kill).start()  # deferred pause
threading.Timer(1.5, sim_kill).start()  # actual pause
threading.Timer(3, complete_flyers).start()  # actual pause

RE(pausing_plan())


```

which gives me:

```
09:50:22.407130 open_run          -> None            args: (), kwargs: {}, run: None
At top of do_scan with 1
09:50:22.408054 kickoff           -> flyer           args: (), kwargs: {'group': 'kickoff'}, run: None
09:50:22.408481 kickoff           -> flyer           args: (), kwargs: {'group': 'kickoff'}, run: None
09:50:22.408868 wait              -> None            args: (), kwargs: {'group': 'kickoff', 'timeout': None}, run: None
09:50:22.409070 complete          -> flyer           args: (), kwargs: {'group': 'complete'}, run: None
09:50:22.409237 complete          -> flyer           args: (), kwargs: {'group': 'complete'}, run: None
09:50:22.409364 wait              -> None            args: (), kwargs: {'group': 'complete', 'timeout': None}, run: None
A 'deferred pause' has been requested. The RunEngine will pause at the next checkpoint. To pause immediately, hit Ctrl+C again in the next 10 seconds.
Deferred pause acknowledged. Continuing to checkpoint.
trying a second time
Pausing...
trying and failing to fail pause
trying and failing to fail pause
False
ate failed pause
09:50:23.908871 checkpoint        -> None            args: (), kwargs: {}, run: None
At top of do_scan with 2
09:50:23.909168 kickoff           -> flyer           args: (), kwargs: {'group': 'kickoff'}, run: None
09:50:23.910349 kickoff           -> flyer           args: (), kwargs: {'group': 'kickoff'}, run: None
09:50:23.911944 wait              -> None            args: (), kwargs: {'group': 'kickoff', 'timeout': None}, run: None
09:50:23.912935 complete          -> flyer           args: (), kwargs: {'group': 'complete'}, run: None
09:50:23.913641 complete          -> flyer           args: (), kwargs: {'group': 'complete'}, run: None
09:50:23.913857 wait              -> None            args: (), kwargs: {'group': 'complete', 'timeout': None}, run: None
09:50:25.408289 null              -> None            args: (), kwargs: {}, run: None
09:50:25.408541 close_run         -> None            args: (), kwargs: {'exit_status': None, 'reason': None}, run: None
Out[4]: ('e06b7a7f-e012-432a-a097-f722a3816cad',)
```


This does also suggest that the handling around "one of N objects has a busted `pause` method" needs to be improved, but that can be handled without the rest of this.

tacaswell:
Test looks good.

How thoroughly do we want to document this at this point?

bluesky/tests/test_examples.py
```diff
@@ -342,6 +345,106 @@ def sim_kill():
...
+
+    @bpp.run_decorator()
+    def pausing_plan():
+        done = False
+        start_from = 1
+        while not done:
+            try:
+                yield from do_scan(start_from)
+            except FailedPause:
+                yield from bps.checkpoint()
```
coretl, line 401: @tacaswell I missed something here, I've swallowed the pause at the plan level here to allow me to checkpoint and break out, but then I immediately continue on. I guess instead I should wait until the RE says resume. How do I do that?

tacaswell: add a `yield from bps.pause()` after the check point.

We need to make sure that `obj.pause()` is called on _everything_ before we raise

bluesky/tests/test_examples.py
```diff
@@ -342,6 +345,106 @@ def sim_kill():
...
+        def kickoff(self):
+            status = Status()
+            status.set_finished()
+            return status
+
+        def complete(self):
+            self.complete_status = Status()
+            return self.complete_status
+
+        def pause(self):
```
coretl, line 369: Need to handle the case where some devices raise and some don't

bluesky/tests/test_examples.py
```diff
@@ -342,6 +345,106 @@ def sim_kill():
...
+        yield from bps.wait(group="kickoff")
+        for flyer in flyers:
+            yield from bps.complete(flyer, group="complete", wait=False)
+        yield from bps.wait(group="complete")
+        # Just to check we got to the end
+        yield from bps.null()
+
+    @bpp.run_decorator()
+    def pausing_plan():
+        done = False
```
coretl, line 395: Try telling the run engine not to cache the commands and check that we still get the "continue to checkpoint" pause behaviour

coretl:
After talking this over with @evalott100 we have now come to this conclusion:

- Only one flyer should respond with `FailedPause`, and only once, it needs to accept the pause the second time
- The plan then catches this, checkpoints, calls `bps.stop` on the other flyers, then calls `bps.pause` again to actually pause

Because we actually have many flyers, we end up with none of them being pausable, and a special device that exists only to punt the `FailedPause` back to the plan:

```python
class Failer:
    fail = False
    def set(self, value: bool):
        self.fail = value
    def pause(self):
        if self.fail:
            raise FailedPause()

def plan(flyers):
    done = False
    start = 0
    failer = Failer()
    while not done:
        yield from bps.set(failer, True)
        try:
            yield from bps.prepare(flyers, start)
            yield from bps.kickoff(flyers)            
            yield from bps.complete(flyers)
        except FailedPause:
            start = get_stream_counter(yield from bps.checkpoint(), "primary")
            yield from bps.stop(flyers)
            yield from bps.set(failer, False)
            yield from bps.pause()
        else:
            done = True
```

coretl:
With this in mind, could we lift this up as a special pausing mode to the RE?

```python
def plan(flyers):
    done = False
    start = 0
    while not done:
        yield from bps.raise_on_pause()
        try:
            yield from bps.prepare(flyers, start)
            yield from bps.kickoff(flyers)            
            yield from bps.complete(flyers)
        except FailedPause:
            start = get_stream_counter(yield from bps.checkpoint(), "primary")
            yield from bps.stop(flyers)
            yield from bps.pause(raise=False)
        else:
            done = True
```
It's less code, but is it any clearer what's going on?
@tacaswell thoughts?

tacaswell:
which raise do you mean to suppress: the one raised by `RE` on its way out or the one throw into the plan?

If it is the second I'm worried about it being used in a context that is not carefully managed like this and the plan ignoring a devices refusing to pause and then resuming into an undefined state. 

On the first pause _all_ of the flyers are notified (so you would not need the stop) and then they all go into a sate where they will not refuse so on the second pass they are all happy.  I think you need this state in the flyers anyway in case they have finished moving but we have not the final complete yet or the have not been kicked off yet when the `pause` comes in from outside.

I suspect that what we should be doing here is raising Execption Groups (which are a py311 feature)

tacaswell:
The flier should be raising `RefusedToPause` not `FailedPause`

bluesky/run_engine.py
```diff
@@ -1444,16 +1446,28 @@ async def _run(self):
                     await self._stop_movable_objects(success=True)
                     # Notify Devices of the pause in case they want to
                     # clean up.
+                    # track of any devices refuse to pause
+                    refuse_to_pause = False
                     for obj in self._objs_seen:
```
tacaswell, line 1451: This loop probably needs some extra protections to make sure that we hit every object even if it fails.  I guess we need `except BaseException as e: # stash excption ` + finally re-raise anything that was caught and stashed?

coretl:
> which raise do you mean to suppress: the one raised by `RE` on its way out or the one throw into the plan?

I don't understand the question. We raise `RefuseToPause` in the device, catch it in the `RE`, raise `FailedPause`, catch it in the plan, then actually do the pause logic there, ending up calling `bps.pause` so the `RE` can return to the user

> If it is the second I'm worried about it being used in a context that is not carefully managed like this and the plan ignoring a devices refusing to pause and then resuming into an undefined state.

Ah, do you mean that if we add in this new mode then people might catch the `FailedPause` and then not actually cause `bps.pause`? I don't know how that is avoidable in any of the options here. Unless we provide the `RE` with a "pause plan" and have it run that instead of this entire PR?

> On the first pause _all_ of the flyers are notified (so you would not need the stop) and then they all go into a sate where they will not refuse so on the second pass they are all happy. I think you need this state in the flyers anyway in case they have finished moving but we have not the final complete yet or the have not been kicked off yet when the `pause` comes in from outside.

The problem is that `device.pause()` is called serially on every seen device in an unpredictable order, so the first failure will raise. This means that we can't rely on the first `pause` to have actually stopped any/all of the detectors, although I guess the second pause would then get any that didn't get paused the first time...

> I suspect that what we should be doing here is raising Execption Groups (which are a py311 feature)

Yes, or not using exceptions for control flow and passing "what to do on pause" into the RE from the plan...

bluesky/utils/__init__.py
```diff
@@ -86,6 +86,10 @@ class NoReplayAllowed(Exception):
     pass
 
 
+class RefusePause(Exception):
```
coretl, line 89:
```suggestion
class UncleanPause(Exception):
```

tacaswell:
> Ah, do you mean that if we add in this new mode then people might catch the FailedPause and then not actually cause bps.pause? I don't know how that is avoidable in any of the options here. Unless we provide the RE with a "pause plan" and have it run that instead of this entire PR?

Yes, but as we discussed on the call...if you do that you are opting into the behavior so maybe you want than (and we should not prevent an odd choice that some user might want).

> The problem is that device.pause() is called serially on every seen device in an unpredictable order, so the first failure will raise.

Those loops need enough `try...excepts` that we will always call `pause` on all of them (also as discussed on the call).

coretl:
@tacaswell what do you think about raising `UncleanPause` up to the plan? Then we would end up doing something like:

```python
start_from = ...
try:
    do_fly_scan(start_from)
except UncleanPause:
    start_from = calc_where_to_start_from(yield from bps.checkpoint())
    yield from bps.pause()
```

This would only catch the explicit expected case of the flyers raising `UncleanPause`, allowing the plan to cleanly pause, and any other `FailedPause` exception would be unexpected and bubble up

tacaswell:
> what do you think about raising UncleanPause up to the plan

That is a good idea.  Also allows devices to say "I paused but lost some stuff, but the plan might be able to recover" vs "ha ha, I'm broken and there is no way to recover"

evalott100:
@tacaswell 

We edited the test to try and match @coretl's [idea from yesterday](https://github.com/bluesky/bluesky/pull/1634#issuecomment-1931673543), however I ran into issues in `run_engine.py` that I found hard to track down. Specifically this [code](https://github.com/tacaswell/bluesky/blob/26ef337d8cb8ff2a9fc01caa4d545c5b2aedd050/bluesky/run_engine.py#L1460-L1463) causing an abort status and an error when the device tries to pause again after excepting `UncleanPause`.

tacaswell:
rebased this across the formatting work.

PR #1644, opened 2023-12-14
Remove empty complete method from RunBundler

rosesyrett:
## Description
We have an empty complete method in the RunBundler. This simply removes it, and the associated call from the RunEngine._complete method to it.

## Motivation and Context
I think it's good to remove code that seems unnecessary. This is one example of such code. 

## Related Discussions
I have realised that, with this removal, there is no need for Msg("complete") to be sent when a run is open. However, we probably want to enforce a certain order of messages being sent/injested by the RunEngine. I recognise that this has probably been discussed before, and it's difficult (if not just impossible) to introspect what messages the RunEngine will get from generator functions/coroutines. However, maybe there is a way to enforce, with things like "collect", that we only want those to be called e.g. if kickoff has been called too (and therefore inside a run).

### DIFF
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -615,24 +615,6 @@ class RunBundler:
         """
         self._uncollected.add(msg.obj)
 
-    async def complete(self, msg):
-        """
-        Tell a flyer, 'stop collecting, whenever you are ready'.
-
-        The flyer returns a status object. Some flyers respond to this
-        command by stopping collection and returning a finished status
-        object immediately. Other flyers finish their given course and
-        finish whenever they finish, irrespective of when this command is
-        issued.
-
-        Expected message object is::
-
-            Msg('complete', flyer, group=<GROUP>)
-
-        where <GROUP> is a hashable identifier.
-        """
-        ...
-
     def _maybe_format_datakeys_with_stream_name(
         self,
         describe_collect_dict: Union[Dict[str, DataKey], Dict[str, Dict[str, DataKey]]],
--- a/bluesky/run_engine.py
+++ b/bluesky/run_engine.py
@@ -2061,14 +2061,6 @@ class RunEngine:
 
         where <GROUP> is a hashable identifier.
         """
-        run_key = msg.run
-        try:
-            current_run = self._run_bundlers[run_key]
-        except KeyError as ke:
-            raise IllegalMessageSequence("A 'complete' message was sent but no "
-                                         "run is open.") from ke
-
-        await current_run.complete(msg)
         kwargs = dict(msg.kwargs)
         group = kwargs.pop("group", None)
         obj = check_supports(msg.obj, Flyable)


danielballan (approved):

danielballan:
Yes. I think more order checking would be desirable.

@tacaswell Can you confirm I am not forgetting context?

tacaswell:
I suspect that this is an artifact of the path we took to get here and when I pulled `RunBundler` out of the `RunEngine` I was mechanical and competitionist.

It conceptually it makes sense to not tell the run bundler about the `complete` as it has no implications about the data (the same way we do not pass `sets` or `waits` through to the bundler).  On the other hand, we tell the bundler about the rest of the flyer verbs and it is funny to leave this one out (and it might have a use after all).

-------

Given that there can be multiple runs open at once (so multiple bundlers exist) and multiple fliers per run, the question is not "Have we seen a kickoff" it is "have we seen a kickoff for this device".  

If we do want to check that `complete` has a matched `kickoff`, then we should not remove this method and leverage the `RunBundler._uncollected` attribute and verify that the object we are about to collect is indeed kicked off!

We already have all of the state we need to validate available in that the message sequence in illegal because it would drive us to an invalid state / requires RE+bundlers to be in a different state.  I think re-generating that state as a queue of messages and then searching/reprocessing them again would add complexity without much gain.


-----

I'm going to hold off merging if after this discussion we want to agument `collect` rather than remove it.   @rosesyrett can self-merge if we are happy not adding any additional validation to `complete`.

rosesyrett:
I think what I really want, is a `MsgTranslator` class that takes bluesky `Msg` objects and performs checks on them before passing them to the correct RunEngine verbs. This way, the RunEngine can be completely agnostic to what the RunBundler is doing; it can just rely on the fact that the translator will block invalid messages (i.e. a close if no run is opened). And the translator can perhaps have some sort of cache for bluesky commands called on `obj`s, so it can reject a `complete` method if it doesn't find the object in the cache with a matching `kickoff`, for example.

This would also mean we can have a definitely typed codebase, or at least a definitely typed RunEngine, which would be nice. As well as just in general reducing the huge amount of code within it.

prjemian:
This conversation starts to look like a new issue instead of more work for this PR.

tacaswell:
I big part of the power of the RE is that the plans are _not_ static.  It lets you write plans like "re-align the beamline at approximately 30 minute intervals" or "run a dynamic alignment / search / ...".   It also supports writing plans like "no matter what happens, close the shutter".

I do not think an invalid sequence of Msgs is not a type error the same way that a divide-by-zero error or integer overflow are not type errors. 


-----


I'm not sure how `MsgTranslator` would help us though, it would us though.  It would be a re-capitulation of the rules in `RunBundler` and as it can not see the future (and because the plans may change what Msg they yield based on the return values from previous messages) we would still have to be able to deal with invalid sequences in `_run`.   There is also no way of getting out of dealing with "motor failed" or "user canceled" or ... type errors in the RunEngine so the lift to also handle `InvalidSequence` is not too bad.


We could write a `def make_plan_static_and_sad_but_definitely_valid(plan: coroutine) -> List[Msg]` helper, but that would still be a runtime thing unless you want to start parsing AST...

tacaswell:
Sorry, commenting while sleepy and distracted.

- an invalid message sequence is not a type error the same way divide-by-zero is not a type error. 
- drop the second ~, it would us though~

DiamondJoseph:
Since @rosesyrett is ~~now free~~ no longer at Diamond, what should be done with this merge request? I'm happy to adopt whatever change needs to be made, or else if there is no great outcry for this to be merged I'm happy to close with comment of revisiting later if complete ever gains args or more checking of message order.

tacaswell:
I'm going to go ahead and merge this.  It is a no-op function and a belt-and-suspenders conditional.  If we find we do need this it is easy enough to put it back.

PR #1648, opened 2023-12-26
FIX: handle case of rewinding to before the beginning of a stream

tacaswell:
## Description

The issue is that:

 - we track the event count (for seq_num which is non-monotonic and
   can repeat if rewound) in `Bundler._sequence_counters`
 - we snapshot the counters at checkpoints
 - on rewind we clear the _event_counters dictionary and then .update it
   from the copy

In the case of implicitly creating a stream between a checkpoint and a
pause (so there is `declare_stream` message) the above process effectively
drops our tracking in the event counters.

In `_prepare_stream` we ensure that the needed counter exists (and starts at 1)
so replicate that logic here.

## Motivation and Context

We should not crash this way.

## How Has This Been Tested?

Manually + adding a test

### DIFF
--- a/bluesky/bundlers.py
+++ b/bluesky/bundlers.py
@@ -2,7 +2,7 @@ import asyncio
 import inspect
 import time as ttime
 from collections import deque
-from typing import (Any, Deque, Dict, FrozenSet, Iterable, List, Optional, Set,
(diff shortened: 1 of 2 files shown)


dmgav (approved):
This PR resolves the bug #2 (immediate pause) reported in https://github.com/bluesky/bluesky/issues/1647
The bug #1 requires further investigation.

PR #1651, opened 2024-01-11
Enable collect for multiple detectors

### LINKED ISSUE #1649: Collect multiple detectors concurrently
Since the I22 project, a need has arisen for a bluesky level protocol and verb (or modification to existing verb) that collects asset docs for multiple detectors up to a certain index, i.e. so all detectors are in step with one another.

This issue is therefore split into two parts; one, to make the new protocol, and the second, to modify the "collect" method on the run engine bundler.

Because this will be associated with the production of multiple stream resources and datums and is an extension to an existing protocol `WritesExternalAssets`, perhaps this should be called `WritesMultipleExternalStreamAssets` (this is probably a bit of a mouthful, though).

## New Protocol

The new protocol should be defined in `protocols.py`:
```python

class WritesExternalStreamAssets(Protocol):
    def collect_asset_docs(self, index: Optional[int] = None) -> SyncOrAsync[Iterator[Asset]]:
        ...

    def get_index(self) -> int:
        ...
```

This is not necessarily exactly what it'll look like, but gives a general idea. The exact details can be worked out by trying to implement it. The idea is, that this protocol will allow the run engine to "collect" multiple detectors up to a minimum index.

## Modifications to collect

The collect method should be able to:
1. Accept multiple objects in the RunEngine `Msg`,
2. Find the minimum index of all of them using `get_index` from above,
3. Call `collect_asset_docs` for all detectors on that minimum, yielding the resulting asset docs.

abbiemery:
## Description
Changes the collect method in bundler to handle the collection of multiple detectors. Contains the logic to collect stream resources using the maximum common number of frames written, determined by the minimum index.

`Asset` has been altered to only address a `resource` and `datum` and a new `StreamAsset` has been created to address `stream_resource` and `stream_datum`.

## Motivation and Context
This enables logic to be moved out of Ophyd-async, generalising the process
Fixes #1649.

## How Has This Been Tested?
TBC.

### DIFF
--- a/bluesky/tests/test_external_assets_and_paging.py
+++ b/bluesky/tests/test_external_assets_and_paging.py
@@ -5,7 +5,6 @@ from event_model.documents.resource import PartialResource
 from event_model.documents.datum import Datum
 from event_model.documents.stream_datum import StreamDatum
 from event_model.documents.stream_resource import StreamResource
-from bluesky.utils import new_uid
 import bluesky.plans as bp
 import bluesky.plan_stubs as bps
 import pytest
@@ -64,6 +63,7 @@ class Named(HasName):
 
     def __init__(self, name: str) -> None:
         self.name = name
+        self.counter = 0
 
 
 def describe_datum(self: Named) -> Dict[str, DataKey]:
@@ -129,38 +129,32 @@ def collect_asset_docs_stream_datum(
     """Produce a StreamResource and StreamDatum for 2 data keys for 0:index"""
     index = index or 1
     for data_key in [f"{self.name}-sd1", f"{self.name}-sd2"]:
-        stream_resource = StreamResource(
-            resource_kwargs={"dataset": f"/{data_key}/data"},
-            data_key=data_key,
-            root="/root",
-            resource_path="/path.h5",
-            spec="ADHDF5_SWMR_STREAM",
-            uid=new_uid(),
-        )
-        yield "stream_resource", stream_resource
+        uid = f"{data_key}-uid"
+        if self.counter == 0:
+            stream_resource = StreamResource(
+                resource_kwargs={"dataset": f"/{data_key}/data"},
+                data_key=data_key,
+                root="/root",
+                resource_path="/path.h5",
+                spec="ADHDF5_SWMR_STREAM",
+                uid=uid,
+            )
+            yield "stream_resource", stream_resource
 
         stream_datum = StreamDatum(
-            stream_resource=stream_resource["uid"],
+            stream_resource=uid,
             descriptor="",
-            uid=f'{stream_resource["uid"]}/1',
-            indices={"start": 0, "stop": index},
+            uid=f"{uid}/{self.counter}",
+            indices={"start": self.counter, "stop": self.counter + index},
             seq_nums={"start": 0, "stop": 0},
         )
         yield "stream_datum", stream_datum
+    self.counter += index
 
 
 def read_empty(self) -> Dict[str, Reading]:
     """Produce an empty event"""
     return {}
-
-
-class StreamDatumReadable(Named, Readable, WritesStreamAssets):
-    """A readable that produces a single frame from 2 StreamResources"""
-
-    describe = describe_stream_datum
-    read = read_empty
-    collect_asset_docs = collect_asset_docs_stream_datum
-    get_index = get_index
 
 
 def describe_pv(self: Named) -> Dict[str, DataKey]:
@@ -185,7 +179,7 @@ class PvAndStreamDatumReadable(Named, Readable, WritesExternalAssets):
     """An odd device that produces a single event with one datakey from a pv, and one backed by a StreamDatum"""
 
     describe = merge_dicts(describe_pv, describe_stream_datum)
-    read = merge_dicts(read_pv, read_empty)
+    read = read_pv
     collect_asset_docs = collect_asset_docs_stream_datum
     get_index = get_index
 
@@ -301,19 +295,23 @@ def collect_pages_pv(self: Named) -> Iterator[PartialEvent]:
 
 class PvCollectable(Named, EventCollectable):
     """Produces events with 2 data keys backed by PVs"""
+
     describe_collect = describe_collect_pv
     collect = collect_pv
 
 
 class PvPageCollectable(Named, EventPageCollectable):
     """Produces event pages with 2 data keys backed by PVs"""
+
     describe_collect = describe_collect_pv
     collect_pages = collect_pages_pv
 
 
-class StreamDatumCollectable(Named, Collectable, WritesStreamAssets):
-    """Produces no events, but only StreamResources for 2 data keys"""
-    describe_collect = describe_stream_datum
+class StreamDatumReadableCollectable(Named, Readable, Collectable, WritesStreamAssets):
+    """Produces no events, but only StreamResources for 2 data keys and can be read or collected"""
+
+    describe = describe_collect = describe_stream_datum
+    read = read_empty
     collect_asset_docs = collect_asset_docs_stream_datum
     get_index = get_index
 
@@ -332,19 +330,25 @@ def test_datum_readable_counts(RE):
 
 
 def test_stream_datum_readable_counts(RE):
-    det = StreamDatumReadable(name="det")
+    det = StreamDatumReadableCollectable(name="det")
     docs = DocHolder()
-    RE(bp.count([det]), docs.append)
+    RE(bp.count([det], 2), docs.append)
     docs.assert_emitted(
-        start=1, descriptor=1, stream_resource=2, stream_datum=2, event=1, stop=1
+        start=1, descriptor=1, stream_resource=2, stream_datum=4, event=2, stop=1
     )
     assert docs["descriptor"][0]["name"] == "primary"
     assert list(docs["descriptor"][0]["data_keys"]) == ["det-sd1", "det-sd2"]
     assert all(
         sd["descriptor"] == docs["descriptor"][0]["uid"] for sd in docs["stream_datum"]
     )
-    assert docs["event"][0]["data"] == {}
-    assert docs["event"][0]["filled"] == {}
+    assert all(e["data"] == {} for e in docs["event"])
+    assert all(e["filled"] == {} for e in docs["event"])
+    assert [sd["indices"] for sd in docs["stream_datum"]] == [
+        {"start": 0, "stop": 1},
+        {"start": 0, "stop": 1},
+        {"start": 1, "stop": 2},
+        {"start": 1, "stop": 2},
+    ]
 
 
 def test_combinations_counts(RE):
@@ -494,7 +498,7 @@ def test_pv_collectable(RE, cls):
 
 
 def test_stream_datum_collectable(RE):
-    det = StreamDatumCollectable(name="det")
+    det = StreamDatumReadableCollectable(name="det")
     docs = DocHolder()
     RE(collect_plan(det, pre_declare=True), docs.append)
     docs.assert_emitted(
@@ -510,7 +514,7 @@ def test_stream_datum_collectable(RE):
 
 @pytest.mark.parametrize("cls1", [PvCollectable, PvPageCollectable])
 @pytest.mark.parametrize(
-    "cls2", [PvCollectable, PvPageCollectable, StreamDatumCollectable]
+    "cls2", [PvCollectable, PvPageCollectable, StreamDatumReadableCollectable]
 )
 def test_many_collectables_fails(RE, cls1, cls2):
     det1, det2 = cls1(name="det1"), cls2(name="det2")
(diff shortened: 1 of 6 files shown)


bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+    EventPageCollectable,
+    Collectable,
+    WritesStreamAssets,
+    StreamAsset,
 )
 
 
-def read_Readable(self) -> Dict[str, Reading]:
-    return dict(det2=dict(value=1.2, timestamp=0.0))
+class DocHolder(dict):
```
danielballan, line 25: This is clever. 

I'll mention that there is a [utility in the test utils ](https://github.com/bluesky/bluesky/blob/91b4a8a2add41b2cecadc4357b32810437533912/bluesky/tests/utils.py#L31-L50)with a similar functionality, but I like the simplicity of this one.

It _might_ be worth considering using `collections.defaultdict(list)` instead, which I think is a more "mainstream" / familiar way to achieve this, and refactoring `assert_emitted` into a function. On the other hand, I'm not sure that buys much over implementation here. Just a thought.

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+class StreamDatumReadableCollectable(Named, Readable, Collectable, WritesStreamAssets):
+    """Produces no events, but only StreamResources for 2 data keys and can be read or collected"""
+
+    describe = describe_collect = describe_stream_datum
+    read = read_empty
+    collect_asset_docs = collect_asset_docs_stream_datum
+    get_index = get_index
+
+
+def test_datum_readable_counts(RE):
```
danielballan, line 319:
```suggestion
def test_datum_readable_counts(RE):
    "Test that count-ing a datum-producing device results in expected documents."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+    docs.assert_emitted(start=1, descriptor=1, resource=1, datum=1, event=1, stop=1)
+    assert docs["descriptor"][0]["name"] == "primary"
+    assert list(docs["descriptor"][0]["data_keys"]) == ["det-datum"]
+    assert docs["event"][0]["descriptor"] == docs["descriptor"][0]["uid"]
+    assert docs["event"][0]["data"] == {"det-datum": "RESOURCEUID/1"}
+    assert docs["event"][0]["timestamps"] == {"det-datum": 456}
+    assert docs["event"][0]["filled"] == {"det-datum": False}
+
+
+def test_stream_datum_readable_counts(RE):
```
danielballan, line 332:
```suggestion
def test_stream_datum_readable_counts(RE):
    "Test that count-ing a StreamDatum-producing device results in expected documents."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
-       )
 
 
-@pytest.mark.parametrize(
-    'asset_type,collect_asset_docs_fun',
-    [
-        ("datum", collect_asset_docs_Datum),
-        ("stream_datum", collect_asset_docs_StreamDatum),
-        # StreamResource is receieved in the collect of stream_datum
+def test_combinations_counts(RE):
```
danielballan, line 354:
```suggestion
def test_combinations_counts(RE):
    "Test that mixing a StreamDatum- and Datum-producing device in one count works."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+
+def test_old_describe_fails_predeclare(RE):
+    with pytest.raises(
+        RuntimeError,
+        match="Old style describe_collect output with stream name not supported in declare_stream",
+    ):
+        RE(collect_plan(OldPvCollectable(name="det"), pre_declare=True))
+
+
+def test_same_key_in_multiple_streams_fails(RE):
```
danielballan, line 408:
```suggestion
def test_same_key_in_multiple_streams_fails(RE):
    "Test that old-style describe_collect cannot express the same data key appearing in multiple streams."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+    with pytest.raises(
+        RuntimeError,
+        match="Multiple streams ['stream1', 'stream2'] would emit the same data_keys ['pv']",
+    ):
+        RE(collect_plan(MultiKeyOldCollectable(name="det"), pre_declare=False))
+
+
+def test_old_pv_collectable(RE):
```
danielballan, line 416:
```suggestion
def test_old_pv_collectable(RE):
    "Test that collecting a PartialEventPage-producing device results in expected documents."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+    )
+    assert docs["descriptor"][0]["name"] == "stream3"
+    assert list(docs["descriptor"][0]["data_keys"]) == ["det-datum"]
+    assert docs["event_page"][0]["descriptor"] == docs["descriptor"][0]["uid"]
+    assert docs["event_page"][0]["data"] == {"det-datum": ["RESOURCEUID/1"]}
+    assert docs["event_page"][0]["timestamps"] == {"det-datum": [456]}
+    assert docs["event_page"][0]["filled"] == {"det-datum": [False]}
+
+
+def test_old_datum_and_pv_collectable(RE):
```
danielballan, line 453:
```suggestion
def test_old_datum_and_pv_collectable(RE):
    "Test collecting from a device that produces a mixture of PartialEventPage and Datum."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+        "stream1-pv2": [0, 1],
+    }
+    assert docs["event_page"][1]["data"] == {
+        "stream2-pv1": [0, 1],
+        "stream2-pv2": [0, 1],
+    }
+    assert docs["event_page"][2]["data"] == {"det-datum": ["RESOURCEUID/1"]}
+
+
+def test_new_collect_needs_predeclare(RE):
```
danielballan, line 474:
```suggestion
def test_new_collect_needs_predeclare(RE):
    "Test that new-style describe_collect requires pre-declaring the stream."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+    with pytest.raises(
+        RuntimeError,
+        match="New style describe_collect output requires declare_stream before collect",
+    ):
+        RE(collect_plan(PvCollectable(name="det"), pre_declare=False))
+
+
+@pytest.mark.parametrize("cls", [PvCollectable, PvPageCollectable])
+def test_pv_collectable(RE, cls):
```
danielballan, line 483:
```suggestion
def test_pv_collectable(RE, cls):
    "Test that a device producing PartialEvent or PartialEventPage results in emitted EventPages."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+        "det-pv1": [0, 1],
+        "det-pv2": [0, 1],
+    }
+    assert docs["event_page"][0]["timestamps"] == {
+        "det-pv1": [100, 101],
+        "det-pv2": [100, 101],
+    }
+
+
+def test_stream_datum_collectable(RE):
```
danielballan, line 500:
```suggestion
def test_stream_datum_collectable(RE):
    "Test that collecting a StreamDatum-producing device results in expected documents."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
-    x = X()
-    collector = []
-    RE([
-            Msg("open_run", x),
-            Msg("collect", x, name="primary"),
-            Msg("close_run", x),
-        ],
-        lambda *args: collector.append(args)
-       )
+def test_many_collectables_fails(RE, cls1, cls2):
```
danielballan, line 519:
```suggestion
def test_many_collectables_fails(RE, cls1, cls2):
   "Test limitation of one collect-able device ("flyer") per stream."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+    det1, det2 = cls1(name="det1"), cls2(name="det2")
+    with pytest.raises(
+        RuntimeError,
+        match="Cannot collect the output of multiple collect() or collect_pages() into the same stream",
+    ):
+        RE(collect_plan(det1, det2, pre_declare=False))
+
+
+def test_many_stream_datum_collectables(RE):
```
danielballan, line 528:
```suggestion
def test_many_stream_datum_collectables(RE):
    "Test collecting from multiple StreamDatum-producing devices."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+    data_keys = ["det1-sd1", "det1-sd2", "det2-sd1", "det2-sd2"]
+    assert docs["descriptor"][0]["name"] == "primary"
+    assert list(docs["descriptor"][0]["data_keys"]) == data_keys
+    assert [d["data_key"] for d in docs["stream_resource"]] == data_keys
+    assert all(
+        d["descriptor"] == docs["descriptor"][0]["uid"] for d in docs["stream_datum"]
+    )
+
+
+def tomo_plan(*objs):
```
danielballan, line 545:
```suggestion
def tomo_plan(*objs):
    "An applied flyscanning example plan"
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+    bps.trigger_and_read(objs, name="flats")
+    bps.trigger_and_read(objs, name="darks")
+    # do the flyscan
+    bps.collect(*objs, name="projections")
+    # take some flats at the end
+    bps.trigger_and_read(objs, name="flats")
+    yield from bps.close_run()
+
+
+def test_tomography_multi_stream_same_detectors(RE):
```
danielballan, line 560:
```suggestion
def test_tomography_multi_stream_same_detectors(RE):
    "Test tomo_plan applied example"
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+    assert [d["data_key"] for d in docs["stream_resource"]] == data_keys
+    assert [d["descriptor"] for d in docs["stream_datum"]] == (
+        [docs["descriptor"][0]["uid"]] * 4
+        + [docs["descriptor"][1]["uid"]] * 4
+        + [docs["descriptor"][2]["uid"]] * 4
+        + [docs["descriptor"][0]["uid"]] * 4
+    )
+
+
+def change_conf_plan(*objs):
```
danielballan, line 586:
```suggestion
def change_conf_plan(*objs):
    "Re-emit a fresh EventDescriptor after the first Event."
```

bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -1,479 +1,614 @@
...
+    yield from bps.open_run()
+    for _ in range(2):
+        yield from bps.declare_stream(*objs)
+        bps.trigger_and_read(objs)
+    yield from bps.close_run()
+
+
+def test_multiple_prepare_in_same_stream(RE):
```
danielballan, line 594:
```suggestion
def test_multiple_prepare_in_same_stream(RE):
    "Test re-emitting an EventDescriptor while using StreamDatum-producing devices."
```

danielballan (approved):
Looks great. The effort on these tests is hugely valuable. I am suggested additional docstrings and added one comment on `DocHolder` for consideration, but I'm happy for this to go ahead.

bluesky/tests/test_external_assets_and_paging.py
coretl: The choice between:
```python
class DocHolder(dict):
    def append(self, name, doc):
        self.setdefault(name, []).append(doc)

    def assert_emitted(self, **numbers: int):
        assert list(self) == list(numbers)
        assert {name: len(d) for name, d in self.items()} == numbers

def test_plan():
    docs = DocHolder()
    RE(plan(), docs.append)
    docs.assert_emitted(start=1, descriptor=1, resource=1, datum=1, event=1, stop=1)
```
and
```python

def assert_emitted(docs: Dict[str, list], **numbers: int):
    assert list(docs) == list(numbers)
    assert {name: len(d) for name, d in docs.items()} == numbers

def test_plan():
    docs = defaultdict(list)
    RE(plan(), lambda name, doc: docs[name].append(doc))
    assert_emitted(docs, start=1, descriptor=1, resource=1, datum=1, event=1, stop=1)
```

I think I marginally prefer the second, but not enough to change all the tests. @abbiemery I leave the choice up to you...

bluesky/tests/test_external_assets_and_paging.py
coretl: @abbiemery I can't remember what pytest does when you add a docstring to a test, please can you check that it doesn't hide the test name in the terminal...

If the output is still fine then all these are fine to go in

bluesky/tests/test_external_assets_and_paging.py
danielballan: I am in complete agreement. Slight preference, but not sure it's worth the effort to change.

abbiemery: Given the time pressure of these now, i'd rather make that a new issue to clean things up if it is preferred.

coretl (approved):

danielballan (approved):

PR #1653, opened 2024-01-22
ENH: allow passing in an axes factory to LivePlot and friends

tacaswell:
## Description

This is to enable re-use of figures when creating LivePlot objects inside of plans.

## Motivation and Context

Because we can not create figures on background threads we can not directly create the axes to pass into `LivePlot` and friends.  This is handled in BEC because the figures are created as part of processing the descriptor.

By allowing the additional level of indirection here, we can sensibly write plans that include the plotting that BL staff want to see. 

Although it is better long term to push the plotting to a separate process, this makes things a bit nicer to deal with now.

## How Has This Been Tested?

A version of this is on the floor at OPLS https://github.com/NSLS-II-OPLS/profile_collection/commit/d91f7d35ce23720478d24a2431302df994d3c020


evalott100 (commented):
Looks good, though it might be nice to add another unit test in test_callbacks.py where `ax` is `Callable`?

mrakitin:
A rebase would help.

PR #1660, opened 2024-03-02
Callback to write documents to Tiled

genematx:
Adds a callback to enable writing Events, Stream Resource, and Stream Datum messages from RunEngine to Tiled.

## Description
This PR implements a new callback class, `TiledWriter`, which can be subscribed to a RunEngine and processes Events and Stream messages.

## How Has This Been Tested?
`test_tiled_writer.py` tests the streaming capabilities with writing data into an hdf5 file and a tiff sequence.


bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,165 @@
...
+            metadata={"start": doc},
+            specs=[Spec("BlueskyRun", version="1.0")],
+        )
+
+    def stop(self, doc):
+        metadata = dict(self.node.metadata) | {"stop": doc}
+        self.node.update_metadata(metadata=metadata)
+
+    def descriptor(self, doc):
+        descriptor_node = self.node.create_container(key=doc["name"], metadata=doc)
```
danielballan, line 63: In the original, group-coded draft of this, we intentionally made a simplifying assumption here: one Event Descriptor document per stream: i.e. we'll only see one such document with a given `name` key.

In general, the Event Descriptor with a given `name` can be re-emitted if a configuration setting (e.g. exposure time) changes. However, the `data_keys` are _not_ allowed to change. (The bundler code ensures this.)

To start, I might suggest following the structure we use in `databroker.mongo_normalized`, which is to place the Event Descriptor(s) in a list, under the key `descriptors`, as in `metadata={"descriptors": [doc]}` and then appending any additional descriptors with the same `name` to this list. That would be sufficient to have something that works.

In the future, pulling apart the changeable parts and the fixed parts (`data_keys`) may be useful. Needs more thought and dicussion.

bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,165 @@
...
+        self._descriptor_nodes[doc["uid"]] = descriptor_node
+        descriptor_node.create_container(key="external")
+        descriptor_node.create_container(key="internal")
+
+    def event(self, doc):
+        descriptor_node = self._descriptor_nodes[doc["descriptor"]]
+        parent_node = descriptor_node["internal"]
+        for table_key in ["data", "timestamps"]:
```
danielballan, line 71: I'll add a note here as a reminder to ourselves: we discussed making a single table here (not sure what to call it...`"table"`?) that contains both the data and timestmaps columns, potentially prefixed like `data_*` and `ts_*` as a way of namespacing them.

bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,165 @@
...
+    def stream_datum(self, doc):
+        descriptor_node = self._descriptor_nodes[doc["descriptor"]]
+        parent_node = descriptor_node["external"]
+
+        num_rows = (
+            doc["indices"]["stop"] - doc["indices"]["start"]
+        )  # Number of rows added by new StreamDatum
+
+        # Get the Stream Resource node if it already exists or register if from a cached SR document
+        try:
```
danielballan, line 104: Elsewhere, the decision of "condition vs exception" came up. This is another place where using a condition might be cleaner.

genematx: Done!

bluesky/callbacks/tiled_writer.py
genematx: This is something I was going to add last Friday just before I was caught up with flyscanning at the lab. Added now.

genematx: I have named the entire table `/internal/data`, and the column names are taken directly from the corresponding data keys: `{key1}, {key2}, ..., ts_{key1}, ts_{key2}, ...`. This seems the most compact to me, but happy to change it to another schema.

danielballan: I think that's probably best. We technically have a problem if someone names a field `ts_`, but I think this is a time to be pragmatic and say, "Don't do that!" We can disallow it in the schema if we want to.

bluesky/callbacks/tiled_writer.py
danielballan: It could be the same file or a different file. It depends on the details of how the detector I/O works, whether it would open a new file or not. So, as far as Bluesky and Tiled go, we have to handle both possibilities.

If there _is_ a new file, there will be a new Stream Resource (and, in it, a new filepath). Given that, do we need to store this descriptor reference? I'm not sure it's needed, but I'm not sure it's _not_ needed....

> Would it make sense to have this list of descriptors stored in the top-most container (along with the Start/Stop messages) instead of its own node?

I think not, but let's examine this because I'm actually not sure what the best arguments are, and maybe there are better arguments in the other directions. Off the top of my head:

- It's easier to mix things than un-mix them, so I have a general impulse to place metadata on the specific thing it describes, in this case "the stream". Consumers can always hoist and flatten.
- The Event Descriptors can be large. For someone who is looping through a large number of scans and displaying basic metadata, it helps performance to only download the Start and Stop documents that describe the whole thing and defer downloading information about the streams.

bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,169 @@
...
+                f"Stream Resource {SR_uid} is referenced before being declared."
+            )
+
+        # Append StreamDatum to an existing StreamResource (by overwriting it with changed shape)
+        url = SR_node.uri.replace("/metadata/", "/data_source/")
+        SR_node.refresh()
+        ds_dict = SR_node.data_sources()[0]
+        ds_dict["structure"]["shape"][0] += num_rows
+        ds_dict["structure"]["chunks"][0].append(num_rows)
+        is_chunked = SR_node.metadata["resource_kwargs"].get("chunked", True)
```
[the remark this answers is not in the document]
danielballan: Yes, `event_model` is where we keep schemas for the documents. The details of `resource_kwargs` (to be renamed `parameters`) will be different for each `spec` (to be renamed `mimetype`)---that is, each filetype. When we factor the HDF5-specific aspects of this into separate classes, I think those classes should define schemas and/or validation methods.

This idea has been [outstanding for awhile](https://github.com/bluesky/area-detector-handlers/issues/6) and it would be great to pick it up!

Regarding `chunked`: I think this _is_ the right place to put this. I'm wondering whether a boolean is too coarse. I think what we're specifying here is the chunk size along the time axis. Did you consider using `chunk_size: Optional[int]`, where you can give:

- `chunk_size=1` -- one HDF5 chunk per frame
- `chunk_size=N` -- N frames into each HDF5 chunk
- `chunk_size=None` -- all frames into one HDF5 chunk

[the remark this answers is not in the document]
genematx: I actually thought about this too; I agree, boolean might be too restrictive. I wasn't sure if we also need to consider variable-size chunks, but I like the idea of `chunk_size` -- it's way less ambiguous. Perhaps we can assume for now that it's either 1, None, or fixed.

bluesky/callbacks/tiled_writer.py
genematx: Thank you for explaining this, Dan. I had another thought about this last night, and I agree, mixing descriptors and start/stop messages is probably a bad idea.

I'm trying to get a better understanding of the underlying data structure and possible use cases, and so far I have the following image in my head.
Bluesky_Messages_Relations

It seems to me that if different parts of the same stream resource (different Stream Datums: either rows of an HDF5 file, sets of tiffs from a folder, etc) can be related to different descriptors, we need some way to keep track of their indices _within_ the descriptor container, so paths like `/run/descriptor/dataset?slice=5:10` would resolve to the correct rows in the file, which could be e.g. `205:210`, or maybe not even be consecutive. I think `seq_nums` and `indices` in StreamDatum serve this purpose, but currently I am not storing them anywhere. (Should there be a lookup table or something like that?)
On the other hand, can we rely on individual Stream Datums alone to determine this correspondence between resource and descriptor indices? Do we need any kind of validation?

danielballan:
Capturing notes from a chat this morning, remaining steps on this PR are:

- [ ] Multiple descriptors
- [ ] Refactor HDF5-specifics from TiledWriter to "Handlers"
- [ ] Deal with mapping raw datset in HDF5 onto Event semantics (shape, skips)

bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
genematx: Updated this to allow for variable-sized chunks as well.

bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,206 @@
...
+        self.start_stop_node.update_metadata(metadata=metadata)
+
+    def descriptor(self, doc):
+        desc_name = doc["name"]
+        metadata = dict(doc)
+
+        # Extract the variable fileds of the descriptor that can change during the run
+        uid = metadata.pop("uid")
+        conf_dict = {uid: metadata.pop("configuration", {})}
+        time_dict = {uid: metadata.pop("time", None)}
```
danielballan, line 74: All EventDescriptor documents have always had `time`, so it's same to assume this will be present and that the values in `time_dict` will never be `None`.

```suggestion
        time_dict = {uid: metadata.pop("time")}
```

bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,206 @@
...
+    def __call__(self, name, doc):
+        self._run_router(name, doc)
+
+
+class _RunWriter(DocumentRouter):
+    "Write the document from one Bluesky Run into Tiled."
+
+    def __init__(self, client):
+        self.client = client
+        self.start_stop_node = None
```
danielballan, line 49: Maybe `run_node` or `root_node` here? Just an idea, not a strong opinion.

bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,206 @@
...
+        if desc_name in self.start_stop_node.keys():
+            desc_node = self.start_stop_nodes[desc_name]
+        else:
+            desc_node = self.start_stop_node.create_container(
+                key=desc_name, metadata=metadata
+            )
+            desc_node.create_container(key="external")
+            desc_node.create_container(key="internal")
+
+        # Update the metadata with the variable fields
```
danielballan, line 87: I like the way that this separates the variable and fixed information. That aspect is definitely an improvement.

I am not sure about the way this seems to be overwriting the variable information from Descriptor 1 with the variable information from Descriptor 2. If there are N events with one exposure time and then M events with a different exposure time, we need to write down both exposure times.

bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,206 @@
...
+        metadata = deep_update(dict(desc_node.metadata), var_fields)
+        desc_node.update_metadata(metadata)
+        self._descriptor_nodes[uid] = desc_node  # Keep a reference to the descriptor
```
danielballan, line 90: Should this move into the `else` block, for clarity? In the other case I think it is or should be a no-op.

bluesky/callbacks/tiled_writer.py
genematx: oh, `root_node` is so much better, thanks!

bluesky/callbacks/tiled_writer.py
genematx: Great, that's what I thought. Just added `None` to err on the safe side.

bluesky/callbacks/tiled_writer.py
genematx: Yes, I agree. The idea was to keep Configurations and Times from both descriptors and "overwrite" only the fixed information (it's going to be the same, so it doesn't matter -- there will be no overwriting effectively). That's why I define `conf_dict` and `time_dict` that store the variable information indexed with keys that are `uids` of specific descriptors. `pydantic.deep_update` will update these internal dictionaries and keep the previous values that have different keys.
Sorry, it probably wasn't clear because I reused the `metadata` variable and it looked like it was overwritten. I'll refactor this bit and add some comments.

bluesky/callbacks/tiled_writer.py
genematx: Same as above. The idea was that the `else` block would run first, create the Descriptor node, and populate it with (fixed) metadata. Then, the variable part of the metadata will be added after the condition.
Next time, when the descriptor is updated, there's no need to rewrite the fixed metadata (hence we only pull the node in the direct path of the condition `desc_node = self.root_nodes[desc_name]`, but the after the entire `if .. else` block the variable part is added like before).
Again, probably worth expanding this part of the code a bit more; i'll look into this.

pyproject.toml
```diff
@@ -34,6 +34,7 @@ requires-python = ">=3.8"
 
 [project.optional-dependencies]
 dev = [
+    "area-detector-handlers",
```
danielballan, line 37: I think this can be dropped.

pyproject.toml
```diff
@@ -83,6 +84,7 @@ dev = [
     "tifffile",
     "tox-direct",
     "types-mock",
+    "tiled[all]",
```
danielballan, line 87: Not sure this is right; I think this might not play well with `pip install .[dev]` for example. I think our options are:

- Recommend `pip install ,[dev,all]` using `all` explicitly, or
- Copy all the requirements from `all` into `dev.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,381 @@
+from abc import abstractmethod
+from typing import Optional
+
+import numpy as np
+import pandas as pd
+from area_detector_handlers import HandlerBase
```
danielballan, line 6: We don't need this. The methods its provides (`__enter__`, `__exit__`, `close`) are not applicable to this use case.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,381 @@
...
+# - instance state to track things like num_total_rows.
+
+# Different:
+# - The "at rest" use case has a conceptual dependency on
+#   DataSource Asset.
+# - The streaming data use case has an import dependency o
+#   whatever the I/O library is (e.g. h5py).
+
+
+class Thing:
```
danielballan, line 63: We should remember to drop this sketch class.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,381 @@
...
+    def __init__(self, data_uri, swmr):  # TODO Add reshape, skips.
+        ...
+
+    def read_stream_datum(self, *stream_datum_docs):
+        array = ...
+        return array
+
+
+class StreamHandlerBase(HandlerBase):
```
danielballan, line 72: Drop `HandlerBase` as discussed.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,381 @@
...
+        sr_0 = sr_node.data_sources()[0]
+        data_source = handler.get_data_source()
+        data_source.id = sr_0["id"]
+        endpoint = sr_node.uri.replace("/metadata/", "/data_source/", 1)
+        handle_error(
+            sr_node.context.http_client.put(
+                endpoint,
+                content=safe_json_dump({"data_source": data_source}),
+                headers={"Accept": MSGPACK_MIME_TYPE},
+                params={"data_source": 1},
```
danielballan, line 379: This can be dropped.

danielballan:
Per discussion with @maffettone and @genematx:

- The "handlers" in this PR are not doing I/O.
- It seems worth exploring reusing the Adapters in `tiled.adapters` to do I/O, such that a "handler" from this PR can work together an an Adapter to do streaming I/O.
- Given that, we should probably call these something other than "handler" because they are serving a different purpose.

Adapter: DataSource/Asset -> numpy (or pandas or whatever...)
Handler: Documents -> numpy
Thing in this PR: Documents -> DataSource/Asset

src/bluesky/callbacks/tiled_writer.py
genematx: Done

src/bluesky/callbacks/tiled_writer.py
genematx: Done

src/bluesky/callbacks/tiled_writer.py
genematx: Removed

pyproject.toml
genematx: Doing `tiled[all,dev]`

src/bluesky/callbacks/tiled_writer.py
genematx: Sure! It was just a reminder for me to think about the IO.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,378 @@
...
+            dim0_chunk = [self.chunk_size] * int(self._num_rows / self.chunk_size)
+            if self._num_rows % self.chunk_size:
+                dim0_chunk.append(self._num_rows % self.chunk_size)
+
+        return [dim0_chunk] + [[d] for d in self.data_shape]
+
+    def consume_stream_datum(self, doc):
+        """Append StreamDatum to an existing Handler (StreamResource)"""
+        # TODO Add reshape, skips.
+        self._num_rows += doc["indices"]["stop"] - doc["indices"]["start"]  # Number of rows added by StreamDatum
```
DiamondJoseph, line 124: AFAIK _num_rows should be doc["seq_nums"] stop - start, the data pointed to within the h5 file is the indices

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,378 @@
...
+        sr_0 = sr_node.data_sources()[0]
+        data_source = handler.get_data_source()
+        data_source.id = sr_0["id"]
+        endpoint = sr_node.uri.replace("/metadata/", "/data_source/", 1)
+        handle_error(
+            sr_node.context.http_client.put(
+                endpoint,
+                content=safe_json_dump({"data_source": data_source}),
+            )
+        ).json()
```
DiamondJoseph, line 378: Handling for non-Stream Resource/Datum?

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,378 @@
...
+    "Write the document from one Bluesky Run into Tiled."
+
+    def __init__(self, client):
+        self.client = client
+        self.root_node = None
+        self._desc_nodes = {}  # references to descriptor containers by their uid's
+        self._sr_nodes = {}
+        self._sr_cache = {}
+        self._handlers = {}
+
```
DiamondJoseph, line 241: These would benefit from type hints, as would the change generally: is _desc_nodes presumably a dict[Descriptor, list[Event]]?

DiamondJoseph (changes requested):

danielballan:
This diagram shows how this new component (marked as "???") fits in to everything else:

image


From brainstorm with DLS on what to call this:

1. Compactor
2. Condenser
3. Aggregator
4. Assembler
5. Distiller
6. Consolidator
7. Munger

My sense from the discussion is that (2), (3), (5), and (6) were leading contenders, but this is still a wide open race.

prjemian:
All of these names suggest some manipulation of the documents. Suggest
something more on point for what happens.  `Collector` is one Idea.

prjemian:
`receiver` is something I use in other code as the callback method.  It
receives the documents from the `RE`.  If you wish, capitalize it.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,378 @@
...
+from tiled.client.utils import handle_error
+from tiled.structures.array import ArrayStructure, BuiltinDtype
+from tiled.structures.core import Spec, StructureFamily
+from tiled.structures.data_source import Asset, DataSource, Management
+from tiled.structures.table import TableStructure
+from tiled.utils import safe_json_dump
+
+MIMETYPE_LOOKUP = {
+    "hdf5": "application/x-hdf5",
+    "ADHDF5_SWMR_STREAM": "application/x-hdf5",
```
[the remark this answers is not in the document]
genematx: I think, I must have taken this from [here](https://github.com/bluesky/bluesky/blob/add624e1b7f03c9b4b981e78dfc2333ed01e1f0c/src/bluesky/tests/test_external_assets_and_paging.py#L140), but I will double-check what `spec` we use, and possibly just add them both to the lookup table.

src/bluesky/tests/test_tiled_writer.py
```diff
@@ -0,0 +1,206 @@
...
+    HasName,
+    Readable,
+    Reading,
+    StreamAsset,
+    WritesStreamAssets,
+)
+
+
+@pytest.fixture
+def catalog(tmpdir):
```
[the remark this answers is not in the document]
genematx: Done

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,378 @@
...
+        array = ...
+        return array
+
+
+class StreamHandlerBase:
+    mimetype: str = ""
+    _num_rows: int = 0
+
+    def __init__(self):
+        pass
```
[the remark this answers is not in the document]
genematx: Thank you, Joseph! I really like this idea. Have refactored the StreamHandler base class and it's subclasses.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,363 @@
...
+    mimetype: str = ""
+
+    def __init__(self, sres: StreamResource, desc: EventDescriptor):
+        assert (
+            self.mimetype == sres["mimetype"]
+        ), f"A data source of {sres['mimetype']} mimetype can not be handled with {self.__class__.__name__}."
+
+        self.data_key = sres["data_key"]
+        self.uri = sres["uri"]
+        self.sres_parameters = sres["parameters"]
```
genematx, line 57: Do we need to keep a copy of the entire StreamResource document in the handler? I'd say it is unnecessary since we are already parsing all required information.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,363 @@
...
+        - ingesting a new mimetype from Bluesky documents and generating DataSource and Asset with appropriate
+          parameters (the handler's job);
+        - Interpreting those DataSource and Asset parameters to do I/O (the adapter's job).
+    """
+
+    mimetype: str = ""
+
+    def __init__(self, sres: StreamResource, desc: EventDescriptor):
+        assert (
+            self.mimetype == sres["mimetype"]
```
danielballan, line 52: It should be possible for one handler to support more than one mimetype. (Perhaps it should be a `set`?) This could come up if we rename something, or if there are a range of acceptable conventions, such as `application/x-hdf` and `application/x-hdf5`.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,363 @@
...
+    together to support:
+        - ingesting a new mimetype from Bluesky documents and generating DataSource and Asset with appropriate
+          parameters (the handler's job);
+        - Interpreting those DataSource and Asset parameters to do I/O (the adapter's job).
+    """
+
+    mimetype: str = ""
+
+    def __init__(self, sres: StreamResource, desc: EventDescriptor):
+        assert (
```
danielballan, line 51: I have read that `assert` should not be used to validate user inputs. Instead a `ValueError` should be raised.

One good reason for this is that python can be run in a mode where all `assert` statements are ignored, for performance.

src/bluesky/callbacks/tiled_writer.py
danielballan: I'd like to tackle that in a separate PR.

src/bluesky/callbacks/tiled_writer.py
genematx: Thank you for noting this, Dan -- this assertion was something else I wanted to ask you.
I read similar notes too somewhere, but I've also seen `assert` used to raise exceptions throughout the Bluesky code quite often; I thought it was a matter of style, but I might be missing some subtle differences here. Personally, I'd also prefer raising an exception explicitly, and I have refactored this part.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,378 @@
...
+        self.data_shape = data_shape  # Intrinsic data shape (single frame as written in HDF5 file)
+        self.data_type = np.dtype(data_type)
+        self.chunk_size = chunk_size
+        self.adapter_parameters = {"path": self.path}
+
+    @property
+    def shape(self):
+        """Shape including the 0-th dimension -- number of rows"""
+        # TODO Add reshape, skips.
+        return [self._num_rows, *self.data_shape]
```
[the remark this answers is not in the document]
genematx: Great suggestion, thank you! It doesn't even have to be abstract.

src/bluesky/callbacks/tiled_writer.py
genematx: Sure, changed the definition to `mimetype: Union[str, Set[str]]`.
What about 'DataSource' in Tiled? Should it also accept sets of mimetypes or simply strings?

Then, it also seems to me that there should be _two_ mimetype attributes associated with each StreamHandler: one for handler discovery (possibly a `set`), and another -- the actual mimetype determined by the StreamResource (always a string, matching/included in the first mimetype attribute) and passed to DataSource. I will think how to simplify this.

src/bluesky/callbacks/tiled_writer.py
danielballan: Assert to check conditions that could only be false due a fully internal bug is OK. I'm mostly concerned with asserts that could be false due to erroneous user inputs.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,378 @@
...
+MIMETYPE_LOOKUP = {
+    "hdf5": "application/x-hdf5",
+    "ADHDF5_SWMR_STREAM": "application/x-hdf5",
+    "AD_HDF5_SWMR_SLICE": "application/x-hdf5",
+    "AD_TIFF": "multipart/related;type=image/tiff",
+}
+DTYPE_LOOKUP = {"number": "<f8", "array": "<f8", "boolean": "bool", "string": "str", "integer": "int"}
+
+
+# TODO (probably in a future PR?):
```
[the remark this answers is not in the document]
genematx: Created an Issue #1740 and put the link in the code.

src/bluesky/callbacks/tiled_writer.py
genematx: That's right, but `_num_rows` here corresponds to the total number of rows in the Data Source (file), including skips. I have added a comment to clarify this.

src/bluesky/callbacks/tiled_writer.py
genematx: Fixed this and added more type hints throughout.

src/bluesky/callbacks/tiled_writer.py
genematx: I allow a (sub-)class of `StreamHandlerBase` to accept several mimetypes (from a set) but overwrite the same field `self.mimetype` with a specific string value passed as mimetype in the Stream Resource. I'm not completely happy about the overwrite, but in this case we still have a single attribute responsible for mimetype, it contains only the relevant value for a specific instance, and -- before instantiation -- allows for handler discovery based on sets of mimetypes.

src/bluesky/callbacks/tiled_writer.py
genematx: Fixed.

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
genematx: Added both.

danielballan:
To belatedly respond to @prjemian, I really like the word "Collector" for a callback that collects documents. This particular thing we are trying to name does something more specific. I does _not_ retain a collection of the documents it sees. Instead, it extracts some information and consolidates it into a new representation, optimized for random and batch access, meant to complement the streaming representation.

---

I move to choose "Consolidator" as the choice that IMO (1) does not already have a conflicting technical meaning and (2) is not too silly.

DiamondJoseph (commented):
realised I didn't publish these before

pyproject.toml
```diff
@@ -84,6 +84,7 @@ dev = [
     "tifffile",
     "tox-direct",
     "types-mock",
+    "tiled[all,dev]",
```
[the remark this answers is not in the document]
danielballan: I had the same comment in my draft review, not yet submitted. :handshake:

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,391 @@
...
+        self._num_rows: int = 0  # Number of rows in the Data Source (all rows, includung skips)
+        self._has_skips: bool = False
+        self._seqnums_to_indices_map: Dict[int, int] = {}
+
+    def _validate_mimetype(self, mimetype):
+        if isinstance(self.mimetype, str) and (mimetype == self.mimetype):
+            return None
+        elif isinstance(self.mimetype, set) and (mimetype in self.mimetype):
+            self.mimetype = mimetype
+            return None
```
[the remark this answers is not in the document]
danielballan: Ha, I also had exactly this suggestion in my draft review, even down to the name. :100:

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,391 @@
...
+            metadata={"start": doc},
+            specs=[Spec("BlueskyRun", version="1.0")],
+        )
+
+    def stop(self, doc):
+        if self.root_node is None:
+            raise RuntimeError("RunWriter is properly initialized: no Start document has been recorded.")
+
+        doc = dict(doc)
+        doc.pop("start", None)
```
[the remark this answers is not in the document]
danielballan: It's not a requirement. Removing it would serve "one source of truth" but I think that there is a case to be made for retaining the complete document in its original form here, to be make it easy to reconstruct the documents correctly, and also for back-compat.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,391 @@
...
+
+        self._num_rows: int = 0  # Number of rows in the Data Source (all rows, includung skips)
+        self._has_skips: bool = False
+        self._seqnums_to_indices_map: Dict[int, int] = {}
+
+    def _validate_mimetype(self, mimetype):
+        if isinstance(self.mimetype, str) and (mimetype == self.mimetype):
+            return None
+        elif isinstance(self.mimetype, set) and (mimetype in self.mimetype):
+            self.mimetype = mimetype
```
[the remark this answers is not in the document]
danielballan: Similar to @DiamondJoseph's comment

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,391 @@
...
+        self.root_node.update_metadata(metadata=metadata)
+
+    def descriptor(self, doc: EventDescriptor):
+        if self.root_node is None:
+            raise RuntimeError("RunWriter is properly initialized: no Start document has been recorded.")
+
+        desc_name = doc["name"]
+        metadata = dict(doc)
+
+        # Remove variable fields of the metadata and encapsulate them into sub-dictionaries with uids as the keys
```
danielballan, line 291: Just noting that I think we may want to take a critical look at how the information from the Event Descriptor is encoded in `metadata` at some point, but that may best be part of a larger discussion about the whole structure here. I want to say that I think this is a good approach, and probably not the final approach, and that it's too early to know what the right final approach is, so we should move ahead.

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
genematx: Thank you, Joseph, I think it's a great idea. I wasn't happy about overwriting the class fields, but this handles it nicely.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,391 @@
...
+
+    To implement new StreamHandlers for other mimetypes, subclass StreamHandlerBase, possibly expand the
+    `consume_stream_datum` and `get_data_source` methods, and ensure that the returned the `adapter_parameters`
+    property matches the expected adapter signature. Declare a specific mimetype or a set of acceptable mimetypes
+    to allow valiadtion and automated discovery of the subclassed StreamHandler.
+    """
+
+    mimetype: Union[str, Set[str]] = ""
+
+    def __init__(self, sres: StreamResource, desc: EventDescriptor):
```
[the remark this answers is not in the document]
genematx: Done!

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
genematx: Agree. Done!

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,391 @@
...
+
+    def __init__(self, client: BaseClient):
+        self.client = client
+        self.root_node: Union[None, Container] = None
+        self._desc_nodes: Dict[str, Container] = {}  # references to descriptor containers by their uid's
+        self._sres_nodes: Dict[str, BaseClient] = {}
+        self._sres_cache: Dict[str, StreamResource] = {}
+        self._handlers: Dict[str, StreamHandlerBase] = {}
+
+    def _ensure_sres_backcompat(self, sres: StreamResource) -> StreamResource:
```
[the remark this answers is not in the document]
genematx: Sure, no problem. Fixed it.

This was kind of the idea -- to mutate the document bringing it to the new format, but I haven't realized that there might be any _old-style_ consumers ingesting the same document, in no consistent order.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -0,0 +1,391 @@
...
+            num=len(self.assets) + 1,
+        )
+        self.assets.append(new_asset)
+        self.data_uris.append(new_datum_uri)
+
+        super().consume_stream_datum(doc)
+
+    @property
+    def adapter_parameters(self) -> Dict:
+        return {"data_uris": self.data_uris}
```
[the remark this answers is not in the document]
genematx: ah, you are right! `TiffSequenceAdapter` in Tiled needs to be initialized with `data_uris`, but I think it gets them from the `Assets` directly.
I also agree that `indx:05d.tif` is awful, but I think it needs to be generalized _before_ creating `adapter_parameters` (likely when we declare Assests).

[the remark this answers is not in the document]
genematx: I've generalized the filename template for tiff sequences. This relies on the formatting string being passed in the Stream Resource document using Python-specific style (`{:05d}.tif`). This is a bit of limitation of course, but I think it should be okay since the `TIFFStreamHandler` is a custom class and can be redefined for a new use case, if needed.
Please let me know if anyone has a better idea how to handle the sequential filenames in a more general way. Maybe there is a widely accepted standard already (e.g. with %) that we can just convert to Python representation.

src/bluesky/callbacks/olog.py
```diff
@@ -173,5 +173,5 @@ def start(self, doc):
             pformat(doc),
         )
         olog_status = self.client.log(document_content, logbooks=self.logbook)
-        logger.debug("client.log returned %s" % olog_status)
+        logger.debug(f"client.log returned {olog_status}")
```
prjemian, line 176:
```suggestion
        logger.debug("client.log returned %s", olog_status)
```

pyproject.toml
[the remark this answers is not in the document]
genematx: That's a good point, thank you. We definitely need client- and server-side dependencies, but `dev` is probably not necessary. I tried different options gradually including missing packages, and I think after `tiled[minimal-client,minimal-server,formats,xarray,dataframe]`, it probably makes sense just to install it as `tiled[all]` (or maybe create another bundle of tiled dependencies needed for bluesky, 'tiled[bluesky]', or something like that?).
I have removed the `dev` flag though.

src/bluesky/callbacks/olog.py
genematx: I've only made this change because this line (and 5 others) didn't pass the linter checks. I'm happy to just silence it with `# noqa: UP031`, but I am curious what is the reason we can't use f-strings here since, as far as I understand, we declare support for only Python 3.8+? Are there any back-compatibility requirements that I'm missing?

prjemian: See, for example, this [article](https://dev.to/izabelakowal/what-is-the-best-string-formatting-technique-for-logging-in-python-d1d).

The methods of the logging package have different arguments than `print()`.  The first argument is the format string.  Subsequent arguments are various python objects to be used with the format string _if_ the method is to generate its output.  By using an f-string (which is allowed since it suffices as a format string), you are circumventing the decision process in the logging method.  That process can choose not to create an output based on the logging level.

tl;dr:  f-strings in logging defeat an optimization based on the log level

prjemian: BTW, `"%s" % item` is the same as `f"{item}"`, should use `"%s", item` instead.  It's the comma that cooperates with the optimization in logging.

genematx: ah right, thanks! I haven't realized it was in `logging.debug`, not in `print`. totally makes sense!

prjemian: I expect the linter would flag `"%s" % item` for improvement.

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
genematx: Yes, the main reason to pop the uid of the Start document from the stored Stop was to eliminate redundancy, but I also completely agree with the argument in favour of maintaining consistency between StartDocument and StopDocument.

On the other hand, what I personally find somewhat confusing is that same keys refer to different types of values depending on the dictionary/structure. E.g., in the metadata of Tiled's `root_node`, `start` and `stop` keys refer to the corresponding actual documents, but in the StopDocument itself, `start` reference the _id_ of the StartDocument, and by extension -- the entire BlueskyRun. Similarly `descriptor` key in StreamResources really contains `descriptor_id`, and so on. One way to solve this would be by calling the fields in Tiled metadata `start_document`and `stop_document`, but this seems to be overly verbose and wouldn't work for descriptors, where the entire metadata dictionary of the same Tiled node depends on multiple descriptor documents, in general.

I have reinstated the `start` field in the saved StopDocuments to keep it in its original form, as suggested, but this part likely needs a closer look and warrants a further discussion in line with Dan's [comment](https://github.com/bluesky/bluesky/pull/1660#discussion_r1587763803) below. My preference would be to keep as much information from the Bluesky documents as possible, but to parse them before storing in the Tiled database and remove all duplications. I agree that reconstructing the original documents would require some work, but it wouldn't be very difficult.

prjemian:
CI for Py3.9 failed yet the others passed.  How come?

2024-05-02T20:57:03.6772906Z FAILED src/bluesky/tests/test_suspenders.py::test_unresumable_suspend_fail[True] - assert 0.1 < (1714683313.233679 - 1714683313.1350915)
2024-05-02T20:57:03.6774312Z = 1 failed, 1671 passed, 15 skipped, 18 xfailed, 16 xpassed in 1153.81s (0:19:13) =
2024-05-02T20:57:18.3210980Z ERROR: InvocationError for command /opt/hostedtoolcache/Python/3.9.19/x64/bin/pytest --cov=bluesky --cov-report term --cov-report xml:cov.xml (exited with code 1)

genematx:
@prjemian I've seen this before (failing tests for only a single Python version). What helped in the past is just to rerun the failed job, but I haven't yet looked into the root cause of the issue (perhaps something strange with GH runners...).

danielballan (approved):

DiamondJoseph (approved):

prjemian:
As I understand it, this PR provides capability essential to connect a tiled server as the client for the stream of published documents from the bluesky RunEngine.  

Here is an example in the test suite (where `tw` is an object of the new `TiledWriter()` class):
https://github.com/bluesky/bluesky/blob/9c2c2d526b39f007e52d86eafe20279ba889ad46/src/bluesky/tests/test_tiled_writer.py#L187

This provides a clear path for beamlines currently using databroker v1.2.x, that rely on `databroker.catalog.v1.insert()` method to receive such documents.  This method is no longer available in databroker v2.

prjemian (approved):
Unit tests demonstrate how this new code is used.  Then, a tiled client is tested that expected data has been received and made available.  Nice!

https://github.com/bluesky/bluesky/blob/9c2c2d526b39f007e52d86eafe20279ba889ad46/src/bluesky/tests/test_tiled_writer.py#L188-L194

PR #1661, opened 2024-03-05
Pass `md_validator` a deep copy that it can normalize

jmaruland:
This PR adds a temporary fix to let us run the kafka consumers that run the prefect end of run workflow from the beamline


tacaswell:
Seems correct, but we need some updated documentation about what `md_validator` can do.

I might err on the side of expecting the validators to be good about not mutating their input rather than preemptively doing it for them.

I am struggling to remember why we went with "you can only raise" initially, I think it may have been related to the complexities of ensuring that we did not mutate the input!

danielballan:
I definitely did not appreciate initially (~2016) that it is frequently useful for validators to just fix the problem if they can and return normalized version. I may well not have seen that pattern yet at the time.

danielballan:
That's an interesting observation. There precedent for [widely-used "validators" that modify/normalize the input](https://docs.pydantic.dev/latest/concepts/validators/). But there is a risk that this PR could break existing uses in unexpected ways.

I wonder if we should instead just add `md_normalizer` in this PR and leave `md_validator` unchanged. If people agree that this a good path, perhaps @jmaruland could add some commit here to make it so?

tacaswell:
I am :+1: on adding `md_normalizer` (raising if both are passed) and 50/50 on deprecating `md_validator` (eventually?).

prjemian:
I like the pivot from the name `validator` (delivers a judgement) to `normalizer` (intends to take an action).

jmaruland:
@danielballan, @tacaswell, @mrakitin and @prjemian I went through your comments and added an approach that collects your ideas. Thank you for sharing your thoughts about this PR

bluesky/run_engine.py
```diff
@@ -1776,12 +1790,13 @@ async def _open_run(self, msg):
                        'plan_name': plan_name},
                       self.md)  # stateful, persistent metadata
         # The metadata is final. Validate it now, at the last moment.
-        # Use copy for some reasonable (admittedly not total) protection
-        # against users mutating the md with their validator.
         self.md_validator(dict(md))
 
+        # Apply normalizer at the same level of the validator
+        validated = self.md_normalizer(md)
```
dmgav, line 1796: Should `copy.deepcopy` be applied before calling the normalizer?
```
validated = self.md_normalizer(copy.deepcopy(md))
```

bluesky/run_engine.py
```diff
@@ -2707,6 +2722,21 @@ def _default_md_validator(md):
             "BAD: sample=[1, 2] ")
 
 
+def _default_md_normalizer(md):
```
danielballan, line 2725: The validator has this handled, so I think we the default normalizer can just be a no-op, i.e. `return md`. The copying should happen first, agreeing with @dmgav's comment above.

dmgav: My idea was that copying should not be delegated to whoever implements the normalizer.

danielballan: Yeah, we are agreeing.

I meant that the sample validation below here is already handled by the default validator and need not be copied into the normalizer also. Sorry, I wrote this in an unclear way.

bluesky/tests/test_metadata.py
```diff
@@ -8,3 +9,61 @@ def test_blueskyversion(RE):
...
+    assert start_doc["test"] == 1
+
+
+def test_md_mormalizer(RE):
+    """
+    Test the new md_normalizer.
+
+    In current versions of bluesky, md_normalizer is introduced to run
+    alongside md_validator. It returns a normalized valid result,
+    or raise if it is unable to provide one.
```
mrakitin, line 46:
```suggestion
    or raises if it is unable to provide one.
```

PR #1663, opened 2024-03-07
Alter ko and complete plan stubs to take multiple flyables

abbiemery:
This adds the multiple object functionality to the kickoff and complete plan stubs. This will neaten up the plans.

### DIFF
--- a/src/bluesky/plan_stubs.py
+++ b/src/bluesky/plan_stubs.py
@@ -5,7 +5,7 @@ import uuid
 import warnings
 from collections.abc import Iterable
 from functools import reduce
-from typing import Dict
+from typing import List
 
 from cycler import cycler
 
@@ -623,15 +623,13 @@ def prepare(obj, *args, group=None, wait=False, **kwargs):
     return ret
 
 
-def kickoff(obj, *args, group=None, wait=False, **kwargs):
+def kickoff(obj, *, group=None, wait=False, **kwargs):
     """
-    Kickoff one or more fly-scanning devices.
+    Kickoff one fly-scanning device.
 
     Parameters
     ----------
-    obj : fly-able
-        Device with 'kickoff', and 'complete' methods.
-    *args : Any Flyables passed in as args will be kicked off along with obj.
+    obj : fly-able Device with 'kickoff', and 'complete' methods.
     group : string (or any hashable object), optional
         identifier used by 'wait'.
     wait : boolean, optional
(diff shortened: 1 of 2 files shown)


danielballan:
I have been pondering the trade offs of making `msg.obj` a list (in some cases) or packing additional objects into `msg.args`. I think this approach is the less messy option of the two.

bluesky/plan_stubs.py
```diff
@@ -650,15 +652,21 @@ def kickoff(obj, *, group=None, wait=False, **kwargs):
...
-        yield from _wait(group=group)
-    return ret
+    objs = [obj]+[check_supports(arg, Flyable) for arg in args]
+
+    statuses: Dict[Flyable, Status] = {}
+    for obj in objs:
+        ret = yield Msg('kickoff', obj, group=group, **kwargs)
+        statuses[obj] = ret
+        if wait:
+            yield from _wait(group=group)
```
coretl, line 662:
```suggestion
    if wait:
        yield from _wait(group=group)
```
Only wait once when we have kicked them all off

bluesky/plan_stubs.py
```diff
@@ -689,10 +698,16 @@ def complete(obj, *, group=None, wait=False, **kwargs):
...
-        yield from _wait(group=group)
-    return ret
+    objs = [obj]+[check_supports(arg, Flyable) for arg in args]
+
+    statuses: Dict[Flyable, Status] = {}
+    for obj in objs:
+        ret = yield Msg('complete', obj, group=group, **kwargs)
+        statuses[obj] = ret
+        if wait:
+            yield from _wait(group=group)
```
coretl, line 708: As above

coretl (dismissed):

abbiemery:
Changes made, but I feel like we should leave this one until the branch is merged and is in main. There is a log going in which needs a review, a lot of test cases and people who really need #1651 in, so we can make an alpha-release and can therefore make a proper ophyd-async release.

There is no real harm in leaving this, other than it makes the plans slightly messier. Something we can put in a following release.

coretl:
I think we are clear that we want to planstubs to kickoff (and complete) multiple devices at once. The questions are:
- Should `bps.kickoff` to work with both a single and multiple devices?
- If so, should the single case move from returning a single Status to returning a tuple of Statuses?
- If not, should we make a `bps.kickoff_all` like [`bps.stage_all`](https://github.com/bluesky/bluesky/blob/f03451a7e188df867df7fc86c70e3efe65a02709/src/bluesky/plan_stubs.py#L787)

@tacaswell @danielballan I feel like we had the "should `func(obj, *other_objs)` always return a tuple of return values, even if there are no `other_objs`?" conversation before, but I can't remember where...

coretl:
Decided on a `kickoff_all` and `complete_all` and then we can use different default args so it waits by default

PR #1672, opened 2024-03-25
MNT: numpy 2.0 compatibility

tacaswell:
Fixes two issues with numpy 2.0

 - use `np.prod` instead of `np.product`
 - simplify a test (that it was not really clear what it was testing ... or why that was being tested _here_)

### DIFF
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -235,7 +235,6 @@ class SigintHandler(SignalHandler):
         super().__init__(signal.SIGINT, log=RE.log)
         self.RE = RE
         self.last_sigint_time = None  # time most recent SIGINT was processed
-        self.num_sigints_processed = 0  # count SIGINTs processed
 
     def __enter__(self):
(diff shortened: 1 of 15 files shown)


tacaswell:
the single -> double quote change is the most annoying formatting nit.

tacaswell:
I propose that we merge this with the skiptest and then open an issues to debug what is going wrong with that test.

danielballan (approved):

src/bluesky/utils/__init__.py
```diff
@@ -629,8 +629,8 @@ def snake_cyclers(cyclers, snake_booleans):
         lengths.append(len(c))
     total_length = np.prod(lengths)
     for i, (c, snake) in enumerate(zip(cyclers, snake_booleans)):
-        num_tiles = np.prod(lengths[:i])
```
tacaswell, line 632: Somehow despite the commit message I must have messed up the rebase and changed it to the wrong one 🤦🏻

Issue #1693, opened 2024-04-15
Devcontainer: Matplotlib issue with collecting tests


callumforrester:
Using matplotlib with a tk backend, we see an error collecting the tests

The following people have reported, please paste your logs:
@AlexanderWells-diamond @noemifrisina @gilesknap

AlexanderWells-diamond:
```
$ tox -e tests -- --collect-only
...

=========================================================================== 1650 tests collected in 6.13s ===========================================================================
Exception ignored in atexit callback: <bound method Gcf.destroy_all of <class 'matplotlib._pylab_helpers.Gcf'>>
Traceback (most recent call last):
  File "/venv/lib/python3.10/site-packages/matplotlib/_pylab_helpers.py", line 82, in destroy_all
    manager.destroy()
  File "/venv/lib/python3.10/site-packages/matplotlib/backends/_backend_tk.py", line 565, in destroy
    self.canvas._tkcanvas.after_cancel(self.canvas._idle_draw_id)
  File "/usr/local/lib/python3.10/tkinter/__init__.py", line 871, in after_cancel
    data = self.tk.call('after', 'info', id)
RuntimeError: main thread is not in main loop
```

This happens inside of the Devcontainer for me. I haven't tried outside of it.

callumforrester:
Here's a workaround: `MPLBACKEND=agg tox -e tests`

AlexanderWells-diamond:
Another possible fix is to add these libraries to the `Dockerfile`:

```
RUN apt-get update && apt-get install -y --no-install-recommends \
    graphviz ffmpeg libsm6 libxext6 libegl1 libqt5gui5 libxcb-cursor0 -y \
    && rm -rf /var/lib/apt/lists/*
```

Note that this will cause matplotlib to start running using Qt5, rather than Tk, so it doesn't really fix the Tk issue...

callumforrester:
Could also combine this ticket with #1689 and get rid of the GUI during test collection altogether, unsure how @tacaswell will feel about that though.

callumforrester:
Reviewers: @tacaswell @gilesknap

tacaswell:
https://github.com/bluesky/bluesky/pull/1672 should fix the "not main thread" issue.

callumforrester:
Fixed by #1672 and #1700

PR #1695, opened 2024-04-15
fix for #1693

gilesknap:
This is a workaround for the tests failing in the developer container.

To reproduce the issue open the repo in vscode and choose 'reopen in container' then run `pytest` (or `tox -p`).

The changes fix two issues:

- The default tk backend for matplotlib was failing like this:
  ```
  File "/usr/local/lib/python3.10/tkinter/__init__.py", line 871, in after_cancel
    data = self.tk.call('after', 'info', id)
  RuntimeError: main thread is not in main loop
  ```
- The two tests with prefix `test_in_plan_qt_callback` required the system dependency libGL.so.1

FIXES:
-  `MPLBACKEND=agg` environment for headless operation 
- add libgl1-mesa-dev system dependency.

Making this a draft as others have added to the issue since I last looked!

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,7 +7,9 @@
     },
     "remoteEnv": {
         // Allow X11 apps to run inside the container
-        "DISPLAY": "${localEnv:DISPLAY}"
+        "DISPLAY": "${localEnv:DISPLAY}",
+        // set up the MatPlotLib backend
+        "MPLBACKEND": "agg"
     },
     "customizations": {
         "vscode": {
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,7 @@ FROM python:${PYTHON_VERSION} as developer
 # Add any system dependencies for the developer/build environment here
 RUN apt-get update && apt-get install -y --no-install-recommends \
     graphviz \
+    libgl1-mesa-dev \
     && rm -rf /var/lib/apt/lists/*
 
 # Set up a virtual environment and put it in PATH


tacaswell:
someone has to review #1672

tacaswell:
There is some integration with Qt that does not work with the other GUI frameworks (see https://github.com/bluesky/bluesky/blob/b6a2577d315449384b9bbd8c7325c888adb42eb8/src/bluesky/utils/__init__.py#L1633-L1725) but the non-qt tests should work with Agg (and the qt tests should gracefully skip if a qt binding is not available).

I don't have any good theories as to why collection takes much longer.

tacaswell:
I think I figured out why collection can be slow: pytest is importing (and hence running) https://github.com/bluesky/bluesky/blob/main/src/bluesky/tests/interactive/best_effort_cb.py 😱 with Agg I suspect all of the "draw_idle" calls are turning into "draw NOW" an so we are rendering a large number of unviewable plots during test discovery.

tacaswell:
https://github.com/bluesky/bluesky/pull/1653/commits/505a613ba83d1b4f9303a88e7b42580d5f0af82a is the fix for that.  I can charry-pick off to its own PR (or to here) if we want.

gilesknap:
Thanks, Tom. 

I take it that https://github.com/bluesky/bluesky/commit/505a613ba83d1b4f9303a88e7b42580d5f0af82a disables the interactive tests. This devcontainer is capable of doing the interactive tests (at least with X11, not sure about wayland).

I figure that the right answer is to support all of the tests if possible so I propose:

- I have a go at reviewing https://github.com/bluesky/bluesky/pull/1672. And verify that it fixes the "not main thread" issue.
- Verify that now all tests run with the default backend
- probably need to add Qt into this container for the qt specific tests
- Next address https://github.com/bluesky/bluesky/issues/1689.
  - For this I'll make sure we can still run the interactive tests

gilesknap:
OK I just changed my mind. https://github.com/bluesky/bluesky/commit/505a613ba83d1b4f9303a88e7b42580d5f0af82a just stops pytest from looking in the file during collection. Will cherry pick that!

gilesknap:
I'm closing this because
a) I did it in a branch instead of fork
b) I managed to get a root owned commit (via devcontainer issue)

I will make another PR from a fork.

Issue #1667, opened 2024-03-19
Add main module and version CLI


coretl:
We would like a similar structure to [the template on which bluesky's layout is based](https://github.com/DiamondLightSource/python-copier-template).

## Expected Behavior
`python -m bluesky --version` outputs the currently installed version of bluesky. 

## Possible Solution
Mimic the copier template linked above, adding a `__main__.py` that prints the current version when run. Also add a test that ensures it keeps working, similar to this one https://github.com/DiamondLightSource/python-copier-template/blob/main/template/tests/test_cli.py.jinja.

## Context

_Originally posted by @callumforrester in https://github.com/bluesky/bluesky/pull/1664#discussion_r1530567471_

callumforrester:
Reviewers: @coretl @mrakitin

PR #1699, opened 2024-04-15
[ENH] Add __main__.py and entrypoint for checking version info.

jwlodek:
## Description
Adding `__main__.py` with simple main entrypoint that prints version, update `pyproject.toml`.

## Motivation and Context

Solves #1667

## How Has This Been Tested?

```
(venv) [jwlodek@vdi-10-65-15-187 bluesky]$ which python
~/Workspace/bluesky/venv/bin/python
(venv) [jwlodek@vdi-10-65-15-187 bluesky]$ python --version
Python 3.11.5
(venv) [jwlodek@vdi-10-65-15-187 bluesky]$ python -m bluesky --version
1.13.0a4.dev31+g6168e9d0.d20240415
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,6 +96,7 @@ all = ["bluesky[dev,ipython,zmq,common,tools,streamz,plotting,cmd,olog]"]
 
 [project.scripts]
 bluesky-0MQ-proxy = "bluesky.commandline.zmq_proxy:main"
+bluesky = "bluesky.__main__:main"
 
 [project.urls]
 GitHub = "https://github.com/bluesky/bluesky"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,7 +96,6 @@ all = ["bluesky[dev,ipython,zmq,common,tools,streamz,plotting,cmd,olog]"]
 
 [project.scripts]
 bluesky-0MQ-proxy = "bluesky.commandline.zmq_proxy:main"
-bluesky = "bluesky.__main__:main"
 
 [project.urls]
 GitHub = "https://github.com/bluesky/bluesky"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -96,6 +96,7 @@ all = ["bluesky[dev,ipython,zmq,common,tools,streamz,plotting,cmd,olog]"]
 
 [project.scripts]
 bluesky-0MQ-proxy = "bluesky.commandline.zmq_proxy:main"
+bluesky = "bluesky.__main__:main"
```
coretl, line 99: I wouldn't add it as a script, just so we can do `python -m bluesky -v`

src/bluesky/__main__.py
coretl, line 1: Please can you add a test like this to test:
https://github.com/DiamondLightSource/python-copier-template-example/blob/main/tests/test_cli.py

prjemian:
`python -m setuptools_scm` gives the version. Why is more code necessary?

coretl:
> `python -m setuptools_scm` gives the version. Why is more code necessary?

To make it extensible for the future? Also because argparse makes it easy this way...

prjemian:
> To make it extensible for the future? Also because argparse makes it easy this way...

Only adds code that duplicates existing functions.  Please do not add this now.  If it is wanted in the future, reconsider at that time.

prjemian:
Ok, I see the reason. But 100% agree with not adding a `bluesky` script just for this convenience.

jwlodek:
It seems that the python 3.11 test failure is unrelated?

coretl:
> It seems that the python 3.11 test failure is unrelated?

yes, #1707

coretl (approved):

PR #1700, opened 2024-04-16
Fix for #1693 - tests failing in devcontainer

gilesknap:
This is a  fix for the tests failing in the developer container.

To reproduce the issue open the repo in vscode and choose 'reopen in container' then run pytest (or tox -p).

The fixes are:
- rebase over https://github.com/bluesky/bluesky/pull/1672 to fix "main thread is not in main loop"
- add qt library into system depenedencies

Note that this still has pop ups for the interactive tests but that will be looked at in a separate issue https://github.com/bluesky/bluesky/issues/1689. The devcontainer supports X11 for this purpose.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,7 +7,9 @@
     },
     "remoteEnv": {
         // Allow X11 apps to run inside the container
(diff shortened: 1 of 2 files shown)


callumforrester (approved):
LGTM

PR #1703, opened 2024-04-16
Dont collect interactive tests

### LINKED ISSUE #1689: Stop GUI Windows Popping up During Test Collection
When pytest collects the test, it runs some code on import that causes GUI windows to pop up. This is annoying as they can drag the focus of the desktop around and interfere with the user's workflow. 

## Possible Solution
@tacaswell suggests adding `pytest-xvfb` as a dev dependency, which will swallow the GUI windows. Unclear if this will also speed up collection, which is desirable, but can be a separate ticket.

gilesknap:
This fixes #1689.

In vscode pytest collection is run every time changes are made in the test folder and this was causing MPL windows to pop up. This was disruptive.

The change tells pytest not to collect from the interactive subfolder of tests. This has no effect on the number of pytest tests collected as there are no pytest functions in the folder. However, it does take away some checks as exceptions happening when loading these modules would previously have caused pytest to abort and report the issue (thus not running the remaining tests).

@tacaswell was the intention for these tests to be executed in CI - if so a different approach is required.

One thing to note: there are still a couple of popups when executing the tests but this could be suppressed using `MPLBACKEND=agg pytest`. This fix means that using the agg backend no longer slows down test collection.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,11 +112,17 @@ ignore_missing_imports = true # Ignore missing stubs in imported modules
 
 [tool.pytest.ini_options]
 # Run pytest with all our checkers, and don't spam us with massive tracebacks on error
+# Don't collect the interactive directory which is intended for manual execution
 addopts = """
-    --tb=native -vv --doctest-modules --doctest-glob="*.rst"
+    --tb=native -vv --doctest-modules --doctest-glob="*.rst" --ignore src/bluesky/tests/interactive
     """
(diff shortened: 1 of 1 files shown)


tacaswell:
I think something changed in how pytest does its discovery in the last 7 or so years as the offending file is https://github.com/bluesky/bluesky/blob/main/src/bluesky/tests/interactive/best_effort_cb.py 

I think that file was meant to be manually run and looked at with eye-balls (the same with everything in that folder).

I suspect that those are candidates to either be turned into examples in the docs or just removed.

tacaswell (approved):

Issue #1650, opened 2024-01-09
Promote DeprecationWarning to UserWarning so people will see it


coretl:
At the moment no-one sees this warning apart from in tests:
https://github.com/bluesky/bluesky/blob/0943efdc9ba8ecff9b5c59cb0f5a38d7bb8ef1c3/bluesky/utils/__init__.py#L1810

Promote it to a UserWarning so people see it

callumforrester:
Reviewers: @jwlodek

PR #1705, opened 2024-04-16
(#1650) promotes warn_if_msg to a UserWarning

CoePaul:
* utils module API function warn_if_msg_args_or_kwargs is hereby promoted from PendingDeprecationWarning

## Description
Promotes warning type to raise visibility from tests only
to every day coding

## Motivation and Context
Prior to this change, the warning it provides 
would only show up in tests ( and therefore in practice be ignored )

## How Has This Been Tested?
It hasn't yet - but the existing Utils module test
was updated to accommodate the change of warning type.
(Full testing is beyond the scope of the change)
Testing will happen when future code changes hit the new warning 
and it is visible


jwlodek:
Looks good to me

CoePaul:
Dear Reviewer(s), 
can the code coverage complaint be ignored ?
The diff coverage is 100 percent and so it would seem that any (hypothetical) tweak of test coverage more widely, 
would step outside the scope of the requested change.

PR #1707, opened 2024-04-16
TST: relax timing as 2s is failing on CI on py311 only

tacaswell:


src/bluesky/tests/test_run_engine.py
coretl, line 1: Also need the same treatment in `test_sigint_many_hits_cb`

tacaswell: Hit both now.

callumforrester:
@tacaswell is https://github.com/bluesky/bluesky/issues/1713 meant for finding a longer-term solution after merge?

callumforrester (approved):

coretl (approved):

mrakitin (approved):

Issue #1679, opened 2024-04-04
Dependencies: Super State Machine - not maintained


stan-dot:
Last update to the Super State Machine [repository](https://github.com/beregond/super_state_machine) took place 7 years ago. 

There already happened a [discussion in 2017](https://github.com/bluesky/bluesky/issues/723#issuecomment-320517555) to refactor it out of bluesky.

## Expected Behavior
We use dependencies with good maintenance status.

## Current Behavior
Bluesky imports a not-maintained module.

## Possible Solution
Refactor out the super state machine

tacaswell:
short-term we should vendor a copy of that library into bluesky so we can drop the dependency.

medium-to-long term we should either right our own state machine or find a replacement.  This is primarily used to keep track of the state of the RE so we need to be very careful about any changes to it.

stan-dot:
vendoring seems to match was has been already discussed before. this issue could focus on that.

for the longer view a quick search reveals libraries: 
- [fgmacedo/python-statemachine](https://github.com/fgmacedo/python-statemachine)  with 635 stars, 
- [pytransitions](https://github.com/pytransitions/transitions) with 5.3 k stars
- [p-org/P](https://github.com/p-org/P) state machine focused programming language - 2.9k stars
- [mriehl/fysom](https://github.com/mriehl/fysom) with 385 stars
- [statesman](https://github.com/opsani/statesman) with 50 stars.
that's the top of the google search so it's highly unlikely we could find something better. Of course it's at the core so care needs to be taken. Do you think this issue fits the hackathon theme well @tacaswell ? or too fragile of a system element? Maybe still some investigation would be harmless?

Just a quick check in other languages - same situation, finite state machines seem to be a niche use case.
- [looplab/fsm](https://github.com/looplab/fsm?tab=readme-ov-file) finite - state - machine in golang - so that we know it's not just Python
- [stonecypher/jssm](https://github.com/StoneCypher/jssm) 296 stars
- [xstate in the javascript ecosystem](https://github.com/statelyai/xstate) hasn't proven that successful

danielballan:
Agreed with the above. And I will add that our usage is pretty simple, so we might consider just using implementing something simple with dictionaries. Caproto has something like this and it works well.

clintonroy:
I've dabbled with automat, spun out of twisted https://github.com/glyph/automat

genematx:
[...]

PR #1708, opened 2024-04-16
Vendor a copy of Super State Machine

genematx:
Vendor a copy of Super State Machine into the Bluesky Repository

## Description
This PR adds a copy of Super State Machine [source code](https://github.com/beregond/super_state_machine), into the Bluesky repository and removes the external dependency. The automated `vendoring` tool was used for this purpose.

## Motivation and Context
The Super State Machine repo has been updated for more than 7 years, and we would like to use dependencies with good maintenance status. This PR address Issue #1679 .


src/bluesky/super_state_machine/machines.py
```diff
@@ -0,0 +1,275 @@
...
+    @property
+    def actual_state(self):
+        """Actual state as `None` or `enum` instance."""
+        attr = self._meta["state_attribute_name"]
+        return getattr(self, attr)
+
+    @property
+    def as_enum(self):
+        """Return actual state as enum."""
+        return self.actual_state
```
genematx, line 262: Moved these methods here from `utils.py` since property decorators outside of class definitions didn't pass linting checks. (This _might_ cause unnoticed name collisions, but within our restricted usage in Bluesky, it is extremely unlikely.)

tacaswell:
We should make the vendored version private, we definitly want to be able to switch this out with a different implementation without having to go through a deprecation cycle.

https://github.com/pypa/pip/tree/main/src/pip/_vendor is a good reference for how to do this and we probably should follow it here.

tacaswell:
It might be worth to make a LICENSE directory and copying the SSM license into a file called LICENSE_SUPER_STATE_MACHINE following something like https://github.com/matplotlib/matplotlib/tree/main/LICENSE

src/bluesky/super_state_machine/machines.py
```diff
@@ -0,0 +1,275 @@
...
+NotSet = object()
+
+
+class DefaultMeta:
+    """Default configuration values."""
+
+    states_enum_name = "States"
+
+
+class AttributeDict(dict):
```
stan-dot, line 17: is this legacy class?

src/bluesky/super_state_machine/machines.py
```diff
@@ -0,0 +1,275 @@
...
+        except TypeError:
+            proper = False
+
+        if not proper:
+            raise ValueError("Please provide enum instance to define available states.")
+
+    @classmethod
+    def _check_if_states_are_strings(cls):
+        """Check if all states are strings."""
+        for item in list(cls.context.states_enum):
```
stan-dot, line 87: unnecessary conversion to list (?)

src/bluesky/super_state_machine/machines.py
```diff
@@ -0,0 +1,275 @@
...
+        meta = getattr(cls.context.new_class, "Meta", DefaultMeta)
+        cls.context.get_config = partial(get_config, meta)
+
+    @classmethod
+    def _set_up_translator(cls):
+        translator = utils.EnumValueTranslator(cls.context["states_enum"])
+        cls.context.new_meta["translator"] = translator
+
+    @classmethod
+    def _calculate_state_name(cls):
```
stan-dot, line 238: refactor into oneliner?

src/bluesky/super_state_machine/utils.py
```diff
@@ -0,0 +1,132 @@
+"""Utilities for core."""
+
+from enum import Enum, unique
+from functools import wraps
+
+from .errors import TransitionError
+
+
+def is_(self, state):
```
stan-dot, line 9: add full function name

stan-dot (changes requested):
half done, stopped at utils

tacaswell:
I would also prefer that we sush the linters / type checkers rather than fix SSM.

genematx:
I agree with @tacaswell regarding not making changes to the vendored code in order to appease the linter. There were only a couple of bits related to the compatibility with Python 2 that I removed, but otherwise I tried to keep the original super-state-machine code intact; definitely, lots of opportunities for improvement though -- thank you @stan-dot for going through it and making notes.

Also, I might have misunderstood the meaning of `vendoring` in this context. I will make the version private and let you know once it's done.

src/bluesky/super_state_machine/machines.py
genematx: Obsolete: Vendored entire package using `vendoring`.

src/bluesky/super_state_machine/utils.py
genematx: Obsolete: Vendored entire package using `vendoring`.

src/bluesky/super_state_machine/machines.py
genematx: Obsolete: Vendored entire package using `vendoring`.

genematx:
Thank you for the suggestions regarding vendoring, @tacaswell! I've used the pip's `vendoring` tool, and the result looks more standardized and definitely easier to manage (the code is copied with absolutely no modifications). Does this look like something you had in mind?

stan-dot (approved):
for vendoring option all looks good

stan-dot:
Still the library did not have its own testing. Unit testing for the core of bluesky is ideally something exhaustive. 

I don't feel 100% convinced that vendoring is the optimal long term choice, `pip` does it for *very specific* reasons, and the vendored-in repos they reference change often upstream. That is not the case with `super_state_machine` repo. https://github.com/beregond/super_state_machine . It is a stale 'last update 7 years ago' package.

I am not aware if right now we are running into any limitations with this library. If are not and we can accept the lack of test or write our own that would be not in the library itself, this PR could be merged and the vendored in library code left alone. Should that not be the case, alternatives to 'non-modify-vendoring' in the order of least to  most work are:
- refactor the existing vendored code and add tests
- choose some library 
- writing a state machine from scratch ourselves

genematx:
Thank you, @stan-dot, for your comments. Yes, I agree that our case for vendoring differs from that of pip's, and they actually the use of their `vendoring` tool is discouraged. As far as I understand, an please correct me if I'm missing something, the main reasons for us to vendor `super-state-machine` is (1) to reduce the risk of the package being removed from pypi -- not sure if this happens very often if at all, and (2) major breaking changes being suddenly introduced.

I didn't know what to do with the tests. I can certainly just keep their existing tests, but this approach seems to be rather formal -- it would be better to have some sort of integration tests that specifically address the functionality of SSM used in bluesky (and I think we already have a couple). One reason in favour of keeping the existing tests that I can see is checking that SSM works with newer versions of Python. Happy to add them if this makes sense.

I personally would be keen to explore options 2, and especially 3, but this definitely would fall out of scope for this sprint/hackathon.

stan-dot:
@genematx option 2 (a rewrite) is only 3k lines, so a refactored and tested MVP could be theoretically done with 2 weeks of engineering time

danielballan:
My two cents: Vendor for _right_ now, but prioritize a rewrite in the medium term. I agree @stan-dot that this should not be a heavy lift.

stan-dot (approved):

tacaswell:
Can we squash this to one commit?

genematx:
Sure, one commit is easy. Should I just merge it?

PR #1709, opened 2024-04-16
Added class and decorator to warn users if plan is not iterated over

### LINKED ISSUE #1632: Add warning when plan not iterated
## Expected Behavior

At the moment, it is very easy to forget to `yield from` in the middle of the plan, and there don't appear to be any warnings that appear when a plan stub is created but not iterated. To contrast, `asyncio` will warn on garbage collect of any coroutine that is not awaited. It would be nice to do the same.

## Current Behavior

```python
from bluesky import RunEngine

RE = RunEngine()

def my_plan():
    yield from bps.null()
    bps.sleep(10)

RE(my_plan())
```

Nothing is printed, vs the asyncio:
```python
import asyncio

async def my_func():
    asyncio.sleep(1)

asyncio.run(my_func())
```
which prints
```
RuntimeWarning: coroutine 'sleep' was never awaited
  asyncio.sleep(1)
```

## Possible Solution

We could decorate the function with something that tracks iteration and raises on `__del__`:

```python
# utils.py
import traceback

class EnsureIterated:
    def __init__(self, f, *args, **kwargs):
        self._iter = f(*args, **kwargs)
        self._stack = traceback.format_stack()
        self._stack += [f"RuntimeWarning: plan `{f.__name__}` was never iterated"]

    def __iter__(self):
        self._stack = None
        yield from self._iter

    def __del__(self):
        if self._stack:
            print(*self._stack)

def ensure_plan_iterated(f):
    # TODO: wrap with functools.wrap and sort out type hints
    def wrapper(*args, **kwargs):
        return EnsureIterated(f, *args, **kwargs)
    
    return wrapper
    

# plan_stubs.py
from bluesky import Msg

@ensure_plan_iterated
def sleep(time):
    return (yield Msg('sleep', None, time))


# my_plan.py
from bluesky import RunEngine

RE = RunEngine()

def my_plan():
    yield from sleep(0.1)
    sleep(10)

RE(my_plan())
```
will output
```
  File "/usr/lib/python3.10/threading.py", line 973, in _bootstrap
    self._bootstrap_inner()
   File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
   File "/usr/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
   File "/usr/lib/python3.10/asyncio/base_events.py", line 600, in run_forever
    self._run_once()
   File "/usr/lib/python3.10/asyncio/base_events.py", line 1896, in _run_once
    handle._run()
   File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
   File "/dls/science/users/tmc43/repos/bluesky/bluesky/run_engine.py", line 1532, in _run
    msg = self._plan_stack[-1].send(resp)
   File "/repos/bluesky/my_plan.py", line 41, in my_plan
    sleep(10)
   File "/repos/bluesky/my_plan.py", line 21, in wrapper
    return EnsureIterated(f, *args, **kwargs)
   File "/repos/bluesky/my_plan.py", line 7, in __init__
    self._stack = traceback.format_stack()
 RuntimeWarning: plan `sleep` was never iterated
```
This is a bit wordy, but demonstrates the principle...

@callumforrester @rosesyrett we talked about this issue a few days ago, maybe something like this would help?

vshekar:
Adds a decorator to be used on plans that warns users if the `yield from` statement is not used

## Motivation and Context
Fixes #1632 

## How Has This Been Tested?
Tested locally using the example described in the issue


vshekar:
Can also add tests to this PR (or a separate one) and apply the decorator to built in plans if that is the intention

tacaswell:
Might as well do it all in one go unless you think there is enough work to be done in the decoration that you want to split it up.

danielballan:
We can imagine use cases for this decorator beyond just ensuring iteration. For example, providing a nice repr. This and other possibilities (which I'm not necessarily still in favor of...) [were discussed back in 2019](https://github.com/bluesky/bluesky/issues/1239). To pave the way for broader uses cases for this decorator, let's give it a more generic name.

Per discussion, there was rough consensus to rename the class to `Plan` and decorator to `@plan`. @danielballan, @callumforrester, @coretl, @abbiemery in favor; @tacaswell willing to go with it. :-D

prjemian:
For example, the bluesky user wants to learn what plans are available (such as [`apstools.utils.listplans()`](https://bcda-aps.github.io/apstools/latest/examples/ho_list_control_objects.html#listplans()-function)).  This decorator could be used to make that discovery process much easier and [better informed](https://github.com/BCDA-APS/apstools/blob/83206b9fab067f9a80bd4f9fa732b9c0c51d12fa/apstools/utils/list_plans.py#L25-L29).  The current [assumption ](https://github.com/BCDA-APS/apstools/blob/83206b9fab067f9a80bd4f9fa732b9c0c51d12fa/apstools/utils/list_plans.py#L68)is far too general:

```py
        if inspect.isgeneratorfunction(obj):  # TODO: bluesky.isplan(obj)
```

vshekar:
Sorry I wasn't able to attend the discussion. But I changed the class to Plan and decorator to @plan. If someone wants to take a look at it

prjemian:
`bluesky/utils/__init__.py` is approaching 2,000 lines with dozens of classes and functions at the module level.  Does anyone else believe it is time to refactor this file into its component modules?  Good fodder for the Hackathon.

src/bluesky/tests/test_utils.py
```diff
@@ -499,3 +501,33 @@ def kickoff(self, *args, **kwargs):
...
+def non_iterating_plan():
+    yield from sample_plan()
+    sample_plan()
+
+
+@pytest.mark.parametrize("gen_func, iterated", [
+    (non_iterating_plan, False),
+    (iterating_plan, True)
+])
+def test_warning_behavior(gen_func, iterated):
```
coretl, line 524: Please can I see an example of the warning that is created? A slightly more detailed regex passed to `match` would be ideal...

vshekar: Sure, here's an example warning thats created, happy to add your suggestion
```
/Users/vshekar/Code/bluesky/src/bluesky/utils/__init__.py:1892: RuntimeWarning:   File "/Users/vshekar/Code/misc/run-engine-tests/bluesky_tests.py", line 20, in my_plan
    sleep(10)
RuntimeWarning: plan `sleep` was never iterated, did you mean to use `yield from`?
  warnings.warn(warning_message, RuntimeWarning, stacklevel=1)
  ```

Comment (coretl) [reply src/bluesky/tests/test_utils.py:524]: Can we put one more newline in please?
```
/Users/vshekar/Code/bluesky/src/bluesky/utils/__init__.py:1892: RuntimeWarning:
File "/Users/vshekar/Code/misc/run-engine-tests/bluesky_tests.py", line 20, in my_plan
    sleep(10)
RuntimeWarning: plan `sleep` was never iterated, did you mean to use `yield from`?
  warnings.warn(warning_message, RuntimeWarning, stacklevel=1)
```

Comment (vshekar) [reply src/bluesky/tests/test_utils.py:524]: Added in https://github.com/bluesky/bluesky/pull/1709/commits/fa2131335c8cff73907d70fddc0df556b4c93eba

File: src/bluesky/utils/__init__.py [hunk 2]
```diff
@@ -1868,3 +1869,43 @@ async def maybe_await(ret: SyncOrAsync[T]) -> T:
...
+
+
+class Plan:
+    def __init__(self, f, *args, **kwargs) -> None:
+        self._iter = f(*args, **kwargs)
+        self._stack = traceback.format_stack()
+        # Find the index in the stack trace where the function `f` is called
+        func_index = next((i for i, s in enumerate(self._stack) if f.__name__ in s), None)
+        if func_index is not None:
+            self._stack = self._stack[func_index:func_index+1]
```
tacaswell, line 1881:
```suggestion
        self._stack = self._stack[:-2]
```

Talking about this in the hackathon room we came to consusus that accepting very long tracebacks to start to err on the side of giving context of "why was I creating this un-iterated plan" not just the exact code that created it.

We have a very old issue about generally filtering tracebacks for RE exceptions (no one wants to see the N levels of locally defined closure plans made by plan_mutator!) and filtering these tracebacks and those tracebacks should probably follow the same rules.  Thus, we should put off the filtering to do those togethor (or until we get real world feedback that verbose warnings are completely untenable.

The `-2` is to chop off the wrapper and `Plan.__init__` layers.

vshekar: Seems like I have a knack for leaving the zoom calls just before this PR gets discussed 😆 
Removed filtering in https://github.com/bluesky/bluesky/pull/1709/commits/88ccaf84a296dcf7f958f041728bcba7ae4d39cb

tacaswell: You can use this power to force discussion on your PRs 🙃

coretl (approved):

src/bluesky/utils/__init__.py
```diff
@@ -1868,3 +1869,40 @@ async def maybe_await(ret: SyncOrAsync[T]) -> T:
...
+    def __init__(self, f, *args, **kwargs) -> None:
+        self._iter = f(*args, **kwargs)
+        self._stack = traceback.format_stack()
+        self._stack = self._stack[:-2]
+        self._stack += [f"RuntimeWarning: plan `{f.__name__}` was never iterated"
+                        ", did you mean to use `yield from`?"]
+
+    def __iter__(self):
+        self._stack = None
+        yield from self._iter
```
tacaswell, line 1884:
```suggestion
        return (yield from self._iter)
```

src/bluesky/callbacks/olog.py
```diff
@@ -173,5 +173,5 @@ def start(self, doc):
             pformat(doc),
         )
         olog_status = self.client.log(document_content, logbooks=self.logbook)
-        logger.debug("client.log returned %s" % olog_status)
+        logger.debug(f"client.log returned {olog_status}")
```
tacaswell, line 176:
```suggestion
        logger.debug(f"client.log returned %s", olog_status)
```

It is best-practice to defer stringy-ifying things passed to loggers to avoid paying the cost if there is nothing consuming the log.

prjemian: Isn't `logger` the exception to use of f strings? The point of *no f strings in logger methods* is that the method chooses whether to evaluate the arguments based on the log level. If an f string is used, the arguments are always evaluated, which defeats one of the features of the `logger`.

prjemian:
```suggestion
        logger.debug("client.log returned %s", olog_status)
```

prjemian: Deleted my suggestion.  When first observed this, my phone showed it differently.  My comments and suggestion restated @tacaswell 's point.

vshekar: Ah then, we may have to change the ruff linter config maybe? The current version of ruff `0.4.2` does not like percent strings and fails, `0.4.1` passes. Any suggestion @tacaswell?

vshekar: Or I guess add `# noqa` to these lines

vshekar: One more question, should I leave other f string conversions as is? i.e. print statements or stick a noqa there as well?

vshekar: Great then I think we are all set. Checks and tests passed!

src/bluesky/utils/__init__.py
```diff
@@ -1914,3 +1915,49 @@ async def maybe_await(ret: SyncOrAsync[T]) -> T:
...
+        self._stack = self._stack[:-2]
+        self._stack += [
+            f"RuntimeWarning: plan `{f.__name__}` was never iterated" ", did you mean to use `yield from`?"
+        ]
+
+    def __iter__(self):
+        self._stack = None
+        return (yield from self._iter)
+
+    def __next__(self):
```
tacaswell, line 1933: Why do we need `__next__`?  Do we also need to forward `.send` and `.throw`?

vshekar: The test_repeater calls next() on a plan: https://github.com/bluesky/bluesky/blob/add624e1b7f03c9b4b981e78dfc2333ed01e1f0c/src/bluesky/tests/test_new_examples.py#L641

`.send` and `.throw` would get forwarded here:
https://github.com/bluesky/bluesky/blob/3173c97e8017892a9f0a1f5c57a0661ed92a4b7b/src/bluesky/utils/__init__.py#L1941-L1942
Which would behave transparently as if it were a real plan, unless you want me to explicitly forward them

danielballan: I think this is conflating an Iterator and an Iterable. A plan should not implement `__next__`.

```py
In [1]: def g():
   ...:     yield 1
   ...:     yield 2
   ...: 

In [2]: next(g)
TypeError                                 Traceback (most recent call last)
Cell In[2], line 1
----> 1 next(g)

TypeError: 'function' object is not an iterator

In [3]: gi = g()

In [4]: next(gi)
Out[4]: 1
```

danielballan: ^ That comment might be overconfident. I should take more time to dig into this if need be, but something smells suspicious about implementing both `__iter__` and `__next__` on this object.

tacaswell: Ah, I somehow missed `__getattr__` being added, it is probably better if we explicitly forward the ones we need.

vshekar: Explicit forwarding of `send()` and `throw()` added in https://github.com/bluesky/bluesky/pull/1709/commits/1b981b7e34958cb3285fb709b14c8b792d1cc5ea

danielballan: I now understand what I was confused about.

danielballan (approved):

Issue #1697, opened 2024-04-15
Fix `fuzz.py`


DominicOram:
`fuzz.py` is currently un-importable as it relies on `setup_test_run_engine`, which no longer exists.

We should either fix it or remove it if it's no longer used. I don't have enough background to decide which one but happy to do the work.

CoePaul:
Target destroyed

PR #1710, opened 2024-04-16
(#1697) removes fuzz in toto

CoePaul:
* Expunges obsolete fuzz - formerly used in some old tests

## Description
Removed obsolete fuzz directory, old python script and associated README

## Motivation and Context
It was no longer in use - so must be removed

## How Has This Been Tested?
Hard to test an absence other than letting the CI tests do their thing


callumforrester (approved):

DominicOram (approved):
Thank you!

Issue #1613, opened 2023-09-13
Allow plans to accept lazy iterables of detectors


callumforrester:
Most plans in `plans.py` currently accept lists or tuples.

## Expected Behaviour
For example, you should be able to pass

```python
detectors = [det1, det2]
generator = (detector for detector in detectors)
RE(count(generator))
```

## Current Behavior
This will run but all detectors will be absent from the event documents. You have to pass a concrete list or tuple such as
```python
RE(count(detectors))
```

## Possible Solution
Some of the plans consume the generator more than once, which is why they behave oddly. One solution is to make sure all plans call

```python
detectors = list(detectors)
```
Or similar at the start.

tacaswell:
Seems reasonable.

callumforrester:
Reviewers: @tacaswell @callumforrester

tacaswell:
Changing my mind (as discussed on the hackathon call) this opens up something dangerous

```
def my_plan():
   generator = (detector for detector in detectors)
   # would do the right thing
   yield from count(generator)
   # would do a very wrong thing!
   yield from count(generator)
```

Instead of making sure the plans refuse to take generators / only take `Sequence` types.

PR #1712, opened 2024-04-16
Only allow plans to accept Sequences of detectors 

noemifrisina:
Allow the plans to only accept lists or tuples of hardware devices eg. detector to avoid unexpected behaviour.

## Description
- Adds a check on the type of the `detector` argument passed to the plan and raises an error if not a Sequence.
- Adds type checking to plans and plan stubs.

Pulled in [PR#1610](https://github.com/bluesky/bluesky/pull/1610) which had started the work on type checking.

## Motivation and Contex
Currently plans will still run to completion if passed an iterable different from a generator as input, but will do the wrong thing in the background.
See #1613 .

## How Has This Been Tested?
- Type annotations should not change the functionality of the code.
- Small additions to unit tests to verify that plans raise a TypeError if a generator is passed instead of a list.

### DIFF
--- a/src/bluesky/plan_stubs.py
+++ b/src/bluesky/plan_stubs.py
@@ -5,9 +5,21 @@ import uuid
 import warnings
 from collections.abc import Iterable
 from functools import reduce
-from typing import List
+from typing import (
+    Any,
+    Awaitable,
+    Callable,
+    Hashable,
+    List,
+    Mapping,
+    Optional,
+    Tuple,
+    Union,
+)
 
 from cycler import cycler
+
+from bluesky.suspenders import SuspenderBase
 
 try:
     # cytools is a drop-in replacement for toolz, implemented in Cython
@@ -15,9 +27,28 @@ try:
 except ImportError:
     from toolz import partition
 
-from .protocols import Flyable, Locatable, Status, Triggerable, check_supports
+from event_model import ComposeEvent
+from event_model.documents import EventDescriptor
+
+from .protocols import (
+    Configurable,
+    Flyable,
+    Locatable,
+    Movable,
+    PartialEvent,
+    Readable,
+    Reading,
+    Stageable,
+    Status,
+    Stoppable,
+    Triggerable,
+    check_supports,
+)
 from .utils import (
+    CustomPlanMetadata,
     Msg,
+    MsgGenerator,
+    ScalarOrIterableFloat,
     all_safe_rewind,
     ensure_generator,
     get_hinted_fields,
@@ -28,15 +59,23 @@ from .utils import (
     short_uid as _short_uid,
 )
 
+#: Any plan function that takes a reading given a list of Readables
+TakeReading = Callable[[List[Readable]], MsgGenerator[Mapping[str, Reading]]]
 
-def declare_stream(*objs, name: str, collect=False):
+
+def declare_stream(
+    *objs: Readable, name: str, collect=False
+) -> MsgGenerator[Tuple[EventDescriptor, ComposeEvent]]:
     """
     Bundle future readings into a new Event document.
 
     Parameters
     ----------
-    name : string
-        name given to event stream, used to convenient identification
+    args :
+        objects whose readings will be present in the stream
+    name : string, optional
+        name given to event stream, used for convenient identification
+        default is 'primary'
     collect : bool, optional
         collect as well as describe when declaring the stream
         default is `False`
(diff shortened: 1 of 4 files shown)


src/bluesky/plan_stubs.py
```diff
@@ -28,15 +59,23 @@
     short_uid as _short_uid,
 )
 
+#: Any plan function that takes a reading given a list of Readables
+TakeReading = Callable[[List[Readable]], MsgGenerator[Mapping[str, Reading]]]
```
callumforrester, line 63: For consistency

```suggestion
TakeReading = Callable[[Sequence[Readable]], MsgGenerator[Mapping[str, Reading]]]
```

src/bluesky/plan_stubs.py
```diff
@@ -28,15 +59,23 @@
...
 
-def declare_stream(*objs, name: str, collect=False):
+
+def declare_stream(
+    *objs: Readable, name: str, collect=False
```
callumforrester, line 67:
```suggestion
    *objs: Readable, name: str, collect: bool =False
```

src/bluesky/tests/test_plans.py
callumforrester, line 1: May already be covered, but if not could do with a test to make sure we can pass a list and tuple. Also a copy/paste job of the count test for the other plans would be ideal, but can be turned into a separate issue if you want to move on

callumforrester (changes requested):

src/bluesky/utils/__init__.py
```diff
@@ -87,6 +91,29 @@ def __repr__(self):
...
+
+#: Object usually returned from plan functions that is fed to the RunEngine
+MsgGenerator = Generator[Msg, Any, Optional[P]]
+
+#: Metadata passed from a plan to the RunEngine for embedding in a start document
+CustomPlanMetadata = Dict[str, Any]
+
+
+#: Return type of a plan, usually None. Always optional for dry-runs.
+P = TypeVar("P")
```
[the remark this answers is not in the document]
noemifrisina: Sorry, leftover from merge conflict. Removed now.

src/bluesky/tests/test_plans.py
noemifrisina: Not completely covered I'd say, all the tests pass lists to the plans but no tuples. Is it worth adding a small test passing both or would changing some of the existing ones to pass tuples be enough?

noemifrisina:
Docs build is failing because `numpydocs` doesn't accept having both `Yields` and `Returns` sections in the docstring. It seems it never has even in older versions and adding type hints may have triggered this check

danielballan:
Numpydoc PR to address this: https://github.com/numpy/numpydoc/pull/542

src/bluesky/plans.py
```diff
@@ -18,16 +20,54 @@
...
+PerStep = Union[PerStep1D, PerStepND]
+
+
+def _check_detectors_type_input(detectors):
+    if not isinstance(detectors, Sequence):
+        raise TypeError("The input argument must be either as a list or a tuple of Readable objects.")
+
+
+def count(
+    detectors: Sequence[Readable],
```
callumforrester, line 58: Slowly resurrecting this change, @tacaswell and @dmgav, how would you feel about making these `set[Readable]` instead of `Sequence[Readable]`? We have use cases that would be easier if it was made clear that these were unique-valued sequences. The counter argument, I suppose, is if they need to be ordered.

danielballan: They do need to be ordered. But maybe we can find another way. Would you describe the use case?

callumforrester: We generate json schemas from plans args using pydantic. Using sets leads to `"uniqueitems": true` in the schema, whereas sequences do not. We're investigating feeding the schemas into [json forms](https://jsonforms.io/), which presents very different interfaces for collections with and without `uniqueitems`.

The alternative, I guess, is to have some custom annotation somewhere and post-processing of the JSON schema, which is definitely an option! Or just wrap the plans.

But yes, I can see they need to be ordered, annoying python has no ordered set type...

tacaswell: There is internal logic to de-duplicate devices so it is OK if they are not unique.

The thing that is a problem is if there are over-lapping keys in the read between multiple devices and I do not think there is a way to express that in the typing.

tacaswell:
There is a numpydoc release now (which is why I tried to rebase #1610 because I thought that one was this one).

danielballan:
I see that @noemifrisina has CI passing. Were your concerns above adequately addressed, @callumforrester?

danielballan (approved):

Issue #1638, opened 2023-12-11
Create a simulated run engine for unit testing


DominicOram:
For unit testing plans it would be useful to have a simulated RunEngine.

## Expected Behavior
The simulator should be able to:
* Run bluesky plans
* Provide data to the plan (as fed in via a unit test)
* Provide the ability to assert which messages have been produced by the plan

## Possible Solution
DLS MX have produced something like this already (see [here](https://github.com/DiamondLightSource/hyperion/blob/5b901baca29a9d19f0626df10ec318f390941a67/tests/unit_tests/experiment_plans/conftest.py#L103)) but it would be good to feed it back into the community.

tacaswell:
We have https://github.com/bluesky/bluesky/blob/9b0d72953cae867d8312cde3da2ed9cc4caf0788/bluesky/tests/utils.py#L20-L50 that will capture the messages that go in and the documents that come out for testing.

However, I'm not sure that it will be worth writing a full simulator of the RE as it will end up looking more-or-less like the RE.  It seems easier to me to put the smart simulation in (mock) ophyd objects or to replace the async coroutines for a given message on the "real" RE. 

With a mock RE you are as much testing your test harness as the plans you want to test.  Early on we had a clone of ophyd's API in bluesky that we used for testing, but after the Nth time where we discovered things were actually broken because our ophyd clone was not exactly ophyd we gave up and added ophyd as a test dependency.  I have a concern that this will go down a similar road.

tacaswell:
Following up on this, I am still not sure that there is an advantage of a mock RE over the real RE + mock devices.

prjemian:
> * Provide data to the plan (as fed in via a unit test)

Why is this the responsibility of the `RE`?

danielballan:
Yes, using real RE with mock devices to provide data is the route we have taken up to this point. It would be interesting to learn more about the DLS MX use case.

rtuck99:
Many things can be tested using mock devices in dodal but there are some things where it's easier to test at the message API level. Particularly where it concerns order of submitted operations and things like correct command grouping, waiting for completion etc. For instance, we want to ensure that the detector move and shutter open happens whilst other things are being setup as it is slow, but we only need to validate that it is complete just before we start collecting data.The sequential nature of the message protocol lends itself well to this kind of introspection. 

[...]

PR #1714, opened 2024-04-16
1638 create a simulated run engine for unit testing

rtuck99:
## Description
Provides a new class called `RunEngineSimulator` added to the `bluesky.simulators` module that assists with testing bluesky experiment plans by invoking the plan generator function and generating the sequence of 'Msg' instances that can then be `assert`ed.

## Motivation and Context
bluesky experiment plans can be quite complex and interact with significant amounts of hardware which must normally all be mocked. opyhd allows for simulated devices but often these may have significant internal complexity that is less easily simulated. In addition there may be callbacks that may need to be invoked and it may be desirable to be able to control the content of the callback documents in testing situations.
Being able to test at the message protocol level has the many advantages including:
* Many interactions with hardware that are not of interest do not have to be mocked since most messages do not yield a return value.
* The sequence of operations is easy to verify since the generated message stream is sequential and can be asserted
* Only those callbacks which are of interest need to be triggered as it's possible to fire them selectively at the individual event level
* Operations such as `wait` and `sleep` that don't directly interact with hardware but are necessary for correct functioning can be tested.
* Tests at the message protocol level may be more robust than tests at the unit level which might instead directly mock and verify invocations of `bluesky.plan_stubs` methods, especially since the order of messages is insensitive to refactoring whereas using `patch` to intercept library calls often breaks when code is refactored

See also discussion in #1638

## How Has This Been Tested?
See additional tests in `test_simulators`. This code has largely already been in use in`hyperion` within our unit testing framework, and this change effectively moves this class into `bluesky`.
The changes made between hyperion and bluesky consist of additional documentation of the methods and also some minor changes to method signatures in order to make them easier to use. In addition, as in hyperion this code was effectively part of the unit test fixtures, it was not itself tested directly and now it has some tests.

It would be good to have some additional feedback, on whether the documentation needs further improvement, or additional tests. I am conscious that as this is now in bluesky it will slightly increase the effort to make changes if in hyperion we want further changes or fixes to this infrastructure. Also in hyperion the callback facility isn't currently in use, as the tests which used it were subsequently refactored out.
Additionally it would be good to know if there are additional features that it needs to be useful to others as in hyperion I think we are only using a subset of all the RunEngine features. For example I don't think we make any use yet of `input`, or the checkpoints, pausing or rewinding. In general though I think it should be possible to add handlers for arbitrary message commands.

### DIFF
--- a/src/bluesky/simulators.py
+++ b/src/bluesky/simulators.py
@@ -1,3 +1,4 @@
+from itertools import dropwhile
 from time import time
 from typing import (
     Any,
@@ -17,7 +18,7 @@ from bluesky.preprocessors import print_summary_wrapper
 from bluesky.run_engine import call_in_bluesky_event_loop, in_bluesky_event_loop
 from bluesky.utils import Msg, maybe_await
 
-from .protocols import Checkable, Reading
+from .protocols import Checkable, Readable, Reading
 
 END = "end"
 
@@ -214,20 +215,12 @@ class RunEngineSimulator:
             ),
         )
 
-    def add_read_handler_for(self, obj_name: str, value: Any, dict_key: str = "values"):
+    def add_read_handler_for(self, obj: Readable, value: Optional[Any]):
         """
-        Convenience method to register a handler to return a result from a 'read' command.
-
-        Parameters
-        ----------
-        obj_name
-            Name of the object to intercept reads for
-        value
-            The value to return. If value is a dict then dict_key is ignored and the result of the read
-            command is the supplied value. Otherwise, the result of the read command is a dict of the form
-            { dict_key: { "value": value } }
-        dict_key
-            The key in the result dictionary to populate with the value, by default this is 'values'
+        Convenience method to register a handler to return a result from a
+        single-valued 'read' command.
+        The behaviour is equivalent to
+        add_read_handler_for_multiple(obj, **{obj.name: value})
 
         Examples
         --------
@@ -243,11 +236,43 @@ class RunEngineSimulator:
         >>> msgs = sim.simulate_plan(trigger_and_return_pin_tip())
         >>> assert sim.return_value == (8, 5)
         """
+        self.add_read_handler_for_multiple(obj, **{obj.name: value})
+
+    def add_read_handler_for_multiple(self, obj: Readable, **kwargs):
+        """
+        Convenience method to register a handler to return a result from a multi-valued
+        'read' command.
+
+        Parameters
+        ----------
+        obj
+           The object to intercept reads for
+        kwargs
+            The handler will return a dict with the corresponding key-value pairs.
+            If the value is a dict the result of the read command for the key is the supplied dict.
+            Otherwise, the value of the read command for the key is
+            a value of Reading(value=value, timestamp=time())
+
+        Examples
+        --------
+        In order to exercise a plan containing the following:
+
+        >>> def read_all_values():
+        >>>     data = yield from bps.read(hw.ab_det)
+        >>>     return data
+        >>> sim = RunEngineSimulator()
+        >>> sim.add_read_handler_for_multiple(hw.ab_det, a=11, b=12)
+        >>> sim.simulate_plan(read_all_values())
+        >>> assert sim.return_value["a"]["value"] == 11
+        >>> assert sim.return_value["b"]["value"] == 12
+        >>> assert isclose(sim.return_value["a"]["timestamp"], time(), abs_tol=1)
+        >>> assert sim.return_value["b"]["timestamp"] == 1717071719
+        """
 
         def handle_read(_):
-            return value if isinstance(value, dict) else {dict_key: Reading(value=value, timestamp=time())}
+            return {k: v if isinstance(v, dict) else Reading(value=v, timestamp=time()) for k, v in kwargs.items()}
 
-        self.add_handler("read", handle_read, obj_name)
+        self.add_handler("read", handle_read, obj.name)
 
     def add_wait_handler(self, handler: Callable[[Msg], None], group: str = GROUP_ANY) -> None:
         """Add a wait handler for a particular message.
@@ -285,17 +310,12 @@ class RunEngineSimulator:
     def add_callback_handler_for(
         self,
         command: str,
-        *,
-        document_name: Optional[str] = None,
-        document: Optional[dict] = None,
-        docs: Optional[Sequence[Sequence[Tuple[str, dict]]]] = None,
+        document_name: str,
+        document: dict,
         msg_filter: Optional[Callable[[Msg], bool]] = None,
     ):
-        """Add a handler to fire one or more callbacks in sequence when a matching command is encountered.
-
-        To fire a single callback just once, specify `document_name` and `document` parameters.
-        To fire a sequence of callbacks, across one or more message invocations, use the `docs` parameter.
-
+        """Add a handler to fire a callback when a matching command is encountered.
+        Equivalent to add_callback_for_multiple(command, [[(document_name, document)]], msg_filter)
         Parameters
         ----------
         command
@@ -304,15 +324,31 @@ class RunEngineSimulator:
             Name of the document to match when used in single shot mode.
         document
             The document to fire in single shot mode.
-        docs: Sequence of Sequence of document_name, document tuples
+        msg_filter
+            Optional predicate on the message which triggers the callback
+        """
+        self.add_callback_handler_for_multiple(command, [[(document_name, document)]], msg_filter)
+
+    def add_callback_handler_for_multiple(
+        self,
+        command: str,
+        docs: Sequence[Sequence[Tuple[str, dict]]],
+        msg_filter: Optional[Callable[[Msg], bool]] = None,
+    ):
+        """Add a handler to fire callbacks in sequence when a matching command is encountered.
+
+        Parameters
+        ----------
+        command
+            Name of the command to match.
+        docs
+            Sequence of Sequence of document_name, document tuples
             On each receipt of a matching message, the sequence contained in the next element of `docs`
             will be iterated over and each (document_name, document) pair will be fired to subscribed callbacks
             in order.
+        msg_filter
+            Optional predicate on the message which triggers the callback
         """
-        if not docs:
-            if not document_name or not document:
-                raise ValueError("No document sequence specified and document or document_name not specified")
-            docs = [[(document_name, document)]]
         it = iter(docs)
 
         def handle_command(_):
@@ -357,7 +393,7 @@ class RunEngineSimulator:
             while msg := gen.send(send_value):
                 send_value = None
                 messages.append(msg)
-                LOGGER.debug(f"<{msg}")
+                LOGGER.debug("<%s", msg)
                 if handler := next((h for h in self.message_handlers if h.predicate(msg)), None):
                     send_value = handler.runnable(msg)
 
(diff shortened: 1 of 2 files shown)


src/bluesky/simulators.py
```diff
@@ -104,3 +106,195 @@ async def check_limits_async(plan):
...
+        self,
+        handler: Callable[[Msg], object],
+        commands: Union[str, Sequence[str]],
+        obj_name: Optional[str] = None,
+        predicate: Optional[Callable[[Msg], bool]] = None,
+    ):
+        """Add the specified handler for a particular message. The handler is prepended so that
+        newer handlers override older ones.
+        Args:
+            handler: a lambda that accepts a Msg and returns an object; the object is sent to the current yield
```
tacaswell, line 161: Can we please use numpydoc format?

src/bluesky/simulators.py
```diff
@@ -104,3 +106,195 @@ async def check_limits_async(plan):
...
+                lambda msg: msg.command == "subscribe",
+                lambda msg: self._add_callback(msg.args),
+            )
+        )
+
+    def add_handler(
+        self,
+        handler: Callable[[Msg], object],
+        commands: Union[str, Sequence[str]],
+        obj_name: Optional[str] = None,
```
tacaswell, line 155: maybe make `obj_name` and `predicate` key-word only?  Would it make sense to merge them into one kwarg called "filter" which which is either a string or a `Callable[[Msg], bool]`.  It is not obvious to me that `or` is unambigiously the correct way to combine the two filters so my merging into one input we avoid having to make that choice.

src/bluesky/simulators.py
```diff
@@ -104,3 +106,195 @@ async def check_limits_async(plan):
...
+            ...     assert done == 1
+
+            >>> sim = RunEngineSimulator()
+            ... def simulate_detector_motion():
+            ...     sim.add_read_handler("detector_motion_shutter", 1)
+            ...
+            ... sim.add_wait_handler(simulate_detector_motion, "my_group")
+            ... sim.simulate_plan(close_shutter())
+        """
+        self.message_handlers.append(
```
tacaswell, line 230: I was thrown that this does not use `add_handler` which I see is because it does something that you can not do with the public API which makes me think that should be part of the public API.

src/bluesky/simulators.py
```diff
@@ -104,3 +106,195 @@ async def check_limits_async(plan):
...
+            _MessageHandler(
+                lambda msg: msg.command == "wait"
+                and (group == RunEngineSimulator.GROUP_ANY or msg.kwargs["group"] == group),
+                handler,
+            )
+        )
+
+    def fire_callback(self, document_name, document) -> None:
```
tacaswell, line 238: Does this get called as part of the simulation or would the caller have to call this manually?  I would have expected this to be called from inside of `simulate_plan`.

src/bluesky/simulators.py
tacaswell: oh, reading the tests I understand now.  It seems very fiddily that you have to pass a function closing over a method on sim into sim.  Is there a way to have the handlers for document generating messages return the documents that should be emited?

src/bluesky/simulators.py
tacaswell: oh, reading the tests I understand now. It seems very fiddily that you have to pass a function closing over a method on sim into sim. Is there a way to have the handlers for document generating messages return the documents that should be emited?

src/bluesky/simulators.py
```diff
@@ -104,3 +106,195 @@ async def check_limits_async(plan):
...
+            _MessageHandler(
+                lambda msg: msg.command in commands
+                and (
+                    (predicate and predicate(msg)) or (obj_name is None or (msg.obj and msg.obj.name == obj_name))
+                ),
+                handler,
+            ),
+        )
+
+    def add_read_handler(self, obj_name, value, dict_key="values"):
```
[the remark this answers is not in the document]
rtuck99: I've tried to make the signature of this method such that it results in the least amount of typing in most use cases. One of the things that is kind of annoying at the moment is that the value required to specify for `dict_key` whether it's "values" or the object name is not obvious (to me at any rate) and usually determined by trial and error looking at the message stream. I would much prefer it if the caller never has to specify it and only has to specify the value, which is usually (but not always) just a scalar.

So I would be against making the basic use case require more arguments, and would prefer if there was even some way it could be automatically determined (or can we even safely ignore it and check both?). But we could make it an option to return the entire dictionary if needed.

As far as including the timestamp, I think it's a bad idea to include in unit tests as in most cases the timestamp is not sensitive, and in those cases we do care, in unit tests you almost always want to be able to specify the exact value of the timestamp.

src/bluesky/simulators.py
rtuck99: I think it's more likely that it was originally written before the other methods had reached their current form. I have made it delegate to `add_handler` but changed so that it has the same prepend behaviour to make it more consistent.

src/bluesky/simulators.py
rtuck99: I've made some changes and have added a new method which should make it simpler to fire callbacks.

I realised also that the test was somewhat unrealistic as in reality the descriptor and the event are fired off the same save message. 

Also in iterative scenarios you may want to generate several save messages expecting different documents at each.

So the new method caters for this additional complexity but also supports simpler one-shot invocation.

src/bluesky/simulators.py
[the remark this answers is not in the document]
tacaswell: I expect that this code will be read more often than it will be written (in the context of "this test is now failing? why is it failing? what is it even doing?!) so I think we should prioritize clarity on reading someone else's tests over brevity of input.

I think returning data structures with different sets of keys is a pretty big divergence from the actual RE.  Some strawman code that would pass without `'timestamps'` but fail in actual usage is

```python
def sneak_plan(obj):
    ret = yield from bps.read(obj)
    for k, v in ret.items:
        # yes, this is a bad idea
        reading, = v.values()
```

Going back to a concern I raised earlier, in the early days we had a shadow-ophyd in bluesky that we used for testing.  The 4th or 5th time we had all our tests passing and then discovered that there was some subtle implementation difference between actual ophyd and our test-ophyd when running on the floor we gave up and just accepted a test-dependency on ophyd.

src/bluesky/simulators.py
```diff
@@ -104,3 +106,279 @@ async def check_limits_async(plan):
...
+            A lambda that accepts a Msg and returns an object; the object is sent to the current yield
+            statement in the generator, and is used when reading values from devices, the structure of the
+            object depends on device hinting.
+        msg_filter
+            This can either be a string corresponding to the Msg obj.name property to match, or it can
+            be a predicate on the Msg which returns a bool. The default is None which will match all
+            messages.
+        index
+            An optional integer indicating where to insert the handler, the default is 0 which
+            prepends it so that newer handlers override older ones. Specify -1 to append the handler.
```
tacaswell, line 183: Any negative number puts it at the end.  Should `-3` put it 3 in from the back?

src/bluesky/simulators.py
tacaswell: There is now a `declare_stream` message which will generate the descriptor, but it is optional.

src/bluesky/simulators.py
rtuck99: I think negative array indexing other than to append is an edge case, and would also be slightly confusing as it would differ from `list.insert()` where -1, -2, -3 inserts counting from just before the end instead of after, (hence I can't just delegate to `insert()`)
I think we can be fairly confident most common use cases are going to be append (default) and prepend, and positive-index as maybe a distant 3rd place. When writing tests you don't want to be worrying about keeping track of indexes in fixture code. I am going to make it raise an exception rather than leave as undefined behaviour.

src/bluesky/simulators.py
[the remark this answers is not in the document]
rtuck99: It now returns an instance of Reading with the timestamp initialised to the current time

src/bluesky/simulators.py
tacaswell: I would expect this to behave exactly like the `list.insert` as that is easy to document and explain ("it does the normal thing") rather than trying to document that it does something different (I had an off-by-one bug in my previous comments).

rtuck99: but then you can't append to the handlers without keeping track of the indices, which makes it much less useful.

tacaswell: I am surprised that `insert` can not put something last!  Could probably do that by taking `int | 'end'` and switch to `append` for 'end'.

src/bluesky/simulators.py
```diff
@@ -102,5 +117,285 @@ async def check_limits_async(plan):
...
+    def add_read_handler_for(self, obj_name: str, value: Any, dict_key: str = "values"):
+        """
+        Convenience method to register a handler to return a result from a 'read' command.
+
+        Parameters
+        ----------
+        obj_name
+            Name of the object to intercept reads for
+        value
+            The value to return. If value is a dict then dict_key is ignored and the result of the read
```
tacaswell, line 226: Ignoring inputs of one parameter based on the value of another is generally a poor pattern.  Could this be a good case for `**kwargs` where if the value is a dict, it is anything else, gets a timestamp automatically appended?

```python
sim.add_read_handler("pin_tip_detection-triggered_tip", a=(8, 5), b=5, c={'value': 'bob', 'timestamp': 0})
```

which would return

```python
{
 'a': {'value': (8, 5}, 'timestamp': XXX},
 'b': {'value': 5, 'timestamp': XXX},
 'c': {'value': 'bob', 'timestamp': 0}
}
```

on `'read'`?

src/bluesky/simulators.py
```diff
@@ -102,5 +117,285 @@ async def check_limits_async(plan):
...
+
+        Parameters
+        ----------
+        command
+            Name of the command to match.
+        document_name
+            Name of the document to match when used in single shot mode.
+        document
+            The document to fire in single shot mode.
+        docs: Sequence of Sequence of document_name, document tuples
```
tacaswell, line 307: Again, having two different call patterns squashed into one signature is a less than great pattern.

This might be better as two methods (one of which does indeed call the second with two layers of nesting).  I was going to suggest `add_callback_handler_for(..., *namedoc, msg_filter=None)` but the second level of nesting makes that not work.

src/bluesky/simulators.py
```diff
@@ -102,5 +117,285 @@ async def check_limits_async(plan):
...
+        list[Msg]
+            list of the messages generated by the plan
+        """
+        messages = []
+        send_value = None
+        try:
+            while msg := gen.send(send_value):
+                send_value = None
+                messages.append(msg)
+                LOGGER.debug(f"<{msg}")
```
tacaswell, line 360: Best practice is to let the logging framework do the templating to avoid paying the cost of generating the string if the logging level is such that this log never gets emitted. 
```suggestion
                LOGGER.debug("<{%s}", msg)
```

(and similar below)

src/bluesky/simulators.py
```diff
@@ -102,5 +117,285 @@ async def check_limits_async(plan):
...
+):
+    """Find the next message matching the predicate, assert that we found it.
+
+    Returns
+    -------
+    list[Msg]
+        All the remaining messages starting from the matched message which is included as a
+        convenience to capture information that may be used in subsequent calls.
+    """
+    indices = [
```
tacaswell, line 388: There is probably a cute way to write this as

```
try:
    inxd = next(i for i, msg in enumerate(mesages) for predicate(msg))
except StopIteration:
    raise ValueError("no match found")
return messages[i:]
```

but that might be too cute

src/bluesky/simulators.py
```diff
@@ -102,5 +117,285 @@ async def check_limits_async(plan):
...
+        i
+        for i in range(len(messages))
+        if (not group or (messages[i].kwargs and messages[i].kwargs.get("group") == group))
```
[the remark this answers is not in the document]
rtuck99: I've changed the code to handle unnamed groups. We were writing a number of tests which were asserting various operations performed in groups that were subsequently`wait`ed and it seemed like this would be useful.

src/bluesky/simulators.py
rtuck99: I've added `add_read_handler_for_multiple` in order to separate the two cases, likewise `add_callback_handler_for` now has `add_callback_handler_for_multiple`

tacaswell (approved):

rtuck99:
> It looks like we will have to pin back pyepics on py3.8 ?

Possibly. I'm not sure who has the dependency on python 3.8.

PR #1717, opened 2024-04-17
Pin Sphinx to <7.3

callumforrester:
7.3 appears to be breaking the build, this PR pins it to unblock people. Issue to investigate further: https://github.com/bluesky/bluesky/issues/1716


mrakitin (approved):
Looks good to me.

Issue #1646, opened 2023-12-15
Replace all try/except blocks that are used for logic parsing wherever easy/possible in the RunEngine


rosesyrett:
In general it seems most people (from the NSLS-II, Diamond collaboration meetings) are in agreement that try/except blocks shouldn't be used for parsing logic. I understand sometimes it's inevitable and actually useful to use try/except blocks in this way (especially when it relates to handling exceptions with generators and passing exceptions around). However there are a lot of relatively simple use cases where try/except could be replaced with if/else for example, and it would improve the readability of the codebase.

## Current Behavior
We have instances of try/except being used in a few places where it's probably not the best tool to use. For example, often try/except is used to catch KeyErrors:

```python
        try:
            current_run = self._run_bundlers[run_key]
        except KeyError as ke:
            raise IllegalMessageSequence(...)
```
This sort of pattern is found almost everywhere. Instead, why not replace it with:
```python
        current_run = self._run_bundlers.get(run_key)
        if not run_key:
            raise IllegalMessageSequence(...)
```

Another obvious place where this would improve readability is in `RunEngine._unsubscribe`:
```python
    async def _unsubscribe(self, msg):
        ...
        _, obj, args, kwargs, _ = msg
        try:
            token = kwargs['token']
        except KeyError:
            token, = args
        ...
```

Should be replaced with:

```python
    async def _unsubscribe(self, msg):
        ...
        _, obj, args, kwargs, _ = msg
        token = kwargs.get("token", args[0])
```

There's probably a lot more examples of this sort of thing, but these were the easiest I could find.

tacaswell:
I am ambivalent to this.  

It does shorten the code a bit, but it lifts error checking that Python was doing for us up to our code (that is if we break the try-execpt `__getitem__` still raises `KeyError` and will still fail (but with a less-good error message/type) but if we are using `get` we can drop the return value checking and end up with `None` breaking in weird ways farther down the line).  There may have also been a performance argument to dropping try/except, but recent Python versions have made them much cheaper (free?).

I'm not sure the second example is correct.  I think it will fail if `len(args) == 0` (as the default value to `get` is evaluated before `get` is called) and will accept (but drop on the floor) any additional args past the first one.  Something like:

```python
In [4]: def test(args, kwargs):
   ...:     if (token := kwargs.get('token')) is None:
   ...:         token, = args
   ...:     return token
   ...:
```

is probably a better replacement.  

[...]

PR #1718, opened 2024-04-17
(#1646) replaces KeyError handling logic

CoePaul:
* Switches out KeyError handling with walrus operator based alternative logic in the run_engine script

## Description

Each clause which used to catch raised KeyError instances has 
had its logic replaced by a suggested walrus operator based flow control
checking for a key absence sentinel in the getter return from a dict key look up

## Motivation and Context

The older, replaced logic was considered clunky and inappropriate for flow control

Fixes issue #1646 

## How Has This Been Tested?

This has been tested locally in development and 
via the usual CI suite.  

No new tests are added as this is a like for like refactoring replacement, 
not a development of new features nor new methods.

### DIFF
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -1638,12 +1638,11 @@ class RunEngine:
                         self._msg_cache.append(msg)
 
                     # try to look up the coroutine to execute the command
-                    try:
-                        coro = self._command_registry[msg.command]
-                    # replace KeyError with a local sub-class and go
-                    # to top of the loop
-                    except KeyError:
-                        # TODO make this smarter
+                    if (
+                        coro := self._command_registry.get(msg.command, key_absence_sentinel := object())
+                    ) is key_absence_sentinel:
+                        # flag invalid command
+                        # and return to the top of the loop
                         new_response = InvalidCommand(msg.command)
                         continue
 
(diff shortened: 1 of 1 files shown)


tacaswell:
@callumforrester Can we sort out why the auto-formatters are fighting each other?  I would really like to avoid the code churn when some other tool puts it back....

Can we only commit the relevant lines in this PR?

tacaswell:
mostly 👍🏻 on the content of the changes, but I have a concern that we are sure that in all of these dictionaries than `None` will never be a valid value.  It may be safer to do something like

```python
_RE_EMPTY_MARKER = object()

...

if (val := dictionary.get(key, _RE_EMPTY_MARKER)) is _RE_EMPTY_MARKER:
   ...

....
```

which checks for a value that can never (well, if it does it is a bug) end up in a dictionary.

CoePaul:
> ... but I have a concern that we are sure that in all of these dictionaries than `None` will never be a valid value. 
> It may be safer to do something like...

I think of the dozen or so instances where KeyError handling has been touched by the change, **most** of them go on to use the dict value as an instance of an object and try to invoke a method via the dot operator.  In that majority of examples - I think `None` is (or should not be) permitted as a value ( even if the dict was storing value `None` against the key )

There are one or two where None is fine and the next pull request should tackle that (I'll submit it once I'm done typing this)

tacaswell:
Even in the cases where `None` is a bad value, debugging "How did a None end up where it should not?" is a very different journey than debugging "Why is a key missing?".

CoePaul:
> re: different debugging journeys...

Good point - and well made!

src/bluesky/run_engine.py
```diff
@@ -1638,12 +1638,11 @@ async def _run(self):
...
 
                     # try to look up the coroutine to execute the command
-                    try:
-                        coro = self._command_registry[msg.command]
-                    # replace KeyError with a local sub-class and go
-                    # to top of the loop
-                    except KeyError:
-                        # TODO make this smarter
+                    if (
+                        coro := self._command_registry.get(msg.command, key_absence_sentinel := object())
```
tacaswell, line 1642: double walrus!

tacaswell (approved):

tacaswell (approved):
The code missing is the sad-paths that were not covered before.

Adding coverage is not useless, but not high enough priority for me to hold the PR over it.

PR #1720, opened 2024-04-17
Add collect_while_completing plan stub and test

jsouter:
Addressing https://github.com/bluesky/bluesky/issues/1641
Tests are currently trivial, should add in more to probe the behaviour of _wait_and_move_on


jsouter:
I believe this logic is correct for the plan stub, though I have not yet added any logic to handle pauses, as mentioned in the original issue.

jsouter:
Have moved the wait_and_move_on logic into a move_on kwarg for the existing _wait method, though the logic for handling pauses is yet to be implemented.

src/bluesky/tests/test_flyer.py
```diff
@@ -4,17 +4,80 @@
...
+        def kickoff(self):
+            return NullStatus()
+
+        def complete(self):
+            completed.append(self)
+            return self.custom_status
+
+    class TrivialDetector:
+        name = "trivial-detector"
+
```
coretl, line 61: Please could you add a `collect` method to this that makes an incrementing number each time it is called, then check that you get the right number of `Event` documents out of the RE by adding a callback.

Something along the lines of this might help:
```python

def assert_emitted(docs: Dict[str, list], **numbers: int):
    assert list(docs) == list(numbers)
    assert {name: len(d) for name, d in docs.items()} == numbers

def test_plan():
    docs = defaultdict(list)
    RE(plan(), lambda name, doc: docs[name].append(doc))
    assert_emitted(docs, start=1, descriptor=1, resource=1, datum=1, event=1, stop=1)
```

coretl:
@tacaswell I commented on the tests, but I'm happy with the code. Are you happy with it too?

src/bluesky/tests/test_new_examples.py
```diff
@@ -697,13 +697,15 @@ def test_trigger_and_read(hw):
     for msg in msgs:
         msg.kwargs.pop("group", None)
         msg.kwargs.pop("timeout", None)
+        msg.kwargs.pop("move_on", None)
```
tacaswell, line 700: Can we add this to the expected rather than popping here?  In the case of `group` it is random so we can not write down what it _should_ be, but `move_on` is a bool so we should be able to!  We probably should be checking that the group is internally self-consistent (the blame on that goes back to 2016).

I think it is fair to ask for `move_on` to be handled better, but bonus points for fixing timeout and group as well!

jsouter:
Okay, have added in a test to check that collect is called the right number of times/the right events are emitted and explicitly check the value of move_on in test_new_examples.py

src/bluesky/plan_stubs.py
```diff
@@ -813,6 +814,43 @@ def collect(obj, *args, stream=False, return_payload=True, name=None):
...
+    msg : Msg
+        A 'complete' message or 'collect' message
+
+    See Also
+    --------
+    :func:`bluesky.plan_stubs.complete`
+    :func:`bluesky.plan_stubs.collect`
+    """
+    group = short_uid(label="complete")
+    for flyer in flyers:
```
abbiemery, line 846: There is now a `complete_all plan` stub, maybe this could be used here? Or would we prefer to keep it with less calling to other plans? (Opinions wanted)

coretl: Yeah, I'd probably use `complete_all(*flyers, wait=False, group=group)`

coretl:
@jsouter do you have time to look at the last few comments on this please?

jsouter:
> @jsouter do you have time to look at the last few comments on this please?

Yep, will try and get those changes pushed through today

PR #1724, opened 2024-04-17
Add OpenTelemetry tracing to some RunEngine methods

### LINKED ISSUE #1682: Add metrics on how long waits took to the RunEngine
In general we would like to be able to get tracing and metric information out of the RunEngine. 

The specific MX DLS use case is that we would like to have good data on where the plans are waiting for hardware to finish doing things. 

## Expected Behavior
As a plan runs through information about how long `wait` messages took should be easily externally available This should be invisible to the plan and turned on via some flag before running the plan e.g. in the RunEngine.

## Possible Solution
From some discussion it seems that https://opentelemetry.io/docs/languages/python/ is the agreed module to use for this. Further discussion may be needed for specifics

dperl-dls:
## Description

Submitted together with https://github.com/bluesky/ophyd/pull/1189

- adds `opentelemetry-api` as a dependency
- wraps some potentially long-running `RunEngine` methods (`wait()`,`set()`,`complete()`...) in tracing spans
- creates spans at `_open_run` which are ended on `_close_run` (or abort, halt etc)
- attaches some metadata about each of the above

## Motivation and Context
Fixes #1682 

## How Has This Been Tested?
The `api` component of `OpenTelemetry` is designed to not really do anything on its own - I welcome suggestions about tests to include here.
This was tested by setting up a `TracingProvider` in Hyperion and exporting the results from running our unit tests to a Jaeger container.

Making use of these features in Hyperion is currently in [this branch](https://github.com/DiamondLightSource/hyperion/tree/bluesky_1682_improve_tracing), and an example of how to set up a tracing provider can be found [here](https://github.com/DiamondLightSource/hyperion/blob/bluesky_1682_improve_tracing/src/hyperion/tracing.py).

screenshot-2024-04-17-16:04:56

screenshot-2024-04-17-16:06:23

screenshot-2024-04-17-16:30:56


callumforrester:
Power cycle

src/bluesky/tests/test_tracing.py
```diff
@@ -0,0 +1,26 @@
...
+    assert isinstance(_test_plan(), Generator)
+
+
+def test_trace_plan_wrapper_opens_span_but_doesnt_do_anything_without_trace_provider():
+    @trace_plan(tracer, "test_plan_2")
+    def test_plan_2():
+        yield from sleep(0)
+        assert not get_current_span().is_recording()
+
+    RE = RunEngine()
```
mrakitin, line 25: Why not to use the `RE` fixture?

mrakitin (commented):
Looks good overall. I have a couple of questions here:
1) How easy it will be to opt-out from tracing?
2) I wonder if we could apply the traces as a `RE.msg_hook`?

dperl-dls:
Thanks for the review @mrakitin !
1. Opting out from tracing means just doing nothing, to actually do anything with these tracing spans is opt-in. Each of these tracing spans is essentially a no-op (well, an `if false`), unless some application code has used the `opentelemetry-sdk` library to set up some tracing provider and exporter. They're basically just hooks for application tracing code to inject into. If you run some code as in https://github.com/DiamondLightSource/hyperion/blob/bluesky_1682_improve_tracing/src/hyperion/tracing.py to set up exporters, then those hooks will be used to determine what info to export.
2. I will look in to it - my immediate guess is that it might be difficult because these traces are meant to wrap functions or be created and destroyed according to some logical unit of work the program is doing, which means we need some corresponding hook at the end of message processing. It's a little more complex than I really wanted this PR to be, but if there is interest I could look at making a `msg_context_hook` or something like that to wrap message processing in some given context manager, which would allow that to work and would be quite elegant IMO.

danielballan:
I am not so sure this is a good fit for `msg_hook`. The `msg_hook` receives a generic input (`Msg`) and performs a generic action (log a line of text). Telemetry Traces are richer things: each _type_ of Msg (each `msg.command`) may suggest custom `attributes` and so on that are best handled inside the specific coroutine where the relevant context is held.

---

1. Should Msg types be differentiated by a part of a span name that has to be parsed out of a string (`Bluesky RunEngine set`, as currently implemented) or via an `attribute` (`set`)? Or both?
2. Should the message content by logged as a single attribute or as separate attributes for each piece (`msg.command`, `msg.args`, `msg.kwargs`, `msg.obj`)?
3. I notice the use of `repr` for generating strings to place in the spans. In the past I've regretting using `repr` for this when the string may be parsed by code in other languages. A more strictly defined serialization (e.g. JSON) can be better.
4. Could a documentation page be added describing the schemas of the events emitted? This could serve as a both a reference for a developers and a helpful design document to aid discussions about how to improve/extend our use of OpenTelemetry.

dperl-dls:
Hi @danielballan , sorry I left this for so long - I've added a bunch of finer grained attributes, which I agree should be much more useful. I left in the `repr`s because they are pretty handy human readable things when looking through the traces by hand. I've also added some docs, mainly with a view of being a reference for devs.

danielballan (approved):
This addresses all my questions and suggestions. Seems solid. Thanks for including docs.

Issue #1701, opened 2024-04-16
Drop support for Python3.8 and 3.9


callumforrester:
Should remove references to unsupported versions from CI and pyproject.toml

              We should follow https://numpy.org/neps/nep-0029-deprecation_policy.html / https://scientific-python.org/specs/spec-0000/ as closely as we can without causing critical operational problems for any of the facilities who are actively involved in the collaboration.

I'm not sure if we have that formalized anyplace though.

_Originally posted by @tacaswell in https://github.com/bluesky/bluesky/issues/1670#issuecomment-2043051456_

danielballan:
As @prjemian recently raised on Mattermost, this is something the Technical Steering Committee should weigh in on.

prjemian:
I added an issue to the governance repo

mrakitin:
Python 3.8 will be EOL in October 2024. I am not concerned about dropping its support. For 3.9 - we can probably consider it once 3.13 is out?

danielballan:
Oh, I sent an email to the TSC before I saw @prjemian's comment.

prjemian:
I still believe supporting on the two most recent versions is an aggressive stance for this toolchain.  Lots of repositories to update to that standard.

tacaswell:
Python is on a 12mo release cadence now so in effect every year we gain a version and lose a version.  Under NEP29 we at 3 versions (3.10, 3.11, 3.12) and in the fall we will be at (3.11, 3.12, 3.13).

maffettone:
I have no strong opinions on what is or isn't supported. But the PR above is ready for review pending that decision point. At the very least, this will create issues when we are out of date using a cron job GHA.

danielballan:
Discussion is ongoing among the Technical Steering Committee. This perspective (longer support window for old version of Python) is represented.

prjemian:
The point raised by @padraic-shafer is new, we must consider the beam line staff as participating developers in this process.  By the time new versions become available in the full software suite available to beam line developers, the 2-version window for supported versions becomes vanishingly short.

danielballan:
Fair, let's document that as another factor to consider when choosing the SPEC 0 + time lag window on a per project basis.

tacaswell:
I think we also need to make sure we keep separate local support (e.g. I am very much against deleting old conda envs from NSLS-IIs systems so anything that works now still works ~forever) with the minimum version of Python supported by new releases of the libraries.

The case where this gets jammed is when you need both:
[...]

PR #1726, opened 2024-04-18
1636 hidden xray vision dependency is still not dealt with anywhere

### LINKED ISSUE #1636: Hidden `xray-vision` dependency is still not dealt with anywhere
There's still a [`xray-vision`](https://pypi.org/project/xray-vision/) hidden dependency [in the LiveImage callback](https://github.com/bluesky/bluesky/blob/9b0d72953cae867d8312cde3da2ed9cc4caf0788/bluesky/callbacks/broker.py#L80), mentioned in previous issues (https://github.com/bluesky/bluesky/issues/371 https://github.com/bluesky/bluesky/issues/823). This dependency is not mentioned anywhere in the `README`, the docs, or the requirements, so you only encounter it if you're reading the code, or if you happily stumble upon a nice `ImportError` on your bluesky journey :slightly_smiling_face: 

## Expected Behavior
No error should be raised on normal operation, if you install a package directly from where it is made available oficially.

## Possible Solution
1. Vendor-in the relevant code from `xray-vision` into Bluesky.
2. Add the `xray-vision` dependency to the project requirements (possibly optional, like the `matplotlib` dependency, though it should not make the callback accessible when it is not installed).

stan-dot:
closes #1636


stan-dot:
some (`test_simulators /old modulenae, plot ratser path, test_live_plotter,  test_disable,  test_many-grids`) tests are failing for reasons seemingly not connected with the task at hand.

Discussion needed to validate file structure choices.

src/bluesky/plans.py
```diff
@@ -21,6 +21,11 @@
 from .utils import Msg, get_hinted_fields
 
 
+def hide_last_line():
```
callumforrester, line 24: This probably shouldn't be public

```suggestion
def _hide_last_line():
```

src/bluesky/callbacks/broker.py
```diff
@@ -1,117 +1,13 @@
-import os
 import time as ttime
 
-import numpy as np
-
 from ..utils import ensure_uid
-from .core import CallbackBase
-
-
-class BrokerCallbackBase(CallbackBase):
```
tacaswell, line 10: Why did this get moved to the other file?

src/bluesky/plans.py
```diff
@@ -1505,8 +1510,7 @@ def tweak_core():
             sys.stdout.flush()
             yield Msg("wait", None, "A")
             clear_output(wait=True)
-            # stackoverflow.com/a/12586667/380231
-            print("\x1b[1A\x1b[2K\x1b[1A")
+            hide_last_line()
```
tacaswell, line 1513:
```suggestion
            _hide_last_line()
```

tacaswell:
- I think we should make `mpl_package` private so that we do not get anyone depending on it (or if they do we are not obligated to feel bad about breaking them)
 - if we are going to move classes out of `callbacks/broker.py` than we have to import them back so we do not break anyone who was using them.

src/bluesky/plans.py
stan-dot: here I tripped on github auto suggestions

src/bluesky/callbacks/broker.py
stan-dot: made sense to me at a time, I'm not opinionated about this now, as I mentioned above ('Discussion needed to validate file structure choices.')

stan-dot:
@tacaswell  how to make the package private?

could you point me to an example of 'importing back' inside this repository?

now I am working on fixing the tests

danielballan:
> how to make the package private?

Rename it with a leading underscore like `_mpl_package`. Just a signal to developers that backward-compatibility is not guaranteed.

> could you point me to an example of 'importing back' inside this repository?

Before:

```py
# a.py
def f():
    ...
```

After:
```py
# b.py
def f():
    ...

# a.py
from .b import f  # for back-compat
```

stan-dot:
ok will do, thanks

stan-dot:
only the tests for 3.8 and 3.9 are failing due to typing annotations. action options
1) refactor them out
2) await until the [discussion](https://diamondlightsource.slack.com/archives/C04TCA0AF2P/p1713433889825369) finishes and support for 3.9 is dropped

stan-dot:
discussion is also in #1701

tacaswell:
How annoying is it to make the types work with 3.8 and 3.9?

stan-dot:
@tacaswell not very much, would take me 30 minutes

stan-dot:
just fixing it now

fuzz/fuzz.py
```diff
@@ -0,0 +1,460 @@
+import asyncio
```
tacaswell, line 1: I don't think this file should have comeback?

callumforrester: Indeed, probably just needs a rebase

src/bluesky/callbacks/olog.py
```diff
@@ -173,5 +173,5 @@ def start(self, doc):
             pformat(doc),
         )
         olog_status = self.client.log(document_content, logbooks=self.logbook)
-        logger.debug("client.log returned %s", olog_status)
+        logger.debug(f"client.log returned {olog_status}")
```
tacaswell, line 176:
```suggestion
        logger.debug("client.log returned %s", olog_status)
```

Calling it this way defers paying the cost of turning `olog_status` into a string until we actually need it.  Even if it is cheap here, it is best-practice do logging calls this way.

src/bluesky/callbacks/_mpl_image_cross_section/cross_section.py
```diff
@@ -0,0 +1,469 @@
+import logging
```
tacaswell, line 1: Given that we dropped all of the rest of the code, we might as well consolidate all of this code into a single `_mpl_image_cross_section.py` ?

src/bluesky/callbacks/_mpl_image_cross_section/cross_section.py
```diff
@@ -0,0 +1,469 @@
...
+        aspect="equal",
+    ):
+        self._cursor_position_cbs: List[Any] = []
+        self._interpolation = interpolation
+        # used to determine if setting properties should force a re-draw
+        self._auto_redraw = auto_redraw
+        # clean defaults
+        if limit_func is None:
+            # note - cannot import this in the defaults init
+            limit_func = fullrange_limit_factory()
```
tacaswell, line 102: Take this opportunity to change the default to percentile autolimits?

src/bluesky/callbacks/olog.py
stan-dot: the change was with an autoformatting tool, likely ruff. maybe we should add a comment that we do for performance reasons the change into using %s and add a noqa comment

fuzz/fuzz.py
stan-dot: deleted fuzz

src/bluesky/callbacks/_mpl_image_cross_section/cross_section.py
stan-dot: consolidating now

src/bluesky/callbacks/_mpl_image_cross_section/cross_section.py
stan-dot: done

stan-dot: @tacaswell turns out this method (percentile_limit_factory) needs parameters, what should be the value?

Also I am not sure what is the preferred type for this, should it be a tuple of floats, an ndarray or a list?

src/bluesky/callbacks/olog.py
stan-dot: done

stan-dot:
update: check again on this in a week

src/bluesky/callbacks/_mpl_image_cross_section/cross_section.py
tacaswell: I would go with 5, 95

stan-dot:
todo: fix the type of the parameter

stan-dot:
across the 3 limit factories, each describes the 'limit_args' differently - one as array, one as object, one as tuple. that is inconsistent  and we cannot proceed further. what to do @tacaswell ?

tacaswell:
two of them can be homogenized to "array like" (absolute and percentile) and the third can be set the same (it is ignored so it can be `Any` but making it stricter seems fine?).

I also noticed that the docstring for `limit_args` is on the returned function on the factory in all three cases which is wrong.

src/bluesky/callbacks/_mpl_image_cross_section.py
```diff
@@ -0,0 +1,583 @@
...
+           of the image.
+
+        Returns
+        -------
+        climits : tuple
+           length 2 tuple to be passed to `im.clim(...)` to
+           set the color limits of a ColorMappable object.
+
+        """
+        return limit_args
```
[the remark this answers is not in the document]
stan-dot: done

stan-dot:
I have no idea how to test this drawing logic at the moment

stan-dot:
will monkey patch the matplotlib method

src/bluesky/tests/test_cross_section_image.py
```diff
@@ -0,0 +1,248 @@
...
+    def test_function_call_and_return_value(self):
+        result = self.obj.some_function(2, 3)
+        self.assertEqual(result, 5, "The decorated function should return the sum of inputs")
+
+    def test_redraw_called_when_auto_redraw_true(self, mock_cross_section: CrossSection):
+        mock_cross_section.draw_idle()
+        self.obj.some_function(2, 3)
+        self.obj._fig.canvas.assert_not_called()
+        self.obj._update_artists.assert_called_once()
+        self.obj._draw.assert_called_once()
```
DominicOram, line 77: A cleaner way to mock this is:
```suggestion
@auto_redraw
def some_function(obj, x, y):
    return x + y

class TestAutoRedrawDecorator(TestCase):
    def test_function_call_and_return_value(self):
        result = some_function(MagicMock(), 2, 3)
        self.assertEqual(result, 5, "The decorated function should return the sum of inputs")

    def test_redraw_called_when_auto_redraw_true(self):
        mock_obj = MagicMock()
        mock_obj._auto_redraw = True
        some_function(mock_obj, 2, 3)
        mock_obj._fig.canvas.assert_not_called()
        mock_obj._update_artists.assert_called_once()
        mock_obj._draw.assert_called_once()
```

And so on...

src/bluesky/tests/test_cross_section_image.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+class TestCrossSection(TestCase):
+    def setUp(self):
+        self.figure = MagicMock(spec=Figure)
+        self.cross_section = CrossSection(figure=self.figure, colormap="gray", auto_redraw=True)
+        self.cross_section._imdata = MagicMock(shape=(100, 100))  # Dummy image data
+        self.cross_section._cursor_position_cbs.append(MagicMock())  # Add mock callback
+
+    @patch("mpl_toolkits.axes_grid1.make_axes_locatable")
```
DominicOram, line 106: You're right that you need to patch this function but:
* When you're patching you actually need to patch where the function has been imported to, not where it was imported from
* You need to patch this before `CrossSection` is instantiated in `setUp`
* Annoyingly this is a bit fiddly https://stackoverflow.com/questions/15821465/how-to-properly-use-mock-in-python-with-unittest-setup

The following gives an instantiatable `CrossSection` and passes this test:
```python
class TestCrossSection(TestCase):
    def setUp(self):
        patch_divider = MagicMock()
        patch_divider.append_axes.return_value.plot.return_value = (0, ) # Patch for _ln_v and _ln_h
        patch("bluesky.callbacks._mpl_image_cross_section.make_axes_locatable", return_value=patch_divider).start()
        self.figure = MagicMock()
        self.cross_section = CrossSection(figure=self.figure, colormap="gray", auto_redraw=True)
        self.cross_section._imdata = MagicMock(shape=(100, 100))  # Dummy image data
        self.cross_section._cursor_position_cbs.append(MagicMock())  # Add mock callback

    def test_move_cb_outside_axes(self):
        event = MagicMock()
        event.inaxes = None
        self.cross_section._move_cb(event)
        self.cross_section._cursor_position_cbs[0].assert_not_called()
```

src/bluesky/tests/test_cross_section_image.py
```diff
@@ -0,0 +1,248 @@
...
+        # Mocking necessary matplotlib components
+        self.mock_figure = MagicMock()
+        self.mock_im = MagicMock()
+        self.mock_axes = MagicMock()
+        self.mock_ln_v = MagicMock()
+        self.mock_ln_h = MagicMock()
+        self.mock_canvas = MagicMock()
+
+        # Instantiate CrossSection with mocked figure
+        self.cross_section = CrossSection(figure=self.mock_figure)
```
DominicOram, line 149: As above this will fail w/o `make_axes_locatable` being patched

DominicOram (commented):
Some guidance on patching

src/bluesky/tests/test_cross_section_image.py
stan-dot: ah I forgot about this

stan-dot:
thanks for the guidance @DominicOram

stan-dot:
closing as not planned. `LiveImage` is the only place where this dependency is used and that does not have broad use.

All that could be found is a [2021 side branch of pdfstream project](https://github.com/NSLS-II-PDF/pdfstream/blob/bde056106f6231e684c778b937674bf38db1ca71/pdfstream/callbacks/basic.py#L169).

`LiveImage` was a prototype and does not feature full controls. Therefore a third party project will likely be much better at the task.
A strong candidate is [Napari](https://napari.org/stable/) . 

Further action suggested : `LiveImage` should be deprecated and if no concerns are raised, can be removed after some time.
@danielballan  what should be the timeframe? 6 months?

discussion done in [slack](
https://diamondlightsource.slack.com/archives/C04TCA0AF2P/p1718972759791919?thread_ts=1718901191.342649&cid=C04TCA0AF2P)

PR #1729, opened 2024-04-18
update tutorial

### LINKED ISSUE #1288: Refactor built-in plans to use contingency_wrapper
The `contingency_wrapper` is better than the `finalize_wrapper` because the "final plan" takes in the exception that was raised as the argument, which allows the clean up to target specific failure modes if desired.

Even where that capability isn't used, it would be an improvement I think to model using `contingency_wrapper` in our code so that users can easily discover it and find examples of its use.

stan-dot:
fixes #1288

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1904,6 +1904,9 @@ This is the general idea:
             # Do this even if an Exception is raised.
             yield from cleanup_plan()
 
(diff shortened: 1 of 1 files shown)


prjemian (approved):
LGTM!

danielballan:
Test failure looks like a fluke, and clearly unrelated:

```
FAILED src/bluesky/tests/test_suspenders.py::test_unresumable_suspend_fail[False] - assert 0.1 < (1713449243.831193 - 1713449243.7326467)
```

Issue #1727, opened 2024-04-18
Replace use of tmpdir fixture with tmp_path


DiamondJoseph:
tmp_path returns a standard pathlib Path object, while tmpdir uses a legacy implementation. The docstring of the tmpdir fixture recommends replacing its use wtih tmp_path.

## Expected Behavior
Tests should continue to pass but using the new tmp_path fixture instead.

## Current Behavior
Tests use legacy tmpdir fixture.

## Possible Solution
Replace use of tmpdir in the arguments of fixtures with `tmp_path: Path`

PR #1730, opened 2024-04-18
Use tmp_path instead of tmpdir fixture

jwlodek:
## Description

Addresses #1727. This was the only occurrence of the `tmpdir` fixture in the `tests` directory currently.

Tests seem to run OK.


danielballan (approved):

DiamondJoseph (approved):

DiamondJoseph:
Nice, didn't realise there was only the one instance. I just saw it being used on the new code and wanted to come back to it and investigate later.

PR #1733, opened 2024-04-18
(#1646) replaces KeyError handling logic

CoePaul:
* Switches out KeyError handling with walrus operator based alternative logic in the run_engine script

* Uses sentinel fallback values on each key look up to detect (in an inline fashion) when the key is absent

## Description

Completes review requests from PR  #1718 all in one commit
(where the previous PR had gotten strung out into several commits, 
owing to a formatting snafu unrelated to the PR topic )

## Motivation and Context

Tackles #1646 which requests a refactoring of KeyError handling code in RunEngine

## How Has This Been Tested?
Only tested using the CI and locally run pytests


CoePaul:
Killing redundant PR

Issue #1723, opened 2024-04-17
Refactor bluesky/utils/__init__.py into component modules


prjemian:
With PR #1709,  `bluesky/utils/__init__.py` is approaching 2,000 lines with dozens of classes and functions at the module level.  It is time to refactor this file into its component modules

_Originally posted by @prjemian in https://github.com/bluesky/bluesky/issues/1709#issuecomment-2061444739_

prjemian:
For example, this will make it easier for me to find the code that provides the [`Msg()`](https://github.com/bluesky/bluesky/blob/a054fab2e3432534a3bebdee906f162eccb0cfe8/src/bluesky/utils/__init__.py#L66) objects.

callumforrester:
Duplicate of #1677

PR #1734, opened 2024-04-18
first attempt: moved everything to utils.py

skarakuzu:
My initial attempt is to move all staff in `__init__.py` into `utils.py` in utils directory. This draft PR is to be improved further.


danielballan:
@callumforrester Do you think we should merge this as "forward progress" or keep splitting up `utils.py` until there is nothing left? That might not be a "good first issue".

prjemian:
Already it is easier to find things.  (per #1723)

Issue #1681, opened 2024-04-04
Dependencies - Transitioning to Redis for Enhanced Persistent Storage Solution


stan-dot:
Transitioning to Redis for Enhanced Persistent Storage Solution

Expected Behavior

The project should have a robust, secure, and efficient mechanism for persistent storage across IPython sessions. An ideal solution would be widely adopted, supported, and capable of meeting our scalability and performance needs without compromising security.

Current Behavior

Currently, we utilize a makeshift solution involving historydict for syncing RE.md to a SQLite database, and have explored zict as an alternative for disk-based dict representation. However, zict's use of pickle files for long-term data persistence poses security concerns. [source](https://github.com/bluesky/bluesky/issues/1227#issuecomment-511559612) Moreover, [zict](https://github.com/dask/zict) has relatively low community adoption, indicated by its 68 GitHub stars.

Possible Solution

Given the challenges we've faced, I wonder if we might explore the possibility of integrating Redis into our workflow. Redis is well-regarded for its performance and has a large community (64.4k stars on GitHub).It has a straightforward [Python SDK](https://redis.io/docs/connect/clients/python/).  However, I'm aware that adopting Redis would be a significant change, and there are considerations we'd need to address, such as operational complexity, dependency management, and ensuring data security.

Potential Risks:

    Operational Complexity: Integrating Redis could introduce new complexities in deployment and maintenance.
    Dependency Management: Adding Redis as a dependency may impact the project's setup and installation process for new users.
    Data Security: We must ensure that using Redis does not introduce security vulnerabilities, particularly in how data is accessed and stored.

SMART Goal:
To determine the viability of Redis for our project, I propose a pilot project with the following SMART goal: "By [specific date, within 3-6 months], conduct a evaluation of Redis by implementing a prototype that addresses the current limitations of historydict and zict, assessing operational complexity, dependency management, and data security concerns, with the aim of presenting findings to the team."

Steps to Reproduce (for bugs)

N/A
Context

Moving to a more stable and secure method would greatly benefit the project's long-term sustainability and ease of use for both new and existing users.

Your Environment

N/A

prjemian:
Persistence across reboot of the host is vital.

stan-dot:
that is a common use case for redis https://docs.redis.com/latest/kubernetes/recommendations/persistent-volumes/

danielballan:
I love this framing of the process. I would very much like to see Bluesky support and  recommend a Redis-based solution.

[...]

PR #1736, opened 2024-04-19
[WIP] Add Redis-backed persistent dictionary for metadata

hyperrealist:
This PR addresses #1681.

- [ ] Add RedisPersistentDict to utils
- [ ] Write tests
- [ ] Update documentation

### DIFF
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -17,25 +17,16 @@ from collections import namedtuple
 from collections.abc import Iterable
 from functools import partial, reduce, wraps
 from inspect import Parameter, Signature
-from typing import (
-    Any,
-    AsyncIterable,
-    AsyncIterator,
-    Callable,
-    Dict,
-    List,
-    Optional,
-    Tuple,
-    Type,
-    Union,
-)
+from typing import Any, AsyncIterable, AsyncIterator, Callable, Dict, List, Optional, Tuple, Type, Union
 from weakref import WeakKeyDictionary, ref
 
 import msgpack
 import msgpack_numpy
 import numpy as np
+import redis as _redis
 import zict
 from cycler import cycler
+from redis_json_dict import RedisJSONDict as _RedisJSONDict
 from super_state_machine.errors import TransitionError
 from tqdm import tqdm
 from tqdm.utils import _screen_shape_wrapper, _term_move_up, _unicode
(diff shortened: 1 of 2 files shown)


src/bluesky/utils/__init__.py
```diff
@@ -870,6 +861,74 @@ def reload(self):
...
+    Create persistent dictionary using an existing redis client
+    >>> d = RedisPersistentDict(redis_client=redis_client, prefix='my_prefix')
+    """
+
+    def __init__(self, *args, prefix="bluesky_metadata_", redis_client=None, **kwargs) -> None:
+        if redis_client is not None:
+            if not isinstance(redis_client, _redis.client.Redis):
+                raise TypeError("'redis_client' must be a valid Redis client object (redis.client.Redis).")
+            if kwargs:
+                warnings.warn("Keyrord arguments are ignored since 'redis_client' is specified.")  # noqa: B028
```
mrakitin, line 884:
```suggestion
                warnings.warn("Keyword arguments are ignored since 'redis_client' is specified.")  # noqa: B028
```

mrakitin (changes requested):
There was a discussion on a call that this code is not ready and should be converted to "Draft". We seemed to agree that we would not have `redis-json-dict` as a dependency for bluesky, but we'd rather like to have the migration path from the `PersistentDict` to `RedisJSONDict` documented.

@hyperrealist, could you please work on the documentation PR for that instead?

The idea is to not

danielballan:
Per discussion on stand-up call with DLS, mostly with @mrakitin and @stan-dot:

Bluesky should not pick up a dependency on a Redis client. Bluesky ships with `RE.md` being a plain `dict` by default. Users who want stateful metadata can `pip install redis-json-dict` and configure `RE = RunEngine(md=RedisJSONDict(prefix="bluesky_metadata"))` or similar. And course they could equally well experiment with other dict-like implementations, such as one built on [RedisJSON](https://github.com/RedisJSON/RedisJSON).

Therefore we propose instead a PR that:

- Documents how to install and configure `redis-json-dict` to get stateful metadata
- Documents a migration path for loading metadata out of the existing (zict-backed) `PersistentDict` and into Redis
- Marks `PersistentDict` as deprecated, with `redis-json-dict` as the new recommended solution

mrakitin:
@hyperrealist, will you have bandwidth to work on the items Dan identified? Thanks.

hyperrealist:
@mrakitin sorry about the hiatus, I have been away for the past couple of weeks. Pushing on another PR at the moment, but I will re-focus on this as soon as it is out, hopefully later today.

whs92:
We just used this solution and it helped us solve an annoying problem keeping things consistent between two run engines. 

Thanks so much for sharing!!!

PR #1743, opened 2024-05-08
Callback for Plotting from StreamDocuments

genematx:
Callback for Plotting from StreamDocuments

## Description
Intorduces a callback to consolidate stream documents and use them for interactive (automated) updates of matplotlib figures


src/bluesky/consolidators.py
skarakuzu, line 1: Aren't some of these data classes exportable from `tiled` already?

skarakuzu:
Is there an example output of these callback function? It would be very helpful to see it.

src/bluesky/consolidators.py
genematx: They are, you are right. the idea here is to eliminate (or at least, minimize) the dependancy of Bluesky on Tiled. Since these classes are rather simple, it's not hard to declare them again in bluesky code. In fact, I think we should simplify them even further and just leave whatever bare minimum is needed -- just haven't done this yet.

genematx:
@skarakuzu, not yet. The output should be a plot
What I find somewhat helpful is seeing what documents are being passed to the callback when running tests. We did this before with Dan: you can save the docuemnts as a separate json file (modify them and pass to the call back again, if you want), add breakpoints in the code and just inspect what's going on.
This is a piece of code I add to the tests to save the bluesky stream to a file:
```
import json

class JSONWriter:

"""Writer for a JSON array"""

   def __init__(self, filepath):
	self.file = open(filepath, "w")
	self.file.write("[\n")

  

   def __call__(self, name, doc):
	json.dump({"name": name, "doc": doc}, self.file)
	if name == "stop":
		self.file.write("\n]")
		self.file.close()
	else:
		self.file.write(",\n")

wr = JSONWriter('./demo_stream_documents/documents_test.json')
RE.subscribe(wr)    # subscribe it _before_ calling RE with your backend
```

stan-dot:
@genematx  I just learned that ALS had some efforts in this direction too
https://github.com/als-computing/dichroview

src/bluesky/consolidators.py
```diff
@@ -111,6 +112,18 @@ def __init__(self, stream_resource: StreamResource, descriptor: EventDescriptor)
...
+        if isinstance(self.chunk_size, int):
+            # Only the left-most dimension (rows) is chunked; other dimensions are in a single chunk
+            self.chunk_size = (self.chunk_size, *self.datum_shape)
+        else:
+            # Chunk size is specified for each dimension separately as a list or a tuple
+            if len(self.chunk_size) == len(self.datum_shape):
+                self.chunk_size = (0, self.chunk_size)  # Assume the leading dimension is not chunked
+            elif len(self.chunk_size) != len(self.shape):
+                raise RuntimeError(f"Dimensions of chunks {self.chunk_size} do not match the data {self.shape}.")
+
```
jwlodek, line 126: Didn't we settle on `chunk_shape` being in the sres parameters? Rather than `chunk_size`?

genematx: Yes, we did!
Sorry, I haven't realized you were getting notifications and paying attention to this PR. I was just updating my working branch with the current main (not even sure why I pushed it...). I will rebase it again once your `chunk_shape` PR is merged into `main`.

PR #1749, opened 2024-05-24
Update test for revised StreamResource.

danielballan:
The release of event-model 1.21.0 changed the schema of the experimental StreamResource document. This PR updates the tests to match.

This PR also fills in more realistic values for the dummy documents and switches to using keyword arguments. Neither of these changes affect how the code runs, but aim to make it a little easier to follow.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ dev = [
     "networkx",
     "numpydoc",
     "ophyd",
+    "packaging",
     "pandas",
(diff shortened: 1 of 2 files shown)


evalott100 (approved):
Thanks!

danielballan:
Knowing that this is temporary, I would be comfortable relaxing codecov in this case. Feel free to move this forward however you like; I don't want to block DLS progress.

PR #1766, opened 2024-07-15
apply suggestions to fix the xfail error and the case if no fields supplied

stan-dot:
Three lines change - adding a default value for fields, and removing pytest xfail lines

### DIFF
--- a/src/bluesky/callbacks/best_effort.py
+++ b/src/bluesky/callbacks/best_effort.py
@@ -134,7 +134,7 @@ class BestEffortCallback(QtAwareCallback):
         # for each dimension, choose one field only
         # the plan can supply a list of fields. It's assumed the first
         # of the list is always the one plotted against
-        default_field_name = 'time'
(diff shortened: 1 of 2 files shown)


danielballan (approved):

stan-dot:
@danielballan the xfail removal makes the tests fail, I think I forgot the minutiae of that part of the ticket - should the failing tests be removed or modified?

tacaswell:
IIRC those tests should work, but they were xfailed because we broke our simulation tools so we should leave the tests alone and either fix or modify the simulated devices we are handing in.

danielballan:
Discussion on call: Merge the change to BEC to fix that issue, but defer removing the `xfail` and fixing the simulation code to a separate issue.

stan-dot:
limiting the PR to the default fields, to make the new issue for the xfail logic which interlaces with the ophyd.sim. 
`hw.det` (module in ophyd)

PR #1767, opened 2024-07-17
Add bps.locate to plan_stub

### LINKED ISSUE #1753: Add bps.locate plan stub
We have a `Locate` message, but not a corresponding plan stub. Something like:
```python
@typing.overload
def locate(obj: Locatable, squeeze: Literal[True] = True) -> Location: ...
@typing.overload
def locate(*objs: Locatable, squeeze: bool = True) -> List[Location]: ...

def locate(*objs: Locatable, squeeze=True) -> List[Location]:
    return Msg("locate", *objs, squeeze=squeeze)
```
with the docstring from https://github.com/bluesky/bluesky/blob/main/src/bluesky/run_engine.py#L1959

subinsaji:
Closes #1753 

Added `Locate` to `bps` since was already in `RunEngine` (but not `bps`)


coretl (commented):
Please can we have the typing overloads in there too? And a test?

coretl (approved):

PR #1770, opened 2024-07-27
Windows: make plans using default `RunEngine` interruptible using CTRL-C

Tom-Willemsen:
On windows, the following is not interruptible using ctrl-c:

```python
RE = RunEngine()
RE(bps.sleep(999))
```

The same plan is interruptible under linux.

Note: it is interruptible on windows if you have imported `matplotlib` and matplotlib happens to be using one of the `qt` backends - that case has special handling already.

## Description

The underlying problem is that on windows, `threading.Event.wait()` cannot be interrupted by ctrl-c. See CPython issue 80116 for details.

I think the approach in this PR means that:
- `block()` still returns as soon as possible after the event is set
- CTRL-C events are handled reasonably responsively for interactive users (within 0.1 secs)
- The approach on linux is unchanged and continues to use a non-polling approach - which in theory is slightly more efficient

I chose 0.1 seconds fairly arbitrarily as a "reasonably responsive" wait timeout for an interactive user, my feeling is we probably don't need to expose this as a configuration parameter as people can always use a custom `during_task` arg to the runengine if they need special behaviour.

## Motivation and Context

At ISIS we run on windows and are starting to use bluesky.

While we can already solve this for our use-case by passing a windows-specific `during_task` into the run engine constructor, I think it would be better if the default run-engine constructor did this.

## How Has This Been Tested?
- CTRL-C handling is tricky to test in an automated way, but I've added a unit test that checks we call `wait()` with short timeouts on windows and retains existing behaviour on linux.
- Manually tested on windows
- Manually tested on linux in a `python:3.11` container

### DIFF
--- a/src/bluesky/tests/test_utils.py
+++ b/src/bluesky/tests/test_utils.py
@@ -1,6 +1,8 @@
 import operator
+import threading
 import warnings
 from functools import reduce
+from unittest.mock import MagicMock, call, patch
 
 import numpy as np
 import pytest
@@ -9,6 +11,7 @@ from cycler import cycler
 from bluesky import RunEngine
 from bluesky.utils import (
     CallbackRegistry,
+    DuringTask,
     Msg,
     ensure_generator,
     is_movable,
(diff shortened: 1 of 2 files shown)


tacaswell:
Did you mean to close this / was it handled someplace else?

Tom-Willemsen:
We're handling this locally by passing a similar `during_task` to our local `RE` constructors - happy to reopen, just closed it as part of regularly tidying up my stale PRs...

tacaswell:
I think it reasonable to improve the behavior on windows upstream!

tacaswell (approved):

tacaswell:
@Tom-Willemsen I took the liberty of rebasing.

Has your local implementation diverged from this?

PR #1778, opened 2024-08-02
add reset user position stub

stan-dot:
## Description

Adding a new stub representing a correction mechanic to the plan stubs list.

## Motivation and Context

As discussed in [the prior PR](https://github.com/bluesky/bluesky/pull/1301) a stub like this is useful in long-running experiments to prevent the RE from staying in an erroneous state.

## How Has This Been Tested?

Not yet.

### DIFF
--- a/src/bluesky/plan_stubs.py
+++ b/src/bluesky/plan_stubs.py
@@ -1814,3 +1814,37 @@ def repeat(
                     yield Msg("sleep", None, d)
 
     return (yield from repeated_plan())
+
+
+@plan
+def reset_user_position(obj, value, *, verbose=False):
+    """
+    Repeat a plan num times with delay and checkpoint between each repeat.
+    This is different from ``repeater`` and ``caching_repeater`` in that it
+    adds ``checkpoint`` and optionally ``sleep`` messages if delay is provided.
+    This is intended for users who need the structure of ``count`` but do not
+    want to reimplement the control flow.
+    Parameters
+    ----------
+    obj : EpicsMotor
+        Must have position attribute and set_current_position methods
+    value : number
+        What to set the current user position to.
+    verbose : bool, optional
+        If we should print out a log of what we are doing
+    Returns
+    -------
+    old_value : number
+        Where the user setpoint was before
+    new_value : number
+        Where the user setpoint is now
+    """
+    ret = yield Msg("reset_user_position", obj, value)
+    if ret is not None:
+        old_value, new_value = ret
+    else:
+        old_value = None
+        new_value = value
+    if verbose:
+        print(f"{obj.name} reset from {old_value} to {new_value}")
+    return old_value, new_value
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -536,6 +536,7 @@ class RunEngine:
             "RE_class": self._RE_class,
             "stop": self._stop,
             "set": self._set,
+            "reset_user_position": self._reset_user_position,
             "trigger": self._trigger,
             "sleep": self._sleep,
             "wait": self._wait,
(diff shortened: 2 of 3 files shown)


callumforrester:
@stan-dot Please could you add further clarification on the purpose of this plan stub? I'm not sure it should require `RunEngine` changes, if it does, can we find a more general solution than one that has to make the `RunEngine` aware of the `Positioner` API?

stan-dot:
@callumforrester  I updated the issue description

stan-dot:
should the idea be reviewed as ok, a test can be done

callumforrester:
I don't know enough about the background so I'll defer to others' opinions

stan-dot:
todo: add tests for beyond the happy path

stan-dot:
@coretl  given the `Locatable` interface, as expressed in https://github.com/bluesky/ophyd-async/pull/541 maybe this would sooner belong in ophyd-async?

coretl:
> @coretl given the `Locatable` interface, as expressed in [bluesky/ophyd-async#541](https://github.com/bluesky/ophyd-async/pull/541) maybe this would sooner belong in ophyd-async?

Personally I would agree with the approach in https://github.com/bluesky/bluesky/pull/1301#issuecomment-587965339. To mitigate plan interruption we could put motor record back into "SET" if not already there in `stage`, so that we will recover without human intervention.

If that is not acceptable then we need a new protocol to cover what is needed with a new method (which you called `set_current_position`) to set the current position of the motor. This should be `MaybeAsync`. `Locatable` should be used to get the current position of the motor in the plan stub.

stan-dot:
> To mitigate plan interruption we could put motor record back into "SET" if not already there in stage, so that we will recover without human intervention.

@coretl do you mean here?

> put motor record back into "SET" if not already there in stage,

is it in the ophyd_asycn Motor class, in an instance method 'stage' or 'set', or just in a stub, or in a plan when we expect that something *might* go wrong?

coretl:
> > To mitigate plan interruption we could put motor record back into "SET" if not already there in stage, so that we will recover without human intervention.
> 
> @coretl do you mean here?
> 
> > put motor record back into "SET" if not already there in stage,
> 
> is it in the ophyd_asycn Motor class, in an instance method 'stage' or 'set', or just in a stub, or in a plan when we expect that something _might_ go wrong?

https://epics.anl.gov/bcda/synApps/motor/motorRecord.html#Fields_calib

```python
def __init__(self, prefix: str):
    ...
    self.set_use = epics_signal_rw(MotorSetUse, prefix + ".SET")
    ...

async def stage(self):
    await self.set_use.set(MotorSetUse.USE)

async def set(self, value):
    status = self.position.set(value)
    ...
```

Then we can continue doing thing like this in the plan stub:
```python
def set_motor_position(motor: Motor, position: float):
    def make_motor_usable():
        yield from abs_set(motor.set_use, MotorSetUse.SET)

    def inner():
        yield from abs_set(motor.set_use, MotorSetUse.USE)
        yield from abs_set(motor.position, position)

    return (yield from finalize_wrapper(inner(), make_motor_usable()))
```

So unless the RE crashes, or is halted without doing cleanup, then motors will always be in USE mode no matter what happens when running `set_motor_position`, but if the `Motor` Device does encounter a motor in SET mode, it will be set back to USE mode in `motor.stage()`

stan-dot:
I see now, thanks @coretl 

regarding the naming, I am not sure what is the tango equivalent, but I feel that `use_set` is unintuitive and coupled with `EPICS`. The `stage` verb is available right there, could we explore naming the property something like `staging_status`?

coretl:
> regarding the naming, I am not sure what is the tango equivalent, but I feel that `use_set` is unintuitive and coupled with `EPICS`. The `stage` verb is available right there, could we explore naming the property something like `staging_status`?

I assume you mean `set_use`. This is most definitely EPICS specific, because we are referring to the `.SET` field of the motor record which I linked above. It could be named `mode` or `set_use_mode` if that is better, but it must be an enum:
```python
class MotorSetUse(str, Enum):
    use = "Use"
    set = "Set"
```
because that is what the motor record defines it to be

prjemian:
In the apstools package, I named the same operations as
`redefine_motor_position`.

https://bcda-aps.github.io/apstools/latest/api/_utils.html#apstools.utils.misc.redefine_motor_position

stan-dot:
@callumforrester , I didn't construe this (https://github.com/bluesky/bluesky/pull/1778#issuecomment-2321736240) as supportive of using EPICS terminology, and I mentioned the EPICS coupling as a negative feature of the solution, as I don't know what does TANGO use

coretl:
Closing as not planned

Issue #1779, opened 2024-08-03
Plan stubs are no longer pass ``inspect.isgeneratorfunction()``


canismarko:
## Expected Behavior

In version 1.13.0a3, passing plan stubs to [inspect.](`https://docs.python.org/3/library/inspect.html#inspect.isasyncgenfunction`) returns true.

```python
In [3]: from bluesky.plan_stubs import mv

In [4]: inspect.isgeneratorfunction(mv)
Out[4]: True
````

## Current Behavior

In version 1.13.0a3, passing plan stubs to [inspect.](`https://docs.python.org/3/library/inspect.html#inspect.isasyncgenfunction`) returns true.

```python
In [3]: from bluesky.plan_stubs import mv

In [4]: inspect.isgeneratorfunction(mv)
Out[4]: False
````

## Steps to Reproduce (for bugs)

1. ``import inspect``
2. ``from bluesky.plan_stubs import mv``
3. ``assert inspect.isgeneratorfunction(mv)`` 

## Context

The bluesky queueserver uses ``inspect.isgeneratorfunction()`` to determine which plans are available. Since plan stubs fail this test now, they are not available on the queueserver.

## Your Environment

https://gist.github.com/canismarko/39d39055a71b165faba801872162f7ba

tacaswell:
This is due to the decorator they all now carry.

I think the fix is unfortunately on the QS side to _also_ look for instances of wrapper class.

canismarko:
Thanks @tacaswell. I opened an issue with bluesky-queueserver and then realized there's already a PR that fixes it. Impressive!

jwlodek:
There will still need to be a PR to queueserver that uses this new feature instead of just `isgeneratorfunction`, but I can work on that once the first PR is merged.

PR #1780, opened 2024-08-05
Add private attribute for aiding in identifying if a function is a bluesky plan

jwlodek:
## Description
Adds helper functions that can identify if a function is either a generator co-routine, or has been marked with the `@plan` decorator.

## Motivation and Context
Addresses need for checking whether or not a function is a plan as described in: #1779 

## How Has This Been Tested?
Added a parametrized unit test that passes & checks both positive and negative expected results. All other tests pass without issue (locally, waiting on CI).

```
(venv) [jwlodek@dell-rhel8 bluesky]$ git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.

(venv) [jwlodek@dell-rhel8 bluesky]$ python
Python 3.11.9 (main, Jun 19 2024, 10:02:06) [GCC 8.5.0 20210514 (Red Hat 8.5.0-22)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import inspect
>>> from bluesky.plan_stubs import mv
>>> inspect.isgeneratorfunction(mv)
False
>>> exit()

(venv) [jwlodek@dell-rhel8 bluesky]$ git checkout custom-wrapper-identify-plan 
Switched to branch 'custom-wrapper-identify-plan'

(venv) [jwlodek@dell-rhel8 bluesky]$ python
Python 3.11.9 (main, Jun 19 2024, 10:02:06) [GCC 8.5.0 20210514 (Red Hat 8.5.0-22)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from bluesky.utils import is_plan
>>> from bluesky.plan_stubs import mv
>>> is_plan(mv)
True
>>> is_plan(print)
False
>>> exit()
```

### DIFF
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -1986,7 +1986,4 @@ def is_plan(bs_plan):
         True if bs_plan arg is a generator, or the __is_plan__ attribute exists and is True.
     """
 
-    if inspect.isgeneratorfunction(bs_plan) or getattr(bs_plan, "_is_plan_", False):
-        return True
-    else:
-        return False
+    return inspect.isgeneratorfunction(bs_plan) or getattr(bs_plan, "_is_plan_", False)
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -1967,7 +1967,7 @@ def plan(bs_plan):
     def wrapper(*args, **kwargs) -> Plan:
         return Plan(bs_plan, *args, **kwargs)
 
-    wrapper.__is_plan__ = True
+    wrapper._is_plan_ = True
 
     return wrapper
 
@@ -1986,7 +1986,7 @@ def is_plan(bs_plan):
         True if bs_plan arg is a generator, or the __is_plan__ attribute exists and is True.
     """
 
-    if inspect.isgeneratorfunction(bs_plan) or getattr(bs_plan, "__is_plan__", False):
+    if inspect.isgeneratorfunction(bs_plan) or getattr(bs_plan, "_is_plan_", False):
         return True
     else:
         return False
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -2,6 +2,7 @@ import abc
 import asyncio
 import collections.abc
 import datetime
+import functools
 import inspect
 import itertools
 import operator
@@ -1949,12 +1950,12 @@ class Plan:
         return self._iter.throw(typ, val, tb)
 
 
-def plan(plan):
+def plan(bs_plan):
     """Decorator that warns user if a `yield from` is not called
 
     Parameters
     ----------
-    plan : Generator
+    bs_plan : Generator
         Generator coroutine usually a Bluesky plan
 
     Returns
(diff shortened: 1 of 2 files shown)


src/bluesky/utils/__init__.py
```diff
@@ -1963,8 +1964,43 @@ def plan(plan):
         Wrapped plans
     """
 
-    @wraps(plan)
+    @wraps_plan(bs_plan, plan)
     def wrapper(*args, **kwargs) -> Plan:
-        return Plan(plan, *args, **kwargs)
+        return Plan(bs_plan, *args, **kwargs)
```
tacaswell, line 1969: If we are going to do this wouldn't be simpler to directly set it?

```suggestion
        return Plan(bs_plan, *args, **kwargs)

    wrapper.__is_plan__ = True
```

src/bluesky/utils/__init__.py
```diff
@@ -1963,8 +1964,43 @@ def plan(plan):
...
+    bs_plan : Function
+        Typically generator coroutine function / Bluesky plan
+
+    Returns
+    -------
+    boolean
+        True if bs_plan arg is a generator, or the __is_plan__ attribute exists and is True.
+    """
+
+    if inspect.isgeneratorfunction(bs_plan) or getattr(bs_plan, "__is_plan__", False):
```
tacaswell, line 2003: Could we also do this as 
```suggestion
    if inspect.isgeneratorfunction(bs_plan) or isintsance(bs_plan, Plan):
```

without having to pick up any extra state?

src/bluesky/utils/__init__.py
```diff
@@ -1963,8 +1964,43 @@ def plan(plan):
...
-    @wraps(plan)
+    @wraps_plan(bs_plan, plan)
     def wrapper(*args, **kwargs) -> Plan:
-        return Plan(plan, *args, **kwargs)
+        return Plan(bs_plan, *args, **kwargs)
 
     return wrapper
+
+
+def mark_as_plan(wrapper, wrapped, decorator, **kwargs):
```
tacaswell, line 1974: if we do go this route, does this need to be public?

src/bluesky/utils/__init__.py
jwlodek: I had tried this originally, but it doesn't appear to work:

```
(venv) [jwlodek@dell-rhel8 bluesky]$ ipython
Python 3.11.9 (main, Jun 19 2024, 10:02:06) [GCC 8.5.0 20210514 (Red Hat 8.5.0-22)]
IPython 8.23.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from bluesky.plan_stubs import mv

In [2]: from bluesky.utils import is_plan

In [3]: is_plan??

Signature: is_plan(bs_plan)
Source:   
def is_plan(bs_plan):
    if inspect.isgeneratorfunction(bs_plan) or isinstance(bs_plan, Plan):
        return True
    else:
        return False
File:      ~/Workspace/bluesky/src/bluesky/utils/__init__.py
Type:      function

In [4]: is_plan(mv)
Out[4]: False
```

src/bluesky/utils/__init__.py
jwlodek: If I do this, it seems that `mv.__wrapped__` is marked with `__is_plan__`, but not `mv` itself:

```
In [11]: is_plan(mv)
Out[11]: False

In [12]: is_plan(mv.__wrapped__)
Out[12]: True
```

jwlodek: Edit: Actually, I had some weird state. I tried this again, and it appears to work, and removes the need for the customized `wrap` and `mark_as_plan` functions. If we decide to go with the marker this is probably the way to go.

src/bluesky/utils/__init__.py
tacaswell: oh, right 🐑 

The thing returned by the decorator is a normal function, the thing returned by the normal function is `Plan` which than implements the co-routine functionality when iterated.

jwlodek: Indeed, if I do `isinstance(mv(), Plan)` it returns true, but then complains that the plan was not iterated, so maybe just keeping track of the state is easier?

tacaswell: Yes, I think tagging it with some private state we then look for in the right implementation.

danielballan:
Generally, double-underscore is supposed to be reserved for use by the Python language. Some frameworks respect that: IPython has methods like `_ipython_key_completions_` and `_repr_html_`. Some frameworks do not: dask has `__dask_tokenize__`. All else being equal, I'd rather respect it and go with `_is_plan_`.

Otherwise, I'm happy with this. Thanks @jwlodek.

src/bluesky/utils/__init__.py
```diff
@@ -1963,8 +1963,30 @@ def plan(plan):
...
+    Returns
+    -------
+    boolean
+        True if bs_plan arg is a generator, or the __is_plan__ attribute exists and is True.
+    """
+
+    if inspect.isgeneratorfunction(bs_plan) or getattr(bs_plan, "_is_plan_", False):
+        return True
+    else:
+        return False
```
tacaswell, line 1992:
```suggestion
    return inspect.isgeneratorfunction(bs_plan) or getattr(bs_plan, "_is_plan_", False)
```

tacaswell (approved):

tacaswell:
@danielballan The failures look tiled related (internal errors on `starlette`), hopefully "just" miss-matched versions?

canismarko:
> @danielballan The failures look tiled related (internal errors on `starlette`), hopefully "just" miss-matched versions?

I ran into this in my beamline's CI as well. starlette <= 0.38.2 works.

jwlodek:
Since test issues appear unrelated, I think I will go ahead and merge?

PR #1784, opened 2024-08-05
add time filter to events

### LINKED ISSUE #1783: Drop events that are out of date
Based on the PR https://github.com/bluesky/bluesky/pull/1505

## Current Behavior

```
if you:

    use monitoring to get an event stream based on camonitor
    the ca events come in at a higher rate than we can emit the event model Events

Then the queue in ophyd gets backed up (because we have no meaningful back pressure, because we grab from the ca stack promptly we are defeating the CA back pressure). Because that queue ensues that they are processed in the order they are received, this means that processing things like "DMOV" also get delayed which causes other issues.

We conceived the monitoring in bluesky for "slow" things, but it gets used for "fast" things. I don't see a way to predict if something is going to be fast or slow apriopi so we can not protectively reject even setting up the bad cases so I think the only option we have is to implement something like this to ensure reliable operation.


```
Originally posted by @tacaswell  https://github.com/bluesky/bluesky/pull/1505#issuecomment-1090355100

This should have (https://github.com/bluesky/bluesky/pull/1505#issuecomment-1030463301)
- [ ] make the 'too old" time be configurable via kwargs in the Msg so that this can be selectively applied (some monitors are fast and likely to be problems, but dropping them is also OK which may be used in tandem with a monitor than is slow and catching everyone is critical)
- [ ] add the ability to thread through a 1 in N logic (either only emitting every Nth or maybe probabilistically keeping emitting 90% of the time) .
- [ ] sort out what happens when both are used (no callback older than drop dead time should go out, but then the question is the 1 in N of all events or of those young enough to be considered)
- [ ] decide if timestamp will be a required kwarg or not
- [ ] max age to be emitted (so once something is too stale, always drop on floor)
- [ ] minimum gap between monitors (if we just emitted something, drop all new updates on the floor for some time)

stan-dot:
Fixes #1783  , based on https://github.com/bluesky/bluesky/pull/1505
## Description

adds an early return logic branch inside the blunders.py/RunBundler/monitor/emit_event
where if any of the readings is too old, the method returns early

### DIFF
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -432,7 +432,7 @@ class RunBundler:
             ), "The callback passed to subscribe() was not called with Dict[str, Reading]"
             now = ttime.time()
             if any(r["timestamp"] - now > timestamp_cutoff_delay for r in readings.values()):  # type: ignore
-                print("WARNING: The timestamps in the readings are too old.")
+                doc_logger.debug("WARNING: The timestamps in the readings are too old.")
                 return
             data, timestamps = _rearrange_into_parallel_dicts(readings)
             doc = compose_event(
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -1,6 +1,5 @@
 import asyncio
 import inspect
-import re
 import time as ttime
 from collections import defaultdict, deque
 from collections.abc import Iterable
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -427,7 +427,7 @@ class RunBundler:
                     f"{readable_obj} has async read() method and the callback "
                     "passed to subscribe() was not called with Dict[str, Reading]"
                 )
-            if readings is not None:
+            if readings is None:
                 raise ValueError("The callback passed to subscribe() was not called with Dict[str, Reading]")
             now = ttime.time()
             if any(r["timestamp"] - now > timestamp_cutoff_delay for r in readings.values()):  # type: ignore
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -431,7 +431,7 @@ class RunBundler:
                 readings is not None
             ), "The callback passed to subscribe() was not called with Dict[str, Reading]"
             now = ttime.time()
-            if any(r.timestamp - now < timestamp_cutoff_delay for r in readings.values()):  # type: ignore
+            if any(r["timestamp"] - now < timestamp_cutoff_delay for r in readings.values()):  # type: ignore
                 print("WARNING: The timestamps in the readings are too old.")
                 return
             data, timestamps = _rearrange_into_parallel_dicts(readings)
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -427,9 +427,9 @@ class RunBundler:
                     f"{readable_obj} has async read() method and the callback "
                     "passed to subscribe() was not called with Dict[str, Reading]"
                 )
-            assert (
-                readings is not None
-            ), "The callback passed to subscribe() was not called with Dict[str, Reading]"
+            if readings is not None:
+                doc_logger.error("The callback passed to subscribe() was not called with Dict[str, Reading]")
+                return
             now = ttime.time()
             if any(r["timestamp"] - now > timestamp_cutoff_delay for r in readings.values()):  # type: ignore
                 doc_logger.debug("WARNING: The timestamps in the readings are too old.")
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -428,8 +428,7 @@ class RunBundler:
                     "passed to subscribe() was not called with Dict[str, Reading]"
                 )
             if readings is not None:
-                doc_logger.error("The callback passed to subscribe() was not called with Dict[str, Reading]")
-                return
+                raise ValueError("The callback passed to subscribe() was not called with Dict[str, Reading]")
             now = ttime.time()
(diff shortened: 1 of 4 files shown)


stan-dot:
there are two plans that get into an error state because of this small change

the first one is `test_monitor` in `tests/test_devices.py`:

```python

@pytest.mark.parametrize("ophyd", ["v1", "v2"])
@requires_ophyd
def test_monitor(RE, ophyd):
    docs = []

    def collect(name, doc):
        docs.append(doc)

    if ophyd == "v1":
        sig = A("", name="a").s1
    else:
        sig = SigNew(name="a_s1")

    def plan():
        yield Msg("open_run")
        yield Msg("monitor", sig)
        sig._run_subs(sub_type="value")
        yield Msg("close_run")

    RE(plan(), collect)
    assert len(docs) == 4
    assert docs[2] == {
        "data": {"a_s1": 0},
        "descriptor": ANY,
        "seq_num": 1,
        "time": pytest.approx(time(), rel=0.1),
        "timestamps": {"a_s1": pytest.approx(time(), rel=0.1)},
        "filled": ANY,
        "uid": ANY,
    }

```

the other one is quite hard to parse and is such:

```python



@requires_ophyd
def test_ramp(RE):
    from ophyd import StatusBase
    from ophyd.positioner import SoftPositioner
    from ophyd.sim import SynGauss

    import bluesky.utils as bsu

    RE.msg_hook = bsu.ts_msg_hook

    tt = SoftPositioner(name="mot")
    tt.set(0)
    dd = SynGauss("det", tt, "mot", 0, 3)

    st = StatusBase()

    def kickoff():
        yield Msg("null")
        for j, v in enumerate(np.linspace(-5, 5, 10)):
            RE.loop.call_later(0.1 * j, lambda v=v: tt.set(v))
        RE.loop.call_later(1.2, st._finished)
        print("YOLO")
        return st

    first = True

    def inner_plan():
        nonlocal first
        if first:
            yield from declare_stream(dd, name="primary")
            first = False

        yield from trigger_and_read([dd])

    g = ramp_plan(kickoff(), tt, inner_plan, period=0.08)
    db = DocCollector()
    RE.subscribe(db.insert)
    rs = RE(g)
    if RE.call_returns_result:
        uid = rs.run_start_uids[0]
    else:
        uid = rs[0]
    assert db.start[0]["uid"] == uid
    assert len(db.descriptor[uid]) == 2
    descs = {d["name"]: d for d in db.descriptor[uid]}

    assert set(descs) == set(["primary", "mot_monitor"])  # noqa: C405

    primary_events = db.event[descs["primary"]["uid"]]
    assert len(primary_events) > 11

    monitor_events = db.event[descs["mot_monitor"]["uid"]]
    # the 10 from the updates, 1 from 'run at subscription time'
    assert len(monitor_events) == 11
```

parsing the test logic now

stan-dot:
maybe it's a detection of a bug, or the new behavior is not correct

src/bluesky/bundlers.py
```diff
@@ -426,6 +426,11 @@ def emit_event(readings: Optional[Dict[str, Reading]] = None, *args, **kwargs):
                     f"{readable_obj} has async read() method and the callback "
                     "passed to subscribe() was not called with Dict[str, Reading]"
                 )
+            now = ttime.time()
+            acceptable_delay = 0.1
```
coretl, line 430: This needs to be configurable.

Passed in as an arg to the `Msg` when we create the monitor

stan-dot: very sensible, will add soon

stan-dot: added

stan-dot:
I just rebuilt the container and all the tests worked

stan-dot:
compared to main, `test_monitor` produces 3 docs instead of 4, start descriptor stop with the `event` missing

src/bluesky/bundlers.py
```diff
@@ -427,6 +427,13 @@ def emit_event(readings: Optional[dict[str, Reading]] = None, *args, **kwargs):
                     f"{readable_obj} has async read() method and the callback "
                     "passed to subscribe() was not called with Dict[str, Reading]"
                 )
+            assert (
+                readings is not None
+            ), "The callback passed to subscribe() was not called with Dict[str, Reading]"
+            now = ttime.time()
+            if any(r["timestamp"] - now > timestamp_cutoff_delay for r in readings.values()):  # type: ignore
+                print("WARNING: The timestamps in the readings are too old.")
```
tacaswell, line 435: Is there a logger we can plug into instead here (I have lost track of the current state of art on that in bluesky 😞 )?

src/bluesky/bundlers.py
```diff
@@ -427,6 +427,13 @@ def emit_event(readings: Optional[dict[str, Reading]] = None, *args, **kwargs):
                     f"{readable_obj} has async read() method and the callback "
                     "passed to subscribe() was not called with Dict[str, Reading]"
                 )
+            assert (
```
tacaswell, line 430: why are we adding asserts in the main code base?  I see there is one above here as well.

tacaswell:
It is worth generating a pathological `Signal` subclass that lies about its timestamps to test this without sleeps.

src/bluesky/bundlers.py
stan-dot: I've seen it used in the repo in other places too

stan-dot: okay will do

tacaswell:
I can not sort out how to make this into a draft...

stan-dot:
image
here

stan-dot:
closing as out of date

PR #1410, opened 2020-10-06
WIP: draft of coro + message to wait on status objects the user has

tacaswell:
## Description

This is responsive to a question from @jklynch about how to write 

## Motivation and Context

```python
# use this function with SubscriptionStatus
# high to low means acquisition is done
def high_to_low(value, old_value, **kwargs):
    if old_value == 1 and value == 0:
        return True

    return False

def a_plan(pe_detector):
    # close the shutter
    ...

    # do the dark frame acquisition
    yield from bps.mv(pe_detector.cam.pe_acquire_offset, "Acquire")
    pe_acquire_offset_status = SubscriptionStatus(
        pe_detector.cam.pe_acquire_offset,
        high_to_low
    )
    yield from bps.wait_for(futures=[pe_acquire_offset_status])

    # open the shutter
    ...

    # acquire some data
    yield from bps.mv(pe_detector.cam.acquire, "Acquire")
    pe_acquire_status = SubscriptionStatus(
        pe_detector.cam.acquire,
        high_to_low
    )
    yield from bps.wait_for(futures=[pe_acquire_status])

    # hurray we are done!

```

but the problem is that `wait_for` expects asyncio futures not StatusBase objects.

## How Has This Been Tested?

not at all yet, more as a long-form comment to @jklynch

PR #1791, opened 2024-08-09
ENH - RunEngine._wait will now accept a list of status objects as a group in-lieu of a group name. See #1410

burkeds:
RunEngine._wait will now accept a list of status objects as a group in-lieu of a group name. See #1410

## Description
RunEngine._wait can be passed a list of status objects. It will then create a new group label called '__status_group__'
and add each status to this group. The method then proceeds as normal.

## Motivation and Context
Being able to wait on a status object is an intuitive use of the 'wait' plan stub. 

## How Has This Been Tested?
In a Jupyter notebook I passed multiple status objects in-lieu of group and the operation seemed to process correctly.
I wrote a test but for some reason I can't seem to run tests properly on my machine as many other tests seems to throw
thread exceptions.

### DIFF
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -2323,12 +2323,18 @@ class RunEngine:
         true when all triggered objects are done. When the keyword argument
         `move_on=<MOVE_ON>` is true, this method can return before all objects are done
         after a flush period given by the `timeout=<TIMEOUT>` keyword argument.
+        
+        Optionally, the `group` keyword argument can be a list of status objects passed
+        in the same way. A temporary group is created for them and the function is 
+        processed as normal.
 
-        Expected message object is:
+        Expected message objects are:
 
             Msg('wait', group=<GROUP>, move_on=<MOVE_ON>)
+            Msg('wait', group=[<STATUS_OBJECTS>], move_on=<MOVE_ON>)
 
-        where ``<GROUP>`` is any hashable key and ``<MOVE_ON>`` is a boolean.
+        where ``<GROUP>`` is any hashable key, ``<MOVE_ON>`` is a boolean, and
+        ``[<STATUS_OBJECTS>]`` is a list of status objects.
         """
         _set_span_msg_attributes(trace.get_current_span(), msg)
         done = False  # boolean that tracks whether waiting is complete
@@ -2338,6 +2344,13 @@ class RunEngine:
         else:
             group = msg.kwargs["group"]
             move_on = msg.kwargs.get("move_on", False)
+            
+        if isinstance(group, list):
+            status_objs = group
+            group = "__status_group__"
+            for status in status_objs:
+                self._add_status_to_group(obj=None, status_object=status, group=group, action="wait")
+                
         if group:
             trace.get_current_span().set_attribute("group", group)
         else:
--- a/src/bluesky/tests/test_run_engine.py
+++ b/src/bluesky/tests/test_run_engine.py
@@ -1957,6 +1957,33 @@ def test_wait_with_timeout(set_finished, RE):
             RE(plan())
 
 
+@pytest.mark.parametrize("set_finished", [True, False])
+def test_wait_with_statuses(set_finished, RE):
+    from ophyd import StatusBase
+    from ophyd.device import Device
+
+    class MockDevice(Device):
+        def stage(self):
+            status = StatusBase()
+            if set_finished:
+                status.set_finished()
+            return status
+
+    mock_device_a = MockDevice(name="mock_device_a")
+    mock_device_b = MockDevice(name="mock_device_b")
+
+    def plan():
+        status_a = yield Msg("stage", mock_device_a)
+        status_b = yield Msg("stage", mock_device_b)
+        yield from wait(group=[status_a, status_b], timeout=0.1)
+
+    if set_finished:
+        RE(plan())
+    else:
+        with pytest.raises(TimeoutError):
+            RE(plan())
+
+
 async def passing_coroutine():
     await asyncio.sleep(0.001)
     return 1080


danielballan:
Python 3.10 failed on an unrelated test:

```
FAILED src/bluesky/tests/test_suspenders.py::test_unresumable_suspend_fail[False] - assert 0.1 < (1723211575.3399906 - 1723211575.2414505)
```

Re-running.

src/bluesky/run_engine.py
```diff
@@ -2338,6 +2343,13 @@ async def _wait(self, msg):
         else:
             group = msg.kwargs["group"]
             move_on = msg.kwargs.get("move_on", False)
+
+        if isinstance(group, list):
```
danielballan, line 2347: The codebase is generally tolerant of different types of collections. For example, if someone accumulated status objects in a `set` or a `tuple` I think there is no reason to reject that. Perhaps this could check for `isinstance(group, collections.abc.Iterable) and not isinstance(group, str)`?

burkeds: @danielballan That makes sense. I changed this to -- if isinstance(group, collections.abc.Iterable): and pushed.

src/bluesky/run_engine.py
```diff
@@ -2338,6 +2344,13 @@ async def _wait(self, msg):
         else:
             group = msg.kwargs["group"]
             move_on = msg.kwargs.get("move_on", False)
+
+        if isinstance(group, collections.abc.Iterable) and not isinstance(group, str):
+            status_objs = group
+            group = "__status_group__"
```
tacaswell, line 2350: because we use the same key every time, if we "move on" and then wait on a different set, will they effectively get merged?

We should probably add the `hash(tuple(group))` to the name or if status objects are not hashable (or we worry there might be non-hashable subclasses in the future...I don't remember off the top of my head...), `hash(tuple(sorted(id(s) for s in group)))`. 

Although using `id()` is a Bad Idea™️ , I think this is a reasonable use case as object identity is the identity we care about on `Status` objects.  

The sorting is so that `group=[sA, sB]` and `group=[sB, sA]` are identified as the same, probably want to do the same thing with teh hash as well....

burkeds: I see, I understand. I made the change to use id() to make a unique hash for the group as suggested.

tacaswell (approved):
It is probably a good idea to add an example to the docs as well, but I'm not going to block the PR over that.

docs/tutorial.rst
```diff
@@ -1376,9 +1376,36 @@ motors in motion at once, including multiple fast motors and one slow motor. We
...
+For Bluesky version v1.13.0 and higher, this approach is recommended:
+
 .. code-block:: python
 
     def staggered_wait(fast_motors, slow_motor):
+        # Start all the motors, fast and slow, moving at once.
+        # Put all the fast_motors in one group...
+        statuses = []
+        for motor in fast_motors:
+            status = yield from bps.abs_set(motor, 5)
```
tacaswell, line 1388: A maybe issue that this example brought to mind: we are also adding these statuses to the default group!

If we have plans that _only_ use this aproach, we will ever clear out those cached statuses or will we have an unbounded list that will (eventually) oom us?

burkeds: @tacaswell I don't think I quite understand what the issue is. Are the statuses not added to their own group labeled with a hash?

tacaswell:
```
 status = yield from bps.abs_set(motor, 5)
```

This will add to the `None` group (as we can not predict how it will be used in the future!)

The 

```
yield from bsp.wait(status)
```

will add it to the (hopefully short lived) group.

The group created via the `wait` will be dropped, but I don't think it will get removed from the `None` group.

burkeds: Why do we maintain this `None` group? If group=None is passed couldn't we simply not associate the status with any group?

tacaswell: It is an early design choice to let

```python
yield from bps.abs_set(mtr1, 1)
yield from bps.bas_set(mtr2, 2)
yield from bps.wait()
```

"just work" with an implicit group.  Dropping that would likely break a bunch of stuff out in the wild so we can not just unilaterally change that.

Maybe we want to add a special group for "do not track me!",  but that would defeat a bunch of the benefits of adding the ability to pass status objects (the reduced boiler plate).  I would also have concerns about that because in addition to being able to wait on the status object the RE does a bunch of book keeping to be able to raise-on-motion-failure and having a loose status object is a bit of an anathema to the design (the first job of the RE is to track when things you asked hardware to do are done! as I said else where if there is something you will never care when it is actually done, then it should return an already completed Status object as that is the right way to nope out of the RE doing its job).

This also makes we wonder if when combined with move on and calling `wait` with different sets of status objects is going to result in status object being in multiple groups?  Something like:

```
st_list = [...]
while len(st_list):
   st_list = [st for st in st_list if not st.done]
   yield from bps.wait(st_list, timeout=.5, move_on=True)
```

so even if we did drop the implicit `None` group, we could still drive ourselves to an unbounded cache (I think).

burkeds: Would it then be possible to check if the statuses exist in the `None` group, pop them, and put them in a dedicated group?

tacaswell: I think possible, but would we want to pop them out forever or put them back if unexpired?

```python
st_list = [...]
yield from bps.wait(st_list, timeout=.5, move_on=True)
yield from bps.wait()  # <- this should wait on anything not yet done!
```

burkeds: I am able to pop the status objects from `self._status_objs[None]' and put them in a new set under a unique group name. The issue is I can't figure out how I can retrieve their futures from `self._groups[None]` to put into a new group of futures.

```python
        if isinstance(group, collections.abc.Iterable) and not isinstance(group, str):
            status_objs = group
            group = str(hash(tuple(sorted(id(s) for s in group))))
            for status in status_objs:
                # self._add_status_to_group(obj=None, status_object=status, group=group, action="wait")
                if status in self._status_objs[None]:
                    self._status_objs[None].remove(status)
                    self._status_objs[group].add(status)
```

tacaswell:
Had a quick chat with @coretl today, this is a quick summary:

 - worry about unbounded accidental cache is probably real (mostly in implicit `None`, but could happen to any grou)
 - the "by group" scheme is going to be hard to deprecate, but even if it were easy the pattern does protect against a high-risk bug (e.g. add a set or trigger to plan but forget to add it to add its status object to the list of other status objects)

It should be possible to add a callback to the status objects that will remove themselves from the lists we hold in the RE that will prevent the leak (we should only remove them on success but leave them there on failure so in the happy-path waiting on the group is faster and in the sad-path you still get the exception bubbling up).

----

Other thought: I don't recall off the top of my head what we do with the lists of status objects on exiting `_run`, if we do not currently do so we should warn if there are un-waited-on status objects hanging out in groups (and maybe clear them?).

tacaswell:
Notes from the developer call:
 - should refactor to go from holding to `dict[str, list[Status]]` and `dict[str, list[Future]]` to a single dictionary that is `dict[str, dict[Status, Future]]` to facilitate the book keeping
 - when handling a list of status we need to pull them out of the groups they are in (`None` or a named one).  On returning from `_wait`
   - if it is done, drop it on the floor (it is done!)
   - if it is not done, put it back where it came from
   - if it errored, raise but do not put it back in the group
 - we need to clear the message cache via `await self._reset_checkpoint_state_coro()` as messages that include (Python) references to `Status` objects can not be safely rewound automatically (you probably need a new status object on resume but rewind can only push the exact messages it saw through as-is).  This needs to be documented as a limitation and is the reason we should not move the planstubs to use the new feature.

Issue #1541, opened 2022-08-22
puzzling exception message: "RuntimeWarning: invalid value encountered in double_scalars"


prjemian:
The message is generated from `callbacks.fitting.center_of_mass()` during a  run using the BestEffortCallback when no signal is on the detector.  User is puzzled what is the real problem since the reported error makes no sense.

The run that triggered the question:

```py
RE(rel_list_scan([scaler1], kphi, [-1, -0.5, 0, 0.5, 1]))
```

and then a warning message at the end:

```py
/home/beams/EPIX33ID/.conda/envs/bluesky_2022_1/lib/python3.9/site-packages/bluesky/callbacks/fitting.py:167: RuntimeWarning: invalid value encountered in double_scalars  np.sum(input * grids[dir].astype(float), labels, index) / normalizer
Out[28]: ('405ff799-d610-4f4b-8507-1c9382a2e684',)
```

## Expected Behavior

Tracing through this, a more appropriate message would be something such as: `ZeroDivisionError: sum(input) is zero` since the problem is a divide-by-zero error.

## Current Behavior
BEC's `center_of_mass()` mis-reports a divde-by-zero exception.

## Possible Solution

After https://github.com/bluesky/bluesky/blob/f51350b5dc839cd442d99a28a4cf9e613bc2b01f/bluesky/callbacks/fitting.py#L163

add this code:

```py
    if normalizer == 0:
        raise ZeroDivisionError("sum(input) is zero")
```

Alternatively, in this case, could avoid the problem entirely with a default normalization:

```py
    if normalizer == 0:
        normalizer  = 1  # special case
```

## Steps to Reproduce (for bugs)

Configure RE with BEC, then run a scan with a scaler and no beam:

```py
RE(rel_list_scan([scaler1], kphi, [-1, -0.5, 0, 0.5, 1]))
```

## Context

The message is misleading which causes delay in understanding the root problem.

prjemian:
User replied (by private email): 

> Yeah, if it is 'divided by zero' kind of thing, I probably wouldn't look at it twice. I at first thought I setup things wrong. But it is only a warning...

tacaswell:
This warning is coming from numpy

```python
In [26]: np.sum([1, 2]) / 0
<ipython-input-26-76ad96894c04>:1: RuntimeWarning: divide by zero encountered in long_scalars
  np.sum([1, 2]) / 0
Out[26]: inf
```

[...]

PR #1792, opened 2024-08-12
handle ZeroDivisionError in fitting.py/center_of_mass()

prjemian:
- #1541

## Description

- Avoids a `ZeroDivisionError` when computing a center of mass.
- Adds first unit tests for this function.

## Motivation and Context

When a signal is all zero, it provokes a `ZeroDivisionError` in the `center_of_mass()` function.  We know why this happens and want to ensure an exception is not raised in such cases.

## How Has This Been Tested?

Unit tests, including example data that provokes the exception.

### DIFF
--- a/src/bluesky/callbacks/fitting.py
+++ b/src/bluesky/callbacks/fitting.py
@@ -166,6 +166,8 @@ def center_of_mass(input, labels=None, index=None):
     [(0.33333333333333331, 1.3333333333333333), (3.5, 2.5)]
     """
     normalizer = np.sum(input, labels, index)
+    if normalizer == 0:
+        normalizer  = 1  # special case to avoid ZeroDivisionError
     grids = np.ogrid[[slice(0, i) for i in input.shape]]
 
     results = [np.sum(input * grids[dir].astype(float), labels, index) / normalizer for dir in range(input.ndim)]
--- a/src/bluesky/tests/test_center_of_mass.py
+++ b/src/bluesky/tests/test_center_of_mass.py
@@ -1,0 +1,18 @@
+import numpy as np
+from scipy import ndimage
+
+from bluesky.callbacks.fitting import center_of_mass
+
+
+def test_center_of_mass():
+    assert True
+
+    arr = np.array(([0, 0, 0, 0], [0, 1, 1, 0], [0, 1, 1, 0], [0, 1, 1, 0]))
+    result = ndimage.measurements.center_of_mass(arr)
+    assert result == (2.0, 1.5), f"{result=}"
+
+    # Calculation of multiple objects in an image
+    b_arr = np.array(([0, 1, 1, 0], [0, 1, 0, 0], [0, 0, 0, 0], [0, 0, 1, 1], [0, 0, 1, 1]))
+    lbl = ndimage.label(b_arr)[0]
+    result = ndimage.measurements.center_of_mass(b_arr, lbl, [1, 2])
+    assert result == [(0.33333333333333331, 1.3333333333333333), (3.5, 2.5)]
--- a/src/bluesky/tests/test_center_of_mass.py
+++ b/src/bluesky/tests/test_center_of_mass.py
@@ -1,18 +1,30 @@
 import numpy as np
-from scipy import ndimage
+import pytest
 
 from bluesky.callbacks.fitting import center_of_mass
 
 
-def test_center_of_mass():
-    assert True
-
-    arr = np.array(([0, 0, 0, 0], [0, 1, 1, 0], [0, 1, 1, 0], [0, 1, 1, 0]))
-    result = ndimage.measurements.center_of_mass(arr)
-    assert result == (2.0, 1.5), f"{result=}"
-
-    # Calculation of multiple objects in an image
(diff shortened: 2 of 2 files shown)


prjemian:
This PR needs to handle this line: https://github.com/bluesky/bluesky/blob/c5f44e87f8e53cf573a312aec301b87a7d695862/src/bluesky/callbacks/fitting.py#L299

Since `center_of_mass()` can now return `None` (instead of raising an exception), this code must not call `np.interp()` when `None` is returned.

prjemian:
Tested this locally with a constant zero signal.  No exception raised now.  `bec.peaks` structure shows `"com" = None`, as expected for a flat signal.

```py
In [1]: from ophyd import Signal

In [2]: flat = Signal(value=0, name="flat")

In [3]: RE(bp.rel_scan([flat], m1, -.1, .1, 5))


Transient Scan ID: 2     Time: 2024-08-13 16:27:58
Persistent Unique Scan ID: 'e61c0c73-b4ff-44d8-8538-747cccee4beb'
New stream: 'label_start_motor'
New stream: 'primary'
|   seq_num |       time |         m1 |       flat |
|         1 | 16:27:59.0 |    -0.1000 |          0 |
|         2 | 16:27:59.5 |    -0.0500 |          0 |
|         3 | 16:27:59.7 |     0.0000 |          0 |
|         4 | 16:28:00.0 |     0.0500 |          0 |
|         5 | 16:28:00.3 |     0.1000 |          0 |
generator rel_scan ['e61c0c73'] (scan num: 2)
Out[3]: ('e61c0c73-b4ff-44d8-8538-747cccee4beb',)

In [4]: peaks
Out[4]: 
{
'com':
    {'flat': None}
,
'cen':
    {'flat': None}
,
'max':
    {'flat': (-0.1,
              0)}
,
'min':
    {'flat': (-0.1,
              0)}
,
'fwhm':
    {'flat': None}
,
}
```

tacaswell:
It looks like when we vendored https://github.com/scipy/scipy/blob/a9fb36bc44bad4bbd2c1a41cb43c6f10925b38ae/scipy/ndimage/measurements.py#L1232-L1283 the bare `sum` in the function body is actually https://github.com/scipy/scipy/blob/a9fb36bc44bad4bbd2c1a41cb43c6f10925b38ae/scipy/ndimage/measurements.py#L541-L582 from this module, not `np.sum`.  I don't think passing `labels` or `index` to `center_of_mass` _ever_ worked.

src/bluesky/tests/test_center_of_mass.py
```diff
@@ -0,0 +1,30 @@
...
+        #     [(1.0 / 3, 4.0 / 3), (3.5, 2.5)],
+        # ],
+        # 1-D
+        [[0, 1, 1, 1, 0], None, None, (2.0,)],
+    ],
+)
+def test_center_of_mass(arr, labels, index, expected):
+    if labels is not None:
+        labels = np.array(labels)
+    result = center_of_mass(np.array(arr), labels=labels, index=index)
```
tacaswell, line 29:
```suggestion
from bluesky.callbacks.fitting import center_of_mass


@pytest.mark.parametrize(
    "arr, expected",
    [
        # example from the source code
        [[[0, 0, 0, 0], [0, 1, 1, 0], [0, 1, 1, 0], [0, 1, 1, 0]], (2.0, 1.5)],
        # signal is all zero, does not raise ZeroDivisionError or other
        [[[0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]],  None],
        [[0, 1, 1, 1, 0], (2.0,)],
    ],
)
def test_center_of_mass(arr, labels, index, expected):
    result = center_of_mass(np.array(arr))
```

The other parameters to `center_of_mass` do not work in the vendored version, don't try to test them.

src/bluesky/callbacks/fitting.py
```diff
@@ -166,6 +170,8 @@ def center_of_mass(input, labels=None, index=None):
     [(0.33333333333333331, 1.3333333333333333), (3.5, 2.5)]
     """
     normalizer = np.sum(input, labels, index)
+    if normalizer == 0:
+        return None
```
tacaswell, line 174: can we do `np.nan` instead?  This should probably also be a tuple of right size for the input?

prjemian:
np nan sounds right

tacaswell:
I can also see a case for special casing the center of mass of all 0 to be the middle.  

Putting on a "practical" hat with the logic that the `COM(data)` and `COM(data + 5)` should be the same.  If data is a non-zero constant, then the COM is the center of the data.  However, if we special case a zero-constant as `None`, then we break the assumption that applying a constant offset to the data does not change the COM which seems not good.  I think the compromise of giving a technically wrong COM in the all 0 case is worth not putting in surprise "the type change from tuple to `None` and there is no COM" from shifting the data.

prjemian:
> I can also see a case for special casing the center of mass of all 0 to be the middle.

Untested.  Yet.

src/bluesky/callbacks/fitting.py
prjemian: Per later discussion, using x.middle instead.

PR #1796, opened 2024-08-14
Update to allow for using dtype_numpy key alongside old dtype_str

jwlodek:
## Description
Allow for using dtype_numpy keyword as specified in ophyd async

## Motivation and Context
Decision was made to use `dtype_numpy` for the machine data type of resources in `ophyd_async` this change allows for using it alongside the older `dtype_str`

### DIFF
--- a/src/bluesky/consolidators.py
+++ b/src/bluesky/consolidators.py
@@ -89,10 +89,22 @@ class ConsolidatorBase:
         data_desc = descriptor["data_keys"][self.data_key]
         self.datum_shape = tuple(data_desc["shape"])
         self.datum_shape = self.datum_shape if self.datum_shape != (1,) else ()
-        self.dtype = data_desc["dtype"]
-        self.dtype = DTYPE_LOOKUP[self.dtype] if self.dtype in DTYPE_LOOKUP.keys() else self.dtype
-        self.dtype = np.dtype(data_desc.get("dtype_str", self.dtype))
-        self.dtype = np.dtype(data_desc.get("dtype_numpy", self.dtype))
+        # Get data type. From highest precedent to lowest:
+        # 1. Try 'dtype_numpy', optional in the document schema.
+        # 2. Try 'dtype_str', an old convention predataing 'dtype_numpy', not in the schema.
+        # 3. Get 'dtype', required by the schema, which is a fuzzy JSON spec like 'number'
+        #    and make a best effort to convert it to a numpy spec like '<u8'.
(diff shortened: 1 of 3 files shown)


tacaswell:
We have a number of ohpyd devices on the floor that are using `'dtype_str'`.  To ease converting them, I think we should apply some normalization in `describe` and convert `'dtype_str'` -> `'numpy_dtype'` silently for now and then loudly in the future.

This means we will stop putting funny data in the databases but do not have to do a flag-day on getting all the devices on the floor fixed and we do not have any "can not roll back" changes made for a while.

jwlodek:
In the case of "announcing it loudly", what is the best mechanism for this? Raising a warning of some kind? Should that be done somewhere here in the bluesky codebase, or in ophyd?

tacaswell:
I think a release where we silently fix it and then `warnings.warn` for a while ("loud") so it would show up in the terminal.

I don't think we ever want to reject `dtype_str` so long term I can see a case of fixing-and-warning forever or warning for a bit and then just letting it pass (once we are sure `dtype_str` has fallen out of memory ;) ).

src/bluesky/consolidators.py
```diff
@@ -85,13 +85,14 @@ def __init__(self, stream_resource: StreamResource, descriptor: EventDescriptor)
...
 
-        # Find data shape and machine dtype; dtype_str takes precedence if specified
+        # Find data shape and machine dtype; dtype_numpy, dtype_str take precedence if specified
         data_desc = descriptor["data_keys"][self.data_key]
         self.datum_shape = tuple(data_desc["shape"])
         self.datum_shape = self.datum_shape if self.datum_shape != (1,) else ()
         self.dtype = data_desc["dtype"]
         self.dtype = DTYPE_LOOKUP[self.dtype] if self.dtype in DTYPE_LOOKUP.keys() else self.dtype
         self.dtype = np.dtype(data_desc.get("dtype_str", self.dtype))
+        self.dtype = np.dtype(data_desc.get("dtype_numpy", self.dtype))
```
genematx, line 95: I had this in my (yet-unmerged refactoring branch of Consolidators), in case you want to use a one-liner:
`self.dtype = np.dtype(data_desc.get("dtype_numpy") or data_desc.get("dtype_str", self.dtype))`

genematx (approved):

danielballan (approved):

danielballan (approved):

genematx (approved):

PR #1802, opened 2024-09-09
Account for chunk size being a tuple instead of just an int

jwlodek:
In https://github.com/bluesky/ophyd-async/pull/544 it was decided to have `chunk_size` be a tuple that describes the actual HDF chunking configuration, rather than just the first dimension. In addition, the default state was set to an empty tuple. As such, this PR changes the code in the consolidator to account for this

## Description

* Default chunk size used if not in stream resource is now `()`, was `None`
* Use first dimension in the chunk size tuple as the equivalent of the old integer chunk size
* Update tests to reflect change.

## Motivation and Context

See https://github.com/bluesky/ophyd-async/pull/544

## How Has This Been Tested?

Running using the test setup in the lab, with a PandA and a 2D areaDetector. Without this change, we get an error on the division of `num_rows` with `chunk_size` (can't divide an int by a tuple...)


genematx:
Thanks, @jwlodek, for putting this together! I like the changes, but I am just wondering if they are, strictly speaking, necessary... I understand the benefits of having the type of `chunk_size` being the same (tuple) throughout for consistency. At the same time, we restrict the api format of bluesky consolidators, which are downstream relative to ophyd. What if instead of ophyd, we have something else that produces `integers` or some other file format that doesn't have chunks at all; i think it would be unreasonable to expect the user to reformat those just to pass the data to bluesky.

Also, somehow it feels a bit off to me to initialize default kwargs in a signature with empty containers instead of `None` (I admit, I must have missed the part when we decided to change the defaults too). I know, it's okay with tuples and can actually make the code shorter, but it also sets a bad example for someone to do the same with mutable types.

Anyway, I'm happy if we merge, but I'd also be very keen to hear your thoughts (and @danielballan's and @mrakitin's too) on the above and have a short discussion about those high-level code design questions.

jwlodek:
Due to the decision to go with tuples in ophyd async, the current consolidator does not work, since it tries to do a division with the chunk size parameter. So for the lab/hex setup to work with the chunk size as it is provided by ophyd async, we will at least need to be able to ingest a tuple. I think we can probably also handle the case where it is an integer or even None as well - as long as we can handle ophyd-async's default which will be an empty tuple the same way.

Will you be on the call on Thursday? Perhaps we can discuss it there w/ the diamond folks.

genematx:
I haven't realized it was an error in Consolidators (not ophyd). I'll try to fix it later, but definitely let's discuss it in person. Unfortunately, I have DUTC this Thursday, and most likely will miss the meeting with Diamond.

jwlodek:
One thing to note - we should probably have some test cases for one dimensional datasets as well? PandA will be 1 D for example. Maybe also worth adding tests for 3D (i.e. color images).

genematx:
> One thing to note - we should probably have some test cases for one dimensional datasets as well? PandA will be 1 D for example. Maybe also worth adding tests for 3D (i.e. color images).

Added! For arrays with `shape = [1]` and also "numbers" with `shape = []` (although I suspect the former is the most relevant case).

genematx (approved):

PR #1803, opened 2024-09-26
made `Reading` generic on the type of `Reading.value`

### LINKED ISSUE #1800: Make `Reading` Generic
It would be good to know what the type of value is for `Reading`. We can do this by changing:
https://github.com/bluesky/bluesky/blob/f25df0b24b30cac20098b487c54d1113695a915f/src/bluesky/protocols.py#L53
to
```python
T = TypeVar("T")

class Reading(Generic[T], ReadingOptional):
```
(you might need to swap `Generic[T]` and `ReadingOptional` if there are errors about MRO)

This means we can pass around a `Reading` in ophyd-async and use type hints to extract the typed value

evalott100:
Closes #1800

I believe this means `Reading` will be typed as `Reading[Any]` but will double check.

### EDIT

Ah pyright sees an unannotated `Reading` as `Reading[Unknown]` which is actually very nice. Whatever `Reading`'s are allowed.

### DIFF
--- a/src/bluesky/consolidators.py
+++ b/src/bluesky/consolidators.py
@@ -95,14 +95,12 @@ class ConsolidatorBase:
         # 3. Get 'dtype', required by the schema, which is a fuzzy JSON spec like 'number'
         #    and make a best effort to convert it to a numpy spec like '<u8'.
         # 4. If unable to do any of the above, pass through whatever string is in 'dtype'.
-        self.dtype = (
-            np.dtype(
-                data_desc.get("dtype_numpy")  # standard location
-                or data_desc.get(
-                    "dtype_str",  # legacy location
-                    # try to guess numpy dtype from JSON type
-                    DTYPE_LOOKUP.get(data_desc["dtype"], data_desc["dtype"])
-                )
+        self.dtype = np.dtype(
+            data_desc.get("dtype_numpy")  # standard location
+            or data_desc.get(
+                "dtype_str",  # legacy location
+                # try to guess numpy dtype from JSON type
+                DTYPE_LOOKUP.get(data_desc["dtype"], data_desc["dtype"]),
             )
         )
         self.chunk_size = self._sres_parameters.get("chunk_size", None)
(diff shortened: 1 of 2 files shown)


evalott100:
# TESTING
Works well!

``` python
from bluesky.protocols import Reading

int_reading = Reading(value=1, timestamp=1.)
str_reading = Reading(value="bar", timestamp=1.)
float_reading = Reading(value=1., timestamp=1.)


def foo_int_reading(int_reading: Reading[int]) -> None:
    print("int_reading", int_reading["value"])

def foo_str_reading(str_reading: Reading[str]) -> None:
    print("str_reading", str_reading["value"])

def foo_float_reading(float_reading: Reading[float]) -> None:
    print("float_reading", float_reading["value"])

def foo_any_reading(reading: Reading) -> None:
    print("any_reading", reading["value"])

# These don't cause pyright errors
foo_int_reading(int_reading)
foo_str_reading(str_reading)
foo_float_reading(float_reading)

# These do
foo_str_reading(int_reading)
foo_int_reading(str_reading)

# These don't
foo_any_reading(int_reading)
foo_any_reading(str_reading)
foo_any_reading(float_reading)

# This does
foo_any_reading("whatabizareinput")
```

Errors are generated correctly:

```
ophyd-async/test_reading_type.py
  ophyd-async/test_reading_type.py:26:17 - error: Argument of type "Reading[int]" cannot be assigned to parameter "str_reading" of type "Reading[str]" in function "foo_str_reading"
    "Reading[int]" is not assignable to "Reading[str]"
      Type parameter "T@Reading" is invariant, but "int" is not the same as "str" (reportArgumentType)
  test_reading_type.py:27:17 - error: Argument of type "Reading[str]" cannot be assigned to parameter "int_reading" of type "Reading[int]" in function "foo_int_reading"
    "Reading[str]" is not assignable to "Reading[int]"
      Type parameter "T@Reading" is invariant, but "str" is not the same as "int" (reportArgumentType)
  test_reading_type.py:35:17 - error: Argument of type "Literal['whatabizareinput']" cannot be assigned to parameter "reading" of type "Reading[Unknown]" in function "foo_any_reading"
    "Literal['whatabizareinput']" is not assignable to "Reading[Unknown]" (reportArgumentType)
3 errors, 0 warnings, 0 informations 

```

coretl (approved):

PR #1804, opened 2024-09-28
Windows: allow use of `bps.input_plan()`

Tom-Willemsen:
## Description

Implement `AsyncInput` using `run_in_executor`, as `loop.add_reader()` is not available for `stdin` on either of the standard windows event loop implementations.

`run_in_executor` may add a small amount of overhead (it uses a thread pool under the hood), but since we're then immediately asking for user input I do not think that is significant (if other facilities feel it _is_ a problem, we can split the implementation into windows-specific and linux-specific versions, but I don't feel it's worth the complexity cost personally...)

## Motivation and Context

At ISIS we're using bluesky on windows. On windows, `bps.input_plan()` doesn't currently work with either the proactor or selector event loop implementations. See https://docs.python.org/3/library/asyncio-platforms.html#asyncio-platform-support for details about which asyncio methods are different/unavailable on windows.

```
Python 3.11.6 (tags/v3.11.6:8b6ee5b, Oct  2 2023, 14:57:12) [MSC v.1935 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from bluesky.run_engine import RunEngine
>>> import bluesky.plan_stubs as bps
>>> RE = RunEngine()
>>> RE(bps.input_plan("foo?"))
Run aborted
  File "C:\Instrument\dev\bluesky\src\bluesky\run_engine.py", line 1580, in _run
    msg = self._plan_stack[-1].throw(stashed_exception or resp)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Instrument\dev\bluesky\src\bluesky\utils\__init__.py", line 1949, in __iter__
    return (yield from self._iter)
            ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Instrument\dev\bluesky\src\bluesky\plan_stubs.py", line 745, in input_plan
    return (yield Msg("input", prompt=prompt))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Instrument\dev\bluesky\src\bluesky\run_engine.py", line 1672, in _run
    new_response = await coro(msg)
                   ^^^^^^^^^^^^^^^
  File "C:\Instrument\dev\bluesky\src\bluesky\run_engine.py", line 2668, in _input
    async_input = AsyncInput(self.loop)
                  ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Instrument\dev\bluesky\src\bluesky\utils\__init__.py", line 1165, in __init__
    self.loop.add_reader(sys.stdin, self.got_input)
  File "c:\instrument\apps\python3\Lib\asyncio\events.py", line 530, in add_reader
    raise NotImplementedError
NotImplementedError
```

## How Has This Been Tested?

Unit testing this is a bit gnarly, but have had a go at implementing unit tests for both the high level `input_plan`, and checking that the low-level `AsyncInput` object works asynchronously and doesn't block the event loop. Suggestions to make the unit tests a bit nicer appreciated...

Have also manually tested `input_plan()` on both windows & linux.

Doesn't play well with ctrl-c, but it's no _worse_ than what was previously on `main` in that regard. May have a look at this in a follow-up PR.

### DIFF
--- a/src/bluesky/tests/test_plans.py
+++ b/src/bluesky/tests/test_plans.py
@@ -2,6 +2,7 @@ import collections
 import inspect
 import random
 import re
+from unittest.mock import patch
 
 import numpy as np
 import numpy.testing as npt
@@ -712,3 +713,12 @@ def test_count_failure(RE, hw):
 def test_no_warning(hw):
     # this should not warn
     list(bps.rel_set(hw.motor, 2))
+
+
+def test_input_plan(RE):
+    def plan():
+        result = yield from bps.input_plan("prompt: ")
+        assert result == "answer"
+
+    with patch("bluesky.utils.sys.stdin.readline", return_value="answer\n"):
+        RE(plan())
--- a/src/bluesky/tests/test_utils.py
+++ b/src/bluesky/tests/test_utils.py
@@ -1,6 +1,9 @@
+import asyncio
 import operator
+import time
 import warnings
 from functools import reduce
+from unittest.mock import patch
 
 import numpy as np
 import pytest
@@ -8,7 +11,9 @@ from cycler import cycler
 
 from bluesky import RunEngine
 from bluesky.plan_stubs import complete_all, mv
+from bluesky.run_engine import WaitForTimeoutError
 from bluesky.utils import (
+    AsyncInput,
     CallbackRegistry,
     Msg,
     ensure_generator,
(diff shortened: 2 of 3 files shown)


tacaswell (approved):

tacaswell:
Makes it way simpler to boot!

I hope that this was not available when we first implemented it.

Issue #1755, opened 2024-06-26
Add mode on the RE that pauses on error


coretl:

tacaswell:
It is not global, but the finalize wrapper has a flag for this: https://github.com/bluesky/bluesky/blob/6116a09dc0a292dc5d08902f6fa7e25816c73b61/src/bluesky/preprocessors.py#L496-L516

coretl:
> It is not global, but the finalize wrapper has a flag for this:
> 
> https://github.com/bluesky/bluesky/blob/6116a09dc0a292dc5d08902f6fa7e25816c73b61/src/bluesky/preprocessors.py#L496-L516

GitHub has a race condition, creating an issue from a task tells people about it before you get a chance to write a description!

tacaswell:
sorry for jumping the gun on an empty issue body!

I'm :+1: on making the default controlled by an ENV.

coretl:
> sorry for jumping the gun on an empty issue body!

No problem! I wish "create issue from task" gave you a UI to fill this in...

> I'm 👍 on making the default controlled by an ENV.

Is the name ok? Any advance on `BLUESKY_PAUSE_BEFORE_CLEANUP`?

tacaswell:
`BLUESKY_PAUSE_ON_PLAN_EXCEPTION` might be better?  "cleanup" is too generic and we have too many phases of cleanup!

One fun side effect of doing this globally might that you pause in _every_ layer on the way out.  Might be worth the complexity of an additional escape hatch.

coretl:
> `BLUESKY_PAUSE_ON_PLAN_EXCEPTION` might be better? "cleanup" is too generic and we have too many phases of cleanup!

That sounds like a good name, let's go with that.

> One fun side effect of doing this globally might that you pause in _every_ layer on the way out. Might be worth the complexity of an additional escape hatch.

Do you mean that if you both `@bpp.stage_decorator` and `@bpp.run_decorator` and raise an exception in the plan then it will pause twice? That's probably ok, as we are likely to kill the plan on the first raise rather than resume.

We probably want to log the exception too on the first debug so that we can see why we paused.

Side note: why is `finalize_wrapper` not implemented as a call to `contingency_wrapper`?

tacaswell:
> Do you mean that if you both @bpp.stage_decorator and @bpp.run_decorator and raise an exception in the plan then it will pause twice? That's probably ok, as we are likely to kill the plan on the first raise rather than resume.

[...]

PR #1805, opened 2024-10-04
add mode pause on error

yxrmz:
This PR introduces control over the behavior of a plan when it encounters an error state. 

## Description
If the environment variable `BLUESKY_PAUSE_ON_PLAN_EXCEPTION` is set to `YES`, the plan will transition to the pause state. If the variable is set to NO or is not set, the plan will proceed with the cleanup, which is the default behavior.
`finalize_wrapper` has been refactored to call `contingency_wrapper` with the `pause_for_debug` argument explicitly passed.

## Motivation and Context
This PR addresses issue #1755. The implementation is based on the discussion outlined in that issue.

## How Has This Been Tested?
The unit test verifies that the RunEngine swiches to `pause` state upon encountering an error if the `BLUESKY_PAUSE_ON_PLAN_EXCEPTION` environment variable is set to `YES`; otherwise, it ensures that the exception is processed without pausing.

### DIFF
--- a/src/bluesky/preprocessors.py
+++ b/src/bluesky/preprocessors.py
@@ -2,6 +2,7 @@ import uuid
 from collections import ChainMap, OrderedDict, deque
 from collections.abc import Iterable
 from functools import wraps
+from os import environ
 
 from bluesky.protocols import Locatable
 
@@ -493,7 +494,7 @@ def configure_count_time_wrapper(plan, time):
         return (yield from finalize_wrapper(plan_mutator(plan, insert_set), reset()))
 
 
-def finalize_wrapper(plan, final_plan, *, pause_for_debug=False):
+def finalize_wrapper(plan, final_plan, *, pause_for_debug=None):
     """try...finally helper
 
     Run the first plan and then the second.  If any of the messages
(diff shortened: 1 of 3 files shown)


callumforrester (changes requested):
Apologies, I wasn't part of #1755. If I were I think I would have suggested some other method than an environment variable. My main concern is debugging difficulties since the behaviour may vary for non-obvious reasons. Especially considering that a plan can still programatically override the setting and the override may be buried in N layers of plans. A slightly more bluesky-standard option might be a setting on the `RunEngine`, passed in at construction time?

I realise the work is already done and I don't expect a complete rewrite based on this opinion, just my 2 cents. The implementation as-is looks okay, I would prefer `True`/`False` or `1`/`0` to `YES`/`NO` as values for the variable, it's a little more standard and can be parsed in a less esoteric way:

```python
pause_for_debug = bool(environ.get("BLUESKY_PAUSE_ON_PLAN_EXCEPTION"))
```

coretl:
I'm ambivalent between an environment variable and some RunEngine `__init__` arg. I am likely to restart the RunEngine in order to enter this mode anyway, so either is as convenient as the other.

On implementation side, the `__init__` arg would end up setting a global anyway, so the differences between that and an environment variable are marginal.

@tacaswell thoughts?

mrakitin:
I agree with @callumforrester's idea! I don't think it's a bunch of work to reimplement it as a kwarg to `RunEngine.__init__`.
@yxrmz, could you please have a look?

PR #1806, opened 2024-10-07
Support ophyd-async signals in Suspenders

### LINKED ISSUE #1789: Suspenders should support ophyd-async SignalRs
## Expected Behavior
A Suspender can be added to the RunEngine that uses the value of an ophyd-async SignalR

## Current Behavior
Suspender must use, or documentation says that it must use, an ophyd Signal object

## Possible Solution
make it work

## Context
In looking to update some code, I noticed that the SuspenderBase takes an instance of ophyd.Signal. It should be possible to use an Ophyd or Ophyd-Async signal, with the same kinds of behaviours supported (suspend when signal over value etc.).

jsouter:
Addresses https://github.com/bluesky/bluesky/issues/1789

Pre-requisite for ophyd-async PR coming soon, setting as draft until I have tested this includes sufficient changes needed for https://github.com/bluesky/ophyd-async/issues/508

---

**Stacked on #1923** — this PR is based on `subscribable-protocol` rather than `main`, so the diff shows only the suspender changes. Review #1923 first; GitHub will retarget this PR to `main` automatically when that one merges.

### DIFF
--- a/src/bluesky/suspenders.py
+++ b/src/bluesky/suspenders.py
@@ -42,6 +42,7 @@ class SuspenderBase(metaclass=ABCMeta):
         self._sig = signal
         self._pre_plan = pre_plan
         self._post_plan = post_plan
+        self._last_value = None
 
     def __repr__(self):
         return "{}({!r}, sleep={}, pre_plan={}, post_plan={}, tripped_message={})".format(  # noqa: UP032
@@ -131,7 +132,7 @@ class SuspenderBase(metaclass=ABCMeta):
             if self.RE is None:
                 return
             loop = self.RE._loop
-
+            self._last_value = value
             if self._should_suspend(value):
                 self._tripped = True
                 # this does dirty things with internal state
@@ -374,7 +375,7 @@ class SuspendFloor(_Threshold):
             return ""
 
         just = (
-            f"Signal {self._sig.name} = {self._sig.get()!r} "
+            f"Signal {self._sig.name} = {self._last_value!r} "
             + f"fell below {self._suspend_thresh} "
             + f"and has not yet crossed above {self._resume_thresh}."
         )
@@ -429,7 +430,7 @@ class SuspendCeil(_Threshold):
             return ""
 
         just = (
-            f"Signal {self._sig.name} = {self._sig.get()!r} "
+            f"Signal {self._sig.name} = {self._last_value!r} "
             + f"went above {self._suspend_thresh} "
             + f"and has not yet crossed below {self._resume_thresh}."
         )
@@ -490,7 +491,7 @@ class SuspendWhenOutsideBand(_SuspendBandBase):
             return ""
 
         just = "Signal {} = {!r} is outside of the range ({}, {})".format(  # noqa: UP032
-            self._sig.name, self._sig.get(), self._bot, self._top
+            self._sig.name, self._last_value, self._bot, self._top
         )
         return ": ".join(s for s in (just, self._tripped_message) if s)
 
@@ -547,7 +548,7 @@ class SuspendOutBand(_SuspendBandBase):
             return ""
 
         just = "Signal {} = {!r} is inside of the range ({}, {})".format(  # noqa: UP032
-            self._sig.name, self._sig.get(), self._bot, self._top
+            self._sig.name, self._last_value, self._bot, self._top
         )
         return ": ".join(s for s in (just, self._tripped_message) if s)
 
@@ -683,7 +684,7 @@ class SuspendWhenChanged(SuspenderBase):
         if not self.tripped:
             return ""
 
-        just = f'Signal {self._sig.name}, got "{self._sig.get()}", expected "{self.expected_value}"'
+        just = f'Signal {self._sig.name}, got "{self._last_value}", expected "{self.expected_value}"'
         if not self.allow_resume:
             just += '.  "RE.abort()" and then restart session to use new configuration.'
         return ": ".join(s for s in (just, self._tripped_message) if s)
--- a/src/bluesky/suspenders.py
+++ b/src/bluesky/suspenders.py
@@ -6,6 +6,8 @@ from datetime import datetime, timedelta
 from functools import partial
 from warnings import warn
 
+from bluesky.protocols import Subscribable
+
 
 class SuspenderBase(metaclass=ABCMeta):
     """An ABC to manage the callbacks between asyincio and pyepics.
@@ -30,12 +32,9 @@ class SuspenderBase(metaclass=ABCMeta):
     tripped_message : str, optional
         Message to include in the trip notification
 
-    is_async : bool, optional
-        Whether the signal is asynchronous and implements `subscribe_value`
-        or is synchronous and implements `subscribe`
     """
 
-    def __init__(self, signal, *, sleep=0, pre_plan=None, post_plan=None, tripped_message="", is_async=False):
+    def __init__(self, signal, *, sleep=0, pre_plan=None, post_plan=None, tripped_message=""):
         """ """
         self.RE = None
         self._ev = None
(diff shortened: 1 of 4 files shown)


src/bluesky/protocols.py
```diff
@@ -464,12 +464,16 @@ def stop(self, success=True) -> SyncOrAsync[None]:
 @runtime_checkable
 class Subscribable(HasName, Protocol):
     @abstractmethod
-    def subscribe(self, function: Callback) -> None:
+    def subscribe(self, function: Callback, run: bool = True) -> None:
```
tacaswell, line 467:
```suggestion
    def subscribe(self, function: Callback, *, run: bool = True) -> None:
```

Do we want to make this keyword-only?

src/bluesky/suspenders.py
```diff
@@ -131,14 +132,17 @@ def __call__(self, value, **kwargs):
...
-
+            if isinstance(value, dict) and len(value) == 1:
+                # handle dict[str, Reading] provided by ophyd-async
+                (reading,) = value.values()
+                value = reading["value"]
             if self._should_suspend(value):
                 self._tripped = True
                 # this does dirty things with internal state
                 if self._ev is None and self.RE is not None:
                     self.__make_event()
```
coretl, line 143: @tacaswell this looks harder than we first thought:

- `SuspenderBase.install` calls `Signal.subscribe(self)`, which is ok, but will only work inside the asyncio event loop for ophyd-async (it does CA calls for instance which need the event loop)
- `SuspenderBase.__call__` expects to be called outside the asyncio event loop and then jumps through hoops to create an `asyncio.Event` inside the asyncio event loop which will fail as we are blocked waiting for a `threading.Event`

There are 2 possible avenues to explore:
1. Doing a bunch of event loop detection in `SuspenderBase` to work out where we are running, and punt the `sig.subscribe` call to the `RE` if `SuspenderBase.install` is called **outside** a plan
2. Make a whole different set of Suspenders for async

Which do you prefer?

coretl: @danielballan suggests we add `SuspenderBase.install_async` that registers `SuspenderBase.invoke_async` on the signal, then make the existing `SuspenderBase.install` register `SuspenderBase.invoke` with `__call__ = invoke` for back compat

src/bluesky/suspenders.py
```diff
@@ -69,7 +75,10 @@ def install(self, RE, *, event_type=None):
         """
         with self._lock:
             self.RE = RE
-        self._sig.subscribe(self, event_type=event_type, run=True)
+        if self._is_async:
+            self.RE.loop.call_soon_threadsafe(partial(self._sig.subscribe_value, self))
```
coretl, line 79: @tacaswell this uses `subscribe_value` which is not in the protocols. Should we:
- Leave it as it is
- Put `subscribe_value` in the protocols
- Revert to using `subscribe` and do the same detection as in `monitor`: https://github.com/bluesky/bluesky/blob/ca4f8aa136c91dccd428936d2d0ebe963b925f4e/src/bluesky/bundlers.py#L415-L435

jsouter: @coretl @tacaswell Bumping this, Devin from DESY was asking about the status of the PR in today's collaboration meeting, can pick this back up if we make a decision about what to do here.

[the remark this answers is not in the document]
tacaswell: There is set of messages to control this so yes, but users should not close plans over RE.

coretl:
@tacaswell I think this is ready

coretl:
@canismarko if you fancy testing this branch with ophyd-async suspenders that would be great. It's not been tried outside tests.

src/bluesky/suspenders.py
```diff
@@ -69,14 +78,54 @@ def install(self, RE, *, event_type=None):
         """
         with self._lock:
             self.RE = RE
-        self._sig.subscribe(self, event_type=event_type, run=True)
+        if self._implements_protocol:
+            self.__on_loop(RE, partial(self._sig.subscribe_reading, self))
```
tacaswell, line 90:
```suggestion
            if event_type is not None:
                 raise RuntimeError(
                     f"Can not specify non-None event_type {event_type=} with "
                     "Subscribable protocol"
                 )
            self.__on_loop(RE, partial(self._sig.subscribe_reading, self))
```

tacaswell: otherwise there is a code path where we ignore user input.

src/bluesky/suspenders.py
```diff
@@ -667,13 +721,19 @@ def __init__(
         tripped_message="",
         **kwargs,
     ):
-        self.expected_value = signal.value if expected_value is None else expected_value
+        if expected_value is None and not isinstance(signal, Subscribable):
```
tacaswell: Lets move this to install time as well.  I would rather that the subtle init vs install time change now rather than have exactly when it latches be determined by a protocol detail.

tacaswell:
Looks good, but I left two comments.  I think both need to be addressed.

canismarko:
@coretl, sorry, I only saw this on Friday.

> @canismarko if you fancy testing this branch with ophyd-async suspenders that would be great. It's not been tried outside tests.

I tested this against some analog and binary inputs on a labjack DAQ. Everything worked well. The only unexpected behavior I came across was when I purposefully put in conflicting suspenders…

```
- suspenders.SuspendFloor(ao, 2.5)
- suspenders.SuspendCeil(ao, 2.5)
```

 and I got an exception when I cross 2.5 that `bluesky._vendor.super_state_machine.errors.TransitionError: Cannot transit from 'suspending' to 'suspending'.`. I guess that error message could be more helpful, but since there's no reason to want suspenders like that, I don't know that it's worth doing anything about.

Other than, that, it worked well. Everything suspended and resumed properly.

Here's the code I used:

```python
from bluesky import RunEngine, plans as bp
from bluesky import suspenders
from ophyd_async.core import init_devices
from ophyd_async.epics.core import epics_signal_r, epics_signal_rw

RE = RunEngine({}, loop=getattr(RE, "loop", None) or None)  # Re-use loop with `%run -i`

with init_devices():
    ao = epics_signal_rw(float, "25idc:LJT7Voltmeter_0:Ao0", name="ao")
    ai = epics_signal_r(float, "25idc:LJT7Voltmeter_0:Ai0", name="ai")
    bi = epics_signal_r(bool, "25idc:LJT7Voltmeter_0:Bi7.VAL", name="bi")

def ticker(name, doc):
    """Just print a tick so we know something is happening."""
    print("•", end="", flush=True)

RE.subscribe(ticker)

suspenders = [
    # suspenders.SuspendBoolHigh(bi),
    # suspenders.SuspendBoolLow(bi),
    # suspenders.SuspendFloor(ao, 1.5),
    # suspenders.SuspendCeil(ao, 2.5, sleep=1.3),
    # suspenders.SuspendWhenOutsideBand(ao, 1.5, 2.5, sleep=1.3),
    suspenders.SuspendWhenChanged(ao, allow_resume=True),
]
for suspender in suspenders:
    RE.install_suspender(suspender)

plan = bp.count([ai], num=50, delay=2)
# RE(plan)
```

(I did do an iteration where I uncommented multiple suspenders)

coretl:
> and I got an exception when I cross 2.5 that `bluesky._vendor.super_state_machine.errors.TransitionError: Cannot transit from 'suspending' to 'suspending'.`. I guess that error message could be more helpful, but since there's no reason to want suspenders like that, I don't know that it's worth doing anything about.

That's a structural issue at the moment, but I'm fixing it as part of the RunEngine refactor https://github.com/coretl/bluesky/pull/3 (not ready for people to review yet unless you like wading through agent slop...)

> Other than, that, it worked well. Everything suspended and resumed properly.

Sounds good!

@tacaswell are you happy to merge this now?

tacaswell (approved):

PR #1810, opened 2024-10-09
Adjust type hinting on mv

DiamondJoseph:


src/bluesky/plan_stubs.py
```diff
@@ -359,9 +360,8 @@ def rel_set(
     return (yield from relative_set_wrapper(abs_set(obj, *args, group=group, wait=wait, **kwargs)))
 
 
-@plan
 def mv(
-    *args: Tuple[Union[Movable, NamedMovable, Any], ...],
+    *args: Unpack[tuple[Movable, Any]],
```
danielballan, line 364: Nice! I hadn't seen `Unpack` yet.

This needs to be capital-T `Tuple` because `bluesky` supports older  Pythons than 3.10.

danielballan (changes requested):

src/bluesky/plan_stubs.py
DiamondJoseph: @dperl-dls correctly points out that pyright isn't actually OK with this =(
https://github.com/bluesky/bluesky/issues/1809#issuecomment-2402393597

PR #1812, opened 2024-10-14
Drop Python 3.8 from CI matrix.

danielballan:
Python 3.8 reached its end-of-life (per [Python upstream](https://devguide.python.org/versions/)) on October 7, a week ago today.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         runs-on: ["ubuntu-latest", "windows-latest"] # can add macos-latest
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
         include:
           # Include one that runs in the dev environment
           - runs-on: "ubuntu-latest"
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -12,14 +12,14 @@ Before You Begin
     software, and you can skip the rest of this section.  Go to
     `https://try.nsls2.bnl.gov <https://try.nsls2.bnl.gov>`_.
 
-* You will need Python 3.8 or newer. From a shell ("Terminal" on OSX,
+* You will need Python 3.9 or newer. From a shell ("Terminal" on OSX,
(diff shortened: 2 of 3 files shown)


danielballan:
There is ongoing discussion among the Technical Steering Committee about which versions of Python Bluesky projects should support:

- Follow CPython upstream?
- Follow SPEC 0?
- More aggressive?

and the answer may be different for different projects. However, even while we wait for that discussion to conclude, I think we can say confidently that Bluesky cannot support versions of Python that are no longer supported by upstream.

callumforrester (changes requested):
@danielballan Could you also drop it from `pyproject.toml` so that PyPi reflects this change?

danielballan:
Good catch! Done.

prjemian:
Note that Py3.13 is [now supported](https://devguide.python.org/versions/).  New work needed for this throughout the framework.

tacaswell:
> New work needed for this throughout the framework.

The biggest problem I expect is waiting for things with rust extensions to release new versions that pin PyO3 high enough to build 3.13 extensions.  I've been doing my (limited) dev work in a 3.13 environment for a few months now and not had any major surprises.

evalott100:
This needs bumping too

https://github.com/bluesky/bluesky/blob/3bce96b85558993499174be7eebcd09f270a9271/pyproject.toml#L35

mrakitin (approved):
Looks good to me. Let's move on to avoid failures in other PRs for the outdated version of Python.

danielballan:
Rebased and addressed @tangkong's comments. Only remaining instances of `3.8` in the codebase:

```
$ git grep "3\.8 "
docs/api_changes.rst:Added support for Python 3.8 and the following for forward-compatibility with
src/bluesky/tests/test_callbacks.py:|        43  |  04:17:23.8  |          0.85  |          0.56  |
```

danielballan:
The linter is failing on unrelated complaints, looks like increased strictness.

tacaswell:
Merged over the lint failure, looks like ruff got stricter about using base types as generics instead of the special ones from typing (e.g. `tuple` vs `Tuple`).

danielballan:
I expect that is related to Python 3.8 going EOL, as Python 3.8 was the last version that did not support using generics as types.

PR #1815, opened 2024-10-15
Refactor the _setup_plots in BEC

stan-dot:
Extracted some logic from _setup_plots into 3 helper functions

### DIFF
--- a/src/bluesky/callbacks/best_effort.py
+++ b/src/bluesky/callbacks/best_effort.py
@@ -16,12 +16,15 @@ from datetime import datetime
 from functools import partial
 from io import StringIO
 from pprint import pformat
+from typing import Any
 from warnings import warn
 
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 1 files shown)


stan-dot:
we'd need to drop Python 3.8 for this (?)

If the numpy conventions are to be followed...
https://github.com/bluesky/bluesky/issues/699#issuecomment-581966554

then?

> On Apr 14, 2023 drop support for Python 3.8 (initially released on Oct 14, 2019) 

https://numpy.org/neps/nep-0029-deprecation_policy.html

tacaswell:
👍🏻 in principle, but

tacaswell:
"power cycled" to re-trigger CI and get new (synthentic) merge commit.

tacaswell (approved):

tacaswell:
I think something got stricter in the underyling typing (and maybe mpl 3.10) so that still needs some work, but 👍🏻 in principal.

PR #1816, opened 2024-10-15
Add JPEG consolidator for JPEG image stacks

jwlodek:
## Description
Makes a superclass `MultipartRelatedConsolidator` that contains the logic for templating based on file number (not attached to the name, perhaps someone could suggest a better one?). Adds `JPEGConsolidator` that goes along with `TIFFConsolidator` to allow for collecting JPEG images.

## Motivation and Context

While not used for scientific data that will be analyzed, beamlines frequently would like the ability to save jpeg snapshots or image series of monitoring cameras or the like.

## How Has This Been Tested?

Tested with jpeg image stack collected with tomo demo setup in the lab. Requires a source install of a specific branch of ophyd async. Data readback requires https://github.com/bluesky/tiled/pull/796


src/bluesky/consolidators.py
```diff
@@ -271,21 +272,22 @@ def adapter_parameters(self) -> Dict:
...
-        """Return a full uri for a datum (an individual TIFF file) based on its index in the sequence.
+        """Return a full uri for a datum (an individual image file) based on its index in the sequence.
 
         This relies on the `template` parameter passed in the StreamResource, which is a string either in the "new"
         Python formatting style that can be evaluated to a file name using the `.format(indx)` method given an
-        integer index, e.g. "{:05d}.tif".
+        integer index, e.g. "{:05d}.ext".
         """
-        return self.uri + self._sres_parameters["template"].format(indx)
+        filename_template = self._sres_parameters["template"]
```
genematx, line 288: I was actually going to move this line to `__init__`, do some extra validation for back-compatibility there, and then just use `self.template` in get_datum_uri. I think you changes here would fall nicely in that pattern.

src/bluesky/consolidators.py
```diff
@@ -271,21 +272,22 @@ def adapter_parameters(self) -> Dict:
         return {"dataset": self._sres_parameters["dataset"].strip("/").split("/"), "swmr": self.swmr}
 
 
-class TIFFConsolidator(ConsolidatorBase):
-    supported_mimetypes = {"multipart/related;type=image/tiff"}
-
-    def __init__(self, stream_resource: StreamResource, descriptor: EventDescriptor):
+class MultipartRelatedConsolidator(ConsolidatorBase):
+    def __init__(self, permitted_extensions: str, stream_resource: StreamResource, descriptor: EventDescriptor):
```
genematx, line 276: shouldn't `permitted_extensions` be an iterable (list) of str?

jwlodek: Yep, originally had it as a string but obviously changed to list since we can allow multiple. Actually, I guess a Set makes more sensr than list here anyway?

genematx: yes, set is perfect!

jwlodek:
I am pretty sure the 3.8 failure is because 3.8 can't handle `type[]` type annotations, and since we are dropping 3.8 with https://github.com/bluesky/bluesky/pull/1812 I will leave it as is for now.

danielballan:
Python 3.8 has been dropped in `main`.

danielballan (approved):

PR #1822, opened 2024-10-22
made relevant protocols generic

evalott100:


src/bluesky/protocols.py
```diff
@@ -352,10 +352,13 @@ def collect_pages(self) -> SyncOrAsyncIterator[PartialEventPage]:
         ...
 
 
+T_CONT = TypeVar("T_CONT", contravariant=True)
```
coretl, line 355: I'm always slightly hazy on when we need to define a TypeVar as contravariant, can you provide a snippet where this fails if it isn't defined as contravariant?

evalott100: Truth be told, I don't really see why this would need to be contravariant explicitly... it just fixed the mypy errors:

```
src/bluesky/protocols.py:358: error: Invariant type variable "T" used in protocol where contravariant one is expected  [misc]
src/bluesky/protocols.py:486: error: Invariant type variable "T" used in protocol where contravariant one is expected  [misc]
```

evalott100: If I had to guess it's because the methods defined on the protocol take `T` so you have to declare that subclasses are allowed.

evalott100:
I think this PEP relates to why we have to define the typevar as contravariant https://peps.python.org/pep-0544/#overriding-inferred-variance-of-protocol-classes

DiamondJoseph (approved):

coretl (approved):

evalott100:
@danielballan Are we okay to merge this?

danielballan (approved):

Issue #1772, opened 2024-07-31
`LiveTable` callback doesn't work with ophyd-async object


Tom-Willemsen:
The `LiveTable` callback crashes if an `ophyd-async` device/signal is passed into it (it works fine with string names, just not with the actual object)

@coretl - as discussed briefly earlier

## Expected Behavior

Expected similar behaviour to an `ophyd` signal:

```python
>>> from ophyd.signal import SignalRO
>>> from bluesky.run_engine import RunEngine
>>> from bluesky.callbacks import LiveTable
>>> import bluesky.plan_stubs as bps
>>>
>>> dev = SignalRO(name="blah")
>>> RE = RunEngine()
>>>
>>> RE(bps.null(), LiveTable([dev]))
()
```

## Current Behavior

With an `ophyd-async` signal:

```python
>>> from bluesky.run_engine import RunEngine
>>> from bluesky.callbacks import LiveTable
>>> from ophyd_async.epics.signal import epics_signal_r
>>> import bluesky.plan_stubs as bps
>>>
>>> dev = epics_signal_r(float, "blah")
>>> RE = RunEngine()
>>>
>>> RE(bps.null(), LiveTable([dev]))
Traceback (most recent call last):
  File "c:\Instrument\dev\bluesky\tmp\.venv\Lib\site-packages\bluesky\callbacks\core.py", line 200, in get_obj_fields
    field_list = sorted(field.describe().keys())
                        ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'coroutine' object has no attribute 'keys'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "c:\Instrument\dev\bluesky\tmp\.venv\Lib\site-packages\bluesky\callbacks\core.py", line 311, in __init__
    self._fields = get_obj_fields(fields)
                   ^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Instrument\dev\bluesky\tmp\.venv\Lib\site-packages\bluesky\callbacks\core.py", line 202, in get_obj_fields
    raise ValueError(  # noqa: B904
ValueError: Fields must be strings or objects with a 'describe' method that return a dict.
```

## Possible Solution

Might just need wrapping in `maybe_await` or similar?

## Steps to Reproduce (for bugs)

See above

## Context

`LiveTable` is useful for debugging. Workaround: pass the signal name as a string rather than the device itself.

[...]

PR #1828, opened 2024-10-29
DOC: Discourage passing a device object to LiveTable

Tom-Willemsen:
## Description

Fixes https://github.com/bluesky/bluesky/issues/1772

As described in #1772, `LiveTable` does not work if you pass in an `ophyd-async` device object.

Although some other possible approaches were discussed on the issue (e.g. guaranteed sync protocols to get column names), the least invasive approach seems to be to just document the limitation and stop _recommending_ passing a device object to `LiveTable` in the docs, but without actually changing the implementation at all.

## How Has This Been Tested?

No code changes.

### DIFF
--- a/docs/callbacks.rst
+++ b/docs/callbacks.rst
@@ -72,7 +72,7 @@ A working example:
     from bluesky.plans import scan
     from bluesky.callbacks import LiveTable
     dets = [det]
-    RE(scan(dets, motor, 1, 5, 5), LiveTable(dets))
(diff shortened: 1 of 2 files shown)


danielballan:
I am on board with this. I remember we added support for this is very early days when we were under pressure from early users to make the syntax as sparse as possible. Now there are higher-level tools (best effort callback, GUIs, …) and users are much less frequently typing this out.

tacaswell (approved):

PR #1830, opened 2024-10-29
Loosen typing in plan_stubs to allow current use of ophyd-async

dperl-dls:
## Description
Loosens/corrects a few type hints

## Motivation and Context
Addresses the part of https://github.com/bluesky/bluesky/issues/1809 which relates to `ophyd-async` and things which are clearly wrong.

Doesn't make any changes which would be needed to improve things for `ophyd` as that is much more complicated and probably nobody actually uses strict type checking with ophyd (?), in any case requires further discussion.

## How Has This Been Tested?
By removing `type: ignore` statements in `dodal` and `mx-bluesky` and checking that pyright no longer flags errors for `bps.mv`, `bps.abs_set`, etc. for `ophyd-async` objects

### DIFF
--- a/src/bluesky/plan_stubs.py
+++ b/src/bluesky/plan_stubs.py
@@ -4,16 +4,9 @@ import time
 import typing
 import uuid
 import warnings
-from collections.abc import Awaitable, Hashable, Iterable, Mapping, Sequence
+from collections.abc import Awaitable, Callable, Hashable, Iterable, Mapping, Sequence
 from functools import reduce
-from typing import (
-    Any,
-    Callable,
-    Literal,
-    Optional,
-    Union,
-    Tuple
-)
+from typing import Any, Literal, Optional, Union
 
 from cycler import cycler
 
@@ -404,7 +397,9 @@ mov = mv  # synonym
 
 
 @plan
-def mvr(*args: Union[Movable, Any], group: Optional[Hashable] = None, **kwargs) -> MsgGenerator[Tuple[Status, ...]]:
+def mvr(
+    *args: Union[Movable, Any], group: Optional[Hashable] = None, **kwargs
+) -> MsgGenerator[tuple[Status, ...]]:
     """
     Move one or more devices to a relative setpoint. Wait for all to complete.
 
--- a/src/bluesky/plan_stubs.py
+++ b/src/bluesky/plan_stubs.py
@@ -12,6 +12,7 @@ from typing import (
     Literal,
     Optional,
     Union,
+    Tuple
 )
 
 from cycler import cycler
@@ -33,7 +34,6 @@ from .protocols import (
     Locatable,
     Location,
     Movable,
-    NamedMovable,
     PartialEvent,
     Preparable,
     Readable,
@@ -263,7 +263,7 @@ def null() -> MsgGenerator:
 @plan
 def abs_set(
     obj: Movable,
-    *args,
+    *args: Any,
     group: Optional[Hashable] = None,
     wait: bool = False,
     **kwargs,
@@ -311,7 +311,7 @@ def abs_set(
 @plan
 def rel_set(
     obj: Movable,
-    *args,
+    *args: Any,
     group: Optional[Hashable] = None,
     wait: bool = False,
     **kwargs,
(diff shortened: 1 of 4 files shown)


coretl (approved):

src/bluesky/plan_stubs.py
```diff
@@ -359,9 +364,11 @@ def rel_set(
     return (yield from relative_set_wrapper(abs_set(obj, *args, group=group, wait=wait, **kwargs)))
 
 
+# The format (device1, value1, device2, value2, ...)
+# is not currently able to be represented in python's type system
 @plan
 def mv(
-    *args: Tuple[Union[Movable, NamedMovable, Any], ...],
+    *args: Any,
```
tacaswell, line 371: why not `Unpack[Tuple[Any]]` here as well?

tacaswell: or why use the fancier spelling in `abs_set` / `rel_set`

dperl-dls: `*args: Unpack[Tuple[Any]]` means "one and only one positional arg of any type with any name", so that wouldn't work here, but `*args: Unpack[Tuple[Any, ...]]`, which would allow `args` to be a tuple of any types of any length, is the widest possible thing `args` can be in the language, and is equivalent to plain `Any` so I think that's more readable.

dperl-dls: @callumforrester @coretl any opinions on that?

coretl: I like:
- typing follows the `bluesky.protocol`, and documents what we want people to use
- runtime allows for args and kwargs for backwards compat

So this really becomes: do we want to encourage the pattern of `set` taking a single argument?

If so, then I would prefer:
- `Movable[T]` having `def set(self, *args: Unpack[Tuple[T]])`
- `def abs_set(obj: Movable[T], *args: Unpack[Tuple[T]], ..., **kwargs)`
- `def mv(*args: Any, ...)` (possibly with some `typing.override`s for the 2, 4, 6 argument case like `asyncio.gather` has)
 
to encourage that close matching at the typing level. It still doesn't break any runtime things, but gives nice type hints when you try to `abs_set` a `float` Signal to a `str`.

callumforrester: I actually have very few opinions: My original hope was to have a set of well-typed, backwards compatible plans that we could use in blueapi. For various reasons (not just this thread) that's not going to be possible and we will be writing wrapper plans with robust argument types. Something like this...

```python
Move[T] = Tuple[Movable[T], T]

def move(moves: list[Move[Any]], wait: bool = True) -> MsgGenerator:
    yield from bps.mv()

...
# in some plan...

yield from move((x, 2.0), (y, 3.0))
yield from move((x, 2.0), (y, "3.0"))  # fails type check
```

So the core bluesky typing should be as loose as is needed to avoid breaking backwards compatibility and maintain a smooth CLI/IPython shell experience.

dperl-dls: Ok, given the above (and us not having `Movable[T]` yet anyway) I'll make the signatures for `abs/rel_set` take anything and we can tighten it in our wrappers

tacaswell: `list[Move[Any]]` is a much nicer API (from a programming point of view, not from a type-it-in-the-terminal point of view).

I would be happy for that to go upstream and `bps.mv` to be a shim that does pair-wise grouping and calls `move` (assuming we can come up with a reasonable way to not have `mv` and `move` both tab-complete in a cli.

dmgav: I just wanted to note that similar changes were made for built-in plans. For example type annotation for `scan` works well with the Queue Server:
https://github.com/bluesky/bluesky/blob/a40a025aef605174bc44980254c3a1ea8f95bfc9/src/bluesky/plans.py#L1189
I am adding support for `NamedMovable`, so the annotation can be extended. If the type is simply set to `Any`, it disables conversion of strings to ophyd objects, so it needs to be a union that includes `Movable` and/or `NamedMovable` for motors. Also `Any` can be replaced with something more specific. I will check this PR branch with QS before merging to make sure everything works.

dperl-dls:
Remaining linting errors are unrelated and for deprecation of % format specifiers in favour of f-strings

coretl:
@genematx please could you have a look at the test failue?

genematx:
> @genematx please could you have a look at the test failue?

Looks like it's a regression in Tiled, but likely caused by an `httpx` update to 0.28.0. My theory is that the issue is somewhere [here](https://github.com/bluesky/tiled/blob/d0e8ac167ac02e5ea9cc2553037bb2b447884f08/tiled/client/container.py#L421); I'll track it down and let you know asap. Interesting that we've caught it by tests in bluesky.

danielballan:
Tiled v0.1.0b12 has been released with an adjustment for the behavior change in httpx.

danielballan:
As promised, @coretl, the tiled release has righted the test jobs. Looks like `codecov/patch` has some remaining objections.

dperl-dls:
My mistake, I had accidentally fixed a UP031 problem before ignoring it, and then codecov wanted all of log.py tested, so I just reverted that

PR #1835, opened 2024-11-01
Migrate code en-masse to fulfill new ruff rules

### LINKED ISSUE #1834: Fix linter on CI
It appears that an update to ruff has become much stricter about using `tuple` vs `Tuple` (and other built in classes).  We need to:

 - run ruff over the code base and mass-migrate to the new scheme (assuming py39 supports generics on the built in classes)
 - tell ruff to ignore those checks

callumforrester:
Fixes #1834 

## Motivation and Context
Ruff has introduced new rules that we can fulfull now that #1812 is merged. In particular, [UP006](https://docs.astral.sh/ruff/rules/non-pep585-annotation/).


tacaswell:
Thank you for cleaning up a mess I made.  I miss-understood the source of the errors and thought it was from ruff updating behind the scenes on us, not having default rules dependent on supported Python versions 🐑 🤦🏻 .

Issue #1838, opened 2024-11-06
return in finally swallows exceptions


iritkatriel:
In https://github.com/bluesky/bluesky/blob/a40a025aef605174bc44980254c3a1ea8f95bfc9/src/bluesky/run_engine.py#L1117 and https://github.com/bluesky/bluesky/blob/a40a025aef605174bc44980254c3a1ea8f95bfc9/src/bluesky/run_engine.py#L1119 there is a return statement in a finally block, which would swallow any in-flight exception.

This means that if an unhandled exception is raised in the try block (including a `BaseException` other than `KeyboardInterrupt`) or an exception is raise in an `except` block (as is explicitly done in the `except Exception` block), this exception would not propagate as expected.

See also https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions.

tacaswell:
Thank you for flagging this!  I have a PR in to fix it.  Fortunately I think that this is far enough down the sad-path that it is unlikely to have been hit is practice.

I am curious what drew your attention?

iritkatriel:
> I am curious what drew your attention?

I'm doing some analysis on how people use/abuse return in finally, so running static analysis on open source code. I report problems when I see them.

PR #1839, opened 2024-11-06
Mnt/move return out of finally

tacaswell:
## Description

Make sure we do not eat exceptions in a finally block 

## Motivation and Context

Bug reported in #1838 

## How Has This Been Tested?

Just the existing test suite.  I'm not sure how to generate the conditions we would need to be in to generate this error and ran out of time before meetings started for the day.

### DIFF
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -1114,9 +1114,11 @@ class RunEngine:
                             plan_return = self._task_fut.result()
                         except concurrent.futures.CancelledError:
                             plan_return = self.NO_PLAN_RETURN
-                        return plan_return  # noqa: B012
                     else:
-                        return self.NO_PLAN_RETURN
+                        plan_return = self.NO_PLAN_RETURN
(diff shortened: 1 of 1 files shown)


tacaswell:
To trigger the bug that this is fixing I think we would have to:
 - have `self._during_task.block` raise something that is not `KeyboardInterupt`
 - have the plan finish successfully before we call `halt()`


You could almost make the case that dropping the exception on the floor _is_ the correct behavior as the exception is coming out of the waiting code, but plan is the thing we really care about.  However, I think that is a edge case where the timing is _juuust_ right to sneak by and in general it means we have a bug in `_during_task.block` that could bring down a otherwise happily running scan and we should fix that.

As for "well, if the during task should not bring down the scan, why don't we eat the error here" is because that can drive us to a place where the user has the prompt back but a plan is running it the background which is a bit of undefined behavior currently so on balance I think it is better to stop things and identify the bug that risk going to an undefined place.

PR #1843, opened 2024-11-12
Running RE and a QTGui under the captive python shell of pycharm with live updating plots 

SHaringT:
## Description
Instead of checking for the mpl qt backend in the block function it is now checked if there is PyQt.QtWidgets or Pyside.QtWidgets imported and then checked if there is a QApplication (QT Gui). Then it proceeds as before.

## Motivation and Context
When using Pycharm the mpl backend is switched and checking for the qt backend isn't a good indicator. This results in the Qt Gui being blocked and freezing while executing a bluesky plan. 

## How Has This Been Tested?
Testing set of bluesky. Running Bluesky and PyQt in a few different setups.

### DIFF
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -1663,8 +1663,8 @@ class DefaultDuringTask(DuringTask):
         """
         if "matplotlib" in sys.modules:
             import matplotlib
-
-            if _qt_is_imported_app_available():
+            backend = matplotlib.get_backend().lower()
+            if "qt" in backend:
                 from bluesky.callbacks.mpl_plotting import initialize_qt_teleporter
 
                 initialize_qt_teleporter()
@@ -1843,15 +1843,11 @@ class DefaultDuringTask(DuringTask):
 
 
 def _qt_is_imported_app_available():
-    qt_modules_names = ["PyQt6.QtCore", "PySide6.QtCore", "PyQt5.QtCore", "PySide2.QtCore"]
-    try:
-        qt_core, *_ = (mod for mod in [sys.modules.get(name) for name in qt_modules_names] if mod is not None)
-    except TypeError:
-        return False
-    qapp = qt_core.QAppliaction.instance()
-    if qapp is None:
-        return False
-    return True
+    qt_modules_names = ["PyQt6.QtWidgets", "PySide6.QtWidgets", "PyQt5.QtWidgets", "PySide2.QtWidgets"]
+    qt_widgets = next((sys.modules.get(name) for name in qt_modules_names if sys.modules.get(name)), None)
+
+    return qt_widgets is not None and qt_widgets.QApplication.instance() is not None
+
 
 def _rearrange_into_parallel_dicts(readings):
     data = {}
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -1664,8 +1664,7 @@ class DefaultDuringTask(DuringTask):
         if "matplotlib" in sys.modules:
             import matplotlib
 
-            backend = matplotlib.get_backend().lower()
-            if "qt" in backend:
+            if _qt_is_imported_app_available():
                 from bluesky.callbacks.mpl_plotting import initialize_qt_teleporter
 
                 initialize_qt_teleporter()
@@ -1683,7 +1682,7 @@ class DefaultDuringTask(DuringTask):
 
             backend = matplotlib.get_backend().lower()
             # if with a Qt backend, do the scary thing
-            if "qt" in backend:
+            if _qt_is_imported_app_available():
                 import functools
 
                 from matplotlib.backends.qt_compat import QT_API, QtCore, QtWidgets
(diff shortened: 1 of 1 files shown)


Tom-Willemsen:
fyi, the reason the `bec` unit tests are failing on windows is that the new check is inconsistent with the check in `bluesky/callbacks/mpl_plotting::QtAwareCallback`. The reason it's failing _only_ on windows is because matplotlib on windows uses qtagg by default as it's backend if available, compared to linux where it defaults to agg.

i.e. where `matplotlib.get_backend() == "qtagg"` but there is no `QApplication.instance()`, previously it would have run [this](https://github.com/bluesky/bluesky/pull/1843/files#diff-f2fea5f6d79b7bee5cc664eab766cf6d40610e946389f2e3f82a8b55be3ec7deR1722) to create a new application, now it no longer does that -- and so [events sent to the qt teleporter in `QtAwareCallback`](https://github.com/bluesky/bluesky/blob/main/src/bluesky/callbacks/mpl_plotting.py#L78) never get received anywhere. It also points to a slightly wider issue - the plots don't pop up successfully for me if matplotlib is configured to use a qt backend but without an existing QApplication.

SHaringT:
Thanks a lot for your insights. That's a quite new topic for me.

Sounds like a check [here](https://github.com/bluesky/bluesky/blob/main/src/bluesky/utils/__init__.py#L1669) for no QApplication being available and creating one could be solving the problem?

I'm not sure what to make of the lint and ubuntu 3.10 test.

Tom-Willemsen:
> I'm not sure what to make of the lint and ubuntu 3.10 test.

For the lint, run `ruff format` & `ruff check --fix` locally and then commit the result (or manually fix any errors ruff flags up)

For the ubuntu 3.10 test, I think I've seen that test / similar tests be flaky before - it's doing timing, so prone to race conditions. Likely nothing to do with your change, you can probably just re-run that job.

I will let others comment on the best place(s) to check for/create qapplications & do the qt magic.

src/bluesky/callbacks/mpl_plotting.py
```diff
@@ -79,6 +80,19 @@ def __call__(self, name, doc, *, escape=False):
         else:
             return CallbackBase.__call__(self, name, doc)
 
+    @staticmethod
```
tacaswell, line 83: we should import these from `blueksy.utils` rather than having a second copy.

tacaswell:
I suggest trying deferring as much work until in the `block` method rather than the `__init__`.

SHaringT:
Did so. Currently I'm mainly trying to get the tests and scenarios I've set up to work and using the test pipeline for that. I hope that's okay.

Totally up for cleanup in the end.

tacaswell:
@SHaringT I took the liberty of pushing commits that may fix the tests, feel free to drop them if you want.

SHaringT:
@tacaswell that's very welcome. I'm a little puzzeled why ubuntu worked before and now fails.

tacaswell:
I think @Tom-Willemsen got it right.  Previously the ubuntu tests don't have X11 set up so it was always falling back to Agg and we were not testing Qt at all on linux, where as on windows there is a graphical environment available and we do use Qt.

I think the failures we are seeing are "can Qt actually make a QAppliaction"

I suspect that we either need to add X11 to the CI jobs or adjust the logic so that we (bluesky) never tries to make a QApplication (which may have some impact on how people set up sessions...)

Tom-Willemsen:
fyi @tacaswell with your latest commits and the following script on windows:

```
from bluesky import RunEngine
from ophyd.sim import det, motor
import matplotlib

print(matplotlib.get_backend())
plt.ion()
RE = RunEngine()
RE(scan([det], motor, 0, 1, num=2), LivePlot(y="det"))
input("done")
```
I get:
```
qtagg
...\.venv\Lib\site-packages\bluesky\callbacks\mpl_plotting.py:151: UserWarning: Starting a Matplotlib GUI outside of the main thread will likely fail.
  fig, ax = plt.subplots()
```
where "will likely fail", does fail, with a segfault. On `main` this does work and brings up a qt plot. So this isn't just to do with unit tests.

SHaringT:
It might be some CL disturbance. Locally the tests work (windows 10), the gui does not freeze and plotting graphs work fine in pycharm and create popups started from cli.

tacaswell:
I think the issue is now that Matplotlib is being lazy about creating the QApplication as well and due to that a whole bunch of stuff not working quite right.

```
from bluesky import RunEngine
from ophyd.sim import det, motor
import matplotlib

print(matplotlib.get_backend())
plt.ion()
plt.figure()
plt.close('all')
RE = RunEngine()
RE(scan([det], motor, 0, 1, num=2), LivePlot(y="det"))
input("done")
```

If I change the example to the above it work (creating the figure forces the import of PyQt and the creation of the QApplication).

tacaswell:
> ERROR src/bluesky/tests/test_bec.py::test_simple[False] - RuntimeError: Invalid DISPLAY variable

That is progress (it broke in a new way) and confirms the "we were not using Qt on linux"!

tacaswell:
ok, psuhed a pretty big refactor (and there is the chance to de-dent a whole lot of code but I did not do that here so the diff is readable).

tacaswell:
Sorry, I thought this would be straight forward. Did not mean to send you into a quagmire and then take over the PR 🤦🏻 .

SHaringT:
@tacaswell no need to apologize really. I have learned things, will look into your refactor tomorrow to learn some more and the problem seems to be solved; I'm happy.

SHaringT:
I tried to reproduce the failing test locally and locally it's passing. Might be a problem because of the timings?

PR #1845, opened 2024-11-13
ENH: Stackable Consolidators

genematx:
This adds a possibility do define Consolidators that, instead of stacking data frames along a new dimension (added to the left) -- e.g. a stack of images, append them at the bottom of the existing left-most dimension -- e.g. a growing CSV table. This behaviour can be controlled by the `stackable` attribute (and Resource parameter).

Moreover, the new `join_chunks` attribute/parameter controls if smaller chunks at the ends of separate consecutive assets need to be joined together in the final consolidated dataset (experimental and debatable...).

## Motivation and Context
This is needed to ingest exisiting data with a developed CSV adapter and keep the shape of the resulting array back-compatible.

## How Has This Been Tested?
New test cases added to the existing testing routines. Also, tested with real data from the beamlines as part of transitioning from Mongo to SQL storage backend.


src/bluesky/callbacks/core.py
```diff
@@ -16,13 +16,20 @@
...
 from ..utils import ensure_uid
 
-MIMETYPE_LOOKUP = {
-    "hdf5": "application/x-hdf5",
-    "AD_HDF5_SWMR_STREAM": "application/x-hdf5",
-    "AD_HDF5_SWMR_SLICE": "application/x-hdf5",
-    "AD_TIFF": "multipart/related;type=image/tiff",
-    "AD_HDF5_GERM": "application/x-hdf5",
-}
+MIMETYPE_LOOKUP = defaultdict(
```
genematx, line 19: we probably need to remove this.

src/bluesky/consolidators.py
```diff
@@ -226,31 +288,23 @@ def get_adapter(self, adapters_by_mimetype=None):
...
-            chunks=self.chunks,
-        )
-        adapter_kwargs = dict(parameters)
-        adapter_kwargs.update(self.adapter_parameters)
+        return adapter_class.from_assets(self.assets, structure=self.structure, **self.adapter_parameters)
 
-        return adapter_factory(**adapter_kwargs)
+
+class CSVConsolidator(ConsolidatorBase):
+    supported_mimetypes: set[str] = {"text/csv;header=absent"}
```
danielballan, line 297: This seems like a good use of MIME type parameters. Are the allowed values present/absent? Using JSON boolean terms `true` and `false` might be slightly more guessable.

src/bluesky/consolidators.py
```diff
@@ -179,6 +237,14 @@ def adapter_parameters(self) -> dict:
         """
         return {}
 
+    @property
+    def structure(self) -> ArrayStructure:
```
danielballan, line 241: I wonder if this should be a method. Since these objects will not generally be handled interactively by users, saving typing `()` is not so important anyway.

The docstring on property above, `adapter_parameters`, says:

> This property is to be subclassed as necessary.

I was taught that overriding properties in a subclass is a bit of an anti-pattern. There is no way to run `super()` on a `property`, so the do not extend well. Maybe all of these (`has_skips`, `adapter_parameters`, `structure`) should be methods.

danielballan (commented):
I made some comments in line.

I have a conceptual question. I think that "How to combine arrays into a larger array" is a two part question:

1. `concatenate` or  `stack`?
2. Which `axis1?

Here are two arrays `a` and `b` combined in four different ways:

```python
In [18]: a
Out[18]:
array([[1., 1.],
       [1., 1.],
       [1., 1.]])

In [19]: b
Out[19]:
array([[2., 2.],
       [2., 2.],
       [2., 2.]])

In [20]: np.concatenate([a, b], 0)
Out[20]:
array([[1., 1.],
       [1., 1.],
       [1., 1.],
       [2., 2.],
       [2., 2.],
       [2., 2.]])

In [21]: np.concatenate([a, b], 1)
Out[21]:
array([[1., 1., 2., 2.],
       [1., 1., 2., 2.],
       [1., 1., 2., 2.]])

In [22]: np.stack([a, b], 0)
Out[22]:
array([[[1., 1.],
        [1., 1.],
        [1., 1.]],

       [[2., 2.],
        [2., 2.],
        [2., 2.]]])

In [23]: np.stack([a, b], 1)
Out[23]:
array([[[1., 1.],
        [2., 2.]],

       [[1., 1.],
        [2., 2.]],

       [[1., 1.],
        [2., 2.]]])
```

As a numpy practioner, if I were present with parameters:

- `mode: Literal["stack", "concatenate"]`
- `axis: int`

I would understand more quickly what I was being asked. "Stackable" requires some reading to grok, and _may_ not be sufficiently expressive (missing `axis`). Or maybe we can assume that `axis=0` always?

src/bluesky/consolidators.py
genematx: yep, this is from [the standard](https://www.ietf.org/rfc/rfc4180.txt#:~:text=The%20%22header%22%20parameter%20indicates%20the%20presence). CSV doesn't seem to account for many optional parameters, but the `header` is handy.

danielballan: Oh! We love a standard! :-D

Bilchreis:
Hi Eugene, 
I see that this PR adds support for structured numpy arrays in `ConsolidatorBase` for external data. I would like to add support for this in internal data. 

 When I tried using `TiledWriter`  I always crashes at this stage when I try to write numpy arrays, even non-structured ones:
``` python
 df = pd.DataFrame(df_dict, index=[0], columns=df_dict.keys())

ValueError: Data must be 1-dimensional, got ndarray of shape (10, 10) instead
```

Is this out of the scope of this PR? 
Should i open a seperate Issue for this?

genematx:
Hi, Peter

Thanks for your interest in our PR!

I think the initial idea was that the "internal" table would hold the data that was received in Bluesky documents themselves (Events and Datums) directly inline; as such, at least in my experience, these were always simple scalars (forming 1D arrays -- table columns -- over the acquisition time). Does your RE write structured numpy arrays and send them as Bluesky Events? Or do your (10x10?) arrays come from some other source (e.g. detector?). In this case, we would normally assume that the data is external, even though it might be filled into the Bluesky documents, it should still be possible to read it directly from files -- which is the default mode for Tiled in this case.

There might be a conceptual limitation to our ability to allow structured numpy data in internal -- namely how the data is stored on disk (currently as csv files, and later -- as tables in an sql database -- this option is still experimental, but we are going to roll it out and make it default very very soon). It is not clear, how to efficiently store sequences of 10x10 data in these tables (hence, better to put them into external).

Hope I understood your question correctly, and it makes sense. @Bilchreis

Bilchreis:
Yes my RE writes structured numpy arrays, and sends them as Bluesky events. 

I use them for structured data provided by SECoP [devices](https://sampleenvironment.github.io/secop-site/specification/datainfo.html#collection-of-named-items-struct). While I am aware, that Bluesky will only handle structured data with a maximum depth of 1 I would really like to have them work as seamless as possible. They are usually small and would not warrant external saving. 

For example somethin glike this:
``` json
{"x": 0.5, "y": 1}
```
But if there is really this conceptual limitation in the future, I will have to think of a better solution.

genematx:
> Are we good to merge this?
I haven't responded to all of @danielballan's comments yet -- will aim to do today. Let's wait until he's back and then merge.
Also, would be good to check it with the struct dtypes, but it could be its own PR.
@jwlodek

src/bluesky/consolidators.py
genematx: good point! Changed.

src/bluesky/consolidators.py
```diff
@@ -86,27 +104,42 @@ def __init__(self, stream_resource: StreamResource, descriptor: EventDescriptor)
...
         )
-        self.dtype = np.dtype(dtype_spec)
+        if dtype_numpy.kind == "V":
+            self.data_type = StructDtype.from_numpy_dtype(dtype_numpy)
+        else:
+            self.data_type = BuiltinDtype.from_numpy_dtype(dtype_numpy)
+
+        # Set chunk (or partition) shape
         self.chunk_shape = self._sres_parameters.get("chunk_shape", ())
         if 0 in self.chunk_shape:
```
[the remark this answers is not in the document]
genematx: Sure, easy!

genematx:
I completely agree with what @danielballan said:

> "Stackable" requires some reading to grok...

I think defining this option as a Literal instead of a bool would make the configuration much smoother. I have changed it to `join_method: Literal["stack", "concat"]`.

In terms of the axis selection, I kept it to `axis=0` by default for now to keep the interface complexity low. It shouldn't be too hard to expand it to other axis in the future, if needed (or I can do it now, if there're such cases already, but from what I've seen we only join on the left-most dimension).

danielballan (approved):

danielballan:
I think this is good to go once CI passes. I am assuming that we will later refactor `MIMETYPE_LOOKUP` to use a plugin system, similar to what we have done with handlers.

But as this is all still rapidly evolving I think that can happen in a separate PR.

I think we must support structured dtypes, too, in order to be compatible with existing data, but that is definitely a separable PR.

danielballan (approved):

PR #1853, opened 2024-12-12
API: make zict an optional dependency and update docs

### LINKED ISSUE #1577: bug: Incompatible with zict 3.0.0
While looking at merging a PR it was found that CI jobs were failing as they were using the latest (breaking) version of zict.

## Expected Behavior
Job should not fail

## Current Behavior
CI Job fails.

## Possible Solution
Someone should investigate why it's failing, exactly; we know it's a zict problem but not sure how/why.

see https://github.com/bluesky/bluesky/pull/1575

## Context
I have a small PR (one line change) open to change something about bluesky FailedStatus tracebacks. The CI jobs for this are all failing. 

Upon closer inspection the culprit is the zict library 3.0.0 upgrade.

## Your Environment
See  https://github.com/bluesky/bluesky/pull/1575

A PR to temporarily pin zict for now has been opened.

tacaswell:
Also xfail the one bad test based on version gating.

closes #1577

## Motivation and Context

zict 3 has a behavior change that breaks one of our tests (and using two bsui sessions), see #1577 for details (thank you @dmgav for sorting out the problem!).

Pinning back zict is causing other conflict with dask and we should start to move away from it all together.

@genematx I know you are working on document RedisJsonDict, hopefully this does not step on what you are doing too much.

## How Has This Been Tested?

locally running tests

### DIFF
--- a/docs/api_changes.rst
+++ b/docs/api_changes.rst
@@ -2,6 +2,21 @@
  Release History
 =================
 
+v1.13.1 (2024-12-12)
+====================
+
+Changed
+-------
+* ``zict`` is no longer a default dependency.  There is a breaking API change
+  between zict2 and zict3 that fixed a race condition between multiple threads
+  using the same instance but broke using multiple instances in the same
+  process or multiple processes sharing the same files.  ``PersistentDict`` is
+  not being removed, but is strongly discouraged for new use.  To get the old pinning use
+  ```pip install bluesky[old_persistentdict]`` or install ``zict<3``.
+
+
+
+
 v1.13.0a4 (2024-07-08)
 ======================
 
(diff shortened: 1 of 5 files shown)


prjemian:
I've got an alternative we're just starting to test. It's file-backed rather than redis. Does not use zict. It uses YAML and MutableMapping.

tacaswell:
If there are better alternatives on the way I still think we should merge this as-is and then update as the new things are ready.

prjemian:
https://github.com/BCDA-APS/bs_model_instrument/blob/main/src/instrument/utils/stored_dict.py

tacaswell:
The code that is not being covered is the error message, however I don't think it is worth setting up an additional full run without zict installed to catch that.

tacaswell:
I marked this as API because it may catch someone out, but I think it is OK to do this in a 1.13.1 release.

danielballan (approved):
Let’s tag and release an RC first and crank the tests on Ophyd and Ophyd-async.

PR #1857, opened 2024-12-12
DEV: run copier update to fix pypi publishing

tacaswell:
## Description

Result of running copier

## Motivation and Context

Primary goal is to fix pypi publishing.


tacaswell:
Squashed this to two commits (we don't need my random walk to fixing everything!) but could go to 1 if we want.

tacaswell:
The decrease in the coverage is the tests of persistent dict. 

I do not think it is worth adding a test run that pins zict back, but could be convinced otherwise.

coretl (approved):

PR #1861, opened 2024-12-28
add StoredDict as alternative to PersistentDict

### LINKED ISSUE #1854: StoredDict- an alternative to PersistentDict
APS has been [motivated](https://github.com/BCDA-APS/bs_model_instrument/issues/14) to create [StoredDict](https://bcda-aps.github.io/bs_model_instrument/api/utils.html#instrument.utils.stored_dict.StoredDict), an alternative to PersistentDict.  

Several of our beamlines are wont to use both IPython console, Jupyter notebook, and possibly queueserver sessions simultaneously.  In such cases, and when PersistentDict is configured to use a common directory for every bluesky session, invariably a report comes back about failure to unpack some file.  The error comes from msgpack,  The root cause is how zict writes a dictionary to storage (one file per key).  When a problem is reported, it is always the case that the file names in the PersistentDict directory are mangled versions of the dictionary keys, such as `versions#23`.  The current workaround is to remove all files in the PersistentDict directory.

- #1853

## Expected Behavior
The `RE.md` dictionary should be saved to storage whenever its contents have changed.  On startup, `RE.md` contents should be restored from storage without errors.

## Requirements to replace `bluesky.utils.PersistentDict`.

- https://github.com/BCDA-APS/bs_model_instrument/issues/14#issuecomment-2504801796

* All dictionary contents must be JSON serializable.
* Contents stored in a single human-readable file.
* Sync to file _shortly_ after dictionary is updated.
* File format should support comments.
* Do not use [zict](https://zict.readthedocs.io/en/latest/index.html).  It uses a directory with one file per dictionary key.  _Zict_ creates an unnecessary problem (recurrent incidents at APS) with how it names these files.  The problem is exacerbated when concurrent sessions are updating via _PersistentDict_.

## Current Behavior
Users report unpacking problems when starting a bluesky session.  They are puzzled since *unpacking*

## Possible Solution

- [StoredDict](https://github.com/BCDA-APS/bs_model_instrument/pull/30) - currently in beamline testing at APS
- Redis-backed dictionary
- Some other idea

## Context
Trying to reduce the number of problems that might occur during routine startup of any bluesky session.

prjemian:
- close #1854

Our motivation is well-outlined in #1854.  The `RE.md` content is a good candidate for YAML storage.  Further, this eliminates the need for `zict`.

This PR provides code, tests, docs, and adds a comment for the API change history.


prjemian:
One advantage with a single-file persistence scheme is that the metadata can be written to the user's PWD.  Often, the `scan_id` should be remembered for a user's experiments, which may include multiple restarts of a bluesky session.  For new users, the `RE.md` automatically resets to the defaults for that instrument.  No carry-over of inappropriate metadata between different users.

I've been using a file name `.re_md_dict` (stored in the session's startup PWD).

prjemian:
CI `zict` failures for testing in Windows OS are unrelated to this PR.

prjemian:
[Such as](https://github.com/bluesky/bluesky/actions/runs/12530026026/job/34946176144):

```
ERROR src/bluesky/tests/test_stored_dict.py::test_repr - PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_3wr9vy1e.yml'
= 1 failed, 1766 passed, 32 skipped, 35 xfailed, 13 errors in 921.33s (0:15:21) =
```

prjemian:
Alternatives to YAML file-backed storage include an ophyd `Signal` (such as an EpicsSignal) and Redis (such as a [Redis Signal](https://github.com/bluesky/bluesky/pull/1861)).  There's some previous history with saving `RE.md` including PersistentDict and HistoryDict.

prjemian:
- https://github.com/bluesky/bluesky/pull/277
- https://github.com/bluesky/bluesky/pull/1228
- https://github.com/bluesky/bluesky/pull/1311

prjemian:
It should be noted that this code could switch to a different storage model by changing these two static methods: https://github.com/bluesky/bluesky/blob/cea114e982da355c31e535bf4fa3fbe2c4ce0a16/src/bluesky/utils/__init__.py#L1047-L1063

docs/metadata.rst
```diff
@@ -381,6 +382,12 @@ of ``RE.md`` can persist between sessions.
     RE.md = StoredDict('some/path/here')  # path is a file
 
 .. rubric:: PersistentDict
+
+Bluesky formerly recommended using :class:`~bluesky.utils.PersistentDict` --- a
+Python dictionary backed by a zict directory. This approach proved
```
tacaswell, line 387: If we are going to provide this commentary it should be correct. 

The issue is that if multiple instances `PersistentDict` work on the same directory (either in one or multiple processes) they would get out of sync and look for files that did not exist.

src/bluesky/utils/__init__.py
```diff
@@ -898,6 +901,168 @@ def reload(self):
...
+        """Number of keys in the dictionary."""
+        return len(self._cache)
+
+    def __repr__(self):
+        """representation of this object."""
+        return f"<{self.__class__.__name__} {dict(self)!r}>"
+
+    def __setitem__(self, key, value):
+        """Write to the dictionary."""
+        outermost_frame = inspect.getouterframes(inspect.currentframe())[-1]
```
tacaswell, line 989: Can you explain a bit more why we need this?

docs/metadata.rst
prjemian: See commit 63228e49ff

prjemian:
Windows CI failed with this exception:

```py
FAILED src/bluesky/tests/test_stored_dict.py::test_StoredDict - AssertionError: assert 1736868181.3347142 > 1736868181.3347142
 +  where 1736868181.3347142 = <built-in function time>()
 +    where <built-in function time> = time.time
 +  and   1736868181.3347142 = <StoredDict {'a': 1}>._sync_deadline
ERROR src/bluesky/tests/test_stored_dict.py::test_StoredDict - PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_zj07vpos.yml'
```

Proves the review point that this code is vulnerable to two (or more) processes trying to write the file at the same time.

prjemian:
> Windows CI failed with this exception:

<details>
> 
> ```python
> FAILED src/bluesky/tests/test_stored_dict.py::test_StoredDict - AssertionError: assert 1736868181.3347142 > 1736868181.3347142
>  +  where 1736868181.3347142 = <built-in function time>()
>  +    where <built-in function time> = time.time
>  +  and   1736868181.3347142 = <StoredDict {'a': 1}>._sync_deadline
> ERROR src/bluesky/tests/test_stored_dict.py::test_StoredDict - PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_zj07vpos.yml'
> ```
> 
> Proves the review point that this code is vulnerable to two (or more) processes trying to write the file at the same time.
</details>

Fails here: https://github.com/bluesky/bluesky/blob/c345f189d068bf32ea373328ccc6e1be6dd15740/src/bluesky/tests/test_stored_dict.py#L39

prjemian:
Look like a typo here: https://github.com/bluesky/bluesky/blob/c345f189d068bf32ea373328ccc6e1be6dd15740/src/bluesky/tests/test_stored_dict.py#L35-L36

prjemian:
```py
2025-01-14T16:26:27.1157972Z src/bluesky/tests/test_stored_dict.py::test_set_exceptions[md5-None-None] PASSED [ 92%]Exception ignored in: <_io.FileIO name='C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_l7kvgkc6.yml' mode='rb+' closefd=True>
2025-01-14T16:26:27.1159009Z Traceback (most recent call last):
2025-01-14T16:26:27.1170299Z   File "C:\hostedtoolcache\windows\Python\3.12.8\x64\Lib\site-packages\_pytest\mark\structures.py", line 607, in __iter__
2025-01-14T16:26:27.1171615Z     if keyword not in self._markers:
2025-01-14T16:26:27.1172015Z        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-01-14T16:26:27.1172886Z ResourceWarning: unclosed file <_io.FileIO name='C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_l7kvgkc6.yml' mode='rb+' closefd=True>
2025-01-14T16:26:27.1174201Z Exception ignored in: <_io.FileIO name='C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_9zlhjw0c.yml' mode='rb+' closefd=True>
2025-01-14T16:26:27.1175045Z Traceback (most recent call last):
2025-01-14T16:26:27.1175938Z   File "C:\hostedtoolcache\windows\Python\3.12.8\x64\Lib\site-packages\_pytest\mark\structures.py", line 607, in __iter__
2025-01-14T16:26:27.1176788Z     if keyword not in self._markers:
2025-01-14T16:26:27.1177159Z        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-01-14T16:26:27.1178000Z ResourceWarning: unclosed file <_io.FileIO name='C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_9zlhjw0c.yml' mode='rb+' closefd=True>
2025-01-14T16:26:27.1179288Z Exception ignored in: <_io.FileIO name='C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_j2_57250.yml' mode='rb+' closefd=True>
2025-01-14T16:26:27.1180103Z Traceback (most recent call last):
2025-01-14T16:26:27.1180942Z   File "C:\hostedtoolcache\windows\Python\3.12.8\x64\Lib\site-packages\_pytest\mark\structures.py", line 607, in __iter__
2025-01-14T16:26:27.1182062Z     if keyword not in self._markers:
2025-01-14T16:26:27.1182418Z        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-01-14T16:26:27.1183250Z ResourceWarning: unclosed file <_io.FileIO name='C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_j2_57250.yml' mode='rb+' closefd=True>
```

from this code: https://github.com/bluesky/bluesky/blob/c345f189d068bf32ea373328ccc6e1be6dd15740/src/bluesky/tests/test_stored_dict.py#L122-L128

Either `sdict.flush()` or `luftpause()`.

prjemian:
Down to one failure:

```py
2025-01-14T17:02:37.1434176Z   File "D:\a\bluesky\bluesky\src\bluesky\tests\test_stored_dict.py", line 80, in test_StoredDict
2025-01-14T17:02:37.1434679Z     assert len(open(md_file).read().splitlines()) == 4  # Still.
2025-01-14T17:02:37.1435008Z AssertionError: assert 0 == 4
2025-01-14T17:02:37.1435281Z  +  where 0 = len([])
2025-01-14T17:02:37.1435626Z  +    where [] = <built-in method splitlines of str object at 0x00000281C4468030>()
2025-01-14T17:02:37.1436148Z  +      where <built-in method splitlines of str object at 0x00000281C4468030> = ''.splitlines
2025-01-14T17:02:37.1436658Z  +        where '' = <built-in method read of _io.TextIOWrapper object at 0x000002819B856B50>()
2025-01-14T17:02:37.1437584Z  +          where <built-in method read of _io.TextIOWrapper object at 0x000002819B856B50> = <_io.TextIOWrapper name='C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_1tnp2_ta.yml' mode='r' encoding='cp1252'>.read
2025-01-14T17:02:37.1438747Z  +            where <_io.TextIOWrapper name='C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\re_md_1tnp2_ta.yml' mode='r' encoding='cp1252'> = open(WindowsPath('C:/Users/RUNNER~1/AppData/Local/Temp/re_md_1tnp2_ta.yml'))
```

Tom-Willemsen:
> Windows CI failed with this exception

Just as a pointer for windows:
- You need to use `with open(something) as f: f.read()` everywhere, not `open(something).read()` otherwise file handles stay open
- You need to explicitly close the `NamedTemporaryFile` in your tests after `yield` and before `unlink` (or use `NamedTemporaryFile` as a context manager which will do that for you)

prjemian:
@Tom-Willemsen : Big thanks! You saved a lot of time for me!

Ready for review again.

prjemian:
> Proves the review point that this code is vulnerable to two (or more) processes trying to write the file at the same time.

In the end, it did _not prove_ this point.  There were problems resolvable by use of context management and longer delays for processing in the unit tests.

tacaswell:
Currently `__delitem__` and `popitem` do not update the cache:

```
In [3]: d = StoredDict('test.yaml', delay=0)

In [4]: ! cat test.yaml
cat: test.yaml: No such file or directory

In [5]: d[0] = 1

In [6]: ! cat test.yaml
# Written by StoredDict.
# Dictionary contents written: 2025-01-16 21:22:37.046256

0: 1

In [7]: del d[0]

In [8]: d
Out[8]: <StoredDict {}>

In [9]: ! cat test.yaml
# Written by StoredDict.
# Dictionary contents written: 2025-01-16 21:22:37.046256

0: 1

```

It can be just as important to remove items as add them.  This is trivially fixable.

-----

May not matter, but iteration order is not preserved through the dump/load cycle 

---


A rather artificial stress test:

```python
import time
import threading

def stress(target, payload):
    target.clear()
    deadline = time.monotonic() + 60
    while time.monotonic < deadline:
        target.update(payload)


d = StoredDict('test.yaml', delay=0)
d2 = StoredDict('test.yaml', delay=0)

th2 = threading.Thread(target=stress, args=(d2, {str(j):j for j in range(10)}))
th1 = threading.Thread(target=stress, args=(d, {j:j for j in range(10)}))
d3 = StoredDict('test.yaml', delay=0)
d3.clear()
d3[0] = 0
th1.start()
th2.start()
deadline = time.monotonic() + 60
res = {'good': 0, 'bad': 0}
while time.monotonic() < deadline:
    try:
        d3 = StoredDict('test.yaml', delay=0)
    except Exception:
        print('failed to read')
        res['bad'] += 1
        continue
    if len(set(type(k) for k in d3) ) >1:
        print(d3)
        res['bad'] += 1
    elif len(d3) != 10:
        print(d3)
        res['bad'] += 1
    else:
        res['good'] += 1
```

comes back with 

```
In [97]: res
Out[97]: {'good': 13291, 'bad': 19258}
```

This is probably fixable using an atomic write scheme (write new results to a temporary file and then use system calls to move it to the correct place) however I'm not sure that even this will save you from non-posix semantics on write (particularly NFS mounts).


-----

This currently does not look past the top level to notice user mutation so if any of the values are mutable objects (like lists or dicts) then `StoredDict` will miss any nested updates:


```
In [99]: d['a'] = {j:j for j in range(5)}

In [100]: d['b'] = list(range(5))

In [101]: d
Out[101]: <StoredDict {'a': {0: 0, 1: 1, 2: 2, 3: 3, 4: 4}, 'b': [0, 1, 2, 3, 4]}>

In [102]: ! cat test.yaml
# Written by StoredDict.
# Dictionary contents written: 2025-01-16 21:56:45.870815

a:
  0: 0
  1: 1
  2: 2
  3: 3
  4: 4
b:
- 0
- 1
- 2
- 3
- 4

In [103]: d['a']['bob'] = 'aardvark'

In [104]: ! cat test.yaml
# Written by StoredDict.
# Dictionary contents written: 2025-01-16 21:56:45.870815

a:
  0: 0
  1: 1
  2: 2
  3: 3
  4: 4
b:
- 0
- 1
- 2
- 3
- 4

In [105]: d['b'].clear()

In [106]: ! cat test.yaml
# Written by StoredDict.
# Dictionary contents written: 2025-01-16 21:56:45.870815

a:
  0: 0
  1: 1
  2: 2
  3: 3
  4: 4
b:
- 0
- 1
- 2
- 3
- 4

In [107]: d
Out[107]: <StoredDict {'a': {0: 0, 1: 1, 2: 2, 3: 3, 4: 4, 'bob': 'aardvark'}, 'b': []}>

In [108]: d['b'].extend('aardvark')

In [109]: d
Out[109]: <StoredDict {'a': {0: 0, 1: 1, 2: 2, 3: 3, 4: 4, 'bob': 'aardvark'}, 'b': ['a', 'a', 'r', 'd', 'v', 'a', 'r', 'k']}>

In [110]: ! cat test.yaml
# Written by StoredDict.
# Dictionary contents written: 2025-01-16 21:56:45.870815

a:
  0: 0
  1: 1
  2: 2
  3: 3
  4: 4
b:
- 0
- 1
- 2
- 3
- 4

In [111]: 
```

This one is also solvable, but you need to do something like https://github.com/NSLS2/redis-json-dict/blob/55bdbb69d9c82fc5cb7febda45e6b71fec97169e/src/redis_json_dict/redis_json_dict.py#L41-L55 to make sure you are notified of any changes and schedule the dump.


----

With the delayed write you probably should register at `atexit` function to make sure there is a last-chance flush and make the worker thread a daemon thread so that there are not extra delays exiting the process, but that is pretty minor.

tacaswell:
Fundamentally the problem is we really _want_ a database without the complexity of actually using a database so there is always going to be limitations of the half-measures we try.

I think the delayed-flush-to-single-file-on-disk is a reasonable approach if we start the pitch that this is a partial best-effort solution and address some of the implementation issues I noted above.

prjemian:
@tacaswell  Thanks for your suggestions!

> iteration order is not preserved

Not a requirement.

> automatically notice if the file on disk has changed and update the Python side cache to account for another agent writing an update

Not only is this sort of IPC is not a requirement, allowing some external process to control the contents of `RE.md` could lead to incorrect metadata in the `start` document.

tacaswell:
> Not only is this sort of IPC not a requirement, allowing some external process to control the contents of RE.md could lead to incorrect metadata in the start document.

As soon as we have two processes looking at the same file you _have_ IPC.  Consider two sessions:

```
>>> d = StoredDict('/tmp/yolo.yaml')
>>> d['a'] = 'bob'
>>> ...
>>> d['c'] = 'other'
>>> ...
>>> sys.exit()
```

```
>>> d = StoredDict('/tmp/yolo.yaml')
>>> d['a'] = 'bill'
>>> ...
>>> d['c'] = 'other
>>> ...
>>> sys.exit()
```

If we make a third session (which would be the only one running) what is the value of `d['a']` ? It will on which process last set _any_ key and flushed to disk.  I think the core of the issue is that doing write-only flushing treats the shared state (file) as cache and the individual state (in-Python dict) as Truth, however this leads to multiple sources of truth.  The user story is something like have two processing running bluesky (which you note is very hard to stop) update some critical metadata in process A, run a scan in process B (which will update scan_id and flush to disk), notice the metadata is wrong, close all the terminals, restart just one and the metadata is still wrong.

The lack of this synchronization was the biggest source of user complaint we had with historydict (open one `bsui` run some scans, open a second `bsui` run some scans in that one, run some more scans in the first one, notice that `scan_id` is repeated between runs, be sad the non-unique thing is not unique).  


-----


In retrospect, I think we made a mistake way back with `historydict` and attempting to automatically persist any values put in which was driven by a desire to not make `scan_id` "special".  I think it would have been better to have had a well-known json/yaml/toml file path to read on init to pre-populate `RE.md` (so the values could be hand edited or manually persisted) and to have had a "get me the next locally ordered count" function for the scan id that was wholly independent (in 2018 we added the ability to do this but I don't think we have ever used it).

prjemian:
> As soon as we have two processes looking at the same file you have IPC

The goal with this code is to save `RE.md` content between sessions.  That is persistence but not IPC.  There is no goal to exchange between multiple, active sessions.  That is IPC.  No intention to implement that here.  It's out of scope.

prjemian:
If an installation has multiple sessions running simultaneously, updating the same file in storage, they should apply some administrative controls to limit that behavior.  Adding the code here to manage such contention is well beyond the design goal.

prjemian:
> then `StoredDict` will miss any nested updates

Acknowledged. Good thing to address here.  How can we detect such changes?

prjemian:
> Anything that mutates the dict in one thread and reads it in another does need some locking

Would this be addressed by changing: https://github.com/bluesky/bluesky/blob/43e61d36a5bd66bf124331d13edda5063a4de464/src/bluesky/utils/__init__.py#L1057

to:

```py
            f.write(yaml.dump(contents.copy(), indent=2)) 
```

prjemian:
This [SO post](https://stackoverflow.com/questions/38165611/detect-value-change-in-custom-mutablemapping-class-with-unkown-depth) discusses how to respond to changes beyond the first level of the `MutableMapping`.  

> ... need to create more communication between container and contained objects ...

To gain this control, we might need to add code to `StoredDict.__setitem__()` that replaces any substructures (list, dict, ...) with our own structures that result in a call to `StoredDict`'s `_queue_storage()` method on changes.

`PersistentDict` also lacks this capability, right?

prjemian:
A scheme to share the `scan_id` across multiple, simultaneous bluesky sessions already exists: https://github.com/bluesky/bluesky/blob/e32b637e9166f67b1f7abe5d7bf0552466bae55b/src/bluesky/run_engine.py#L420

The "scheme" is that one writes a function similar to the default example: https://github.com/bluesky/bluesky/blob/e32b637e9166f67b1f7abe5d7bf0552466bae55b/src/bluesky/run_engine.py#L208-L209

that references and maintains a shared location.  Apstools has code that uses an EPICS PV to share the current `scan_id` between various EPICS clients (including bluesky sessions).  Each client is responsible for reading the PV and incrementing its value for a new scan.  It is known that any client could modify this PV at any time.  That requires some administrative controls.

tacaswell:
> If an installation has multiple sessions running simultaneously, updating the same file in storage, they should apply some administrative controls to limit that behavior. 

But if you do this, then don't we avoid all of the problems with `PeristintDict`?

prjemian:
I believe that APS beamlines currently using PersistentDict will report fewer startup problems when they switch to StoredDict.

prjemian:
Current CI failures (across the board) follow this example:

```py
2025-01-17T17:48:30.0119648Z FAILED src/bluesky/tests/test_tiled_writer.py::test_handling_non_stream_resource[False] - AttributeError: module 'dask.dataframe.core' has no attribute 'DataFrame'
2025-01-17T17:48:30.0120661Z FAILED src/bluesky/tests/test_tiled_writer.py::test_handling_non_stream_resource[True] - AttributeError: module 'dask.dataframe.core' has no attribute 'DataFrame'
```

Not related to this PR.

prjemian:
Why did CI start failing on dask now (`AttributeError: module 'dask.dataframe.core' has no attribute 'DataFrame'`)?  New [version](https://anaconda.org/conda-forge/dask/files)?  Nope, latest conda-forge version is 30 days old.

```py
2025-01-17T17:54:04.9921772Z Collecting dask[array,bag] (from databroker->bluesky==1.13.1rc3.dev29+g7ba0372c)
2025-01-17T17:54:04.9980108Z   Downloading dask-2025.1.0-py3-none-any.whl.metadata (3.8 kB)
```

On PyPI, the newest version is about an hour old:

image

prjemian:
@danielballan : This is a tiled situation:

```py
2025-01-17T18:08:34.5777253Z   File "/opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/site-packages/tiled/client/dataframe.py", line 174, in read
2025-01-17T18:08:34.5777777Z     ddf = dask.dataframe.core.DataFrame(
2025-01-17T18:08:34.5778019Z           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-01-17T18:08:34.5778353Z AttributeError: module 'dask.dataframe.core' has no attribute 'DataFrame'
```

Not sure when this moved, but dask [source ](https://github.com/dask/dask/blob/0a75125ba89ae1ad71dff7805e1c3ebba6de819a/dask/dataframe/__init__.py#L22-L23) shows this might be imported as `dask.dataframe.DataFrame` or `dask.dataframe.expr.DataFrame`:

```py
from dask.dataframe.dask_expr import (
    DataFrame,
```

prjemian:
FYI: The link that tiled uses was correct in [dask 2023.12.0](https://github.com/dask/dask/blob/753b601b27a224ccc2817e227806cf33463bdd81/dask/dataframe/core.py#L4902).

callumforrester:
This looks good, happy to approve when CI is happy.

Probably for a future PR but we should consider decoupling the storage backend from YAML (but maybe leave YAML as the default). See my suggestion for a similar approach in ophyd-async here: https://github.com/bluesky/ophyd-async/issues/16#issuecomment-2505603713 

Might be worth making an issue about it.

prjemian:
@callumforrester : CI is happy now, but for codecov which drops .57%.  According to this [report](https://github.com/bluesky/bluesky/pull/1861/checks?check_run_id=36068444864), 100% of the differences of this PR are covered.

callumforrester:
I'm happy if everything is resolved to @tacaswell's satisfaction

prjemian:
@tacaswell , you asked

> don't we avoid all of the problems with PeristintDict?

If you read #1854, the answer is clearly "No".  Are you able to approve this PR?

prjemian:
Thanks for all the input.

Issue #1847, opened 2024-11-25
Connecting the RunEngine to Qt GUI via psygnal


jacopoabramo:
I'm trying to build a GUI application with Qt and the RunEngine inside it. I talked about this in the past with other people and the major problem was that using Qt signals would require inheriting from QObject which was undesirable.

I'm currently using [psygnal](https://github.com/pyapp-kit/psygnal) as a replacement for the Qt signals. It's a pure python implementation of the observer pattern and it can be used in connection with Qt application.

I'm attaching a minimal working example below:

[bluesky_playground.py.txt](https://github.com/user-attachments/files/17901466/bluesky_playground.py.txt) (remember to change file extension)
[requirements.txt](https://github.com/user-attachments/files/17901481/requirements.txt)

In the example a detector simply counts endlessly and updates a QLabel with it's data content.

I'm sure that there may be problems with this implementation so feel free to provide feedback. One thing I had to do was monkey patching the [install_remote_qt_kicker](https://github.com/bluesky/bluesky/blob/92d2633127e33e51aa1d1b109c80503e74bbe0e0/src/bluesky/utils/__init__.py#L997) function by doing the following:

```py
    _create_qApp()
    # qApp = matplotlib.backends.backend_qt.qApp
    from qtpy.QtWidgets import QApplication
    qApp = QApplication.instance()
```

I tried to write a kicker function without doing so but it wasn't working. This was the only way that I managed to do it.

EDIT: fixed typos

jacopoabramo:
In the end I managed to resolve by reimplementing DuringTask

```python
class ProcessEventsDuringTask:
    """Implementation of DuringTask that emits events to the Qt event loop."""

    def __init__(self):
        super().__init__()
        self.app = QApplication.instance()

    def block(self, blocking_event: threading.Event):
        """Keep processing events until the blocking event is out of the "waiting" state.
        
        In the default implementation, blocking_event.wait() prevents 
        the Qt loop to correctly process events events. With this implementation we
        keep the Qt event loop alive while the task is running.
        """
        while True:
            done = blocking_event.wait(timeout=0.03)
            self.app.processEvents()
            if done:
                break
```

A truly Christmas miracle. Happy holidays everyone. Closing.

tacaswell:
[...]

PR #1863, opened 2024-12-29
[ENH] Add `bluesky.utils.ProcessQtEventsDuringTask`

jacopoabramo:
## Description

- An implementation of `DuringTask` that kicks a Qt event loop with a fixed refresh rate.
- This implementation follows these assumptions:
  - The Qt event loop and the Bluesky event loop are in the same process.
  - The Qt event loop is the main thread.
- Does not implement any logic for checking SIGINT at this time.

## Motivation and Context

The PR was requested in ticket #1847 by @tacaswell. It implements a logic similar to the one described in the ticket, where the only changes are in how the `QApplication.instance` is recovered (in my original example I simply called QApplication.instance() without importing any package; in this `matplotlib` is checked to see if a Qt backend is currently present).

It provides an alternative implementation to `DefaultDuringTask` as it takes the assumption that both event loops are in the same process.

There's currently no way of handling interruption signal. To be honest, I do believe that when using this approach the interruption should be handled by the Qt event loop by injecting a signal that calls the public API when necessary. I'd be happy to receive feedback about this or pointers on how to address this if necessary.

I don't know if this is true but I believe this is somewhat similar to what the original `install_qt_kicker` implementation, just in the perspective of a  `DefaultTask`.

## How Has This Been Tested?

I only managed to test this locally as there are no UTs available for `DuringTask`. My machine specs:

- OS: Windows 11
- Python 3.10 (mamba environment)
- Bluesky version: 1.13
- Frontend: qtpy with PyQt6 installed

### DIFF
--- a/docs/utils.rst
+++ b/docs/utils.rst
@@ -83,3 +83,4 @@ waiting for the plan to complete in the background thread
    DuringTask.block
 
    DefaultDuringTask
+   ProcessQtEventsDuringTask
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -1849,6 +1849,56 @@ class DefaultDuringTask(DuringTask):
                 # We are not using matplotlib + Qt. Just wait on the Event.
                 blocking_event.wait()
 
+class ProcessQtEventsDuringTask(DuringTask):
+    """This class runs the Qt main loop while waiting for the plan to finish.
+    
+    Differently from "DefaultDuringTask", this class directly invokes the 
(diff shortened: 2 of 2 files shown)


jacopoabramo:
mypy fails:

`Module "matplotlib.backends.qt_compat" has no attribute "QtWidgets"`

This is weird since this is identical to what is implemented in `DefaultDuringTask`. I'm guessing that the error is a "don't care"?

Then a test fails on Windows with Python 3.10 and I have no idea why.

PR #1866, opened 2025-01-07
Add watcher group in wait

### LINKED ISSUE #1860: Add support for waiting for multiple groups in `bps.wait`
## Expected Behavior
- We would like to `bps.wait(group=motors)`, but also `bps.wait(group=dets)` at the same time, and return when the motors are complete. 
- Use case 1 in https://github.com/bluesky/bluesky/issues/1849

## Current Behavior
We can't do this

## Possible Solution
Add a new `watch: Sequence[str] = ()` argument to `bps.wait` (and the wait message). This will be the additional groups to watch while the primary group is waited for. The behaviour should be:
- If the primary group or any of the watch groups fails we should raise that exception
- If the primary group completes we should return immediately
- If any of the watch groups completes successfully we should continue waiting for the primary group

ZohebShaikh:
closes #1860


src/bluesky/run_engine.py
```diff
@@ -2339,9 +2340,11 @@ async def _wait(self, msg):
         if msg.args:
             (group,) = msg.args
             error_on_timeout = True
+            watch = ()
         else:
             group = msg.kwargs["group"]
             error_on_timeout = msg.kwargs.get("error_on_timeout", True)
+            watch = msg.kwargs.get("watch", ())
```
coretl, line 2347: I'm not sure I understand the argument handling of the original, would this make more sense?
```suggestion
        if msg.args:
            (group,) = msg.args
        else:
            group = msg.kwargs["group"]
        error_on_timeout = msg.kwargs.get("error_on_timeout", True)
        watch = msg.kwargs.get("watch", ())
```

coretl (changes requested):

src/bluesky/tests/test_run_engine.py
```diff
@@ -1936,8 +1936,16 @@ def foo(name, doc): ...
...
+@pytest.mark.parametrize(
+    "set_finished,watch_finished",
+    [
+        (True, True),
+        (False, False),
+        (False, True),
+        (True, False),
+    ],
+)
+def test_wait_with_timeout(set_finished, watch_finished, RE):
```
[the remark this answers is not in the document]
coretl: Also, please could we have a test case for use case 1 in https://github.com/bluesky/bluesky/issues/1849

src/bluesky/run_engine.py
```diff
@@ -2357,6 +2358,9 @@ async def _wait(self, msg):
                     # bar.
                     self._call_waiting_hook(status_objs)
                 await self._wait_for(Msg("wait_for", None, futs, timeout=msg.kwargs.get("timeout", None)))
+                # Monitor the status objects for any failures and raise an exception if any are detected.
+                self._watch_status_objs(watch)
```
coretl, line 2362: Doesn't this mean that we don't start watching these groups until after the wait for is complete?

src/bluesky/tests/test_flyer.py
```diff
@@ -623,3 +625,109 @@ def fly_plan():
...
     assert flyer2.call_counts["complete"] == 1
+
+
+def collect_data_during_multiple_scan(dets, motors, stream_name: str = "test_stream"):
+    yield from open_run()
+    yield from declare_stream(*dets, name=stream_name, collect=True)
+
+    grid_scan: list = [10, 10, 10]
+
+    yield from prepare(*dets, sum(grid_scan))
```
coretl, line 636: I was expecting to see this prepare the dets for `[10, 10, 10]`, then call kickoff & complete 3 times, otherwise we don't know whether the detectors have produced all the frames needed for a particular inner scan.

Thinking about it, we I don't think my use case 1 in https://github.com/bluesky/bluesky/issues/1849 is quite correct, let's discuss in person...

ZohebShaikh:
I believe [this](https://github.com/bluesky/bluesky/blob/6456606b3eedf455d2e96f578589e74606a3c5ee/src/bluesky/run_engine.py#L1596) line in run_engine is where the stashed exceptions are being raised. I traced the exception from watcher_groups, and it gets stored in self._exceptions. The watcher is unable to raise the exception because the wait plan is still pending.

coretl:
@tacaswell @danielballan 

I thought that wait returned:
- Either when any status completed early
- or when all statuses completed successfully

However it appears to wait until they are all complete, and only then check error status:
https://github.com/bluesky/bluesky/blob/0d25c16ed6e35918fcb48dfc3ab9fe62cdaf5b9a/src/bluesky/run_engine.py#L2364
https://github.com/bluesky/bluesky/blob/0d25c16ed6e35918fcb48dfc3ab9fe62cdaf5b9a/src/bluesky/run_engine.py#L1827

Is that expected/desired?

Because I can see that being a problem if you start 2 flyers off, and one of them fails and the other continues, it might be an hour until you realise your scan is junk...

If it's not desired, should I fix that as part of this PR?

coretl:
Tests are waiting for @danielballan to make a patch release of tiled to add a dependency on `adbc_driver_duckdb` in some way (caused by new release of duckdb).

@evvaaaa FYI

src/bluesky/tests/test_run_engine.py
```diff
@@ -1961,6 +1963,91 @@ def plan():
...
+
+
+def test_set_failed_before_watch_raises_immediately(RE, device_pair):
+    def plan():
+        status = yield Msg("stage", device_pair[0], group="test_group")
+        yield Msg("stage", device_pair[1], group="watch_group")
+        status.set_exception(Exception("Mock device failed"))
+        yield from wait(group="test_group", timeout=0.1, watch=("watch_group",))
+
+    with pytest.raises(FailedStatus):
```
tacaswell, line 2057: This also needs the timing check I think (to make sure it is not failing at the time of the timeout).

PR #1867, opened 2025-01-08
Remove implicit Pydantic v1 dependency

DiamondJoseph:
Removes an implicit dependency on Pydantic v2 which used Pydantic v1 method that caused DeprecationWarnings

### DIFF
--- a/src/bluesky/callbacks/tiled_writer.py
+++ b/src/bluesky/callbacks/tiled_writer.py
@@ -13,7 +13,6 @@ from event_model.documents import (
     StreamDatum,
     StreamResource,
 )
-from pydantic.v1.utils import deep_update
 from tiled.client import from_profile, from_uri
 from tiled.client.base import BaseClient
 from tiled.client.container import Container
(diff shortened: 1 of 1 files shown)


genematx:
I think the idea here was not just to deep copy the dictionary, but also to update it's subdictionaries, i.e.
```
D1 = {'a':1, 'b':{'c':3, 'd':4}}
D2 = {'b':{'e':5}}
D1.deep_update(D2)

>>D1={'a':1, 'b':{'c':3, 'd':4, 'e':5}}  # desired outcome
```
Maybe this is something new they've added, but I don't think `copy.deepcopy` does this, does it?
Of course, there's no deep_update method on dictionaries, and normal update would completely overwrite 'b'.
The pydantic function was one of the first library implementations of this logic, but I agree we need to do something about the deprecation warnings. Perhaps this is the time for us to write this short function ourselves (or steal it from SO) and put it in utils.

DiamondJoseph:
We could just deepcopy then update the copy?

PR #1868, opened 2025-01-08
Deep update

genematx:
Implements our own version of nested dictionary updates to avoid importing it from pydantic.

### DIFF
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -15,7 +15,7 @@ import types
 import uuid
 import warnings
 from collections import namedtuple
-from collections.abc import AsyncIterable, AsyncIterator, Generator, Iterable
+from collections.abc import AsyncIterable, AsyncIterator, Generator, Iterable, Mapping
 from collections.abc import Iterable as TypingIterable
 from functools import partial, reduce, wraps
 from inspect import Parameter, Signature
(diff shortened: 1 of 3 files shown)


jwlodek (approved):

src/bluesky/utils/__init__.py
```diff
@@ -2010,3 +2010,22 @@ def is_plan(bs_plan):
     """
 
     return inspect.isgeneratorfunction(bs_plan) or getattr(bs_plan, "_is_plan_", False)
+
+
+def deep_update(primary: dict, updates: dict) -> dict:
```
tacaswell, line 2015:
```suggestion
def deep_update(primary: Mapping, updates: Mapping) -> Mapping:
```

I think this is what mypy is asking you to do.

DiamondJoseph: The original deep_update in Pydantic had

KeyType = TypeVar('KeyType')
def deep_update(mapping: dict[KeyType, Any], *updating_mappings: dict[KeyType, Any]) -> dict[KeyType, Any]:

Although I can't see why it shouldn't be `(Dict[K, V], Dict[K,V]) -> Dict[K,V]`

DiamondJoseph (approved):

src/bluesky/utils/__init__.py
```diff
@@ -2010,3 +2010,22 @@ def is_plan(bs_plan):
     """
 
     return inspect.isgeneratorfunction(bs_plan) or getattr(bs_plan, "_is_plan_", False)
+
+
+def deep_update(primary: Mapping, updates: Mapping) -> Mapping:
```
tacaswell, line 2015:
```suggestion
def deep_update(primary: MutableMapping, updates: Mapping) -> Mapping:
```

tacaswell:
I think the problem is now that `Mapping` is just the `__getitem__` and other read APIs, we need `MutableMapping` to be able to update.

tacaswell:
This does suggest that almost anyplace we put `dict[...]` we really should be putting `Mapping[...]` so that mypy shouts at us if we start mutating input!

tacaswell:
```
src/bluesky/callbacks/tiled_writer.py:127: error: Argument 2 to "deepcopy" has incompatible type "dict[str, dict[Any, Any]]"; expected "dict[int, Any] | None"  [arg-type]
```

Maybe we need the explicit `[Any, Any]` on everything?

genematx:
I actually don't understand where this typing error is coming from. There's no deepcopy on line 127 in tiled_writer.py

tacaswell:
Remember that GHA runs an a synthetic merge of this branch into main, a rebase may help make things clearer.

PR #1870, opened 2025-01-14
BaseModel RunEngine.md

burkeds:
Prototyping a Pydantic BaseModel as the RunEngine.md for schema and serialization validation. This PR is to discuss the validity of this idea and how the model can be built from the event_model BaseModels.

In its current state, it is simply an example of a dict-like Pydantic BaseModel with tests demonstrating how it can be used with ChainMap.


danielballan:
I support putting something like this in bluesky with pydantic as an optional dep.

evalott100:
> I support putting something like this in bluesky with pydantic as an optional dep.

It feels like this could maybe replace [`DataType`](https://github.com/bluesky/event-model/blob/a8d45752479d512b9a431e33a142174a14988a14/src/event_model/basemodels/run_start.py#L18-L32) and live in event-model?

Then any extra fields on documents permitting metadata can be aggregated into it in the [basemodels](https://github.com/bluesky/event-model/blob/a8d45752479d512b9a431e33a142174a14988a14/src/event_model/basemodels/run_start.py#L252-L263) and the [jsonschema](https://github.com/bluesky/event-model/blob/a8d45752479d512b9a431e33a142174a14988a14/src/event_model/schemas/run_start.json#L113-L122).

PR #1871, opened 2025-01-17
Change to unpack arguments on `Movable.set`

### LINKED ISSUE #1809: Improve typing in plan stubs, generally and to allow using Ophyd devices
Recent additions of type hints have some errors and omissions - importantly they aren't compatible with function signatures and devices from Ophyd 

## Expected Behavior

Scanning with `pyright` on code which uses bluesky library functions correctly applied to ophyd devices should not raise errors

## Current Behavior
Something like
```python
bps.mv(backlight, BacklightPosition.OUT)
```
will give
```
Argument of type "Backlight" cannot be assigned to parameter "args" of type "Tuple[Movable | NamedMovable | Any, ...]" in function "mv"
    "Backlight" is not assignable to "Tuple[Movable | NamedMovable | Any, ...]" (reportArgumentType)
```

or something like 

```python
bps.abs_set(motor, position)
```
will give 
```
 error: Argument of type "EpicsMotor | Motor" cannot be assigned to parameter "obj" of type "Movable" in function "abs_set"
    Type "EpicsMotor | Motor" is not assignable to type "Movable"
      "EpicsMotor" is incompatible with protocol "Movable"
        "set" is an incompatible type
          Type "(new_position: Any, *, timeout: float = None, moved_cb: (...) -> Unknown = None, wait: bool = False) -> StatusBase" is not assignable to type "(value: Unknown) -> Status"
            Parameter name mismatch: "value" versus "new_position"
            "None" is not assignable to "float"
            Type "None" is not assignable to type "(...) -> Unknown"
            Function return type "StatusBase" is incompatible with type "Status" (reportArgumentType)

```
## Possible Solution
- Expand the signatures for the plan stubs to accommodate Ophyd devices and not only ophyd-async
- Fix or remove errors like the above `tuple` definition

## Your Environment
Encountered in dodal CI runs

evalott100:
Partially closes #1809

### DIFF
--- a/src/bluesky/protocols.py
+++ b/src/bluesky/protocols.py
@@ -19,7 +19,7 @@ from event_model.documents.event import PartialEvent
 from event_model.documents.event_descriptor import DataKey, Dtype
 from event_model.documents.event_page import PartialEventPage
 from event_model.documents.resource import PartialResource
-from typing_extensions import TypedDict
+from typing_extensions import TypedDict, Unpack
 
 # Squashes warning
 Dtype = Dtype  # type: ignore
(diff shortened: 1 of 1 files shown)


coretl:
@olliesilvester please could you see if this will break anything for mx?

coretl (approved):

PR #1873, opened 2025-01-21
Fix mypy typing warnings and errors for event-model

thopkins32:
## Description

Correct typing errors found by `mypy` when using the changes in https://github.com/bluesky/event-model/pull/346

## Motivation and Context

This fixes a lot of typing issues related to event-model. It reduces the likelihood of mistakes since static type checking will now work with event-model data types.

Please let me know if any of the typing choices are invalid.

## How Has This Been Tested?

From the root directory, run:
```bash
$ mypy src
```
and notice that there are no typing errors.

I am assuming that unit tests will cover a lot of the cleanup/refactoring I did. Please let me know if any of the choices I made here could cause issues.

### DIFF
--- a/src/bluesky/callbacks/tiled_writer.py
+++ b/src/bluesky/callbacks/tiled_writer.py
@@ -1,5 +1,5 @@
 import copy
-from typing import Any, Optional, Union
+from typing import Any, Optional, Union, cast
 
 import pandas as pd
 from event_model import RunRouter
@@ -76,9 +76,10 @@ class _RunWriter(CallbackBase):
         Will make changes to and return a shallow copy of StreamRsource dictionary adhering to the new structure.
         """
         doc = copy.copy(doc)
+        stream_resource_doc = cast(StreamResource, doc)
         # If the document already adheres to StreamResource schema, return it
         if "mimetype" in doc:
-            return doc
+            return stream_resource_doc
 
         # The document is a `Resource` or a < v1.20 `StreamResource`.
         # Both are converted to latest version `StreamResource`.
@@ -87,12 +88,13 @@ class _RunWriter(CallbackBase):
                 raise RuntimeError(f"`Resource` or `StreamResource` legacy document is missing a '{expected_key}'")
 
         # Convert the Resource (or old StreamResource) document to a StreamResource document
-        doc["mimetype"] = MIMETYPE_LOOKUP[doc.pop("spec")]
-        doc["parameters"] = doc.pop("resource_kwargs", {})
-        file_path = doc.pop("root").strip("/") + "/" + doc.pop("resource_path").strip("/")
-        doc["uri"] = "file://localhost/" + file_path
+        resource_dict = cast(dict, doc)
+        stream_resource_doc["mimetype"] = MIMETYPE_LOOKUP[resource_dict.pop("spec")]
+        stream_resource_doc["parameters"] = resource_dict.pop("resource_kwargs", {})
+        file_path = resource_dict.pop("root").strip("/") + "/" + resource_dict.pop("resource_path").strip("/")
+        stream_resource_doc["uri"] = "file://localhost/" + file_path
 
-        return doc
+        return stream_resource_doc
 
     def start(self, doc: RunStart):
         self.root_node = self.client.create_container(
--- a/src/bluesky/consolidators.py
+++ b/src/bluesky/consolidators.py
@@ -98,7 +98,7 @@ class ConsolidatorBase:
         # 4. If unable to do any of the above, pass through whatever string is in 'dtype'.
         dtype_spec = (
             data_desc.get("dtype_numpy")  # Try dtype_numpy first
-            or data_desc.get("dtype_str", "")  # Then try dtype_str
+            or str(data_desc.get("dtype_str", ""))  # Then try dtype_str
             or DTYPE_LOOKUP.get(data_desc["dtype"])  # Then try mapping from JSON type
             or data_desc["dtype"]  # Finally fall back to raw dtype string
         )
--- a/src/bluesky/tests/test_seq_num_indices_in_sync.py
+++ b/src/bluesky/tests/test_seq_num_indices_in_sync.py
@@ -43,7 +43,8 @@ class ExternalAssetDevice:
         self.compose_stream_resource = ComposeStreamResource()
         if DRAFT_0_STREAM_RESOURCE:
             self.stream_resource_compose_datum_pairs = tuple(
-                self.compose_stream_resource("", "", f"non_existent_{det}.hdf5", det, {}) for det in self.detectors
+                self.compose_stream_resource("", "", f"non_existent_{det}.hdf5", det, {})  # type: ignore[arg-type]
+                for det in self.detectors
             )
         else:
             self.stream_resource_compose_datum_pairs = tuple(
@@ -124,7 +125,7 @@ class ExternalAssetDevice:
             # New stream_resource half way through the run
             if DRAFT_0_STREAM_RESOURCE:
                 self.stream_resource_compose_datum_pairs = tuple(
-                    self.compose_stream_resource("", "", f"non_existent_{det}.hdf5", det, {})
+                    self.compose_stream_resource("", "", f"non_existent_{det}.hdf5", det, {})  # type: ignore[arg-type]
                     for det in self.detectors
                 )
             else:


jacopoabramo:
I see from this PR that you're using the pipe `|` character to describe union types but that isn't supported until Python 3.10. Does this mean that now Bluesky is setting the minimum supported version to 3.10?

thomashopkins32:
> Does this mean that Bluesky is setting the minimum supported version to 3.10?

@jacopoabramo Not to my knowledge. I will be reverting those changes soon, sorry for the confusion.

thomashopkins32:
Some tests were skipped due to a tiled import issue which is why the coverage went down so much. Re-running the tests should resolve it once it picks up the newer version of tiled: https://github.com/bluesky/tiled/releases/tag/0.1.0-b15

evalott100 (changes requested):

src/bluesky/bundlers.py
```diff
@@ -717,26 +749,28 @@ async def _describe_collect(self, collect_object: Flyable):
...
+        duplicates: dict[str, dict[str, DataKey]] = defaultdict(lambda: {})
+        for stream, data_keys in describe_collect_items:
+            for name, data_key in data_keys.items():
+                for other_stream, other_data_keys in describe_collect_items:
+                    for other_name, other_data_key in other_data_keys.items():
+                        if stream != other_stream and name == other_name and data_key == other_data_key:
+                            duplicates[stream][name] = data_key
         if len(duplicates) > 0:
             raise RuntimeError(
                 f"Can't use identical data keys in multiple streams: {duplicates}",
```
evalott100, line 776: This'll also print the whole dict in the first line. Should probably take away 

```suggestion
                f"Can't use identical data keys in multiple streams:",
```

evalott100:
Another small note, is that linting CI uses `tox -e pre-commit`, which uses `ruff` instead of `mypy`. Could you check it works with `ruff` instead of `mypy`?

src/bluesky/bundlers.py
```diff
@@ -717,26 +749,28 @@ async def _describe_collect(self, collect_object: Flyable):
...
-        for stream, data_keys in describe_collect.items():
-            for key, stuff in data_keys.items():
-                for other_stream, other_data_keys in describe_collect.items():
-                    for other_key, other_stuff in other_data_keys.items():
-                        if stream != other_stream and key == other_key and stuff == other_stuff:
-                            duplicates[stream][key] = stuff
+        # Make sure you can't use identical data keys in multiple streams
+        # TODO: Does `DataKey` have a `__eq__` method implemented?
+        # Data structure is assumed to be dict[stream_name, dictionary of data_keys]
+        duplicates: dict[str, dict[str, DataKey]] = defaultdict(lambda: {})
```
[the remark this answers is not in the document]
thomashopkins32: It does complain here unfortunately. `defaultdict` requires a `Callable` which the builtin `dict` apparently is not.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -109,9 +111,12 @@ def descriptor(self, doc: EventDescriptor):
         metadata = dict(doc)
 
         # Remove variable fields of the metadata and encapsulate them into sub-dictionaries with uids as the keys
-        uid = metadata.pop("uid")
-        conf_dict = {uid: metadata.pop("configuration", {})}
-        time_dict = {uid: metadata.pop("time")}
+        del metadata["uid"]
+        del metadata["time"]
+        del metadata["configuration"]
```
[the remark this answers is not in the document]
thomashopkins32: I see, I can try to rework this a bit then. The problem with using `metadata.pop` is that `metadata` is a `dict` without any typing while the `doc` has typing. That's why I preferred the use of the `doc` here and in other places.

See https://github.com/bluesky/bluesky/pull/1873/files#diff-04097102019c6f37f2ddd2e4d79ba285e1efde3ee7876b89f374f19908ca407cR111 where `metadata` is a copy of the `doc` but as a dictionary without any types.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -129,12 +134,14 @@ def descriptor(self, doc: EventDescriptor):
...
         # Keep specifications for external and internal data_keys for faster access
+        if not isinstance(metadata["data_keys"], dict):
+            raise RuntimeError("Expected data_keys to be a dictionary")
         self.data_keys_int.update({k: v for k, v in metadata["data_keys"].items() if "external" not in v.keys()})
         self.data_keys_ext.update({k: v for k, v in metadata["data_keys"].items() if "external" in v.keys()})
 
         # Write the configuration data: loop over all detectors
         conf_node = desc_node["config"]
-        for det_name, det_dict in conf_dict[uid].items():
+        for det_name, det_dict in doc.get("configuration", {}).items():
```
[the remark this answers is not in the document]
thomashopkins32: I did this for typing since `doc.configuration` is a dictionary which has `.items()` but it doesn't know what `conf_dict[uid]` is.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -192,31 +199,37 @@ def event(self, doc: Event):
...
                 else:
                     raise RuntimeError(f"Datum {datum_id} is referenced before being declared.")
 
     def datum(self, doc: Datum):
-        self._docs_cache[doc["datum_id"]] = copy.copy(doc)
+        self._datum_cache[doc["datum_id"]] = copy.copy(doc)
 
     def resource(self, doc: Resource):
-        self._docs_cache[doc["uid"]] = self._ensure_resource_backcompat(doc)
+        self._resource_cache[doc["uid"]] = copy.copy(doc)
```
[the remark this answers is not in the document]
thomashopkins32: Yes, I think this may have been a bug? The backwards compatibility appears to be only for `StreamResource` only, not `Resource`. Let me know if we still need to ensure backwards compatibility here though because maybe I am misunderstanding something.

src/bluesky/consolidators.py
```diff
@@ -96,14 +96,13 @@ def __init__(self, stream_resource: StreamResource, descriptor: EventDescriptor)
...
-        self.dtype = np.dtype(
-            data_desc.get("dtype_numpy")  # standard location
-            or data_desc.get(
-                "dtype_str",  # legacy location
-                # try to guess numpy dtype from JSON type
-                DTYPE_LOOKUP.get(data_desc["dtype"], data_desc["dtype"]),
-            )
+        dtype_spec = (
+            data_desc.get("dtype_numpy")  # Try dtype_numpy first
+            or str(data_desc.get("dtype_str"))  # Then try dtype_str
```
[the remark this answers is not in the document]
thomashopkins32: Ah I wasn't aware that `""` is considered as `False`. I guess that makes sense because this happens for empty collections as well. Nice catch!

I was also unaware that `str(None) -> "None"`.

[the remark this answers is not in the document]
evalott100: If I could go back in time and ask Guido van Rossum to do anything it would be to

``` python
str(None) # fails like int(None)
repr(None) -> "None"
```

:stuck_out_tongue:

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
evalott100: Ah okay, maybe we could clean it up a little bit in a different issue.

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
evalott100: The doc here is an `EventDescriptor` `TypedDict`, which at runtime can take extra non type hinted values (since it's just a dict at runtime). We use a dict for metadata here because it's fields won't be type hinted in `EventDescriptor`.

``` python
metadata = {
    k: v for k, v in doc.items()
    if k not in (
        "uid", "time", "configuration"
    )
}

# The blow should pass since doc is type hinted:
uid = doc["uid"]
conf_dict = {uid: doc.get("configuration", {})}
time_dict = {uid: doc["time"]}

```
work?

src/bluesky/bundlers.py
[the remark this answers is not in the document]
evalott100: Hmmm that's strange :thinking: `dict` in `defaultdict` is mypy complient for the following snippet:

```

defaultdict(list)
defaultdict(dict)
defaultdict(lambda: {})

assert callable(dict) and callable(list) and callable(lambda: {}) # passes
```

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
thomashopkins32: Also, `conf_dict` is used in `var_fields` which is then used in the `deep_update` of the metadata. So maybe this is used elsewhere?

thomashopkins32:
@evalott100 `pyright` (locally) shows 699 errors and 22 warnings whereas `mypy` shows no issues or warnings...it looks like most of these are unrelated to adding `py.typed` to event-model.

thomashopkins32:
In `tiled_writer.py`, test failure seems to be from separating the document cache into separate caches per document type: `Datum`, `Resource`, `StreamDatum`, and `StreamResource`. It looks like `stream_resource` is never called when it needs to be called prior to `stream_datum`? I'm not sure when/where this call to `StreamResource` should be made but I can dig a bit into this to try and figure it out.

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
thomashopkins32: @genematx When I checked the current schema it looked like the backcompat method wouldn't affect the `Resource` but I definitely could have missed something. Sorry, you are correct in that it's not a bug but I think it may be unnecessary.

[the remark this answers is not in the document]
thomashopkins32: I figured out why this was needed before (it's related to the failed test [here](https://github.com/bluesky/bluesky/actions/runs/12993057312/job/36234392746?pr=1873)). We had `self._docs_cache` before and this backcompat method was basically converting a `Resource` into a `StreamResource`. Since we are introducing stricter typing in this PR, we can't do this turning a `Resource` into a `StreamResource` in the same way as before.

There are two options as I see it:
1. Explicitly turn every `Resource` into a `StreamResource` and cache only `StreamResource`s
2. Handle both `Resource` and `StreamResource` separately, even when their use-cases overlap

I don't have the context for what these types are used for or how they differ so please enlighten me :smile:

evalott100:
> pyright (locally) shows 699 errors and 22 warnings whereas mypy shows no issues or warnings...

Yup, apologies, don't worry about it! Bluesky uses mypy, just got my wires crossed

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
evalott100: Okay we do need this. It explicitly converts (the older) `Resource` to `StreamResource`.

If this were the case, this line would fail though:


https://github.com/bluesky/bluesky/blob/5b3ecff831b1f5c3dd88fc2a7679e64a4f81ef2d/src/bluesky/callbacks/tiled_writer.py#L80


... unless `mimetype` is added to any `Resource` document before this, and `spec` is added to any `StreamResource`. It feels like this should be an `or` statement @genematx ?

If so we can fix it here.

evalott100:
> I'm not sure when/where this call to StreamResource should be made but I can dig a bit into this to try and figure it out.

When making a `stream_datum`, a `resource` seems to be being made instead of a `stream_resource`...

```
=========== STREAM RESOURCE CACHE
{}
=========== RESOURCE CACHE
{'9bff1ab9-54ed-4261-a71e-ca66c78d4e87': {'path_semantics': 'posix',
                                          'resource_kwargs': {'chunk_shape': (1,),
                                                              'template': '_{:d}.tif'},
                                          'resource_path': '223d4f',
                                          'root': '/tmp/pytest-of-twj43146/pytest-159/test_handling_non_stream_resou1',
                                          'run_start': '7db036f1-e793-497b-b169-8d47f65d99f7',
                                          'spec': 'AD_TIFF',
                                          'uid': '9bff1ab9-54ed-4261-a71e-ca66c78d4e87'}}
```

It is probably making a resource, running `def resource` in `tiled_writer.py`, which fills `resource_cache` instead of `stream_resource_cache` and doesn't convert using `_ensure_resource_backcompat`.

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
genematx: > If I'm right, _ensure_resource_backcompat should take Resource | StreamResource -> StreamResource
Exactly! This is what I was gonna suggest too; thanks, @evalott100! Essentially convert all `Resource` and `StreamResource` to the _new_ schema of `StreamResource`. Maybe there's a better name for this function (`_enforce_stream_resource_schema` or `_convert_to_stream_resource`) or just a bit more clarity in the docstring...

> It feels like this should be an or statement
No, I think `and` is okay because we have negations in each sub-clause; you could say `not (("mimetype" in doc.keys()) or ("spec" in doc.keys()))`, which of course is the same.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -75,12 +78,16 @@ def _ensure_resource_backcompat(self, doc: StreamResource) -> StreamResource:
 
         doc = copy.copy(doc)
         if ("mimetype" not in doc.keys()) and ("spec" not in doc.keys()):
-            raise RuntimeError("StreamResource document is missing a mimetype or spec")
+            raise RuntimeError("StreamResource document is missing a 'mimetype' or 'spec'")
         else:
-            doc["mimetype"] = doc.get("mimetype") or MIMETYPE_LOOKUP[doc.get("spec")]
+            doc["mimetype"] = doc.get("mimetype") or MIMETYPE_LOOKUP[str(doc.get("spec"))]
```
evalott100, line 83:
```suggestion
            if "mimetype" not in doc:
                doc["mimetype"] = MIMETYPE_LOOKUP[doc.pop("spec")]
```

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -75,12 +78,16 @@ def _ensure_resource_backcompat(self, doc: StreamResource) -> StreamResource:
 
         doc = copy.copy(doc)
         if ("mimetype" not in doc.keys()) and ("spec" not in doc.keys()):
-            raise RuntimeError("StreamResource document is missing a mimetype or spec")
+            raise RuntimeError("StreamResource document is missing a 'mimetype' or 'spec'")
```
evalott100, line 81: Since this fails if true the else below is superfluous, we can remove it and unindent the contents.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -75,12 +78,16 @@ def _ensure_resource_backcompat(self, doc: StreamResource) -> StreamResource:
...
         else:
-            doc["mimetype"] = doc.get("mimetype") or MIMETYPE_LOOKUP[doc.get("spec")]
+            doc["mimetype"] = doc.get("mimetype") or MIMETYPE_LOOKUP[str(doc.get("spec"))]
         if "parameters" not in doc.keys():
             doc["parameters"] = doc.pop("resource_kwargs", {})
         if "uri" not in doc.keys():
+            if "root" not in doc.keys():
+                raise RuntimeError("StreamResource document is missing a 'root' path")
+            if "resource_path" not in doc.keys():
+                raise RuntimeError("StreamResource document is missing a 'resource_path'")
```
evalott100, line 90: Since this method functions as a converter from `Resource | StreamResource -> StreamResource` (not actually acting on a `StreamResource`) we could probably clean things up by putting these checks at the top of the method. It either contains a `"mimetype"` in which case it's a stream resource and we return the copy as is, or it has to contain `"spec"`, `"root"`, `"uri"`, `"resource_path"`, etc (anything we need for the conversion).

We should also change the name to `_convert_resource_to_stream_resource`.

src/bluesky/consolidators.py
```diff
@@ -96,14 +96,13 @@ def __init__(self, stream_resource: StreamResource, descriptor: EventDescriptor)
...
-        self.dtype = np.dtype(
-            data_desc.get("dtype_numpy")  # standard location
-            or data_desc.get(
-                "dtype_str",  # legacy location
-                # try to guess numpy dtype from JSON type
-                DTYPE_LOOKUP.get(data_desc["dtype"], data_desc["dtype"]),
-            )
+        dtype_spec = (
+            data_desc.get("dtype_numpy")  # Try dtype_numpy first
+            or str(data_desc.get("dtype_str", ""))  # Then try dtype_str
```
evalott100, line 101: **NIT**:

Should work without

```suggestion
            or data_desc.get("dtype_str", "")  # Then try dtype_str
```

src/bluesky/tests/test_seq_num_indices_in_sync.py
```diff
@@ -43,7 +43,7 @@ def __init__(
         self.compose_stream_resource = ComposeStreamResource()
         if DRAFT_0_STREAM_RESOURCE:
             self.stream_resource_compose_datum_pairs = tuple(
-                self.compose_stream_resource("", "", f"non_existent_{det}.hdf5", det, {}) for det in self.detectors
+                self.compose_stream_resource("", f"non_existent_{det}.hdf5", det, {}) for det in self.detectors
```
evalott100, line 46: I think this has to be how it was before since if `DRAFT_0_STREAM_RESOURCE`, `event-model < 1.21.0` and the method takes different args:

https://github.com/bluesky/event-model/blob/cb6896fb5033e0ac64fdddde48fa805d3ab91c90/event_model/__init__.py#L2087-L2091

vs

https://github.com/bluesky/event-model/blob/145118722a2e2f15abf32010e3e6b71506398140/event_model/__init__.py#L2083-L2088

Add an ignore comment instead.

src/bluesky/tests/test_seq_num_indices_in_sync.py
```diff
@@ -124,8 +124,7 @@ def collect_stream_datum_repeat_stream_resource(self) -> Iterator[StreamAsset]:
             # New stream_resource half way through the run
             if DRAFT_0_STREAM_RESOURCE:
                 self.stream_resource_compose_datum_pairs = tuple(
-                    self.compose_stream_resource("", "", f"non_existent_{det}.hdf5", det, {})
-                    for det in self.detectors
+                    self.compose_stream_resource("", f"non_existent_{det}.hdf5", det, {}) for det in self.detectors
```
evalott100, line 127: As above, though arguably since this is a test we can always assume the bundled version of event-model. We should take these out in a different PR if that's the case though.

evalott100 (changes requested):
Since all resources are converted to stream resource, if I understand @genematx correctly, we can probably have only one cache for `StreamResource` there and take out the one for `Resource`.

src/bluesky/callbacks/tiled_writer.py
evalott100: Ah nevermind, it does act on `StreamResource` it has to convert the [older version stream resource](https://github.com/bluesky/event-model/blob/v1.20.0/event_model/documents/stream_resource.py) to the [newer one](https://github.com/bluesky/event-model/blob/v1.22.3/src/event_model/documents/stream_resource.py).

The above suggestion still works though, if it has a `"mimetype"` it is a newer `StreamResource`, if it has a `"spec"` it is an older stream resource, or resource, and needs converting.

thomashopkins32: Moving the runtime checks to the top :+1: 

I realize the `"mimetype"` check should be sufficient to assume that `doc` is a _new_ `StreamResource` but do we want to be explicit that it should have all 3 of `"mimetype"`, `"parameters"`, and `"uri"`?

src/bluesky/tests/test_seq_num_indices_in_sync.py
thomashopkins32: Ah I see that now, thanks!

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
thomashopkins32: Changes made, feel free to review. The definition is now `Resource | StreamResource -> StreamResource`

src/bluesky/callbacks/tiled_writer.py
evalott100: Just checking `"spec"` should be fine, I would be tempted to add an

``` python
assert {"mimetype", "parameters", "uri"} <= doc.keys()
```

but I don't think we need to validate that it is a stream resource > v1.20.0 here. We can assume at this point that we have a stream resource or resource adhering to the schema (which will have different definitions for stream resource depending on event-model version). Adding a check like the above would just leave more to change if the definitions of stream resource are altered.

I think all this needs is the current pops for outdated fields if present. Then putting them all in the same `StreamResource` cache outside the method.

src/bluesky/callbacks/tiled_writer.py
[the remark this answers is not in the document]
evalott100: @genematx I just want to verify, tiled is currently taking `stream_datum/stream_resource` as is, and converting `datum/resource` -> `datum/stream_resource` with a classical datum pointing at a converted `stream_resource`?

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -62,27 +63,37 @@ def __init__(self, client: BaseClient):
         self.root_node: Union[None, Container] = None
         self._desc_nodes: dict[str, Container] = {}  # references to descriptor containers by their uid's
         self._sres_nodes: dict[str, BaseClient] = {}
-        self._docs_cache: dict[str, Union[Datum, Resource, StreamResource]] = {}
+        self._datum_cache: dict[str, Datum] = {}
```
genematx, line 66: Should we call it _stream_datum_cache (by analogy with _stream_resource_cache)? And maybe the type should be `dict[str, StreamDatum]`?

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -97,7 +108,6 @@ def stop(self, doc: RunStop):
         if self.root_node is None:
             raise RuntimeError("RunWriter is properly initialized: no Start document has been recorded.")
 
-        doc = dict(doc)
         metadata = {"stop": doc, **dict(self.root_node.metadata)}
```
genematx, line 111: I'm not 100% sure, but there might have been a requirement on the metadata to be json-serializable, hence we converted `doc=dict(doc)`. Maybe it passes now.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -106,12 +116,13 @@ def descriptor(self, doc: EventDescriptor):
...
-        metadata = dict(doc)
+        # Copy the document items, excluding the variable fields
+        metadata = {k: v for k, v in doc.items() if k not in ("uid", "time", "configuration")}
 
-        # Remove variable fields of the metadata and encapsulate them into sub-dictionaries with uids as the keys
-        uid = metadata.pop("uid")
-        conf_dict = {uid: metadata.pop("configuration", {})}
-        time_dict = {uid: metadata.pop("time")}
+        # Encapsulate variable fields into sub-dictionaries with uids as the keys
+        uid = doc["uid"]
```
genematx, line 123: yeah, probably cleaner without `.pop` 👍

src/bluesky/callbacks/tiled_writer.py
evalott100: The doc is already a dict at runtime, so should be json-serializable (or just as serializable as `dict(doc)` given any extra fields). I was assuming that `doc = dict(doc)` was to create a shallow copy?

genematx: Ah, great, thanks! I don't think we need a shallow copy there.

src/bluesky/callbacks/tiled_writer.py
evalott100: Actually... This is confusing me. `StreamDatum` docs are sent off to the handler right away and not cached in `_datum_cache`, however when an `event` is sent off (linked above) it fails if it the `"data"` field isn't in `_datum_cache`. Does this mean `event`s referencing `stream_datum` are currently unsupported?

genematx: From what I always assumed, `Events` are not meant to reference `StreamDatums`, only usual `Datums`. It might have changed since I looked at this last time a few months ago, but what you said before makes sense, @evalott100.

(1) `StreamDatum` -> directly sent to handler and registered in Tiled. Assumes that the corresponding `StreamResource` has been received before and is either in cache or there's already a node for it in Tiled. If no `StreamDatums` are ever received for a `StreamResource`, no Tiled node is created, and `StreamResource` just dies in the cache. No `Events` are sent/received at all.

(2) 'Datum' -> cached; then when `Event` referencing _external_ data is received, `Datum` -> pulled from cache, processed, and registered.

This is the model I had in my head, but please correct me if I misunderstood something.

genematx (approved):

src/bluesky/callbacks/tiled_writer.py
evalott100: Nah that's correct! I'm forgetting how this works even though I implemented it lol, time to clock out I think :laughing:

mrakitin (commented):
Great effort here! Do we run this test in CI anywhere?
```
$ mypy src
```

evalott100:
> Great effort here! Do we run this test in CI anywhere?

Yep [here](https://github.com/bluesky/bluesky/actions/runs/13060641070/job/36442532970?pr=1873#step:4:29)

jacopoabramo:
> This has turned into a much bigger PR than I anticipated

I'm secretly monitoring it because once this is completed I also wanted to maybe make a PR myself to make `run_engine.py` fully type hinted as well, and maybe some code sections in `utils` as well. Since I see that some changes also affected the run engine as well, it might be worth to expand the PR and take a look into that?

evalott100:
> Since I see that some changes also affected the run engine, it might be worth to expand the PR and take a look into that?

Id be happy to review if you do end up working on type hints across the codebase, but we should probably fix the above type errors in this PR. It shouldn't be too extensive, just a couple of casts.

thomashopkins32:
@evalott100 I was wondering why I wasn't getting any errors anymore. I had installed my local event-model with `pip install -e` and I changed event-model branches so the `py.typed` wasn't there anymore :man_facepalming: 

I can make the changes.

src/bluesky/consolidators.py
thomashopkins32: This `str` is actually necessary because `dtype_str` is not in the `DataKey` schema so it could technically be `Any`.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -87,12 +88,13 @@ def _convert_resource_to_stream_resource(self, doc: Union[Resource, StreamResour
...
         # Convert the Resource (or old StreamResource) document to a StreamResource document
-        doc["mimetype"] = MIMETYPE_LOOKUP[doc.pop("spec")]
-        doc["parameters"] = doc.pop("resource_kwargs", {})
-        file_path = doc.pop("root").strip("/") + "/" + doc.pop("resource_path").strip("/")
-        doc["uri"] = "file://localhost/" + file_path
+        resource_dict = cast(dict, doc)
+        stream_resource_doc["mimetype"] = MIMETYPE_LOOKUP[resource_dict.pop("spec")]
+        stream_resource_doc["parameters"] = resource_dict.pop("resource_kwargs", {})
+        file_path = resource_dict.pop("root").strip("/") + "/" + resource_dict.pop("resource_path").strip("/")
+        stream_resource_doc["uri"] = "file://localhost/" + file_path
```
thomashopkins32, line 95: @evalott100 This was the only way I could get this part to work nicely, let me know if you can think of something better!

evalott100: Nah that's a good way to do it!

evalott100 (approved):
Fab, thanks very much!

Good to squash and merge this and the event-model PR.

PR #1876, opened 2025-01-30
Updated ConsolidatorBase to support frames_per_event as first dim of descriptor shape

thomashopkins32:
## Description

In https://github.com/bluesky/ophyd-async/pull/726, we are changing the `.describe().shape` to always have the number of frames per event (or index) as the first dimension.

This requires changes to how the consolidators interpret the descriptor shape to chunk the data prior to writing it.

### Some context for the purpose of `frames_per_event`
Detectors can collect data at different frequencies. If the maximum frequency of at least one detector is lower than the maximum of the slowest (lowest frequency) detector by a factor >= 2, then we may want to collect that much more data per event for the faster detector(s) compared to the slowest detector.

Collecting multiple `frames_per_event` allows us to do this.

### Context for why we moved this information into the datum shape

Before, this information was separate from the shape which can be misleading. For example, if a detector is collecting 256x256 images, before the shape would be `(256, 256)`. However, this information alone does not tell us how many frames are being collected per event. We would have to track a separate piece of information to get this. After this change in https://github.com/bluesky/ophyd-async/pull/726, the new shape is `(frames_per_event, 256, 256)`.

The slowest detector will typically write a single frame per event: `(1, 256, 256)`.

## How Has This Been Tested?

Testing is underway on this. We will document testing results in this PR when it relates to the changes being made here.

cc. @jwlodek


src/bluesky/consolidators.py
```diff
@@ -141,30 +141,30 @@ def chunks(self) -> tuple[tuple[int, ...], ...]:
...
 
-        # Number of indices needs to be multiplied by the number of frames per index
-        num_chunks = self._num_rows * self.datum_shape[0]
-        # Retain the shape of the signal coming from the device
-        raw_signal_shape = self.datum_shape[1:]
-        shape = (num_chunks, *raw_signal_shape)
-        if len(self.chunk_shape) == 0:
-            return (num_chunks,), *[(d,) for d in raw_signal_shape]
+        # Calculate total shape based on number of rows and datum shape
+        total_shape = self.shape
```
genematx, line 152: we probably don't need another variable here; I'd just use `self.shape`

src/bluesky/consolidators.py
```diff
@@ -88,8 +88,12 @@ def __init__(self, stream_resource: StreamResource, descriptor: EventDescriptor)
 
         # Find data shape and machine dtype; dtype_numpy, dtype_str take precedence if specified
         data_desc = descriptor["data_keys"][self.data_key]
-        self.datum_shape = tuple(data_desc["shape"])
-        self.datum_shape = self.datum_shape if self.datum_shape != (1,) else ()
+        self.datum_shape: tuple[int, ...] = tuple(data_desc["shape"])
```
genematx, line 91:
```suggestion
        self.datum_shape: tuple[int, ...] = tuple(data_desc["shape"]) or (1, )
```
I would not raise an error here, since it would prevent us from ingesting old data that may not have `datum_shape` properly set. I'd suggest just forcing it to be `(1,)`, and if it happens that the file has more dimensions, we can correct the shape later in Tiled once the assets are registered. You may instead opt to raising a warning here, so that the user could fix the ophyd config, but I'm not sure it's necessary.

src/bluesky/consolidators.py
```diff
@@ -136,25 +140,30 @@ def chunks(self) -> tuple[tuple[int, ...], ...]:
...
             # Generate a list with repeated b summing up to A; append the remainder if necessary
             return tuple([b] * (A // b) + ([A % b] if A % b > 0 else [])) or (0,)
 
-        if len(self.chunk_shape) == 0:
-            return (self._num_rows,), *[(d,) for d in self.datum_shape]
-
-        elif len(self.chunk_shape) == 1:
-            return list_summands(self._num_rows, self.chunk_shape[0]), *[(d,) for d in self.datum_shape]
+        # Calculate total shape based on number of rows and datum shape
+        total_shape = self.shape
```
genematx, line 152:
```suggestion
```

src/bluesky/consolidators.py
```diff
@@ -136,25 +140,30 @@ def chunks(self) -> tuple[tuple[int, ...], ...]:
...
 
-        elif len(self.chunk_shape) == len(self.shape):
-            return tuple([list_summands(ddim, cdim) for cdim, ddim in zip(self.chunk_shape, self.shape)])
+        # If chunk shape is less than or equal to the total shape dimensions, chunk each specified dimension
+        # starting from the leading dimension
+        if len(self.chunk_shape) <= len(total_shape):
```
genematx, line 156:
```suggestion
        if len(self.chunk_shape) <= len(self.shape):
```

src/bluesky/consolidators.py
```diff
@@ -136,25 +140,30 @@ def chunks(self) -> tuple[tuple[int, ...], ...]:
...
+                list_summands(ddim, cdim)
+                for ddim, cdim in zip(total_shape[: len(self.chunk_shape)], self.chunk_shape)
+            ) + tuple((d,) for d in total_shape[len(self.chunk_shape) :])
 
+        # If chunk shape is greater than the total shape dimensions, raise an error
         else:
             raise ValueError(
-                f"The shape of chunks, {self.chunk_shape}, is not consistent with the shape of data, {self.shape}."
+                f"The shape of chunks, {self.chunk_shape}, should be less than or equal to the shape of data, "
+                f"{total_shape}."
```
genematx, line 166:
```suggestion
                f"{self.shape}."
```

src/bluesky/consolidators.py
```diff
@@ -136,25 +140,30 @@ def chunks(self) -> tuple[tuple[int, ...], ...]:
...
 
-        elif len(self.chunk_shape) == len(self.shape):
-            return tuple([list_summands(ddim, cdim) for cdim, ddim in zip(self.chunk_shape, self.shape)])
+        # If chunk shape is less than or equal to the total shape dimensions, chunk each specified dimension
+        # starting from the leading dimension
+        if len(self.chunk_shape) <= len(total_shape):
+            return tuple(
+                list_summands(ddim, cdim)
+                for ddim, cdim in zip(total_shape[: len(self.chunk_shape)], self.chunk_shape)
+            ) + tuple((d,) for d in total_shape[len(self.chunk_shape) :])
```
genematx, line 160:
```suggestion
                for ddim, cdim in zip(self.shape[: len(self.chunk_shape)], self.chunk_shape)
            ) + tuple((d,) for d in self.shape[len(self.chunk_shape) :])
```

PR #1878, opened 2025-01-30
Changes CoM behaviour & adds/adjusts unit tests

### LINKED ISSUE #1877: PeakStats CoM Unexpected Value
When doing an asymmetrical scan along data that has a background that is non-zero, the calculated centre of mass is skewed.

We found this issue when scanning a laser which could return a positive to negative range of intensity, we were expecting a centre of mass around the centre of the gaussian but got a value far off. 

## Steps To Reproduce
```
ps = PeakStats("x", "y")
    
ps.start({})
ps.event({"data": {"x": 0, "y": -1}})
ps.event({"data": {"x": 1, "y": 0}})
ps.event({"data": {"x": 2, "y": -1}})
ps.event({"data": {"x": 3, "y": -1}})
ps.stop({})

print(ps["com"])
```

This simulates the documents for a very simple asymmetric scan that has a background of -1, demonstrating this issue. The existing peakstats callback returns a centre of mass of x=1.66... when we would expect a CoM of x=1.

Image

## Acceptance Criteria
- CoM is correctly calculated for data that has a non-zero background
- Unit tests are created that account for these changes

jackbdoughty:
## Description
When calculating CoM, use `y - min(y)` instead of `y`. This is necessary so that the new background of the data is zero but the correlation between data points is maintained. The background must be made zero because otherwise all points of y value less than zero are forced positive in the calculation, meaning that the correlation between data is lost and the CoM is skewed.

Fixes #1877

## How Has This Been Tested?
Tested by running plans with data with non-negative backgrounds and checking that CoM is returned as expected.
Added a parameterised unit test that checks for data with non-negative backgrounds returns correct CoM.
Updated `get_ps` from `test_scientific.py` to reflect the changes made in `PeakStats` of `fitting.py`.
This should not affect other areas of code,

### DIFF
--- a/src/bluesky/callbacks/fitting.py
+++ b/src/bluesky/callbacks/fitting.py
@@ -290,7 +290,7 @@ class PeakStats(CollectThenCompute):
 
         fields["min"] = (x[argmin_y], y_orig[argmin_y])
         fields["max"] = (x[argmax_y], y_orig[argmax_y])
-        (fields["com"],) = np.interp(center_of_mass(y), np.arange(len(x)), x)
+        (fields["com"],) = np.interp(center_of_mass(y - np.min(y)), np.arange(len(x)), x)
         mid = (np.max(y) + np.min(y)) / 2
         crossings = np.where(np.diff((y > mid).astype(int)))[0]
         _cen_list = []
--- a/src/bluesky/tests/test_scientific.py
+++ b/src/bluesky/tests/test_scientific.py
@@ -16,7 +16,7 @@ def get_ps(x, y, shift=0.5):
     x = np.array(x)
     y = np.array(y)
 
-    COM = np.sum(x * y) / np.sum(y)
+    COM = np.sum(x * (y - np.min(y))) / np.sum(y - np.min(y))
     ps["com"] = COM
 
     # from Maksim: assume this is a peak profile:
@@ -134,3 +134,18 @@ def test_peak_statistics_with_derivatives(RE):
     assert len(ps.derivative_stats.x) == num_points - 1
     assert len(ps.derivative_stats.y) == num_points - 1
     assert np.allclose(np.diff(ps.y_data), ps.derivative_stats.y, atol=1e-10)
+
+@pytest.mark.parametrize("com,background", [(1, -1), (1, 1)])
+def test_peak_statistics_with_non_zero_background_and_non_symmetry_scan(RE, com, background):
+    """peak statistics calculation on a gaussian function with a non zero background and a non symmetrical scan"""
+
+    ps = PeakStats("x", "y")
+    
+    ps.start({})
+    ps.event({"data": {"x": 0, "y": background}})
+    ps.event({"data": {"x": 1, "y": background+1}})
+    ps.event({"data": {"x": 2, "y": background}})
+    ps.event({"data": {"x": 3, "y": background}})
+    ps.stop({})
+
+    assert ps["com"] == com


mrakitin (commented):
Great contribution, @jackbdoughty! Do you mind fixing the linter issues relevant to formatting?

mrakitin:
Also, docs building issue seems to be new:
```py
/home/runner/work/bluesky/bluesky/src/bluesky/callbacks/fitting.py:171: RuntimeWarning: invalid value encountered in scalar divide
  results = [np.sum(input * grids[dir].astype(float), labels, index) / normalizer for dir in range(input.ndim)]
```

jackbdoughty:
> Also, docs building issue seems to be new:
> 
> ```python
> /home/runner/work/bluesky/bluesky/src/bluesky/callbacks/fitting.py:171: RuntimeWarning: invalid value encountered in scalar divide
>   results = [np.sum(input * grids[dir].astype(float), labels, index) / normalizer for dir in range(input.ndim)]
> ```

Thanks for this @mrakitin . I believe this is an already existing problem which relates to [this PR](https://github.com/bluesky/bluesky/pull/1792). What I've done to avoid this build error is just change in the docs which devices/dets it uses in a couple example scans so that the data gathered isn't flat. I think that when the above PR is merged it will fix this problem but to avoid having this PR rely on the other, this might be the diplomatic thing to do. Hope this is OK, cheers.

src/bluesky/callbacks/fitting.py
```diff
@@ -290,7 +290,7 @@ def _calc_stats(x, y, fields, edge_count=None):
 
         fields["min"] = (x[argmin_y], y_orig[argmin_y])
         fields["max"] = (x[argmax_y], y_orig[argmax_y])
-        (fields["com"],) = np.interp(center_of_mass(y), np.arange(len(x)), x)
+        (fields["com"],) = np.interp(center_of_mass(y - np.min(y)), np.arange(len(x)), x)
```
prjemian, line 293: This change decreases the generality.  It imposes an assumption that `y` data is positive.  

PeakStats cannot assume the units of `y` (which includes the sign).  PeakStats must be usable with scaler counts, photocurrents, and any other `float` or `int` value.

prjemian: The challenge to performing any background removal is to identify which background model is to be applied.  This change assumes the model of a positive peak with an additive, constant offset (most common case, encountered when using a scaler).

Other models might consider positive peak against a sloped background, negative peak versions of these two, or even `y` data that has comparably strong positive and negative values.

For PeakStats to be general, it cannot perform background removal assuming only one of these models, even if that model is common.

prjemian:
If accuracy is COM determination is important, a subsequent `rel_scan` with range balanced around the peak is recommended.  COM determination will always be skewed by the presence of more data on one side of the peak.

Tom-Willemsen:
Thanks for the comments. Based on those comments, and with @jackbdoughty 's permission, I've updated his PR with a refactor which:
- Leaves the existing implementation of `ps["com"]` untouched
- Adds a new argument, `com_method`, to `PeakStats` - which allows choosing an alternative com implementation, which might use a different interpretation of what "com" should conceptually mean. Defaults to the existing implementation.
- Adds a `com_method`, "area_under_curve", hopefully unambiguously-enough named to make it clear that it's always calculating the COM of an area *under* a curve, i.e. a positive peak. 
  * For the negative-peak case, I think a negating `LiveDispatcher` is sufficient, for those who may want that.
  * Just for clarity, it's trying to compute the com of the 2d shape bounded by `min(y)`, `min(x)`, `max(x)`, and straight-line segments joining `(x, y)` points to their nearest-neighbours along x.
  * Hopefully the naming, along with leaving the existing implementation in place and used by default, is sufficient to ease the above concerns about generality for existing users - since anyone who doesn't _explicitly_ choose to use this new option won't see any change.
  * New option also addresses some other cases we also had in our data, e.g. x data that isn't constantly-spaced & monotonically increasing - as happens from adaptive/continuous scans etc. But as above, leaves existing implementation exactly as-is and used by default. We were originally going to make a follow-up PR for these other cases - but if we're adding an entirely different com method anyway, then makes more sense to make all the changes together.
- Adds a number of unit tests with minimal data sets which try to show:
  * Where the two implementations give the same answer
  * Where the two implementations differ from each other (perhaps due to different interpretations/assumptions/implementation choices about what "center of mass" should mean)
  * Hopefully this helps future plan authors/callers select the implementation which makes most sense for their actual data

---

> If accuracy is COM determination is important, a subsequent rel_scan with range balanced around the peak is recommended

<summary> context

<details>
Thanks, we are already well aware of this. For reference the data that caused us to write this issue/pr looked like this:

image

```
ps[cen]: 498.6816161343001
ps[com]: 507.4936482348703
ps[area_under_curve_com]: 498.6772505554155
lmfit trapezoid <Parameter 'cen', value=498.6626993320932 +/- 0.00264, bounds=[-inf:inf]>
```

Total scan range `487.64 - 509.64` - i.e. very close to centered about the peak, depending which centering method is to be believed most.

And just to pre-empt the obvious question: we were testing COM on this data hoping to "prove it worked" & gave an alignment number comparable to our old scanning software; for _this specific_ data, either a trapezoid fit or `ps[cen]` do give usable results, but other data we'll need to deal with in future is not quite as friendly as this.

</details>
</summary>

PR #1881, opened 2025-02-07
Add doc references that ophyd-async depends on

coretl:


stan-dot (approved):
LGTM but it seems it's in correspondence to some Event-Model changes which I am not aware of what happened there

PR #1883, opened 2025-02-14
[MNT] Remove non 3.9 dependencies and adjust some typing mechanisms

jacopoabramo:
## Description
Drop dependencies requiring `Python < 3.9`, use `TypeIs` in `utils.maybe_await` and use `sys.version_info` to import `ParamSpec` in `protocols`.

## Motivation and Context
- Since Bluesky supports `Python >= 3.9`, I removed some dependencies from `pyproject.toml` required for lower versions.
- [PEP 742](https://peps.python.org/pep-0742/) introduced `TypeIs` to narrow types and addresses the [issue](https://github.com/python/mypy/issues/15520) originally mentioned in `utils.maybe_await`. For `Python < 3.13`, this is backported in `typing_extensions >= 4.10.0`, which is bumped in `pyproject.toml`.
- Rather than using `try-except` to import `ParamSpec`, I used `sys.version_info` which is suggested in [this guide](https://learn.scientific-python.org/development/patterns/backports/#conditional-usage) from the Scientific Python community.

## How Has This Been Tested?
I ran

```
mypy src\bluesky
```

locally on my machine.

Note that it also reported the following errors in the `tracing` module:

```
src\bluesky\tracing.py:12: error: The first argument to Callable must be a list of types, parameter specification, or "..."  [valid-type]
src\bluesky\tracing.py:12: note: See https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas
src\bluesky\tracing.py:15: error: The first argument to Callable must be a list of types, parameter specification, or "..."  [valid-type]
src\bluesky\tracing.py:15: note: See https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas
src\bluesky\tracing.py:21: error: The first argument to Callable must be a list of types, parameter specification, or "..."  [valid-type]
src\bluesky\tracing.py:21: note: See https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas
```

### DIFF
--- a/src/bluesky/protocols.py
+++ b/src/bluesky/protocols.py
@@ -1,3 +1,4 @@
+import sys
 from abc import abstractmethod
 from collections.abc import AsyncIterator, Awaitable, Iterator
 from typing import (
@@ -24,10 +25,10 @@ from typing_extensions import TypedDict, Unpack
 # Squashes warning
 Dtype = Dtype  # type: ignore
 
-try:
+if sys.version_info >= (3, 10):
     from typing import ParamSpec
-except ImportError:
-    from typing_extensions import ParamSpec  # type: ignore
+else:
+    from typing_extensions import ParamSpec
 
 
 # TODO: these are not placed in Events by RE yet
--- a/src/bluesky/protocols.py
+++ b/src/bluesky/protocols.py
@@ -1,4 +1,3 @@
-import sys
 from abc import abstractmethod
 from collections.abc import AsyncIterator, Awaitable, Iterator
 from typing import (
@@ -20,15 +19,10 @@ from event_model.documents.event import PartialEvent
 from event_model.documents.event_descriptor import DataKey, Dtype
 from event_model.documents.event_page import PartialEventPage
 from event_model.documents.resource import PartialResource
-from typing_extensions import TypedDict, Unpack
+from typing_extensions import ParamSpec, TypedDict, Unpack
 
 # Squashes warning
 Dtype = Dtype  # type: ignore
-
-if sys.version_info >= (3, 10):
-    from typing import ParamSpec
-else:
-    from typing_extensions import ParamSpec
 
 
 # TODO: these are not placed in Events by RE yet
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -1927,8 +1927,10 @@ async def maybe_collect_asset_docs(
         async for doc in iterate_maybe_async(obj.collect_asset_docs(*args, **kwargs)):
             yield doc
 
+
 def _isawaitable(value: SyncOrAsync[T]) -> TypeIs[Awaitable[T]]:
     return inspect.isawaitable(value)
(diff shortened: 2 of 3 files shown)


jacopoabramo:
@evalott100 could you review this?

src/bluesky/protocols.py
```diff
@@ -24,10 +25,10 @@
 # Squashes warning
 Dtype = Dtype  # type: ignore
 
-try:
+if sys.version_info >= (3, 10):
     from typing import ParamSpec
-except ImportError:
-    from typing_extensions import ParamSpec  # type: ignore
+else:
+    from typing_extensions import ParamSpec
```
evalott100, line 31: It seems that ParamSpec is also available in typing_extensions in later python versions (I tested 3.11). Can't we just import it from there for all versions?

src/bluesky/utils/__init__.py
```diff
@@ -1926,14 +1927,14 @@ async def maybe_collect_asset_docs(
         async for doc in iterate_maybe_async(obj.collect_asset_docs(*args, **kwargs)):
             yield doc
 
+def _isawaitable(value: SyncOrAsync[T]) -> TypeIs[Awaitable[T]]:
+    return inspect.isawaitable(value)
```
evalott100, line 1932: Add empty lines here to make `tox -e pre-commit` happy (it will do it itself if you run it locally).

evalott100 (changes requested):
One thing to note, 3.8 is still required on v1.13, only the prereleases have deprecated 3.8.

src/bluesky/protocols.py
jacopoabramo: I think it makes sense to keep the conditional to avoid installing typing_extensions when not necessary. But I made a mistake since I removed the upper bound. Not sure what the upper bound would be now

To be completely exhaustive also `TypeIs` should be wrapped in an `if sys.version < 3.13`, but at the moment Bluesky doesn't support that version.

src/bluesky/utils/__init__.py
jacopoabramo: Done, let's see if ruff is happy now

src/bluesky/protocols.py
evalott100: > I think it makes sense to keep the conditional to avoid installing typing_extensions when not necessary.

But `typing-extensions>=4.10.0` is a production requirement anyway, we're also importing from it for the other python versions [here](https://github.com/jacopoabramo/bluesky/blob/6f51ce3f4bb3c17fb7a8d06b49f41ae8bce3b9c1/src/bluesky/protocols.py#L23)?

> but at the moment Bluesky doesn't support that version.

Yup, no need for that now.

jacopoabramo: Fair enough, marking as resolved with last commit.

evalott100 (approved):
Thanks!

PR #1887, opened 2025-02-19
Remove debug(?) print statements from `plot_peak_stats`

Tom-Willemsen:
## Description

Removes some debug-style prints from `plot_peak_stats` - I can't see that these would be useful for a typical end-user?

Example of what it would have printed:

```
{'color': 'k'} cen
{'color': 'r'} com
{'color': 'b'} max
{'color': 'm'} min
```

## Motivation and Context

Cleaning up the output for an interactive session, after `PeakStats` have been plotted using `plot_peak_stats`

## How Has This Been Tested?

Tested interactively. Unit test not added for a `print` statement deletion.

### DIFF
--- a/src/bluesky/callbacks/mpl_plotting.py
+++ b/src/bluesky/callbacks/mpl_plotting.py
@@ -721,14 +721,12 @@ def plot_peak_stats(peak_stats, ax=None):
     vlines = []
     styles = iter(cycler("color", "krgbm"))
     for style, attr in zip(styles, ["cen", "com"]):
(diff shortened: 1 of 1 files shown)


danielballan:
These prints were introduced by [my commit](https://github.com/bluesky/bluesky/commit/a36eb96414260193ecded4b4591c9aa05a5bef20) so I feel safe claiming that they were probably a mistake. Thanks @Tom-Willemsen, and sorry for the slow review!

danielballan (approved):

PR #1888, opened 2025-02-23
API: add `configure` to `Configurable`

### LINKED ISSUE #1862: Question about `Configurable` protocol
I was looking over the `Configurable` protocol and compared it to the current Bluesky's documentation. I noticed that `Configurable` only provides the `read_configuration` and `describe_configuration` methods. But the message protocol documentation also reports a [`configure`](https://blueskyproject.io/bluesky/main/msg.html#configure) which I imagine it's the equivalent of the `set` method but for slow-changing parameters. Also in the `plan_stubs` the `configure` message expects a `Configurable` object as an input. Is this a missing method somehow or is it intentional?

jacopoabramo:
## Description
- Add the `configure` method to `Configurable` protocol
- Remove the generic `T` from `Configurable` definition
- `read_configuration` and `describe_configuration` now return `Any` instead of `T`

Closes #1862 

## Motivation and Context
Although documented as a [message](https://blueskyproject.io/bluesky/main/msg.html#configure) and a [plan stub](https://blueskyproject.io/bluesky/main/generated/bluesky.plan_stubs.configure.html#bluesky.plan_stubs.configure), there is no  proper definition in the `Configurable` protocol for said method. This PR introduces it as part of the protocol for better clarity and enhanced type hinting.

Furthermore, as configuration parameters are not tied to a specific type, `Configurable` now is a non-generic protocol and all the methods return type `Any`.

## How Has This Been Tested?

Tested locally via

```
pytest
```

Test reported a couple of failures. I wanted to run them on the CI in order to attempt to resolve them since they're related to the `test_flyer` module.

### DIFF
--- a/src/bluesky/protocols.py
+++ b/src/bluesky/protocols.py
@@ -188,9 +188,21 @@ class WritesStreamAssets(Protocol):
 
 
 @runtime_checkable
-class Configurable(Protocol[T]):
+class Configurable(Protocol):
     @abstractmethod
-    def read_configuration(self) -> SyncOrAsync[dict[str, Reading[T]]]:
+    def configure(self, *args: Any, **kwargs: Any) -> SyncOrAsync[tuple[Reading[Any], Reading[Any]]]:
+        """Configure an object.
+
+        The method returns a tuple of ``Reading``.
+        The first element is the old configuration value
+        and the second element is the new configuration value.
+
+        This can be a standard function or an ``async`` function.
+        """
+        ...
+
+    @abstractmethod
+    def read_configuration(self) -> SyncOrAsync[dict[str, Reading[Any]]]:
         """Same API as ``read`` but for slow-changing fields related to configuration.
         e.g., exposure time. These will typically be read only once per run.
 


jacopoabramo:
It appears that this change causes a failure alltogether. I'm not exactly sure why though. I'm leaving this PR open in case anyone has suggestions and/or reasons as for why the protocol shouldn't be changed - there might be something I'm unfamiliar with.

src/bluesky/protocols.py
```diff
@@ -188,9 +188,21 @@ def get_index(self) -> SyncOrAsync[int]:
 
 
 @runtime_checkable
-class Configurable(Protocol[T]):
+class Configurable(Protocol):
```
tacaswell, line 191: this drops the generics which I suspect causes a run-time typecheck to fail and not call the right methods.

coretl:
Suggestion is use `set()` or run a plan instead of the `configure()` method that has side effects in the run bundler of re-emitting event descriptors

Issue #1884, opened 2025-02-17
Flakey test on windows


evalott100:
On the [test run](https://github.com/bluesky/bluesky/actions/runs/13308159093/job/37163837737), the following test failed on windows with python 3.9:

```
 =========================== short test summary info ===========================
FAILED src/bluesky/tests/test_utils.py::test_CallbackRegistry_1[class_method-True-True] - AssertionError: Incorrect number of signals
assert 2 == 3
 +  where 2 = len({'sig2': <WeakKeyDictionary at 0x1d9db084cd0>, 'sig3': <WeakKeyDictionary at 0x1d9db084e50>})
 +    where {'sig2': <WeakKeyDictionary at 0x1d9db084cd0>, 'sig3': <WeakKeyDictionary at 0x1d9db084e50>} = <bluesky.utils.CallbackRegistry object at 0x000001D9DB257BE0>._func_cid_map
 +  and   3 = len({'sig1': 5, 'sig2': 4, 'sig3': 3})
= 1 failed, 1745 passed, 41 skipped, 18 xfailed, 16 xpassed in 1018.24s (0:16:58) =
Exception ignored in: <function BaseEventLoop.__del__ at 0x000001D99C0F28B0>
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\asyncio\base_events.py", line 686, in __del__
ResourceWarning: unclosed event loop <ProactorEventLoop running=False closed=False debug=False>
```

jacopoabramo:
I lazily tried running the tests locally.

- OS: Windows 11
- Python: 3.9.21

I ran `pytest` and `tox -e tests` (you never know), but couldn't reproduce it locally - all tests are passing. Is there any other information available about this?

Tom-Willemsen:
This commit https://github.com/bluesky/bluesky/commit/d49b7419d66d3437152e1ec713ecfa82e3bfe433 demonstrates the failure consistently, on both windows & linux (by forcing a gc collection at the relevant time). I'm not sure I understand enough about the use-cases of the `BoundMethodProxy` to know whether it's the test or implementation that needs adjusting though.

evalott100:
If I had to guess it's an issue with the `weakref.ref` in `_BoundMethodProxy` then 🤔

PR #1889, opened 2025-02-24
[ENH] `bps.locate` typing

jacopoabramo:
## Description
Fix type hints for `bps.locate`.

## Motivation and Context
Originally it returned `Location | list[Location]`, when it should be returning `MsgGenerator[Location] | MsgGenerator[list[Location]]`. The return type is hinted `Location[T]` (`T` being imported from `bluesky.protocols`). The overloaded signatures have been swapped to make mypy happy.

## How Has This Been Tested?

Ran locally

```
mypy .
```

And then

```
pytest
```

to ensure no regression.

### DIFF
--- a/src/bluesky/plan_stubs.py
+++ b/src/bluesky/plan_stubs.py
@@ -34,6 +34,7 @@ from .protocols import (
     Stageable,
     Status,
     Stoppable,
+    T,
     Triggerable,
     check_supports,
 )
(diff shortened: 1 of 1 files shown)


jacopoabramo:
@evalott100 could you confirm that the failing test here is due to #1884 ? It seems like it.

evalott100:
>  It seems like it.

Yup, I can. It's interesting this is on 3.9 only :thinking:

PR #1898, opened 2025-04-09
Include ophyd_async version in start doc version metadata

jwlodek:
## Motivation and Context

As we start using ophyd async more in prod, it would be good to include its version in the start doc metadata.

## How Has This Been Tested?

Ran scan with change:

```
========= Emitting Doc =============
name = 'start'
{
    "uid": "3f79be77-edb9-4410-90bf-c24b0a9a3674",
    "time": 1744208682.0387623,
    "versions": {
        "ophyd": "1.10.0",
        "ophyd_async": "0.8.0a7.dev54+g46be3bf",
        "bluesky": "1.13.2.dev44+g0ed50c97"
    },
    "scan_id": 1,
    "plan_type": "generator",
    "plan_name": "software_flyscan"
}
============ Done ============
```

### DIFF
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -466,6 +466,13 @@ class RunEngine:
         except ImportError:
             self.log.debug("Failed to import ophyd.")
 
+        try:
+            import ophyd_async
+
+            self.md["versions"]["ophyd_async"] = ophyd_async.__version__
+        except ImportError:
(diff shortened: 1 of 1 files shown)


prjemian (commented):
This makes ophyd_async a required package for bluesky? Is it necessary to raise an exception?

jwlodek:
I think like ophyd we'd want to make ophyd-async only a dev dependency at most, so we need the out if the package is not installed

PR #1902, opened 2025-05-06
FIX: regenerate the configuration cache before the descriptors

tacaswell:
Doing it in the opposite order results in the old configuration value going in the new descriptor.

## Description

I broke this in f1b3179329f7af91acc3846b805a47f3403308f0 .  It would have been caught by tests in databroker (which the new test is a modified version of), but it was xfailed due to other the temorary too-strict rules for pre-declare and it was never un-xfailed.

## Motivation and Context

Found a bluesky regression while working on tiled/databroker support

## How Has This Been Tested?

Added a test adapted from databroker tests.

### DIFF
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -1172,6 +1172,7 @@ class RunBundler:
             object.configure(*args, **kwargs)
         """
         obj = msg.obj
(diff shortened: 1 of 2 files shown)


tacaswell:
This does make a compelling argument that this feature is not actually used.

PR #1904, opened 2025-05-07
Mongo Migration Follow-Up

genematx:
A follow-up PR on Mongo-SQL migration.

- Drop metadata revisions when updating in `TiledWriter`


danielballan (approved):

danielballan:
Needs a Tiled release before CI will pass.

danielballan (approved):

danielballan:
Patch coverage was at 86% because an error condition isn't covered by tests. Might be worth follow up but I won't bump this from the release over it.

PR #1905, opened 2025-05-08
Add `timeout` argument to `mv` and `mvr`

flowln:
This change is intended a shorthand for the following operation (and for `rel_set` / `mvr`):

```python
yield from bps.abs_set(obj, X, group=Y)
yield from bps.wait(group=Y, timeout=Z)
```

, so we can instead simply write

```python
yield from bps.mv(obj, X, timeout=Z)
```

## Motivation and Context
I've come across many cases like this, where I'm moving regular signals (i.e not Positioners or anything of that sort), but the IOC or the hardware sometimes gets stuck, usually during some pre-scan configuration. In these cases, having a timeout is useful to give the user a meaningful indication of what has failed.

### DIFF
--- a/src/bluesky/tests/test_new_examples.py
+++ b/src/bluesky/tests/test_new_examples.py
@@ -113,16 +113,20 @@ from bluesky.utils import IllegalMessageSequence, all_safe_rewind
             mv,
             ("motor1", 1, "motor2", 2),
             {"group": "A"},
-            [Msg("set", "motor1", 1, group="A"), Msg("set", "motor2", 2, group="A"), Msg("wait", None, group="A")],
+            [
+                Msg("set", "motor1", 1, group="A"),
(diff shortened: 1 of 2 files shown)


src/bluesky/tests/test_new_examples.py
```diff
@@ -285,7 +304,8 @@ def test_mvr_with_timeout(hw):
     actual = list(mvr(hw.motor1, 1, hw.motor2, 2, timeout=42))
     for msg in actual[:2]:
         msg.command == "set"  # noqa: B015
```
tacaswell, line 306: Might as well fix this test while here!

```suggestion
        assert msg.command == "set"  # noqa: B015
```

tacaswell (approved):

tacaswell:
Thank you @flowln !

PR #1908, opened 2025-05-19
Configure Access Policy for Tiled Nodes

genematx:
This uses Tiled `access_tags` to configure access policy to nodes created with TiledWriter.


danielballan:
Oh, I just understood your suggestion with `Container.new`. That does seem more ergonomic.

danielballan:
@nmaytan and I discussed this for ten minutes. We convinced ourselves that propagating access tags by default raises some subtle trade-offs. We'd like to time more time with it.

danielballan (approved):

PR #1911, opened 2025-05-23
Refactoring TiledWriter

genematx:
Some refactoring changes to TiledWriter.

## Description
Split TiledWriter into two callbacks, `_RunNormalizer` and `_RunWriter` with the former intended to convert any encountered Bluesky documents that follow legacy schema conventions to the most recent `EventModel` schema. `_RunWriter` is subscribed to the `_RunNormalizer` in chain and can only accept documents that adhere to the new schemas. Furthermore, any non-stream documents are converted to `StreamResource` and `StreamDatums`. `_RunNormalizer` creates deep copies of the documents to prevent modifying them for other callbacks subscribed to the same RE.

Furthermore, a backup capability has been added to TiledWriter: if writing to Tiled fails at any point, the entire BlueskyRun (all previous and all future documents) would be written to a local JSONLines file.

## Motivation and Context
The introduced separation of functionality between schema conversion and document writing would help apply any beamline-specific patches and transforms, if needed, before ingesting the documents to Tiled.

## How Has This Been Tested?
* Introduced a new test file, `external_assets_legacy.json`, containing a detailed example of legacy external asset documents to validate backward compatibility.
* Added a new test, `test_bad_document_order`, to ensure `TiledWriter` can handle documents emitted in an unexpected order, such as `datum` documents appearing late.

### DIFF
--- a/src/bluesky/tests/test_consolidators.py
+++ b/src/bluesky/tests/test_consolidators.py
@@ -430,6 +430,7 @@ def test_tiff_and_jpeg_chunks(
     assert len(cons.assets) == 5 * frames_per_datum / expected_chunks[0][0] if join_method == "concat" else 5
 
 
+# Tuples of (filename, original_template, expected_template, formatted)
 template_testdata = [
     ("", "img_{:06d}", "img_{:06d}", "img_000042"),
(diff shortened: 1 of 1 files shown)


genematx:
Renamed the branch.

PR #1914, opened 2025-05-28
Threaded Buffer to Allow Non-Blocking Callbacks

genematx:
Add buffering capability to Bluesky callbacks

## Description
An implementation of a simple wrapper for general bluesky callbacks that allows the downstream processing to happen in a separate thread without blocking the `RunEngine`. This could be useful for slow I/O operations (e.g. `TiledWriter`).

Intended usage:
```
tw = TiltedWriter(client)
buff_tw = BufferingWrapper(tw)
RE.subscribe(buff_tw)
```

Note that the `BufferingWrapper` must be used with great caution: the wrapped callback should be thread-safe and must not be subscribed to the RE directly. The errors encountered in the wrapped callback will not be propagated to the RE, and hence there is a possibility for data loss.

## Motivation and Context
This is mainly a development feature to allow subscribing (potentially many) experimental callbacks to a `RunEngine` without the risk of blocking the experiment. The use in production is not encouraged; for critical workflows, it is generally recommended to have a backup (blocking) writing callback.

## How Has This Been Tested?
* `src/bluesky/tests/test_buffering.py`: Added tests for `BufferingWrapper` covering:
  - Delegation of calls to the wrapped callback.
  - Graceful shutdown, ensuring no new items are processed post-shutdown.
  - Proper handling of signal-triggered shutdowns.
  - Validation of shutdown behavior with and without waiting.
  - Thread safety and robustness against double shutdowns

### DIFF
--- a/src/bluesky/callbacks/buffer.py
+++ b/src/bluesky/callbacks/buffer.py
@@ -2,6 +2,7 @@ import atexit
 import logging
 import threading
 from queue import Empty, Queue
+from typing import Callable
 
 logger = logging.getLogger(__name__)
 
@@ -38,7 +39,7 @@ class BufferingWrapper:
         RE.subscribe(buff_tw)
     """
 
-    def __init__(self, target, queue_size=1_000_000):
+    def __init__(self, target: Callable, queue_size: int = 1_000_000):
         self._wrapped_callback = target
         self._queue = Queue(maxsize=queue_size)
         self._stop_event = threading.Event()
@@ -72,7 +73,7 @@ class BufferingWrapper:
             except Exception as e:
                 logger.exception(f"Exception in {self._wrapped_callback.__class__.__name__}: {e}")
 
-    def shutdown(self, wait=True):
+    def shutdown(self, wait: bool = True):
         if self._stop_event.is_set():
             return
         self._stop_event.set()
--- a/src/bluesky/callbacks/buffer.py
+++ b/src/bluesky/callbacks/buffer.py
@@ -44,9 +44,7 @@ class BufferingWrapper:
         self._thread = threading.Thread(target=self._process_queue, daemon=True)
         self._thread.start()
 
-        atexit.register(self.shutdown)
-        signal.signal(signal.SIGINT, self._signal_handler)
-        signal.signal(signal.SIGTERM, self._signal_handler)
+        self._register_handlers()
 
     def __call__(self, name, doc):
         with self._shutdown_lock:
@@ -73,6 +71,8 @@ class BufferingWrapper:
             self._stop_event.set()
             self._queue.put(None)
 
+            self._unregister_handlers()
+
         if wait:
             self._thread.join()
         print(f"{self._wrapped_callback.__class__.__name__} shut down gracefully.")
@@ -81,3 +81,26 @@ class BufferingWrapper:
         print(f"Signal {signum} received. Shutting down {self._wrapped_callback.__class__.__name__}...")
         self.shutdown()
         raise SystemExit(0)
+
+    def _register_handlers(self):
+        if threading.current_thread() is threading.main_thread():
+            try:
+                atexit.register(self.shutdown)
+                signal.signal(signal.SIGINT, self._signal_handler)
+                signal.signal(signal.SIGTERM, self._signal_handler)
+            except Exception as e:
+                print(f"Failed to register signal handlers: {e}")
+
+    def _unregister_handlers(self):
+        try:
+            atexit.unregister(self.shutdown)
+        except Exception:
+            pass
+
+        try:
+            if signal.getsignal(signal.SIGINT) == self._signal_handler:
+                signal.signal(signal.SIGINT, signal.SIG_DFL)
+            if signal.getsignal(signal.SIGTERM) == self._signal_handler:
+                signal.signal(signal.SIGTERM, signal.SIG_DFL)
+        except Exception:
+            pass
--- a/src/bluesky/callbacks/buffer.py
+++ b/src/bluesky/callbacks/buffer.py
@@ -1,5 +1,4 @@
 import atexit
-import signal
 import threading
 from queue import Empty, Queue
 
@@ -86,8 +85,8 @@ class BufferingWrapper:
         if threading.current_thread() is threading.main_thread():
             try:
                 atexit.register(self.shutdown)
-                signal.signal(signal.SIGINT, self._signal_handler)
-                signal.signal(signal.SIGTERM, self._signal_handler)
+                # signal.signal(signal.SIGINT, self._signal_handler)
+                # signal.signal(signal.SIGTERM, self._signal_handler)
             except Exception as e:
                 print(f"Failed to register signal handlers: {e}")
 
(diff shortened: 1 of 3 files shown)


src/bluesky/callbacks/buffer.py
```diff
@@ -0,0 +1,106 @@
...
+    Usage
+    -----
+        tw = TiltedWriter(client)
+        buff_tw = BufferingWrapper(tw)
+        RE.subscribe(buff_tw)
+    """
+
+    def __init__(self, target):
+        self._wrapped_callback = target
+        self._queue = Queue()
```
danielballan, line 39: As the trio documentation teaches us, all queues have a size limit, whether you declare one or not. Better to get limited at the application level—and handle it—than to get OOM-ed by the operating system.

src/bluesky/callbacks/buffer.py
```diff
@@ -0,0 +1,106 @@
...
+            try:
+                if item := self._queue.get(timeout=1):
+                    self._wrapped_callback(*item)  # Delegate to wrapped callback
+                else:
+                    break  # Received sentinel value to stop processing
+            except Empty:
+                if self._stop_event.is_set():
+                    break
+
+    def shutdown(self, wait=True):
```
danielballan, line 66: It's interesting that [upstream has a solution for this now](https://docs.python.org/3/library/queue.html#queue.Queue.shutdown) but we'll have to wait for Python 3.13 to be our minimum support version to use it.

src/bluesky/callbacks/buffer.py
```diff
@@ -0,0 +1,106 @@
...
+            self._unregister_handlers()
+
+        if wait:
+            self._thread.join()
+        print(f"{self._wrapped_callback.__class__.__name__} shut down gracefully.")
+
+    def _signal_handler(self, signum, frame):
+        print(f"Signal {signum} received. Shutting down {self._wrapped_callback.__class__.__name__}...")
+        self.shutdown()
+        raise SystemExit(0)
```
danielballan, line 83: To compose well with any other signal handling in the application, we can stash the original signal handlers (the ones that `_register_handlers` overrides) and then fire _those_ here.

src/bluesky/callbacks/buffer.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def _register_handlers(self):
+        if threading.current_thread() is threading.main_thread():
+            try:
+                atexit.register(self.shutdown)
+                signal.signal(signal.SIGINT, self._signal_handler)
+                signal.signal(signal.SIGTERM, self._signal_handler)
+            except Exception as e:
```
danielballan, line 91: Under what conditions would this fail?

src/bluesky/callbacks/buffer.py
```diff
@@ -0,0 +1,106 @@
...
+                print(f"Failed to register signal handlers: {e}")
+
+    def _unregister_handlers(self):
+        try:
+            atexit.unregister(self.shutdown)
```
danielballan, line 96: Same question as above. I _think_ this is safe, and there's no need to introduce unreachable lines uncovered by tests. (Happy to learn I'm wrong, though.)

src/bluesky/callbacks/buffer.py
```diff
@@ -0,0 +1,106 @@
...
+        except Exception:
+            pass
+
+        try:
+            if signal.getsignal(signal.SIGINT) == self._signal_handler:
+                signal.signal(signal.SIGINT, signal.SIG_DFL)
```
danielballan, line 102: Better to stash whatever was registered before. It could be something custom. (The RunEngine itself registers custom handlers, so this is not a hypothetical.)

danielballan (commented):
Some "fit and finish" suggestions. I think the model is great; I like that it composes with _any_ callback, not just `TiledWriter`. This would have been useful in several situations I can think of. :+1:

src/bluesky/callbacks/buffer.py
genematx: done!

src/bluesky/callbacks/buffer.py
genematx: I am still very unsure about handling exits. On the one hand, we want to make sure that if bsui is exited (or RE crashes), whatever is left in the queue would still be processed. On the other hand, when RE is stopped with Ctrl-C, for example, we don't want to shutdown the BufferedWrapper, the thread, and its wrapped callback.

src/bluesky/callbacks/buffer.py
genematx: Yeah, try-except here is probably not necessary -- was an extra precaution to make sure the RE is not blocked. Will remove them.

src/bluesky/callbacks/buffer.py
genematx: Removed.

src/bluesky/callbacks/buffer.py
genematx: Not needed anymore; SIGINT should not terminate the callback, unless the entire interpreter exits.

src/bluesky/callbacks/buffer.py
genematx: Made a comment in the code.

src/bluesky/callbacks/buffer.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __call__(self, name, doc):
+        if self._stop_event.is_set():
+            raise RuntimeError("Cannot accept new data after shutdown.")
+            # TODO: This can be refactored using the upstream functionality (in Python >= 3.13)
+            # https://docs.python.org/3/library/queue.html#queue.Queue.shutdown
+        try:
+            self._queue.put((name, doc))
+        except Full as e:
+            logger.exception(
```
genematx, line 61: Do we need to log this error here, or is the assumption that it would be captured and logged later by the RE itself?

danielballan: I guess the plan now is to raise?

genematx: Yep, I raise it below. Just not sure if we need to log it here as well or not. (I think no, but Garrett was concerned about missing log messages, so just want to doublecheck)

src/bluesky/callbacks/buffer.py
genematx: refactored.

src/bluesky/callbacks/buffer.py
jacopoabramo: Sorry for the intrusion; would it be possible to add a logger input to `__init__` in case an user would want to provide a custom logger instance? This feature is very interesting to me and I would want to be able to log any error with my own logging interface.

genematx: Not an intrusion at all! :) And welcome the discussion, @jacopoabramo! Thank you for the suggestion; we actually discussed almost the same idea with @gwbischof recently.
If I understand you correctly, you mean specifying a logger as an _instance_attribute_ via dependency injection as opposed to the module-level logging? Something like this:
```
    def __init__(self, logger=None):
        self.logger = logger or logging.getLogger(__name__)
```
I see that this may offer extra flexibility and customization, but I'm not completely convinced that instance-level logging is strictly necessary here (I couldn't find a prior example of this in Bluesky or Tiled). I always thought that it was more idiomatic to use `logger = logging.getLogger("module or class name")` and then attach handler to it or customize it in another way if desired.  Of course, if there are multiple callbacks with the same name subscribed to the RE it would be handy to differentiate between them with per-instance loggers, but from what I understand this would be uncommon in the Bluesky codebase. Wonder what @danielballan thinks.

jacopoabramo: @genematx

> If I understand you correctly, you mean specifying a logger as an instance_attribute via dependency injection as opposed to the module-level logging?

Yes precisely!

> I always thought that it was more idiomatic to use `logger = logging.getLogger("module or class name")` and then attach handler to it or customize it in another way if desired.

Yes now that you mention it, it makes a lot more sense to approach it in this manner.

In my case I'm using a `LoggerAdapter` which customizes the format of the logging message to provide additional information about the class caller and/or if there's a `name` attribute the object instance, but this is an implementation detail irrelevant to the topic so I would go for the option you suggested.

Probably, for completion, the main fallback logger could be the `bluesky` one or a dedicated `bluesky-callbacks` added for the occasion.

danielballan: I agree that using a `LoggerAdapter` is a nice way to do this.

danielballan (approved):

PR #1916, opened 2025-06-04
Refactoring TiledWriter

genematx:
Some refactoring changes to TiledWriter.

## Description
Split TiledWriter into two callbacks, `_RunNormalizer` and `_RunWriter` with the former intended to convert any encountered Bluesky documents that follow legacy schema conventions to the most recent `EventModel` schema. `_RunWriter` is subscribed to the `_RunNormalizer` in chain and can only accept documents that adhere to the new schemas. Furthermore, any non-stream documents are converted to `StreamResource` and `StreamDatums`. `_RunNormalizer` creates deep copies of the documents to prevent modifying them for other callbacks subscribed to the same RE.

Furthermore, a backup capability has been added to TiledWriter: if writing to Tiled fails at any point, the entire BlueskyRun (all previous and all future documents) would be written to a local JSONLines file.

## Motivation and Context
The introduced separation of functionality between schema conversion and document writing would help apply any beamline-specific patches and transforms, if needed, before ingesting the documents to Tiled.

## How Has This Been Tested?
* Introduced a new test file, `external_assets_legacy.json`, containing a detailed example of legacy external asset documents to validate backward compatibility.
* Added a new test, `test_bad_document_order`, to ensure `TiledWriter` can handle documents emitted in an unexpected order, such as `datum` documents appearing late.

Related: [EventModel PR-223](https://github.com/bluesky/event-model/pull/223)

### DIFF
--- a/docs/tiled-writer.rst
+++ b/docs/tiled-writer.rst
@@ -48,11 +48,11 @@ An example of the Tiled catalog structure for a Bluesky run might look like this
 Callback Architecture
 =====================
 
-Structurally, TiledWriter consists of two main parts: `_RunNormalizer` and `_RunWriter`.
+Structurally, TiledWriter consists of two main parts: `RunNormalizer` and `_RunWriter`.
 
-The former is responsible for converting legacy document schemas to their latest version; this ensures that existing Bluesky code that relies on older versions of the Bluesky Event Model can still function correctly with TiledWriter. For example, while TiledWriter natively works with the modern `StreamResource` and `StreamDatum` documents commonly used in asynchronous plans, the `Resource` and `Datum` documents are automatically converted to their modern counterparts prior to being written to the Tiled catalog. The schema normalization is mostly done by renaming and restructuring certain document fields, but `_RunNormalizer` also allows the user to invoke use-case-specific patches for each type of documents and achieve high flexibility.
+The former is responsible for converting legacy document schemas to their latest version; this ensures that existing Bluesky code that relies on older versions of the Bluesky Event Model can still function correctly with TiledWriter. For example, while TiledWriter natively works with the modern `StreamResource` and `StreamDatum` documents commonly used in asynchronous plans, the `Resource` and `Datum` documents are automatically converted to their modern counterparts prior to being written to the Tiled catalog. The schema normalization is mostly done by renaming and restructuring certain document fields, but `RunNormalizer` also allows the user to invoke use-case-specific patches for each type of documents and achieve high flexibility.
 
-The simplified flowchart of the `_RunNormalizer` logic is shown below. It illustrates how the input documents (top) are processed and emitted as output documents (bottom) after specific transformations or caching operations.
+The simplified flowchart of the `RunNormalizer` logic is shown below. It illustrates how the input documents (top) are processed and emitted as output documents (bottom) after specific transformations or caching operations.
 
 .. mermaid::
 


src/bluesky/callbacks/tiled_writer.py
```diff
@@ -30,11 +39,16 @@
...
 
 from ..consolidators import ConsolidatorBase, DataSource, StructureFamily, consolidator_factory
+from ..run_engine import Dispatcher
 from ..utils import truncate_json_overflow
 from .core import MIMETYPE_LOOKUP, CallbackBase
+from .json_writer import JSONLinesWriter
 
 # Try to aggregare the table rows in batches of this size before writing to Tiled
 TABLE_UPDATE_BATCH_SIZE = 10000
+RESERVED_DATA_KEYS = ["time", "seq_num"]  # Reserved words that should not be used as data_keys identifiers
```
danielballan: It might be worth referencing this long-stalled PR, which would do this at the schema level. https://github.com/bluesky/event-model/pull/223

src/bluesky/consolidators.py
```diff
@@ -126,8 +128,8 @@ def __init__(self, stream_resource: StreamResource, descriptor: EventDescriptor)
         #    and make a best effort to convert it to a numpy spec like '<u8'.
         # 5. If unable to do any of the above, pass through whatever string is in 'dtype'.
         self.data_type: Optional[Union[BuiltinDtype, StructDtype]]
-        dtype_numpy = np.dtype(
-            list(map(tuple, data_desc.get("dtype_descr", [])))  # fileds of structural dtype
+        dtype_numpy = np.dtype(  # type: ignore
+            list(map(tuple, data_desc.get("dtype_descr", [])))  # type: ignore # fileds of structural dtype
```
danielballan:
```suggestion
            list(map(tuple, data_desc.get("dtype_descr", [])))  # type: ignore # fields of structural dtype
```

src/bluesky/consolidators.py
```diff
@@ -106,7 +106,9 @@ def __init__(self, stream_resource: StreamResource, descriptor: EventDescriptor)
 
         # Find datum shape and machine dtype; dtype_numpy, dtype_str take precedence if specified
         data_desc = descriptor["data_keys"][self.data_key]
-        self.datum_shape: tuple[int, ...] = tuple(data_desc["shape"])
+        if None in data_desc["shape"]:
+            raise NotImplementedError(f"Consolidator for {self.mimetype} does not support variable-sized data")
```
danielballan, line 107: Support for this might be coming soon. I wondering if it's worth putting this into a separate method (`validate_shape`) so it is easy to override. Maybe, maybe not.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -61,62 +75,157 @@ def concatenate_stream_datums(*docs: StreamDatum):
...
+    ----------
+        client : `tiled.client.BaseClient`
+            The Tiled client to use for writing data. This client must be initialized with
+            the appropriate credentials and connection parameters to access the Tiled server.
+    """
+
+    def __init__(
+        self,
+        client: BaseClient,
+        normalize: bool = True,
```
danielballan: I expect we'll be iterating on `_RunNormalizer` quite a bit, as we discover new edge cases. Our peer facilities may as well. Should we make it pluggable so that it is easy to pass in a custom `_RunNormalizer`?

```py
normalizer: CallbackBase = RunNormalizer
```

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -61,62 +75,157 @@ def concatenate_stream_datums(*docs: StreamDatum):
...
+                    try:
+                        bcb(name, doc)
+                    except Exception as e:
+                        logger.warning(
+                            f"Backup callback {bcb.__class__.__name__} failed with error: {e}", stacklevel=2
+                        )
+            self._buffer.clear()
+
+
+class _RunNormalizer(CallbackBase):
```
danielballan: Going along with my comment above, we might have to accept that this is public API (as much as it's also a closet full of skeletons).

docs/tiled-writer.rst
```diff
@@ -0,0 +1,155 @@
...
+
+The time dimension (or the sequence of measurements) is typically shared between the internal and external data, which in Tiled is enforced by writing all data from the same Bluesky stream into a specialized `Composite` container (node in the Tiled catalog). The metadata on each stream node contains the specifications for the related data keys and the relevavt configuration parameters supplied in the `EventDescriptor` document.
+
+Finally, nodes for multiple streams are groupped together and placed into a container for the entire run; its metadata contains the `Start` and `Stop` documents. While the strcuture of the `streams` container is fixed, the parent Run allows for optional user-controlled namespaces within `views` and `aux` containers. The Run container created by TiledWriter is designated with the `BlueskyRun_v3.0` spec to enable its back-compatibility with legacy code via bluesky-tiled-plugins.
```
danielballan:
```suggestion
Finally, nodes for multiple streams are groupped together and placed into a container for the entire run; its metadata contains the `Start` and `Stop` documents. While the strcuture of the `streams` container is fixed, the parent Run allows for optional user-controlled namespaces within `views` and `aux` containers. The Run container created by TiledWriter is designated with the `BlueskyRun` version `3.0` spec to enable its back-compatibility with legacy code via bluesky-tiled-plugins.
```

docs/tiled-writer.rst
```diff
@@ -0,0 +1,155 @@
...
+    RE = RunEngine()
+    tw = TiledWriter(tiled_client)
+    RE.subscribe(tw)
+
+    # Run an experiment collecting internal data
+    uid, = RE(bp.count([det], 3))
+    data = tiled_client[uid]['streams/primary/det'].read()
+
+    # Run an experiment collecting external data
+    save_path = str(tiled_server.directory / "data")
```
danielballan: I am genuinely not sure how I feel about directing "external" detectors to write in Tiled's internal writable area like this. I _think_ it might be better to do this:

```py
save_path = "some/path"  # maybe use tempfile here?
tiled_server = SimpleTiledServer(readable_storage=[save_path])  # soon-to-be-released feature from Jessica
```

Thoughts? I can argue this both ways.

danielballan (commented):
The separation of legacy is a huge improvement to the readability of this code.

Just a couple comments.

src/bluesky/callbacks/tiled_writer.py
genematx: Oh, definitely! Thanks for pointing this out.

src/bluesky/callbacks/tiled_writer.py
genematx: hmm, it does make sense! and just use the one from the module, by default, right? I was imagining that `patches` can be used for any fine-tuning of the documents before running them through `_RunNormalizer`, but I see that deeper involvement might be necessary in certain cases. will make it an extra argument.

src/bluesky/callbacks/tiled_writer.py
genematx: Do you mean we should call it `RunNormalizer` (without the underscore)? My main motivation for doing so was to emphasize that it is only intended to be used within `TiledWriter`, although other callbacks may also benefit from it.

docs/tiled-writer.rst
genematx: I like this new option, but likewise, I'm not sure if it's bad to write to the Tiled writable area. Probably it is, especially if the data created by the detectors is very similar to what Tiled might be writing (i.e. zarr). At the same time, it's a SimpleTiledServer... but we need to use the "best practices" and not provide bad examples of course.
I think I'm leaning towards declaring the separate `readable_storage`.

genematx: fixed

src/bluesky/callbacks/tiled_writer.py
genematx: changed; now instead of `normalize = True`, the user can pass `normalizer = _MyNormalizer` or `None` to disable this step altogether. By default the `_RunNormalizer` will be used.

src/bluesky/consolidators.py
genematx: Are you thinking the `validate_shape` method should just raise the error (or not, if there are no issue), or should it be more like a setter for the `datum_shape` attribute? I've only added that exception after I noticed that the EvenModel schema has been updated recently to allow for `None`s in the shape, and mypy was complaining.

I personally prefer setting the attributes directly in `__init__` instead of in separate methods, so it's more visible what attributes the class has (we can do `self.datum_shape = self.validate_shape(...)` here of course). I also see that there's quite a bit of other back-compat things going on while we setting this shape attribute (notably checking for the multiplier in StreamResource); ideally I think, it should be done in `_RunNormalizer`, but it means delaying the emittance of Descriptors even further...

I need to think more about this, but currently I'm thinking about just leaving it as is.

src/bluesky/callbacks/tiled_writer.py
danielballan: Yes. If it's public, people can safely subclass/extend it to add any facility-specific normalizations.

genematx: Done!

Issue #1895, opened 2025-03-18
Allow RE to accept an async function in the scan_id_source hook


abbiemery:
In blueapi we have created a client for our numtracker to get unique file names and scan numbers. Creating a new scan this way is currently sync, but it could be made async by altering the `scan_id_source`.

@callumforrester for more details?

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

callumforrester:
Yes, supporting an async function to fetch the scan id from an external service would be useful for us and can be added in a backwards-compatible way using `sync_or_async`, similar to what's already happened to other parts of the `RunEngine`

PR #1919, opened 2025-06-16
ENH: Support async scan_id_source functions

tpoliaw:
Allow scan_id_source functions to be async

## Description
If the scan id is determined by an external service (as is the case at DLS), it
would be useful to not block on the request.

## Motivation and Context
See #1895

## How Has This Been Tested?
New tests asserting scan id is updated correctly on `open_run` with both sync
and async source functions

### DIFF
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -38,6 +38,7 @@ from .protocols import (
     Status,
     Stoppable,
     T,
+    SyncOrAsync,
     Triggerable,
     check_supports,
 )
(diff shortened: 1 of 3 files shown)


tacaswell (approved):

src/bluesky/run_engine.py
```diff
@@ -269,7 +270,7 @@ class RunEngine:
         completely up to the user.
         Expected return: normalized metadata
 
-    scan_id_source : callable, optional
+    scan_id_source : callable
```
coretl:
```suggestion
    scan_id_source : callable, optional
```

src/bluesky/run_engine.py
```diff
@@ -269,7 +270,7 @@ class RunEngine:
         completely up to the user.
         Expected return: normalized metadata
 
-    scan_id_source : callable, optional
```
tacaswell, line 272: Please leave as optional and update to note that the function can be sync or async.

PR #1923, opened 2025-06-24
Update subscribable protocol to use subscribe_reading

### LINKED ISSUE #1915: Subscribeable Protocol changes to support async suspenders
Discussion needed for https://github.com/bluesky/bluesky/pull/1806

Currently, the `Subscribeable` protocol has an abstractmethod `subscribe(self, function: Callback[T])`
where `Callback = Callable[[dict[str, Reading[T]]], None]`. In reality, the callbacks in `EpicsSignalBase` from ophyd expect the value, not a Reading, as its argument. In ophyd-async, `SignalR` implements a `subscribe` which does match the protocol, and a `subscribe_value`, which is not described by any protocol.

When installing suspenders, we would like to use `subscribe_value` for both ophyd and ophyd-async signals. How do we update the Subscribeable Protocol, and how do we update sync ophyd accordingly?

Comments from the collaboration call say 
"Remove subscribe from the protocols; add `subscribe_value`; make up a new thing for calling `read()` on a device ("subscribe reading"?)"

Do we add a `subscribe_value` to the ophyd signals that just calls the existing `subscribe`? Is it necessary to also add `subscribe_reading` to sync ophyd signals? Should both methods belong to the same protocol?

@coretl @danielballan @tacaswell

jsouter:
Closes https://github.com/bluesky/bluesky/issues/1915, prerequisite for https://github.com/bluesky/bluesky/issues/1789

ophyd signals explicitly fail to implement Subscribable now (previously, signature on `subscribe` method was wrong, but still passed an `isinstance` check)
monitor updated to work with Subscribable, and ophyd signals which implement `subscribe`.

Also adds ophyd-async as an optional dev dependency, though the test is currently skipped.

ophyd-async will need updating to pin to a released bluesky, then to be updated to implement the new Subscribable protocol (rename subscribe to `subscibe_reading`, implement `subscribe_value` where necessary) after this is merged/released

dodal doesn't seem to require any changes to its codebase but may also need to be pinned to a release of bluesky and ophyd-async until all those PRs are through.

### DIFF
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -470,18 +470,11 @@ class RunBundler:
         where kwargs are passed through to ``obj.subscribe()``
         """
         obj = msg.obj
-        subscribable = False
-        try:
-            check_supports(obj, Subscribable)
-            subscribable = True
-        except AssertionError:
-            assert callable(getattr(obj, "subscribe", None)), (
-                "%s does not implement Subscribable protocol or adhere to ophyd subscription pattern." % obj
-            )
         if msg.args:
             raise ValueError("The 'monitor' Msg does not accept positional arguments.")
         kwargs = dict(msg.kwargs)
         name = kwargs.pop("name", short_uid("monitor"))
+        assert not kwargs, "If subscribe callback called with readings, kwargs are not supported."
         if obj in self._monitor_params:
             raise IllegalMessageSequence(f"A 'monitor' message was sent for {obj} which is already monitored")
 
@@ -490,21 +483,7 @@ class RunBundler:
         stream_bundle = await self._prepare_stream(name, {obj: self._current_stream_cache.describe_cache[obj]})
         compose_event = stream_bundle[1]
 
-        def emit_event(readings: dict[str, Reading] | None = None, *args, **kwargs):
-            if readings is not None:
-                # We were passed something we can use, but check no args or kwargs
-                assert not args and not kwargs, (
-                    "If subscribe callback called with readings, args and kwargs are not supported."
-                )
-            else:
-                # Ignore the inputs. Use this call as a signal to call read on the
-                # object, a crude way to be sure we get all the info we need.
-                readable_obj = check_supports(obj, Readable)  # type: ignore
-                readings = readable_obj.read()  # type: ignore
-                assert not inspect.isawaitable(readings), (
-                    f"{readable_obj} has async read() method and the callback "
-                    "passed to subscribe() was not called with Dict[str, Reading]"
-                )
+        def emit_event(readings: dict[str, Reading]):
             data, timestamps = _rearrange_into_parallel_dicts(readings)
             doc = compose_event(
                 data=data,
@@ -512,12 +491,29 @@ class RunBundler:
             )
             self.emit_sync(DocumentNames.event, doc)
 
-        self._monitor_params[obj] = emit_event, kwargs
-        # TODO: deprecate **kwargs when Ophyd.v2 is available
-        if subscribable:
-            obj.subscribe_reading(emit_event, **kwargs)
+        if isinstance(obj, Subscribable):
+            self._monitor_params[obj] = emit_event, kwargs
+            obj.subscribe_reading(emit_event)
+        elif callable(getattr(obj, "subscribe", None)) is not None:
+
+            def emit_event_readable(readings: Optional[dict[str, Reading]] = None, *args, **kwargs):
+                if readings is None:
+                    # Ignore the inputs. Use this call as a signal to call read on the
+                    # object, a crude way to be sure we get all the info we need.
+                    readable_obj = check_supports(obj, Readable)  # type: ignore
+                    readings = readable_obj.read()  # type: ignore
+                    assert not inspect.isawaitable(readings), (
+                        f"{readable_obj} has async read() method and the callback "
+                        "passed to subscribe() was not called with Dict[str, Reading]"
+                    )
+                emit_event(readings)
+
+            self._monitor_params[obj] = emit_event_readable, kwargs
+            obj.subscribe(emit_event_readable, **kwargs)
         else:
-            obj.subscribe(emit_event, **kwargs)
+            raise RuntimeError(  # must be a better error
+                "%s does not implement Subscribable protocol or adhere to ophyd subscription pattern." % obj
+            )
 
     def record_interruption(self, content):
         """
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -511,7 +511,7 @@ class RunBundler:
                 emit_event_from_readings(readings)
 
             self._monitor_params[obj] = emit_event_for_subscribe, kwargs
-            obj.subscribe(emit_event_for_subscribe, **kwargs)
+            obj.subscribe(emit_event_for_subscribe)
         else:
             raise RuntimeError(
                 "%s does not implement Subscribable protocol or adhere to ophyd subscription pattern." % obj
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -806,9 +806,9 @@ class RunBundler:
         def is_data_key(obj: Any) -> bool:
             return isinstance(obj, dict) and {"dtype", "shape", "source"}.issubset(frozenset(obj.keys()))
 
-        assert all(
-            not is_data_key(value) for value in describe_collect.values()
-        ), "Single nested data keys should be pre-declared"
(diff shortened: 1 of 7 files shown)


src/bluesky/tests/test_devices.py
```diff
@@ -116,6 +120,34 @@ def plan():
     }
 
 
+@pytest.mark.skip("ophyd-async must be updated to implement Subscribable protocol")
```
coretl: This should work at the moment shouldn't it? As ophyd-async signal has a `subscribe` method, but passes its readings down to it, so should work as it used to?

jsouter: I think when we last discussed this the plan was to rename `Signal.subscribe` to `Signal.subscribe_reading` in ophyd-async. We would have to pin ophyd-async's bluesky to a release, then merge this, then merge the PR to update ophyd-async to adhere to the new protocol and make a release of it before this test would work

coretl: But the code you've written here looks like it would work for `ophyd-async` as it currently is, with the `subscribe()` method. I think that is preferable, then you only need to put in a minimum bound as part of https://github.com/bluesky/ophyd-async/pull/950

jsouter: The error message right now is:
```
TypeError: Can't instantiate abstract class SignalRW without an implementation for abstract method 'subscribe_reading'
```
as Signal already inherits from the Subscribable protocol

coretl: That's annoying, I thought inheriting from protocols didn't runtime check the contents, but I guess it does:
```python
>>> from typing import Protocol
>>> from abc import abstractmethod
>>> class Foo(Protocol):
...     @abstractmethod
...     def foo(self) -> None: ...
... 
>>> class Bar(Foo):
...     ...
... 
>>> bar = Bar()
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class Bar without an implementation for abstract method 'foo'
```

As part of https://github.com/bluesky/ophyd-async/pull/950 could we set `subscribe = subscribe_reading` for all the changes you made please? Then we can merge and release that first, and it will work for old and new bluesky releases, then we can get this protocol change in and the suspender one in after

jsouter: Just clearing my tickets. With ophyd-async updated this should be good to go now.

src/bluesky/bundlers.py
```diff
@@ -426,11 +426,12 @@ async def monitor(self, msg):
 
         where kwargs are passed through to ``obj.subscribe()``
         """
-        obj = check_supports(msg.obj, Subscribable)
+        obj = msg.obj
         if msg.args:
             raise ValueError("The 'monitor' Msg does not accept positional arguments.")
         kwargs = dict(msg.kwargs)
         name = kwargs.pop("name", short_uid("monitor"))
+        assert not kwargs, "If subscribe callback called with readings, kwargs are not supported."
```
coretl: We assert it is empty here, but still use it below?

tacaswell: Don't use asserts in library code.  Depending on the optimization level these can be elided out.  Please change this to as `if ...`.

tacaswell: Oh, I see we had `asserts` in here already, there should not have been.

src/bluesky/bundlers.py
```diff
@@ -439,31 +440,37 @@ async def monitor(self, msg):
...
+            obj.subscribe_reading(emit_event)
+        elif callable(getattr(obj, "subscribe", None)):
+
+            def emit_event_readable(readings: Optional[dict[str, Reading]] = None, *args, **kwargs):
+                if readings is None:
+                    # Ignore the inputs. Use this call as a signal to call read on the
+                    # object, a crude way to be sure we get all the info we need.
+                    readable_obj = check_supports(obj, Readable)  # type: ignore
+                    readings = readable_obj.read()  # type: ignore
+                    assert not inspect.isawaitable(readings), (
```
tacaswell: Please also change this to an `if ...`

src/bluesky/bundlers.py
jsouter: Will update!

src/bluesky/bundlers.py
```diff
@@ -439,31 +440,37 @@ async def monitor(self, msg):
...
+        if isinstance(obj, Subscribable):
+            self._monitor_params[obj] = emit_event, kwargs
+            obj.subscribe_reading(emit_event)
+        elif callable(getattr(obj, "subscribe", None)):
+
+            def emit_event_readable(readings: Optional[dict[str, Reading]] = None, *args, **kwargs):
+                if readings is None:
+                    # Ignore the inputs. Use this call as a signal to call read on the
+                    # object, a crude way to be sure we get all the info we need.
+                    readable_obj = check_supports(obj, Readable)  # type: ignore
```
tacaswell: Why the `#type: ignore`?

src/bluesky/bundlers.py
```diff
@@ -439,31 +440,37 @@ async def monitor(self, msg):
...
-            else:
-                # Ignore the inputs. Use this call as a signal to call read on the
-                # object, a crude way to be sure we get all the info we need.
-                readable_obj = check_supports(obj, Readable)  # type: ignore
-                readings = readable_obj.read()  # type: ignore
-                assert not inspect.isawaitable(readings), (
-                    f"{readable_obj} has async read() method and the callback "
-                    "passed to subscribe() was not called with Dict[str, Reading]"
-                )
+        def emit_event(readings: dict[str, Reading]):
```
tacaswell: This is minor, but can we call this `emit_event_for_subscribe_reading` and the other `emit_event_for_subscribe`? 

Can we also move the other function up here?  I don't see any reason it needs to be defined nested, but I may be missing something.

src/bluesky/bundlers.py
```diff
@@ -504,7 +511,14 @@ async def unmonitor(self, msg):
 
             Msg('unmonitor', obj)
         """
-        obj = check_supports(msg.obj, Subscribable)
+        try:
+            obj = check_supports(msg.obj, Subscribable)
```
tacaswell: How did this ever work before?

tacaswell: oh, I see we change the `Protocol`.

Can we just do the `callable(getattr(obj, "clear_sub", None)` check and skip the protocol?

jsouter: `check_supports` asserts `isinstance(obj, protocol)`, which checks the obj has all the attributes of the Protocol without verifying the signature, ophyd signals previously passed this check because they have a method called `subscribe`, but now fail because they don't have a method called `subscribe_reading`.

src/bluesky/tests/test_protocols.py
```diff
@@ -38,6 +38,7 @@ def test_pausable():
     assert isinstance(sim.det1, bs_protocols.Pausable)
 
 
+@pytest.mark.skip("ophyd signals do not properly implement Subscribable protocol")
```
tacaswell:
```suggestion
@pytest.mark.xfail("Subscribable Protocol was not written to match ophyd's"
                   "implementation.  Future ophyd will implement new Protocol")
```

tacaswell: This may need to be re-formatted to appease the linter.

src/bluesky/bundlers.py
jsouter: That was in there before but seems the type checker doesn't complain without it.

src/bluesky/bundlers.py
jsouter: It can be moved up but it won't be used in the first case of the if branch. The other function (`emit_event_reading`) also calls `emit_event`, so maybe `emit_event_for_subscribe_reading` isn't an appropriate name in that case.

jsouter: Have changed to `emit_event_from_readings` and `emit_event_for_subscribe`

src/bluesky/tests/test_protocols.py
```diff
@@ -38,6 +38,10 @@ def test_pausable():
     assert isinstance(sim.det1, bs_protocols.Pausable)
 
 
+@pytest.mark.xfail(
```
tacaswell: I would rather fully remove this test.

coretl:
@tacaswell I think this is ready to merge now

tacaswell (approved):

tacaswell:
A 3.11 test failed 2x in a row with different failures so I kicked them all off again to see if we gained some flaky tests.

tacaswell:
The docs failed to fetch the intresphinx from scipy, merging over that.

PR #1927, opened 2025-07-10
Add typing to RunEngine __call__ method

tpoliaw:
Since Python 3.8 we have been able to define positional only args (see
PEP 570). This means we no longer need the varargs and custom signature
combination and the types can be included in the definition as normal.

### DIFF
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -15,7 +15,7 @@ import types
 import uuid
 import warnings
 from collections import namedtuple
-from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Generator, Iterable
+from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Generator, Iterable, Sequence
 from collections.abc import Iterable as TypingIterable
 from functools import partial, reduce, wraps
 from inspect import Parameter, Signature
@@ -23,7 +23,6 @@ from typing import (
     Any,
     Callable,
     Optional,
-    Sequence,
     TypedDict,
     TypeVar,
     Union,


src/bluesky/utils/__init__.py
```diff
@@ -93,6 +97,18 @@ def __repr__(self):
 #: Scalar or iterable of values, one to be applied to each point in a scan
 ScalarOrIterableFloat = Union[float, TypingIterable[float]]
 
+# Single function to be used as an event listener
+Subscriber = Callable[[str, DocumentType], Any]
```
tpoliaw: Open to name bikeshedding. `EventListener`? `EventHandler`?

src/bluesky/utils/__init__.py
```diff
@@ -93,6 +97,18 @@ def __repr__(self):
...
+
+# List of event listeners
+SubscriberList = List[Subscriber]
+
+# Mapping from event type to listener or list of listeners
+SubscriberMap = Mapping[Literal["all", "start", "descriptor", "event", "stop"], Union[Subscriber, SubscriberList]]
```
tpoliaw: At the moment this is a map of event type to subscribers of any event type so that something like 

```python
def start_listener(name, start_doc: RunStart):
    ...

RE(plan, {"start": start_listener})
```
would fail type-checking.

Is there a defined list of which document types are sent to which handler types? If so this might be better as a TypedDict along the lines of

```python
class SubscriberMapping(TypedDict):
    all: Subscriber[RunStart] | SubscriberList[RunStart] = []
    start: ...
    # etc
```

src/bluesky/utils/__init__.py
```diff
@@ -93,6 +95,24 @@ def __repr__(self):
 #: Scalar or iterable of values, one to be applied to each point in a scan
 ScalarOrIterableFloat = Union[float, TypingIterable[float]]
 
+# Single function to be used as an event listener
+Subscriber = Callable[[str, P], Any]
```
tacaswell, line 99: Is it kosher to re-use a TypeVar like this?

tacaswell:
This seems reasonable to me and I am 👍🏻 on dropping the complexity of the signature munging.

tacaswell (approved):
Modulo the question about re-using a TypeVar.

I'm also not sure about the typing on the callbacks, but as it does not affect run time I'm happy to have it merged and iterated on.

src/bluesky/utils/__init__.py
tpoliaw: You mean re-using the `P` for both `MsgGenerator` and `Subscriber`? I think so, when you check the types they resolve to `Generator[Msg, Any, P@MsgGenerator]` and `(str, P@Subscriber) -> Any` respectively so there's no link when type-checking.

src/bluesky/run_engine.py
```diff
@@ -873,7 +864,13 @@ def _create_result(self, plan_return):
...
         return rs
 
-    def __call__(self, *args, **metadata_kw):
+    def __call__(
+        self,
+        plan: typing.Union[MsgGenerator, typing.Iterable[Msg]],
+        subs: typing.Optional[Subscribers] = None,
+        /,
+        **metadata_kw: typing.Any,
+    ):
```
DiamondJoseph:
```suggestion
    def __call__(
        self,
        plan: typing.Iterable[Msg],
        subs: typing.Optional[Subscribers] = None,
        /,
        **metadata_kw: typing.Any,
    ):
```

A Generator is an Iterator is an Iterable, but it may be clearer to read for us that it explicitly accepts a MsgGenerator (if we made use of the P that it returns it definitely would)

src/bluesky/utils/__init__.py
```diff
@@ -93,6 +95,24 @@ def __repr__(self):
...
+OneOrMany = Union[P, list[P]]
+
+
+# Mapping from event type to listener or list of listeners
+class SubscriberMap(TypedDict, total=False):
+    all: OneOrMany[Subscriber[DocumentType]]
+    start: OneOrMany[Subscriber[RunStart]]
+    stop: OneOrMany[Subscriber[RunStop]]
+    event: OneOrMany[Subscriber[Event]]
+    descriptor: OneOrMany[Subscriber[EventDescriptor]]
```
DiamondJoseph, line 110: This looks correct for what the docstring in the RunEngine says, do the subscribers not support the other doc types (Resource[Stream], Datum[Stream]) without going through `all`?

DiamondJoseph (approved):

src/bluesky/utils/__init__.py
DiamondJoseph: I think 

```python
def my_plan() -> MsgGenerator[int]:  # P@MsgGenerator = int
  yield from []
  return 0

def my_subscriber(name: str, doc: RunStart) -> None:  # P@my_subscriber = RunStart
  return None

RE(my_plan(), {"start": my_subscriber})  # These are two different Ps
```

If the `__call__` method was instead:

```
  def __call__(plan: MsgGenerator[P], subscriber: Subscriber[P]):
# or
  def __call__(plan: MsgGenerator[P], subscriber: Subscriber[P]) -> P
```

Only then would they need to be matching

src/bluesky/run_engine.py
tpoliaw: Makes sense, I thought it would be useful to include that plans can be coroutines that accept feedback but given it's currently typed as `Any` I don't think it adds much for now.

src/bluesky/utils/__init__.py
jacopoabramo: Technically, when implementing a callback inheriting from [`CallbackBase`](https://github.com/bluesky/bluesky/blob/103accac0ebe3d2f0efcd0c5f7da87e8172f5b16/src/bluesky/callbacks/core.py#L122) (which it's simply a [`DocumentRounter`](https://github.com/bluesky/event-model/blob/678bf04f756fe0f6f54b8dc742594cf82d3b77e7/src/event_model/__init__.py#L124) with a different name at this point), there's the possibility to define methods to handle **all** document types. I think that the the docstring is a bit misleading.

tpoliaw: So should this have fields for all the types in `DocumentNames`?

src/bluesky/run_engine.py
```diff
@@ -873,7 +863,13 @@ def _create_result(self, plan_return):
...
         return rs
 
-    def __call__(self, *args, **metadata_kw):
+    def __call__(
+        self,
+        plan: typing.Iterable[Msg],
+        subs: typing.Optional[Subscribers] = None,
+        /,
+        **metadata_kw: typing.Any,
+    ) -> RunEngineResult | tuple[str, ...]:
```
jacopoabramo:
```suggestion
    ) -> typing.Union[RunEngineResult, tuple[str, ...]]:
```

For >= 3.9 compatibility

src/bluesky/utils/__init__.py
jacopoabramo: I believe it should; correct me if I'm wrong but documents are processed in the [`Dispatcher`](https://github.com/bluesky/bluesky/blob/103accac0ebe3d2f0efcd0c5f7da87e8172f5b16/src/bluesky/run_engine.py#L2695), which already accepts `DocumentNames` as callback signals. To be thorough, both the docstring and the documentation should reflect this... but maybe it can be postponed to another PR.

src/bluesky/utils/__init__.py
```diff
@@ -93,6 +95,24 @@ def __repr__(self):
...
+
+
+# Single listener, list of listeners or mapping of listeners by event type
+Subscribers = Union[Subscriber[DocumentType], list[Subscriber[DocumentType]], SubscriberMap]
```
jacopoabramo:
```suggestion
Subscribers = Union[Subscriber[DocumentType], Sequence[Subscriber[DocumentType]], SubscriberMap]
```

I'm undecided between `Sequence` and `Iterable` but it should make sense that tuples or custom-written containers should be acceptable inputs

EDIT: given this [answer](https://stackoverflow.com/a/72158086/4437552), `Sequence` should be fine

src/bluesky/utils/__init__.py
jacopoabramo: I think `Subscriber` makes more sense; the public APIs for registering callbacks are `subscribe` and `unsubscribe`, so the definitions match all accross the board.

src/bluesky/utils/__init__.py
tpoliaw: Dispatcher can handle everything in DocumentNames but the `RE.__call__` method calls out to `normalize_subs_input` which restricts the names to the five in the docstrings (via `SUB_NAMES`). If we want to extend the typing to the other `subscribe` methods it might be worth removing (or extending) that check and updating all the docstrings in one go. I will leave it for this PR though as it is correct for the current state of where it's used.

It looks like this is referenced in docstrings in several places (for my benefit if/when I get back to this: RE.subscribe,  RE.__call__, RE._subscribe, Dispatcher.process, Dispatcher.subscribe, subs_context, subs_wrapper SUBS_NAMES)

jacopoabramo: Then I'm curious of what happens if one defines methods of `CallbackBase` with events that are not part of `SUBS_NAMES`. Are these discarded? Or this behavior is specific to function callbacks only?

tpoliaw: I think if you add a single callback to `subscribe` it would work as expected with all `DocumentNames` being supported. It's only the methods that accept dicts and pass them to `normalize_subs_input` that restrict them to the four types.

tacaswell:
Asking the RE to filter documents at subscription time was a likely a mistake, but I don't think it is worth the effort to try and back out of it at this point.

PR #1933, opened 2025-07-17
WIP: Fixes of TiledWriter

genematx:
Some minor fixes and improvements of TiledWriter:

## Description

1. Add missing "/" in the Asset uri of MultipartRelatedConsolidator.
2. Improved readability of Error and Warning messages.
3. Added a test for JSON Writer.

### DIFF
--- a/src/bluesky/callbacks/tiled_writer.py
+++ b/src/bluesky/callbacks/tiled_writer.py
@@ -599,7 +599,7 @@ class _RunWriter(CallbackBase):
                     consolidator.validate(fix_errors=True)
                 except Exception as e:
                     msg = f"{type(e).__name__}: " + str(e).replace("\n", " ").replace("\r", "").strip()
-                    warn(f"Validation of StreamResource {sres_uid} failed with error: {msg}", stacklevel=2)
+                    warn(f"Validation of StreamResource {sres_uid} failed with {msg}", stacklevel=2)
                 self._update_data_source_for_node(sres_node, consolidator.get_data_source())
 
         # Write the stop document to the metadata
--- a/src/bluesky/callbacks/tiled_writer.py
+++ b/src/bluesky/callbacks/tiled_writer.py
@@ -599,7 +599,7 @@ class _RunWriter(CallbackBase):
                     consolidator.validate(fix_errors=True)
                 except Exception as e:
                     msg = f"{type(e).__name__}: " + str(e).replace("\n", " ").replace("\r", "").strip()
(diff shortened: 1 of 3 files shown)


src/bluesky/consolidators.py
```diff
@@ -450,7 +455,7 @@ def get_datum_uri(self, indx: int):
         integer index, e.g. "{:05d}.ext".
         """
         assert os.path.splitext(self.template)[1] in self.permitted_extensions
-        return self.uri + self.template.format(indx)
+        return self.uri.rstrip("/") + "/" + self.template.format(indx).lstrip("/")
```
jwlodek, line 458: I thought we were including the trailing slash, at least in ophyd-async devices, is it being stripped out due to conversion to a `Path` object somewhere?

Edit: Oh I see - this approach covers both possibilities.

jwlodek (approved):

src/bluesky/consolidators.py
genematx: I thought so too, but we've run into an issue at SMI this morning, which didn't include the trailing slash. I also thought that this would cover both possibilities, but now seeing the failing tests, I've realized that a big exception is when the first half of the filename is provided in the `resource_path` parameter of the Resource document and template covers the second half (there might have been a case like this in the wild somewhere, I'm a bit confused now).
This needs some further thought. The convention to always include the trailing slash is good though!

jwlodek: Ah yes, this is a fairly common setup on the floor. In ophyd async the convention is always include trailing slash and full dir path in uri, just filename template in parameters.

src/bluesky/callbacks/tiled_writer.py
```diff
@@ -599,7 +599,7 @@ def stop(self, doc: RunStop):
                     consolidator.validate(fix_errors=True)
                 except Exception as e:
                     msg = f"{type(e).__name__}: " + str(e).replace("\n", " ").replace("\r", "").strip()
-                    warn(f"Validation of StreamResource {sres_uid} failed with error: {msg}", stacklevel=2)
+                    warn(f"Validation of StreamResource {sres_uid} failed with {msg}", stacklevel=2)
```
prjemian: Would it be OK to add !r here in both cases?

```suggestion
                    warn(f"Validation of StreamResource {sres_uid!r} failed with {msg!r}", stacklevel=2)
```

genematx: Of course! That's a great idea; thank you, @prjemian!

PR #1937, opened 2025-07-22
Runengine metadata type from dict to MutableMapping[str, Any]

kivel:
Change the RunEngine metadata type to `MutableMapping[str, Any]`

## Description
1. type alias `RunEngineMetadata = MutableMapping[str, typing.Any]`
2. init method adjusted to use the type alias.
3. signature of default validator and normalizer adjusted.
4. linting error in Dockerfile fixed
5. platform locked to linux/amd64 in devcontainer.json

## Motivation and Context
The current type of the RunEngine metadata is `dict`, which causes static type checkers to fail when more complex objects are provided that present themself as the more general `MutableMapping`.
Based on the requirement for the metadata to be JSON serializable, the type is modified to be `MutableMapping[str, Any]`

## How Has This Been Tested?
- The existing tests have been run.
- The RunEngine was provided with a class identifying as the new type with meaningful data and a call `RE([Msg("open_run")])` yielded the expected outcome.


src/bluesky/run_engine.py
```diff
@@ -197,7 +198,12 @@ def __get__(self, instance, owner):
             return super().__get__(instance, owner)
 
 
-def default_scan_id_source(md):
+# type alias
+# TODO(kivel): check with core devs where to put this
+RunEngineMetadata = MutableMapping[str, typing.Any]
```
tacaswell, line 203: I think here is fine.  Makes sense to me to import this from the same place you import RunEngine from.

kivel: todo removed [240f3a3](https://github.com/bluesky/bluesky/pull/1937/commits/240f3a3e9516fd0e1a5e9d6d23c2b16b589b09fb)

tacaswell (approved):

tacaswell:
I do not fully trust my judgement on typing yet so this should get a second review please.

kivel:
I found a line that can break with a type `MutableMapping[str, Any]`
https://github.com/bluesky/bluesky/blob/d1314258863ad9a97d1270d398c9a5e8ef69bd0b/src/bluesky/run_engine.py#L1869
The `Any` would have to be changed to something is guaranteed to be JSON serializable.
```
JsonValue = Union[str, int, float, bool, None, dict[str, 'JsonValue'], list['JsonValue']]
RunEngineMetadata = MutableMapping[str, JsonValue]
```
Another issue is my limit knowledge of the pre 3.10 typing system. The input from a more a experienced developer regarding the use of `TypeAlias` is highly appreciated.

jwlodek:
I'd like to get back to this PR. @kivel I think 3.9 has been dropped as of 1.15.0, so should be good to stick with 3.10 style unions. I think that was the only remaining thing?

Also, in the linked PR I added a test that used a MutableMapping as the RE.md, which makes sure using that works, and via the CI type checking makes sure the type checker is happy with it. Perhaps something similar could be integrated here as well?

kivel:
@jwlodek After looking into this again, I conclude that a specific JSON-type is not the way to go. We'd need to satisfy the Union in each of the cases that accesses the `RE.md`, just to satisfy the type checker. That's - at least to me - no-value overhead.
Let me know if the PR is now in a state that your happy with.

jwlodek:
I am happy with `str, Any`, we can always revisit if we decide we need to tighten the typing.

jwlodek:
The CI failures are unrelated, and the changes are minor type fixes. I'm going to approve and merge tomorrow unless there are any last minute objections.

jwlodek (approved):

PR #1939, opened 2025-07-23
Correct `Subscriber` type-hints

Tom-Willemsen:
## Description

Fixes https://github.com/bluesky/bluesky/issues/1938

## Motivation and Context

We would like code like the following to pass a pyright type-check:

```python
from bluesky.callbacks import LiveTable

RE = RunEngine()
RE(bps.null(), LiveTable([]))
```

We ran into this in downstream CI.

## How Has This Been Tested?

This is a change to type-hints only; tested by running pyright manually on the code sample above and checking it now passes.

### DIFF
--- a/src/bluesky/tests/test_tiled_writer.py
+++ b/src/bluesky/tests/test_tiled_writer.py
@@ -290,7 +290,7 @@ def test_stream_datum_readable_counts(RE, client, tmp_path):
     det = StreamDatumReadableCollectable(name="det", root=str(tmp_path))
     RE(bp.count([det], 3), tw)
     stream = client.values().last()["streams"]["primary"]
-    keys = sorted(set(stream.parts).difference({"internal"}))
+    keys = sorted(set(stream.base.keys()).difference({"internal"}))
 
     assert stream[keys[0]].shape == (3,)
     assert stream[keys[1]].shape == (15, 10, 15)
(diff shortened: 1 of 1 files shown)


tpoliaw:
I think the simplest fix for this would be to replace `DocumentType` with `Union[DocumentType]` in the `Subscribers` definition. Alternatively, the same type could be defined in event_model (see PR 365), and then imported here. 

I don't think relaxing the type to accept any dict is useful - the callbacks should only ever handle events.

tpoliaw:
> > replace `DocumentType` with the equivalent of `Union[DocumentType]` in the `Subscribers` definition
> 
> I don't understand this suggestion - how does `Union[X]` differ from `X`?

I think I meant `Union[ALL_BASEMODELS]`, ie an instance one of the the various bluesky documents rather than `DocumentType` which is one of the document types themselves.

> > the callbacks should only ever handle events
> 
> I don't understand what you mean here either - e.g. using LiveTable as an example, it handles RunStart and descriptors as well as Events?

By `Event` I meant any of the models defined in `event_model` rather than the specific `Event` type - maybe `Document` would have been clearer. Does raise the question why the package isn't called `document_model`.

tacaswell:
I am OK with any typing approach that works (so long as runtime behavior is not affected).

DiamondJoseph:
re-running CI as Tiled failures possibly fixed by https://github.com/bluesky/tiled/pull/1115 included in https://github.com/bluesky/tiled/releases/tag/v0.1.0-b39 ?

DiamondJoseph:
Alright, this is now blocked by fallout from https://github.com/bluesky/tiled/pull/1093 - can we pin Tiled to a previous version to get this fix in and worry about Tiled later?

danielballan:
The rebase should do it, I think.

danielballan:
@genematx and I can sort this out today...so to have slowed this down.

PR #1940, opened 2025-07-29
Cast Numpy values in Msg.args to Python scalars in OpenTelemetry span attribute

flowln:
The OpenTelemetry API package does not support assigning a Numpy object, or a list containing them, as a Span attribute, giving errors like:

`Invalid type float64 in attribute 'msg.args' value sequence. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or None`

This is particularly important for the `args` argument, as it a lot of the times contains a Numpy scalar, generated from linspace or similar, including in the default plans.

Ref.: https://github.com/open-telemetry/opentelemetry-python/blob/f9ca4755af4a0e3cf4941a44948f73d5fea24d72/opentelemetry-api/src/opentelemetry/attributes/__init__.py#L42


src/bluesky/run_engine.py
```diff
@@ -2823,7 +2824,7 @@ def ignore_exceptions(self, val):
 
 def _set_span_msg_attributes(span, msg):
     span.set_attribute("msg.command", msg.command)
-    span.set_attribute("msg.args", msg.args)
+    span.set_attribute("msg.args", np.array(msg.args).tolist())
```
tacaswell: Can we use `bluesky.utils.sanitize_np` instead?

flowln: I didn't know about that, thanks!

tacaswell: json and bson also do not like numpy objects.

tacaswell (approved):

src/bluesky/run_engine.py
```diff
@@ -1908,8 +1909,8 @@ def _close_run_trace(self, msg: Msg):
         reason = msg.kwargs.get("reason", self._reason)
         try:
             _span: Span = self._run_tracing_spans.pop()
-            _span.set_attribute("exit_status", exit_status)
-            _span.set_attribute("reason", reason)
+            _span.set_attribute("exit_status", exit_status or "None")
+            _span.set_attribute("reason", reason or "None")
```
tacaswell:
```suggestion
            _span.set_attribute("exit_status", exit_status if exit_status is not None else "None")
            _span.set_attribute("reason", reason if reason is not None or "None")
```

Just using "truthy" would convert `''` -> `'None'` which does not seem right to me.

src/bluesky/tests/test_tracing.py
```diff
@@ -1,11 +1,31 @@
...
 
+from bluesky import plans as bp
 from bluesky.plan_stubs import sleep
+from bluesky.tests import requires_ophyd
 from bluesky.tracing import trace_plan, tracer
 
 
+@pytest.fixture(scope="session")
+def with_otl_instrumentation():
+    from opentelemetry import trace
```
tacaswell, line 16:
```suggestion
    pytest.importorskip('opentelemetry')
    from opentelemetry import trace
```

I think this will make the tests fail gracefully if the dependency is missing.  Even though it is now in the `[dev]` group, it is still better to have the tests gracefully skip on optional dependencies than to hard-fail at test collection.

tacaswell (approved):

PR #1949, opened 2025-09-02
ENH: add type variable `StatusType` for status objects with additional API

### LINKED ISSUE #1948: Support different status types using Generics
`ophyd-async` currently supports a [different type of status](https://github.com/bluesky/ophyd-async/blob/9d2016b893c6df0662b57273bf69f3e267a54137/src/ophyd_async/core/_status.py#L132) which has extra behaviour to that of the standard Bluesky status (which is represented by [AsyncStatus](https://github.com/bluesky/ophyd-async/blob/9d2016b893c6df0662b57273bf69f3e267a54137/src/ophyd_async/core/_status.py#L91)).

In Bluesky, plan stubs like `bps.complete` return `MsgGenerator[Status]`, so if I have a function which requires a `WatchableAsyncStatus`, I have to explicitly cast the status returned from Bluesky, and static type checking can't be used. 

If we had a status in Bluesky to represent the `WatchableAsyncStatus`, and had something like

```
StatusType = TypeVar("StatusType", bound=Status)


@plan
def complete(
    obj: Flyable[StatusType],
    *,
    group: Optional[Hashable] = None,
    wait: bool = False,
    **kwargs,
) -> MsgGenerator[StatusType]:
```
then we can have the benefits of static typing here.

## Acceptance Criteria
- Typing added in Bluesky to support different kinds of statuses

jacopoabramo:
## Description

Closes #1948

Incidentally annotates the return type of `kickoff_all` and `complete_all` plan stubs which were previously not annotated.

## Motivation and Context

I called it `StatusType` to make it more clear. Previously I named it `S` but it seemed a bit too generic.

## How Has This Been Tested?

Currently untested. I'm letting the CI do it's thing.

### DIFF
--- a/src/bluesky/plan_stubs.py
+++ b/src/bluesky/plan_stubs.py
@@ -827,7 +827,7 @@ def kickoff(
 
 @plan
 def kickoff_all(
-    *args: Sequence[Flyable[StatusType]], group: Optional[Hashable] = None, wait: bool = True, **kwargs
+    *args: Flyable[StatusType], group: Optional[Hashable] = None, wait: bool = True, **kwargs
 ) -> MsgGenerator[tuple[StatusType, ...]]:
     """
     Kickoff one or more fly-scanning devices.


jacopoabramo:
Should the plan stubs be updated with this type hinting as well?

olliesilvester:
Thanks, I won't have a chance to look properly in the next few days, but will get round to it eventually!

jacopoabramo:
@coretl I'm starting to think that this PR may have bad implications for ophyd and ophyd-async if there's strict inheritance from these protocols

coretl:
> @coretl I'm starting to think that this PR may have bad implications for ophyd and ophyd-async if there's strict inheritance from these protocols

I think it is backwards compatible unless using pyright in strict mode. That's probably ok for ophyd-async, but not sure about blueapi. @DiamondJoseph thoughts?

DiamondJoseph:
For blueapi this should be fine- we added some handling around `Movable[T]` that doesn't quite do what we want, but I can't imagine someone would query for a `Flyable[MySpecialStatusType]`. Plans would still be written to expect `MySpecificDevice(Flyable[MySpecialStatusType])`

I'm not 100% on whether the Protocols need to also extend Generic, because Protocols aren't as strict as I would like them to be. I will try adding a git dependency later to try it out if I get time and will report back- I'm on leave after today.

src/bluesky/plan_stubs.py
```diff
@@ -918,7 +921,9 @@ def complete(
 
 
 @plan
-def complete_all(*args, group: Optional[Hashable] = None, wait: bool = False, **kwargs):
+def complete_all(
+    *args, group: Optional[Hashable] = None, wait: bool = False, **kwargs
```
DiamondJoseph:
```suggestion
    *args: Flyable[StatusType], group: Optional[Hashable] = None, wait: bool = False, **kwargs
```

I think from a Generics perspective we only can/should limit the returned Status to a StatusType when we are passing in the other bound. For all of the stubs that take Movable above this, we don't know any more precisely than Status what type they return even with the new type hint.

src/bluesky/plan_stubs.py
```diff
@@ -1074,7 +1079,7 @@ def stage(
     *,
     group: Optional[Hashable] = None,
     wait: Optional[bool] = None,
-) -> MsgGenerator[Union[Status, list[Any]]]:
+) -> MsgGenerator[Union[StatusType, list[Any]]]:
```
DiamondJoseph, line 1082: as above, obj should become `Stageable[StatusType]`

src/bluesky/plan_stubs.py
```diff
@@ -1159,7 +1164,7 @@ def unstage(
...
```
DiamondJoseph, line 1167: As above

PR #1955, opened 2025-09-30
BUG: RunEngine._wait fails when waiting for a status times out

thomaslima:
I have encountered an error when I used the wait plan stub with a timeout that actually triggered. This is a simple bug.

## Description
In the RunEngine._wait method, the following line is incorrect
```python
# We might wait to call wait again, so put the futures and status objects back in
self._groups[group] = futs
```
because other parts of the same code relies on `self._groups[group]` being a `set` (e.g. in `self._groups[group].add(p_event.wait)`). This causes an Attribute error.

I fixed by changing the above line to 
```python
self._groups[group].update(futs)
```

Thanks! Big fan of the project so far. Really useful.

### DIFF
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -2334,7 +2334,7 @@ class RunEngine:
                 await status_task
             except WaitForTimeoutError:
                 # We might wait to call wait again, so put the futures and status objects back in
-                self._groups[group] = futs
+                self._groups[group].update(futs)
                 self._status_objs[group] = status_objs
                 if error_on_timeout:
                     raise


src/bluesky/run_engine.py
```diff
@@ -2334,7 +2334,7 @@ def cancel_status_task_if_error(fut: asyncio.Future[list[asyncio.Future]]):
                 await status_task
             except WaitForTimeoutError:
                 # We might wait to call wait again, so put the futures and status objects back in
-                self._groups[group] = futs
+                self._groups[group].update(futs)
```
tacaswell: Above in L2290 we have pop'd group out so I think we do need to do set item.

```suggestion
                self._groups[group] = set(update(futs))
```

However, that said the better fix may be to make the types stable at the top of this block and make the default value popped from the dictionary to be a `set` rather than an empty list.

tacaswell: wait, this is a default dict so it does work 🤦🏻 .

I'm surprised that the type-checkers did not flag this is well as `._groups` is typed as 

```python
        self._groups: defaultdict[typing.Any, set[typing.Any]] = defaultdict(set)  # sets of Events to wait for
```

I still lean to it being better to drop the cast-through-list and have it be a set everywhere.

thomaslima: I agree. This is a bigger change that I don't feel comfortable making. Your call. In this particular case, I think the defaultdict and `.update` is a better pattern.

coretl:
@tacaswell the reason mypy didn't pick up the issue is because it only checks type annotated functions, which makes it easier to adopt gradually but gives you a false sense of security. 

Changing the signature to:
```python
async def _wait(self, msg) -> None:
```
triggers checking of the function correctly:
```
type-checking run-test: commands[0] | mypy src
src/bluesky/run_engine.py:2322: error: Need type annotation for "watch_futs" (hint: "watch_futs: list[<type>] = ...")  [var-annotated]
src/bluesky/run_engine.py:2337: error: Incompatible types in assignment (expression has type "list[Any]", target has type "set[Any]")  [assignment]
src/bluesky/run_engine.py:2356: error: No return value expected  [return-value]
Found 3 errors in 1 file (checked 88 source files)
```

I suggest we move to pyright that checks everything, however there will be a lot of things to fix:
```
767 errors, 17 warnings, 0 informations
```
@thomaslima if I make a branch with the type checker switched to pyright, would you be interested in contributing your typing changes to that?

@tacaswell I'll have a go at a test for this and add to this PR

src/bluesky/run_engine.py
```diff
@@ -2353,7 +2353,7 @@ def cancel_status_task_if_error(fut: asyncio.Future[list[asyncio.Future]]):
                     if done:
                         self._call_waiting_hook(None)
                         self._seen_wait_and_move_on_keys.remove(group)
-            return done
+        return done
```
tacaswell, line 2358: just verifying that this is intentional and fixes a lingering bug where sometimes we would return `None`?

tacaswell, line 2358: If there are no futures, are we done?

```suggestion
        else:
            done = True
        return done
```

coretl: Yes

rtuck99:
Would be good to have a release with this fix as it is currently affecting our unit tests in mx-bluesky
https://github.com/DiamondLightSource/mx-bluesky/actions/runs/18313358716/job/52147121033?pr=1335

PR #1959, opened 2025-10-06
Mnt/drop py39

tacaswell:
## Description

Python 3.9 goes EOL upstream this month. 

## Motivation and Context

We should not support versions of Python that are no longer getting security updates (leaving aside if we should adopt a more aggressive schedule (which I think we should))

Used ruff to fix most of the use of `Union`, but not clear if using `|` is valid when defining type aliases.

## How Has This Been Tested?

CI


jacopoabramo:
> Used ruff to fix most of the use of Union, but not clear if using | is valid when defining type aliases.

If you mean

```pyt
MyType = int | None
```

then yes it is accepted.

PR #1960, opened 2025-10-06
Remove the 'streams' Namespace

genematx:
Remove the 'streams' namespace (container) from the container structure created by TiledWriter.


tacaswell:
How does this have to be coordinated with other changes?  Do we want to take this opportunity to deprecate `TiledWriter` in bluesky and move it to databroker or tiled (or where ever)?

danielballan:
Clients using bluesky-tiled-plugins will need. https://github.com/bluesky/databroker/pull/863

> Do we want to take this opportunity to deprecate TiledWriter in bluesky and move it…?

No, we aim to release and deploy this in maint (the next two days) to get it done ahead of work that would increase the blast radius of this change.

I will make a new issue.

PR #1950, opened 2025-09-08
Update ZMQ classes to support IPC transport

mrakitin:

PR #1961, opened 2025-10-06
Enh/ Add support for CURVE encryption for ZMQ proxy in TCP mode. Add unit tests for zmq proxy and zmq proxy CLI. Remove depracated options for ZMQ Proxy.

tacaswell:
## Description

Add curve support to TCP connections in the zmq callbacks and allow the zmq-proxy process to either bind or connect for either the in our out port.

## Motivation and Context

Enabling Curve allows for secure communication between hosts on the network.  This is based on https://github.com/zeromq/pyzmq/blob/main/examples/security/ironhouse.py (and the other examples in that directory).

Allowing the connections on the proxy to connect rather than bind allows them to act in both a "proxy" and "bridge" roll.  This allow us to create topologies like

```

RE --ipc--> proxy1 <--|tcp+curve|-- proxy2 --ipc--> proxy3 <--ipc-- consumer

```

where only proxy1 and proxy2 have to be exposed to the network / know curve secrets and the both the RE and the consumers only have to know about local named pipes (`ipc://`) for connections.

This builds on #1950 which should be merged first.

## How Has This Been Tested?

Manual testing, still needs automated testing.  Still needs docs.


jwlodek:
The CI failure on windows/3.12 appears to be unrelated.

src/bluesky/tests/test_zmq.py
```diff
@@ -368,8 +435,203 @@ def test_zmq_RD_ports_spec(host):
...
+    [
+        (True, False, False),
+        (False, True, False),
+        (False, False, True),
+    ],
+)
+def test_proxy_fails_to_configure_sockets(
+    mocker: MockerFixture, context_err: bool, frontend_err: bool, backend_err: bool
+):
+    """Checks to make sure we are properly cleaning up resources. on failure"""
```
mrakitin:
```suggestion
    """Checks to make sure we are properly cleaning up resources on failure"""
```

mrakitin (approved):
Approved. Minor docstring correction would not hurt.

Issue #1965, opened 2025-10-13
Enhancing type annotations. Using pyright as type checker


thomaslima:
I am creating this issue to track progress in the type annotations in the pyright branch.

Per @coretl 's request in #1955 , I spent some time fixing some type annotations that I had in my local repo. The main issues that I had was that implementing a protocol (say Stageable and Readable) was not sufficient to use the `count` plan. Implicit in that plan was that devices need parent and names etc. I had no warning in my IDE until I was ready to launch an experiment.

It turns out there might be other issues hidden in the codebase that pyright may be able to catch.

PR #1966, opened 2025-10-13
Enhancing type annotations. Using pyright as type checker

thomaslima:
This PR is a work in progress at this stage. A first attempt to tackle #1965 .

@coretl Is this what you had in mind regarding improvements? I spent some hours studying the code and implementing some issues I encountered in my usage of bluesky. (We use bluesky without ophyd or suitcase in my organization, we made our own library of devices/database hooks).

Please let me know if the edits are going in the direction you had in mind, before I can proceed.

Thanks!

### DIFF
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import asyncio
 import concurrent
 import copy
@@ -16,6 +18,7 @@ from datetime import datetime
 from enum import Enum
 from inspect import iscoroutine
 from itertools import count
+from typing import TYPE_CHECKING, Protocol
 from warnings import warn
 
 import event_model
@@ -173,6 +176,37 @@ class RunEngineStateMachine(StateMachine):
         ]
 
 
+if TYPE_CHECKING:
+
+    class StateHook(Protocol):
+        """Type stub for the state_hook variable of RunEngine."""
+
+        def __call__(self, new_state, old_state) -> None: ...
+
+    class _RunEngineState(str):
+        """Type stub for the state returned by RunEngine._state / RunEngine.state."""
+
+        is_idle: bool
+        is_running: bool
+        is_pausing: bool
+        is_paused: bool
+        is_halting: bool
+        is_stopping: bool
+        is_aborting: bool
+        is_suspending: bool
+        is_panicked: bool
+        can_pause: bool
+        can_be_idle: bool
+        can_be_running: bool
+        can_be_pausing: bool
+        can_be_paused: bool
+        can_be_halting: bool
+        can_be_stopping: bool
+        can_be_aborting: bool
+        can_be_suspending: bool
+        can_be_panicked: bool
+
+
 class LoggingPropertyMachine(PropertyMachine):
     """expects object to have a `log` attribute
     and a `state_hook` attribute that is ``None`` or a callable with signature
@@ -181,7 +215,7 @@ class LoggingPropertyMachine(PropertyMachine):
     def __init__(self, machine_type):
         super().__init__(machine_type)
 
-    def __set__(self, obj, value):
+    def __set__(self, obj: RunEngine, value):
         own = type(obj)
         old_value = self.__get__(obj, own)
         with obj._state_lock:
@@ -193,7 +227,7 @@ class LoggingPropertyMachine(PropertyMachine):
         if obj.state_hook is not None:
             obj.state_hook(value, old_value)
 
-    def __get__(self, instance, owner):
+    def __get__(self, instance: RunEngine | None, owner):  # pyright: ignore[reportIncompatibleMethodOverride]
         if instance is None:
             return super().__get__(instance, owner)
         with instance._state_lock:
@@ -380,12 +414,13 @@ class RunEngine:
         "remove_suspender",
         "_start_suspender",
     ]
+    state_hook: StateHook | None
 
     RunBundler = RunBundler
 
     @property
-    def state(self):
-        return self._state
+    def state(self) -> _RunEngineState:
+        return self._state  # pyright: ignore[reportReturnType]
 
     @property
     def deferred_pause_requested(self):
@@ -708,11 +743,13 @@ class RunEngine:
 
     @property
     def verbose(self):
-        return not self.log.disabled
+        # TODO Include test coverage for this property.
+        return not self.log.logger.disabled
 
     @verbose.setter
     def verbose(self, value):
-        self.log.disabled = not value
+        # TODO Include test coverage for this property.
+        self.log.logger.disabled = not value
 
     @property
     def call_returns_result(self):
@@ -925,7 +962,7 @@ class RunEngine:
                 raise RuntimeError(text)
 
         # If we are in the wrong state, raise.
-        if not self._state.is_idle:
+        if not self.state.is_idle:
             raise RuntimeError(f"The RunEngine is in a {self._state} state")
 
         futs = []
@@ -1005,9 +1042,9 @@ class RunEngine:
             raise RuntimeError("The RunEngine is panicked and cannot be recovered. You must restart bluesky.")
 
         # The state machine does not capture the whole picture.
-        if not self._state.is_paused:
+        if not self.state.is_paused:
             raise TransitionError(
-                f"The RunEngine is the {self._state} state. You can only resume for the paused state."
+                f"The RunEngine is the {self.state} state. You can only resume for the paused state."
             )
 
         self._interrupted = False
@@ -1230,7 +1267,7 @@ class RunEngine:
                 self._interrupted = True
                 with self._state_lock:
                     self._exception = FailedPause()
-                was_paused = self._state == "paused"
+                was_paused = self.state == "paused"
                 self._state = "aborting"
                 if not was_paused:
                     self._task.cancel()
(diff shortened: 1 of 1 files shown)


src/bluesky/protocols.py
```diff
@@ -219,7 +220,7 @@ def trigger(self) -> Status:
 @runtime_checkable
 class Preparable(Protocol):
     @abstractmethod
-    def prepare(self, value) -> Status:
+    def prepare(self, *args, **kwargs) -> Status:
```
coretl, line 208: Although the code is written to support `*args` and `**kwargs` I would prefer to keep the protocol narrower so it only takes a single positional arg like:
https://github.com/bluesky/bluesky/blob/8adfe3ab1266fbb2555de7e3f906f79cdeb2d704/src/bluesky/protocols.py#L347-L352

src/bluesky/protocols.py
```diff
@@ -521,13 +531,64 @@ def hints(self) -> Hints:
...
+
+
+@runtime_checkable
+class HintedMovable(Movable[T_co], HasHints, Protocol):
     """A movable object that has a name and hints."""
 
-    ...
 
+# TODO if this is too ugly, just include HasParent into all devices
+# Adding this here to make it easy to search the codebase for places where it is needed.
```
coretl, line 546: Hmm, I wonder if we should include `HasParent` and `HasName` into all the protocols? Would that remove all the ugliness here?

coretl: We could really do with an `Intersection`, but the PEP for that isn't written yet: https://github.com/python/typing/issues/213

src/bluesky/protocols.py
```diff
@@ -537,9 +598,24 @@ def check_supports(obj: T, protocol: type[Any]) -> T:
...
+
+    For multiple protocols, use protocol inheritance instead (see examples
+    in this file like ReadableAndTriggerable).
     """
     assert isinstance(obj, protocol), "%s does not implement all %s methods" % (obj, protocol.__name__)  # noqa: UP031
-    return obj
+    # The isinstance check above ensures obj implements the protocol at runtime.
+    # This cast informs the type checker that obj now has the protocol's type.
+
+    return cast(T, obj)
```
coretl: Is this cast actually required? I thought the `isinstance` above did the type inference for us?

src/bluesky/utils/__init__.py
```diff
@@ -790,6 +795,7 @@ def separate_devices(devices):
     """
     result = []
     for det in devices:
+        check_supports(det, HasParent)
```
coretl, line 795: Was this for typing or runtime? If for typing, wouldn't you need to do this to get the type narrowing?
```suggestion
    for d in devices:
        det = check_supports(d, HasParent)
```

src/bluesky/plans.py
```diff
@@ -64,7 +87,7 @@ def derive_default_hints(motors: list[Any]) -> dict[str, Sequence]:
 
 
 def count(
-    detectors: Sequence[Readable],
+    detectors: Sequence[ChildReadableAndStageable],
```
coretl, line 90: I think `Stageable` is optional, so this is just `ChildReadable`

src/bluesky/plans.py
```diff
@@ -130,8 +153,8 @@ def inner_count() -> MsgGenerator[str]:
 
 
 def list_scan(
-    detectors: Sequence[Readable],
-    *args: tuple[Union[Movable, Any], list[Any]],
+    detectors: Sequence[ChildReadableAndStageable],
```
coretl, line 156: ditto for all other instances of `ChildReadableAndStageable`

src/bluesky/plans.py
```diff
@@ -671,16 +697,16 @@ def inner_relative_log_scan():
 
 
 def adaptive_scan(
-    detectors: Sequence[Readable],
+    detectors: Sequence[ChildReadableAndStageable],
     target_field: str,
-    motor: NamedMovable,
+    motor: NamedChildMovableAndStageable,
```
coretl, line 702: And this

src/bluesky/plans.py
```diff
@@ -1866,9 +1895,9 @@ def inner_relative_spiral_fermat():
 
 
 def spiral(
-    detectors: Sequence[Readable],
-    x_motor: NamedMovable,
-    y_motor: NamedMovable,
+    detectors: Sequence[ChildReadableAndStageable],
+    x_motor: HintedMovable,
+    y_motor: HintedMovable,
```
coretl, line 1904: Again, I think the hints are optional

thomaslima:
@coretl Thanks for getting back to me. I will try to work on these when I have free time next. `Stageable` was a key issue in my codebase. Some plans (like `count`) `check_supports` for `Stageable` and then proceeds to `separate_devices` which then requires a parent. This tripped me up several times where I made a `stage()` method but forgot to add a dummy `parent` attribute.

src/bluesky/utils/__init__.py
thomaslima: This is for runtime. I propose to fail early with a good message (`check_supports` has a nice error message). The current Exception just complains that the object has no parent when it reaches `ancestry`. This would be removed of course if Stageable had a parent by default. Or if `separate_devices` is fixed such that it treats devices with no parents as root by default.

src/bluesky/protocols.py
thomaslima: the isinstance does nothing for typing outside the function's scope. This said, this particular cast may be unnecessary because we annotate the return type with `T`. I need to test this. (btw cast is very fast at runtime).

src/bluesky/protocols.py
thomaslima: I miss Intersection too! But my issue is that some plans and plan stubs are not consistent with the protocols. I like the clean small protocols like you, so I think there could be three phases to this refactoring:
1. Manually create all typing Intersections.
2. See where pyright detects errors within plans and plan stubs and change the annotations there (it will be ugly in the meantime).
3. Once pyright clears all errors, then we can investigate where the codebase makes too strong assumptions. Like the `stage_wrapper` -- it doesn't really need a parent. We fix the assumptions and revert the annotation back to the atomic protocols.
4. Once the cleanup is fixed, we can delete the temporary Intersection types.

In this PR, my plan was to start with 1 and 2, but only if some of the core developers will invest time in step 3.

src/bluesky/plans.py
thomaslima: I see. I just looked again and this tripped me up. the stage_decorator will attempt to stage every detector but it will give up at the last minute in the `RunEngine._stage` method. Question before I proceed. Following your suggestion, the `stage_all` method or the `stage_wrapper` shouldn't be typed to require Stageable devices, right? https://github.com/bluesky/bluesky/blob/6875901f0662d5bd340c41509a85ab1ba8737a07/src/bluesky/plan_stubs.py#L1080-L1086

I propose to check if a device is stageable at the plan decorator level before creating a stage message in the plan.

coretl: I think this is fine, but @tacaswell may have other opinions

src/bluesky/protocols.py
coretl: That seems like a good plan to me, I'm on board as long as @tacaswell will review the changes...

src/bluesky/protocols.py
coretl: I think the cast is unnecessary here:
https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKcANFACqICmAagIYhkAK4MYAxmADYBQP5UALwVq9EAAoAROUkBKPgBMqwKOwAWVdgGsA%2BgGcArghx494sACMAVgC5CxMglYdud%2BAioBtcgF1ZUAFoAPgobHigIqDo9PSo8TD1UPRg6FHYqc2tHZ04uWTJJAFI9KAUwKhKUMHwsBC4qCCoUfDouLihiqEaYNTAFPUkOqEyrbOqXLgA6HR0UOkaZ%2BUioECoYAxAUKEsrPnYuaJKAJSo6BToLevEWcdzZMOWlFVWz8ViuYHuoSZ%2B9g5ioABZOAnM4XeoPSJPFanBRvKgfL4-SZ-Q5AuAAOWqoPOlyokIi0KqMB0Lzh70%2BdmRfAAHkJVBptPojCYYGZgTjwRl8lBOXj5KsAG6nLg6dwZGnyHh04TqTS6QzGXBs8TArEwPlXHmaqgCqjC1pi6jiSU8IA

src/bluesky/plans.py
tacaswell: That seems reasonable to me.  I am also OK with adding a deprecation to the inside of `_stage` to move the filtering higher, but I don't think we should do that in this PR.

src/bluesky/plans.py
thomaslima: This was not my doing. Previously, this was called `NamedMovable`, but the protocol for `NamedMovable` was `NamedMovable(Movable[T_co], HasHints, Protocol)`, I just renamed it to `HintedMovable(Movable[T_co], HasHints, Protocol)` for consistency because I needed an actual `NamedMovable`.

src/bluesky/protocols.py
thomaslima: I don't understand. In your example, `set` can take multiple arguments.

src/bluesky/protocols.py
thomaslima: Resolved in b514408

Issue #1962, opened 2025-10-07
Should TiledWriter move out of bluesky?


danielballan:
@tacaswell raised the question whether `bluesky.callbacks.tiled_writer` should move somewhere else.

There has been some friction, but as everything is stabilizing (tiled 0.1.0 is minted, bluesky-tiled-plugins 2.0.0 is around the corner) I think that will be less of an issue than it has been.

The `bluesky.callbacks` package has always been a home for code integrating the document model and external libraries: Matplotlib, ZeroMQ, etc. Tiled (much!) less mature than these other examples, but it is now settling.

The TiledWriter is tightly connected to "consolidation" utilities, also developed in the Bluesky repo, that could have applications beyond tiled. Moving TiledWriter outside blueksy would increase the coordination required to make changes to those utilities.

Where else could we put TiledWriter? 

- tiled: We have avoided putting the word "bluesky" anywhere in the codebase, to make it generic to arrays and tables and to make that feature very clear to potential users. There is no event-model dep and no direct integration with documents.
- databroker: Not a good home because it is destined for deprecation. The project can be put into maintenance mode once the MongoDB-to-SQL migration is done, only needed for reading clients who want the back-compat wrapper for Tiled.
- bluesky-tiled-plugins: This is probably the best alternative. It is a lean package, currently developed within the databroker repo, providing client conveniences for looking at Bluesky data is tiled.

My view is that, in spite of past churn, TiledWriter is OK where it is, and it is not worth adding more churn by moving it. But these are our other options.

PR #1968, opened 2025-10-28
Data source patch

genematx:


danielballan:
Note: This is older work which was used in July at SMI, opened as PR so we don't lose track of it. It may be that this never gets merged, as work is in progress to move TiledWriter and Consolidator development into a different repository, as discussed in #1962.

danielballan:
Superseded by https://github.com/bluesky/databroker/pull/867 as TiledWriter is moving from bluesky to bluesky-tiled-plugins.

PR #1972, opened 2025-11-20
Squash "Future exception was never retrieved" at teardown

coretl:
Fixes https://github.com/bluesky/ophyd-async/issues/1140


tacaswell:
I continue to worry that this is masking bugs other places in our management of the futures, but for the way we are using the future (an event that can also carry a success/fail flag).

tacaswell:
I think this is an acceptable risk.

PR #1978, opened 2026-01-07
Read configuration for a device once per stream

oliwenmandiamond:
## Description

Rewritten cache for a device (`Readable`, `Collectable`, `Configurable`) to be per stream rather than per run. This is done by grouping relevant cache into a dataclass to make managing more easy and logical.

## Motivation and Context

Fixes https://github.com/bluesky/bluesky/issues/1899

## How Has This Been Tested?
- Added additional tests to check that new `read_configuration` is called once per stream to have the desired behaviour linked in issue. 
- Added additional test to check that the cache is used the expected number of times. 
- Checked existing tests are still passing.

### DIFF
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -1217,7 +1217,7 @@ class RunBundler:
                 await self._prepare_stream(name, obj_set)
                 continue
 
-    async def _cache_describe_collect(self, obj: Collectable):
+    async def _cache_describe_collect(self, obj):
         "Read the object's describe_collect and cache it."
         if obj not in self._describe_collect_cache:
             obj = check_supports(obj, Collectable)


src/bluesky/bundlers.py
```diff
@@ -250,13 +250,21 @@ async def _prepare_stream(
             list(objs_dks),
         )
 
+    def _needs_new_config_cache(self, obj) -> bool:
+        """Check if the bundle name is in the descriptors to check if we have done this before with this stream"""
+        # _bundle_name is stream name for read, is None on save. Don't need new cache on save.
+        if self._bundle_name is None:
```
tacaswell: This feels brittle.  `_ensure_cached` is private, lets add another flag to its siganture or split it up into a couple of helper functions.

oliwenmandiamond: Would something like this be better?

```Python
    async def _ensure_cached(self, obj, collect=False):
        coros = self._get_readable_and_collectable_obj_cache(obj, collect)
        coros = coros + self._get_configurable_obj_cache(obj)
        await asyncio.gather(*coros)

    def _get_readable_and_collectable_obj_cache(self, obj, collect: bool) -> list[Coroutine]:
        coros = []
        if not collect and obj not in self._describe_cache:
            coros.append(self._cache_describe(obj))
        elif collect and obj not in self._describe_collect_cache:
            coros.append(self._cache_describe_collect(obj))
        return coros

    def _get_configurable_obj_cache(self, obj) -> list[Coroutine]:
        coros = []
        is_new_stream = self._bundle_name is not None and self._bundle_name not in self._descriptor_objs
        if obj not in self._config_desc_cache or is_new_stream:
            coros.append(self._cache_describe_config(obj))
            coros.append(self._cache_read_config(obj))
        return coros
```
If not, could you please be more specific?

oliwenmandiamond:
I see, because if we do stream1, then stream2, then back to stream1, the cache again would be wrong because the cache should be pointing to stream1, it would now stay at the last stream (stream2) where as the original issue it was only the first stream it was used for (stream1).

src/bluesky/bundlers.py
```diff
@@ -317,8 +349,8 @@ async def create(self, msg):
                 "bundle is closed with a 'save' or "
                 "'drop' message."
             )
-        self._read_cache.clear()
-        self._asset_docs_cache.clear()
```
coretl, line 321: Is this the problem?

oliwenmandiamond: Yeah, I believe it was! Thank you

src/bluesky/bundlers.py
oliwenmandiamond: I have rewritten this to implement Tom Cobb's suggestion. Can you please have another look?

src/bluesky/bundlers.py
```diff
@@ -295,6 +360,14 @@ async def declare_stream(self, msg):
 
         return await self._prepare_stream(stream_name, objs_dks)
 
+    def _set_current_stream_cache(self, stream_name: str):
```
coretl, line 354: Unfortunately I think holding the current stream cache is still fragile...

src/bluesky/bundlers.py
```diff
@@ -744,9 +783,9 @@ async def _describe_collect(self, collect_object: Flyable):
             }
 
         """
-        await self._ensure_cached(collect_object, collect=True)
+        await self._current_stream_cache.ensure_cached(collect_object, collect=True)
```
coretl: ...because then we do things like use it in collect, where there is no concept of the current stream.

Edge case:
- predeclare obj1 in stream1
- predeclare obj2 in stream2
- collect obj1, but find that _current_stream_cache is stream2

The only fix I can think of is:
- Don't use _current_stream_cache in collect
- Keep a _multi_stream_collect: dict[Any, list[str]] mapping collect objects to the list of stream names their describe_collect published
- If stream_name given then lookup the cache using that stream_name
- Elif obj in _multi_stream_collect then get the stream names from that
- else call describe_collect, check it is doubly nested and create the stream caches from it, adding the object to _multi_stream_collect

This is getting messy...

@tacaswell any other suggestions?

oliwenmandiamond: Okay, so I now `describe_collect_cache` which is not in StreamCache now. 

I think this also "works" but I have left some comments in the code in parts I'm not entirely sure about. Also just to your _multi_stream_collect doesn't seem needed as we already have declared_streams which does the same thing

src/bluesky/bundlers.py
tacaswell: It is not legal to open more than one create/save pair at the same time  so I think this is OK (we have to have some state the lives across a create / read / read / ... / save set of messages.  If we are going to have the per-stream caching of descriptor etc then this is about the only option.

src/bluesky/bundlers.py
```diff
@@ -1083,6 +1119,7 @@ async def collect(self, msg: Msg):
                         "the objects first and provide the stream name"
                     )
                 else:
+                    print(f"stream_name: {stream_name}")
```
tacaswell:
```suggestion
```

src/bluesky/bundlers.py
```diff
@@ -1107,6 +1144,7 @@ async def collect(self, msg: Msg):
         indices_difference = await self._pack_external_assets(
             collected_asset_docs, message_stream_name=stream_name
         )
+        print(f"stream_name: {stream_name}")
```
tacaswell:
```suggestion
```

PR #1984, opened 2026-02-04
STY: use modern type syntax

tacaswell:
This will unbreak linting on all PRs.


danielballan (approved):

tacaswell:
I don't get it locally and then got distracted by other things.

Issue #1989, opened 2026-03-12
_set_span_msg_attributes causes OTel WARNING for numpy scalar types in msg.args


Alexj9837:
When a plan moves a motor to a position derived from a numpy array (e.g. np.linspace scan positions), bluesky's _set_span_msg_attributes produces the following warning on every set message:
  WARNING opentelemetry.attributes Invalid type float64 in attribute 'msg.args' value sequence. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or None
## Acceptance Criteria
   The above warning no longer appears when running a motor scan plan with numpy scan positions

current 
```
def _set_span_msg_attributes(span, msg):
    span.set_attribute("msg.command", msg.command)
    span.set_attribute("msg.args", msg.args)
    span.set_attribute("msg.kwargs", json.dumps(msg.kwargs, default=repr))
    span.set_attribute("msg.obj", repr(msg.obj)) if msg.obj else span.set_attribute(
        "msg.no_obj_given", True
    )
```

fix 
```
def _set_span_msg_attributes(span, msg):
    span.set_attribute("msg.command", msg.command)
    span.set_attribute("msg.args", json.dumps(msg.args,default=repr))
    span.set_attribute("msg.kwargs", json.dumps(msg.kwargs, default=repr))
    span.set_attribute("msg.obj", repr(msg.obj)) if msg.obj else span.set_attribute(
        "msg.no_obj_given", True
    )
```

PR #1990, opened 2026-03-12
fix: serialise msg.args with json.dumps to avoid OTel WARNING for num #1989

Alexj9837:
## Description
Serialise msg.args using json.dumps(..., default=repr) to match the existing handling of msg.kwargs

## Motivation and Context
msg.args contains numpy.float64 values. OTel validates attribute values using exact type checks and rejects numpy.float64 even though it behaves like a float, causing the following warning on every set message and silently dropping the attribute:
https://github.com/bluesky/bluesky/issues/1989

## How Has This Been Tested?
Ran the full bluesky test suite, also ran the changes again blueapi tests and the changes removed the error message.


tacaswell:
I'm going to close this a (pre-) superseded.

Issue #1654, opened 2024-01-23
Separate user interaction logic of RE from task running logic


callumforrester:
The `RunEngine` currently executes a plan through its `__call__` dunder which handles user interaction logic (argument wrangling et al.) _and_ task running logic (actually interpreting the plan and talking to hardware).

## Possible Solution
Create a promoted, public version of `RunEngine._run()` which `__call__` calls _after_ wrangling the arguments. The public `run()` (assuming we use that name) kicks off a separate asyncio task and includes some facility for restricting vocabulary (messages the RE can interpret).

## Context
Separating these two is useful for #1652 because it paves a path to the RE running two asyncio tasks simultaneously. The more machine-friendly interface will also be useful for headless bluesky applications such as queueserver.
The restricted vocabulary is useful for both of the above cases as it facilities ensuring that some plans are run in a headless mode by disallowing "wait for user input"-style messages.

tacaswell:
I think the public `RE.run` should be as plain of a `async def` as possible and leave managing it in a task or simply `await`'d to the caller.

dperl-dls:
I'm having a look at this based on https://github.com/bluesky/ophyd/issues/1121#issuecomment-1579146876 as well as this issue description.
From the linked comment:

> The refactoring I think we need on the RE is:
> 
>     1. wrap `_run` and the message stacks, "continue permit" logic, and  into a class (like we did for the bundlers).  This should mostly be state-shuffling and pushing RE-singleton state into state on an object the RE holds
> 
>     2. making the dictionary of allowed messages a thing that is passed into `_run` rather than pulled from the parent (this will require a bit of management around how the ability to register messages on the fly works, but should still be possible.
> 
>     3. add a message that creates new sub `_run`s on demand.
> 

This issue will just be 1 and 2 of the above.

Some questions:
- should the `_state` be per-run or for the `RE`?
- `_run` makes a lot of reference to the state - should each run be able to access it?
- if the run is paused, which objects should it stop? the ones it has seen itself, or all that anyone has seen?

tacaswell:
[...]

Issue #1652, opened 2024-01-16
Allow RE to execute sub plans in parallel


whs92:
Currently the run engine does one task at a time. There are use cases where we would like to ask the run engine to "run this plan/these plans and tell me when you are done". Example use cases are:

- Running alignment or calibration procedures which are independent of each other while also preparing a robot to select a new sample
- Change beamline energy while a sample robot get's a new sample
- Move/Spin a sample in a given trajectory while running a measurement technique
- Arm multiple detectors at the same time
- https://bliss.gitlab-pages.esrf.fr/bliss/1.10.x/motion_hook.html 

In each of these cases there are different ways to solve the problem. Execution could be handled in Ophyd, or in EPICS/Tango/etc layer or in the device itself. This is discussed in https://github.com/bluesky/ophyd/issues/1121#issuecomment-1579146876. @tacaswell touched on the problem some time ago in https://github.com/bluesky/bluesky/pull/541

Suggested reading by @tacaswell: https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/

## Proposed Changes

- A new message added to the RE which takes as arguments a plan and a set of verbs that are allowed to be used. This allows us to have some control over what plans are allowed and not. `async def _run_sub(self, plan, verbs)` should return a status object
- RE should be modified to be able to run sub_plans in parrallel in a way that is backwards compatible.

DominicOram:
As discussed, for a lot of these cases we can get away with a very small subset of verbs that the `_run_subs` is allowed. I suggest if whoever does the implementation finds that it's more complex to allow the verbs that interact with the Event system they spin those into a new issue

tacaswell:
The Msg should return a status object to the calling plan and it should particpate in the same wait logic that set and trigger do so you can write things like:

```python
def my_plan():
   ...
   yield from run_sub_plan(adjust_energy(), group='setup')
   yield from run_sub_plan(change_sample_with_robot(), group='setup')
   yield from bps.set(detector_pos, 5, group='setup')
   yield from bps.wait(group='setup')
   ...
   yield from bp.count(...)
```

other design questions:
[...]

PR #1992, opened 2026-03-18
MNT: replace `super_state_machine` with custom FSM

jacopoabramo:
## Description
This PR removes the dependency on `super-state-machine` using a custom written FSM which maintains the original behavior with the addition that the descriptor `__get__` returns the actual FSM rather than a `ProxyString`.

## Motivation and Context

I can't find the relevant issue for it (if there's any) but the main drive for the PR is [this comment](https://github.com/bluesky/bluesky/issues/1679#issuecomment-2037301596). Additionally, to allow the `RunEngine` to run concurrent plans (see #1652) in an eventual future class dispatched to execute a single plan (an initial attempt is currently ongoing in #1991 and the state machine is the blocking part...), the state must be globally shared between the `RunEngine` and the task runner (the idea is taken from #1654, in particular [this comment](https://github.com/bluesky/bluesky/issues/1654#issuecomment-2264995978)).

Some pseudo-code on how to do this:

```py

class TaskRunner:
   
   # when creating a new task runner, the FSM
   # must be provided at initialization
   _state = MachineDescriptor()
   
   def __init__(self, fsm: RunEngineStateMachine):
       self._state = fsm

class RunEngine:

    # the actual FSM is created in the RunEngine and passed
    # to the TaskRunner at creation
    _state = MachineDescriptor(RunEngineStateMachine)
    
    def create_runner(self) -> TaskRunner
        return TaskRunner(fsm=self._state)
```

*Note*: I moved the whole state machine logic into a public `fsm.py` module to prevent circular imports. I noticed too late that [`blueapi` uses `super-state-machine` for type hints](https://github.com/DiamondLightSource/blueapi/blob/97b43cec747b22c94de266e05db286ba98f81aba/src/blueapi/worker/event.py#L11), so there are naming mismatches which I can address.

## How Has This Been Tested?

Tested locally on Windows.


src/bluesky/fsm.py
```diff
@@ -0,0 +1,195 @@
...
+        # the class must provide the machine
+        # instance itself before any access
+        state = MachineDescriptor()
+
+        def __init__(self, fsm: MyStateMachine | None = None):
+            self.state = fsm or MyStateMachine()
+    """
+
+    def __init__(self, fsm_type: type[RunEngineStateMachine] | None = None) -> None:
+        self._fsm_type = fsm_type
```
tacaswell, line 149: Why do we need this flexibility?

src/bluesky/fsm.py
```diff
@@ -0,0 +1,195 @@
...
+        try:
+            return self._memory[obj]
+        except KeyError as exc:
+            if self._fsm_type is None:
+                raise RuntimeError(f"{obj!r} has no FSM assigned and no FSM type was provided.") from exc
+            fsm = self._fsm_type()
+            self._memory[obj] = fsm
+            return fsm
+
+    def __get__(self, obj: "RunEngine | None", owner: type) -> RunEngineStateMachine:
```
tacaswell: might be worth a `@override` here to handle the literal `None` case correctly.

src/bluesky/fsm.py
jacopoabramo: To keep the current descriptor approach of declaring the state machine and be able to pass it to a future SingleRunExecutor, see the comments in #1991 

I'm open to suggestions on having a different approach. I wanted to keep it as close as possible to the current FSM implementation.

tacaswell: ok, the lets atleast drop the `None` default so that we don't need the error checking about that later.

src/bluesky/fsm.py
```diff
@@ -0,0 +1,199 @@
...
+    def __get__(self, obj: None, owner: type) -> "MachineDescriptor": ...
+    @overload
+    def __get__(self, obj: "RunEngine", owner: type) -> RunEngineStateMachine: ...
+    def __get__(self, obj: "RunEngine | None", owner: type) -> "RunEngineStateMachine | MachineDescriptor":
+        if obj is None:
+            return self
+        with obj._state_lock:
+            return self._get_or_create(obj)
+
+    def __set__(self, obj: "RunEngine", value: str | REState | RunEngineStateMachine) -> None:
```
tacaswell: Mixing using `__set__` to both change the state and to change the state machine is too much in my view.

Rather than have objects that can change their state machine, we should have objects with a fixed lifespan that we discard.

jacopoabramo: > Mixing using __set__ to both change the state and to change the state machine is too much in my view.

I also agree that it's not ideal, but for a future minimal SRE that only has the contents of _run the FSM has to come from somewhere. I would also much prefer that the descriptor simply manipulates the state.

Maybe adding an additional method to the FSM so that it can be fetched? I don't know...

src/bluesky/fsm.py
jacopoabramo: > ok, the lets atleast drop the `None` default so that we don't need the error checking about that later.

You mean the init should be

```py
def __init__(self, fsm_type: type[RunEngineStateMachine] | None) -> None:
```

or

```py
def __init__(self, fsm_type: type[RunEngineStateMachine]) -> None:
```

just like it is now?

PR #1995, opened 2026-03-23
1994-executable_protocol

burkeds:
Addresses issue https://github.com/bluesky/bluesky/issues/1994.

Introducing `bluesky.protocols.Executable` which calls `execute` to run an RPC.
@coretl


thopkins32 (commented):
Changes look good to me for the most part. I don't have a strong opinion on the Msg.args and Msg.kwargs warning, it is technically not deprecated yet.

I agree with @jacopoabramo that we should create a new `Status` protocol that holds the result, similar to https://docs.python.org/3/library/asyncio-future.html#asyncio.Future which has `set_result()` and `result()` to fetch the value returned (with similar error semantics).

src/bluesky/run_engine.py
```diff
@@ -2244,6 +2246,28 @@ async def _set(self, msg):
...
+        Expected message object is:
+
+            Msg('execute', obj, *args, **kwargs)
+        """
+        obj = check_supports(msg.obj, Executable)
+        args = msg.args
+        kwargs = dict(msg.kwargs)
+        warn_if_msg_args_or_kwargs(msg, obj.execute, args, kwargs)
+        wait = kwargs.pop("wait", False)
+        group = kwargs.pop("group", None)
```
coretl: This means that no RPC server could ever take the arguments `wait` or `group`. This seems like it might bite us. Maybe we should have `execute_args` and `execute_kwargs` as keys in `kwargs` and pass that down?

src/bluesky/run_engine.py
```diff
@@ -2244,6 +2246,28 @@ async def _set(self, msg):
...
+        Execute an RCP on a device and cache the returned status object.
+
+        Expected message object is:
+
+            Msg('execute', obj, *args, **kwargs)
+        """
+        obj = check_supports(msg.obj, Executable)
+        args = msg.args
+        kwargs = dict(msg.kwargs)
+        warn_if_msg_args_or_kwargs(msg, obj.execute, args, kwargs)
```
coretl: This is where the deprecation warning is coming from. Just delete the line and it will go away...

```suggestion
```

src/bluesky/plan_stubs.py
```diff
@@ -604,6 +605,44 @@ def trigger(
...
+    kwargs :
+        passed through to ``obj.execute()``
+
+    Yields
+    ------
+    msg : Msg
+        Msg('execute', obj, *args, **kwargs)
+
+    """
+    return (yield Msg("execute", obj, *args, **kwargs, group=group, wait=wait))
```
coretl: Which would be passed down here

```suggestion
    return (yield Msg("execute", obj, execute_args=args, execute_kwargs=kwargs, group=group, wait=wait))
```

This means we can't specify `wait` or `group` as `execute_kwargs` via the plan stub, but we could construct one manually if we needed to.

src/bluesky/run_engine.py
```diff
@@ -2244,6 +2246,28 @@ async def _set(self, msg):
...
+        wait = kwargs.pop("wait", False)
+        group = kwargs.pop("group", None)
+        if wait:
+            ret = await maybe_await(obj.execute(*msg.args, **kwargs))
+            return ret
+        else:
+            ret = obj.execute(*msg.args, **kwargs)
+            self._add_status_to_group(obj=obj, status_object=ret, group=group, action="execute")
+            return ret
```
coretl: This seems wrong. I was expecting something more like `_trigger`, where `obj.execute` always returns a status...

src/bluesky/protocols.py
```diff
@@ -530,3 +530,17 @@ def check_supports(obj: T, protocol: type[Any]) -> T:
 
 # Descriptor with previous name on imports for backwards compatibility.
 Descriptor = DataKey
+
+@runtime_checkable
+class Executable(Protocol):
+    @abstractmethod
+    def execute(self, *args, **kwargs) -> SyncOrAsync[T]:
```
coretl: ... so we would need to define `ValuedStatus[T]` as a protocol and return it here

src/bluesky/protocols.py
```diff
@@ -530,3 +530,17 @@ def check_supports(obj: T, protocol: type[Any]) -> T:
 
 # Descriptor with previous name on imports for backwards compatibility.
 Descriptor = DataKey
+
+@runtime_checkable
+class Executable(Protocol):
```
coretl: I think this should this be generic on `[P, T]` like the ophyd-async `Command[P, T]`

coretl (changes requested):

jacopoabramo:
> Although we seem to have strayed from Future already with the `done` property.

True, but as @danielballan mentioned in the past, `Status` differs from `Future` only in its ability to be canceled, which a `Status` object shouldn't have. I'd say that it should be fine to implement `result()` as public method. `set_result` should be an implementation detail of those implementing the protocol, there should be no way from the outer scope to call that.

danielballan:
I'm good with following [`concurrent.futures`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result) with `result`. I approved @jacopoabramo's PR adding the protocol.

src/bluesky/plan_stubs.py
burkeds: So something like this @coretl?

```python
@plan
def execute(
        obj: Executable,
        group: Hashable | None = None,
        wait: bool = False,
        execute_args: tuple[Any, ...] | None = None,
        execute_kwargs: dict[str, Any] | None = None,
) -> MsgGenerator[Status]:
    ret = yield Msg("execute", obj, group=group, execute_args=execute_args, execute_kwargs=execute_kwargs)
    if wait:
        yield Msg("wait", None, group=group)
    return ret
    
async def _execute(self, msg):
    obj = check_supports(msg.obj, Executable)
    args = msg.execute_args
    kwargs = dict(msg.execute_kwargs)
    wait = msg.wait
    group = msg.group

    ret = obj.execute(*msg.args, **kwargs)
    self._add_status_to_group(obj=obj, status_object=ret, group=group, action="execute")
    return ret
```

burkeds:
Waiting on https://github.com/bluesky/bluesky/pull/2030

burkeds:
@tacaswell can you take a look at this?

src/bluesky/plan_stubs.py
coretl: The RE looks right, for the plan stub I meant:

```python
@plan
def execute(
        obj: Executable[P, T],
        *args: P.args: 
        group: Hashable | None = None,
        wait: bool = False,
        **kwargs: P.kwargs,
) -> MsgGenerator[StatusWithResult[T]]:
    ret = yield Msg("execute", obj, group=group, execute_args=args, execute_kwargs=kwargs)
    if wait:
        yield Msg("wait", None, group=group)
    return ret
```

Then this will still support type hinting on args, and if we find the corner case where an RPC method takes `wait` as an arg, we can drop down to yielding the `Msg` directly...

src/bluesky/protocols.py
```diff
@@ -530,3 +530,17 @@ def check_supports(obj: T, protocol: type[Any]) -> T:
 
 # Descriptor with previous name on imports for backwards compatibility.
 Descriptor = DataKey
+
+@runtime_checkable
+class Executable(Protocol[P, T]):
+    @abstractmethod
+    def execute(self, *args, **kwargs) -> StatusWithResult[T]:
```
coretl, line 537:
```suggestion
    def execute(self, *args: P.args, **kwargs: P.kwargs) -> StatusWithResult[T]:
```

src/bluesky/protocols.py
```diff
@@ -550,3 +550,17 @@ def check_supports(obj: T, protocol: type[Any]) -> T:
 
 # Descriptor with previous name on imports for backwards compatibility.
 Descriptor = DataKey
+
+@runtime_checkable
+class Executable(Protocol[P, T]):
```
jacopoabramo:
```suggestion
class Executable(Protocol[P, T_co]):
```

According to the typing spec spec, `T` in this case requires to be covariant, so it requires another `TypeVar`:

```python
T_co = TypeVar('T_co', covariant=True)
```

coretl (approved):
Looks good to me

PR #1997, opened 2026-03-24
Protect SIGINT handler against preemption on successive signals

### LINKED ISSUE #1993: SIGINT handler not protected against preemption
The SIGINT handler has a counting mechanism to determine whether to request a deferred pause (count = 1), request a hard pause (count = 2), ignore the input (count = 3 to count = 10), or issue a `KeyboardInterrupt` causing an RE panic (count > 10).

When two SIGINTs arrive within milliseconds of each other. The second one preempts the first one. This means that there is a race condition on incrementing `self.count` by 1 and checking `self.count` to determine if we should request a deferred pause, request a hard pause, ignore the input, or issue a `KeyboardInterrupt`.

I believe this accounts for many such cases where your `Ctrl + C` is seemingly ignored. @tacaswell and I recently observed duplicate SIGINTs being sent within milliseconds of each other on a single press of `Ctrl + C` within IPython. This lead to some strange behavior with pausing the `RunEngine`. 

## Steps To Reproduce
1. Versions of packages are here: https://github.com/NSLS2/six-profile-collection/blob/main/pixi.lock
2. Start IPython
3. Trip a Bluesky suspender (leaves the RunEngine in `running` state)
4. Run `RE(bps.null())`
5. Observe suspender trip
6. Press `Ctrl + C` twice -> **FLAKY** behavior
    - Sometimes the counts line up and we get "normal" behavior
      - First is deferred (count = 1), second is a hard pause (count = 2)
    - Other times we get unexpected behavior
      - First is deferred (count = 1), second is ignored (count = 3)
 7. Otherwise, press `Ctrl + C` once -> **FLAKY** behavior
    - Sometimes it defers a pause (count = 1)
    - Other times it hard pauses immediately (count = 2)

## Acceptance Criteria
- SIGINT handler is protected against preemption (probably should not use a count, but timestamps instead)

## Possible Solution

- Count all SIGINTs within 100ms as 1 pause request (intentional SIGINTs only)
- Any number of requests > 1 and < 11 within 10s should request a hard pause

thopkins32:
## Description

- Reworks how `SigintHandler` works internally
  - **Treats every signal received within 100ms as 1 count (to respond to intentional signals)**
  - Greatly reduces the complexity of the signal handler callback so it can be very fast. It now offloads to a separate thread for printing and pause requests.
- Deprecates `SignalHandler` base class (no longer used for `SigintHandler`)
- Updates existing SIGINT test timing to new 100ms timing window
- Reactivates skipped SIGINT test -> panic test (used to hang forever, no longer does)

## Motivation and Context

Closes #1993

Current `SigintHandler`'s signal callback is slow (uses print and schedules threads). SIGINT signals within ~1-10ms of each other often preempt each other on the main thread. If the operating system or IPython or Pixi duplicates sending SIGINT on a single press of `Ctrl + C`, then you *could* get an immediate hard pause rather than a deferred one.

Interleaving signal handling should be avoided by simplifying the instructions done in the signal handler callback. Treating all SIGINTs received in a 100ms window as the same helps resolve the duplicate interrupt signal problem.

## How Has This Been Tested?

Test coverage is already good on this. It has minimal impact on other code.

I included a few more deterministic SIGINT tests to improve coverage.

### DIFF
--- a/src/bluesky/utils/__init__.py
+++ b/src/bluesky/utils/__init__.py
@@ -325,19 +325,13 @@ class SigintHandler:
                 try:
                     self._RE.request_pause(defer=True)
                 except TransitionError:
-                    print(
-                        "Deferred pause request failed. RunEngine is not in a pausable state. "
-                        "Inspect `RE.state` and try again after 10 seconds..."
-                    )
+                    ...
             elif self._request == PauseRequest.HARD:
                 print("A 'hard pause' has been requested.")
                 try:
                     self._RE.request_pause(defer=False)
                 except TransitionError:
-                    print(
-                        "Hard pause request failed. RunEngine is not in a pausable state. "
-                        "Inspect `RE.state` and try again..."
-                    )
+                    ...
 
             # Block until next request
             self._request_event.wait()
@@ -381,10 +375,11 @@ class SigintHandler:
                 self._count += 1
                 if self._count < 11:
                     self._request = PauseRequest.HARD
+                    self._request_event.set()
                 else:
                     self._released = True
+                    self._request_event.set()
                     self._original_handler(signum, frame)
-                self._request_event.set()
 
         # Install handler callback
         signal.signal(signal.SIGINT, handler)


src/bluesky/tests/test_run_engine.py
```diff
@@ -722,95 +724,114 @@ def simple_plan():
...
+        "Hangs on Python <3.12 due to a CPython bug: after "
+        "PyThreadState_SetAsyncExc the main thread deadlocks and "
+        "never reaches the subsequent blocking_event.wait(). "
+    ),
+)
 @uses_os_kill_sigint
-def test_sigint_many_hits_panic(RE):
-    raise pytest.skip("hangs tests on exit")
-    pid = os.getpid()
+def test_sigint_many_hits_panic(RE, deterministic_sigint):
```
thopkins32, line 808: This test was previously skipped, but it now works on Python != 3.11. I spent a lot of time trying to debug why it wasn't working on Python 3.11 but couldn't figure it out. We send `PyThreadState_SetAsyncExc` but it hangs forever...

See https://github.com/bluesky/bluesky/actions/runs/23557465255/job/68588133124 for a CI run that dumps the stack of each thread when it gets stuck.

src/bluesky/tests/test_run_engine.py
```diff
@@ -722,95 +724,114 @@ def simple_plan():
...
     assert RE.state == "idle"
-    timer.join()
 
 
+@pytest.mark.skipif(
+    sys.version_info < (3, 12),
+    reason=(
+        "Hangs on Python <3.12 due to a CPython bug: after "
+        "PyThreadState_SetAsyncExc the main thread deadlocks and "
+        "never reaches the subsequent blocking_event.wait(). "
```
tacaswell:
```suggestion
        "Hangs on Python <3.12 due to a possible CPython bug: after "
        "PyThreadState_SetAsyncExc the main thread deadlocks and "
        "never reaches the subsequent blocking_event.wait(). "
        "Issue only reproduces on CI."
```

src/bluesky/tests/test_run_engine.py
tacaswell: It looks like I added the skip in 1971117e86b83c8b471a49fade0cc3dd3648fe95 as part of https://github.com/bluesky/bluesky/pull/1672 but did not get back to following up.

src/bluesky/utils/__init__.py
```diff
@@ -267,57 +280,120 @@ def release(self):
...
                     "A 'deferred pause' has been requested. The "
                     "RunEngine will pause at the next checkpoint. "
                     "To pause immediately, hit Ctrl+C again in the "
                     "next 10 seconds."
                 )
+                try:
+                    self._RE.request_pause(defer=True)
+                except TransitionError:
+                    print(
+                        "Deferred pause request failed. RunEngine is not in a pausable state. "
```
tacaswell: I think we should go back to eating this error.  If we get TransitionError raise here I think the only way we got here was that we lost a race (either the SIGINT is align such that was issued before RE.__call__ returns but the thread does not run until after or in the case of the hard pauses below multiple got stacked).

src/bluesky/utils/__init__.py
```diff
@@ -267,57 +280,120 @@ def release(self):
...
+            elif time_diff > 0.1 and self._count > 0:
+                # Second or more pause requests
+                self._last_sigint_time = now
+                self._count += 1
+                if self._count < 11:
+                    self._request = PauseRequest.HARD
+                else:
+                    self._released = True
+                    signal.signal(signal.SIGINT, self._original_handler)
+                    self._original_handler(signum, frame)
```
tacaswell, line 382: I think this will raise an exception meaning the next line never gets run on this code branch leaving a zombie thread.

thopkins32: We did the signal handler switch within the signal handler before this PR as well without an exception. I intended to actually remove this line (got lost during all of my debugging commits):
```python
                    signal.signal(signal.SIGINT, self._original_handler)
```
and leave it up to the context manager `__exit__` to reset the original handler.

src/bluesky/utils/__init__.py
thopkins32: I'll try to write a test to cover this case. I think it is untested.

thopkins32: Removing the try-except for `TransitionError` will cause this watcher thread to die, it doesn't propagate to the main thread. So the experience from the user would be pressing `Ctrl + C` does nothing.

This can happen when multiple hard pause requests are stacked and the RE is "pausing" or is already "paused".

If I remove the try-except for `TransitionError`, a user who does:
```
1. Ctrl + C (deferred pause)
2. Ctrl + C (hard pause)
3. RE enters "pausing" or "paused" state
4. Ctrl + C -> SigintHandler._request_thread dies due to uncaught TransitionError
5. RE.resume()
6. Now pausing doesn't work at all
```
I think catching the `TransitionError` is correct here.

thopkins32: I realize that the RE exits the context managers in "paused" so this could only happen during "pausing", which is a short window to hit (but maybe not short enough for CI runners).

But if someone has additional context manager with a very long running `__exit__` then it may be "pausing" or "paused" for a while during which a `TransitionError` could occur. See the new test in [dcbdbfc](https://github.com/bluesky/bluesky/pull/1997/commits/dcbdbfca4df0c1509fcfbd41b0d67835952ace17)

tacaswell: I meant we should still catch it so it does not propagate, but we do not need to print anything to the user.

src/bluesky/utils/__init__.py
```diff
@@ -267,57 +280,121 @@ def release(self):
...
+
+        # Install handler callback
+        signal.signal(signal.SIGINT, handler)
+        return self
 
-                threading.Thread(target=maybe_prompt_pause).start()
-            self.last_sigint_time = time.time()
+    def __exit__(self, type, value, tb) -> None:
+        signal.signal(signal.SIGINT, self._original_handler)
+        if not self._released:
```
tacaswell, line 390: I think we still have a logic error in the panic case as now we won't go through this path on the way out as `self._release` will be `True`.

src/bluesky/utils/__init__.py
thopkins32: Okay, I misunderstood what you meant by "eating this error." I can remove the print but I still think its useful information to know that a pause request failed, even if it is very rare.

PR #1998, opened 2026-03-25
MNT: refactor RunEngine to extract _run in SingleRunExecutor

jacopoabramo:
## Description

Supersedes #1991
Aims to address #1654

This is the second attempt to refactor the content of the `_run` method into a separate `SingleRunExecutor` class that acts as "glorified" `asyncio.Task`.

## Motivation and Context
The PR is much closer to the original desired behavior of keeping all the code inside the RunEngine and moving out the _run method to separate the execution logic from the public API.

Three new cache objects have been created:

- `_ObjsCache`: all objects manipulated during a plan;
- `_StateCache`: all RE-relevant state variables;
- `_InternalCallbackCache`: methods and functions to propagate functionalities from the RE to the RunEngine (i.e. setting the state).

The SRE is created during `__call__` and discarded when the plan finishes (either naturally or because it's being paused).

The SRE is tracked in a local `sre_objs` set which should in the future be used for when creating multiple SREs objects.

Additionally, we refactored `__call__` in order to have a new public API `run()` that instead of blocking the main thread it returns a the future object tracking the current running task. `_prepare_run()` encapsulates the logic common to both operations.

There's currently no protection on the case of multiple calls to `run()` so I don't consider it stable yet.

There are some open questions:

- where should `task` and `task_fut` live? The SRE or the RE? I believe:
  - `task` -> `SRE`
  - `task_fut` -> `RE`
  but I don't have strong arguments for that yet...
- right now whenever the RE is requested to pause the running plan we're iterating over the - hopefully - single-element set containing the SRE objects; is this the correct approach? In the future we want to be able (I think) to selectively stop a plan leaving the others running, how do we deal with that? Should it be addressed now?
- returning a `concurrent.futures.Future` is risky because it provides access to a `future.cancel()` method that will make a mess. Since we're expecting a plan to be controlled by the RunEngine this is a good place to use a possible `PlanStatus` object that simply tracks the execution of a plan and use it to wrap the actual future object (it should not be visible to the user, I think).
   - it's also how one would naturally deal with subplans inside a plan: a message that builds an SRE and returns a status object tracking it.
- right now we're creating a single `run_permit` while I'm thinking that there should be one run permit per SRE, so this should be removed from the internal cache and instead passed at SRE construction similarly to how we do with the state lock.

## How Has This Been Tested?

Ran tests locally on Windows, letting CI do it's thing.


jacopoabramo:
@tacaswell gentle ping, this should be ready for review

Issue #1987, opened 2026-02-05
SuspendWhenChanged suspender misinterprets falsy expected values


canismarko:
The SuspendWhenChanged suspender takes an optional argument *expected_value* and stashes it for later comparison. If the argument is omitted (i.e. has the default value `None`), the suspender [uses the current value](https://github.com/bluesky/bluesky/blob/a9ed87072e54217973a407ed14fec732ea8c0c29/src/bluesky/suspenders.py#L666) of the signal via:

```python
self.expected_value = expected_value or signal.value
```

This breaks if *expected_value* is a falsy argument, like `0` or `""` but the current value of the signal is not the expected value. This happens because e.g. `bool(0)` is False, and so `self.expected_value` is assigned `signal.value`, which is different from *expected_value*.

A possible solution is to explicitly check for `None` identity. It's longer, but more explicit.

```python
self.expected_value = expected_value if expected_value is not None else signal.value
```

## Steps To Reproduce

Here is an MWE:

```python
from ophyd import Signal
from bluesky.suspenders import SuspendWhenChanged

def test_suspend_when_changed_falsy_expected_value():
    sig = Signal(name="sig")
    sig.set(5)
    assert sig.value == 5
    suspender = SuspendWhenChanged(signal=sig, expected_value=0)
    assert suspender.expected_value == 0  # <-- fails, is actually 5
```

## Acceptance Criteria

The above test should pass.

Mirochill:
For triage: this appears to have been addressed by #2004, which merged on 2026-05-07. On current `main`, `SuspendWhenChanged.__init__` now preserves falsy explicit values with:

```python
self.expected_value = signal.value if expected_value is None else expected_value
```

That matches the fix proposed in this issue and the merged PR also added a regression test for falsy expected values.

danielballan:
@canismarko Do you agree with the assessment above? Good to close this?

canismarko:
Yes, this works as I expect now.

The behavior of the suspender (to use `signal.value`) might warrant changing since `ophyd.Signal.value` raises a user warning and `ophyd_async.core._signal.Signal` doesn't have a `value` attribute.

That's out of the scope of this issue, though, so I'll close this now. Thanks @danielballan and @Mirochill.

PR #2004, opened 2026-04-07
SuspendWhenChanged suspender misinterprets falsy expected values

GreyforgeLabs:
## Summary
- Addresses https://github.com/bluesky/bluesky/issues/1987
- Applies the validated ForgeHub draft for SuspendWhenChanged suspender misinterprets falsy expected values

## Validation
- manual_validation: uv run --extra dev pytest src/bluesky/tests/test_suspenders.py -k preserves_falsy=0

### DIFF
--- a/src/bluesky/suspenders.py
+++ b/src/bluesky/suspenders.py
@@ -663,7 +663,7 @@ class SuspendWhenChanged(SuspenderBase):
         tripped_message="",
         **kwargs,
     ):
-        self.expected_value = expected_value or signal.value
+        self.expected_value = signal.value if expected_value is None else expected_value
         self.allow_resume = allow_resume
         super().__init__(
             signal, sleep=sleep, pre_plan=pre_plan, post_plan=post_plan, tripped_message=tripped_message, **kwargs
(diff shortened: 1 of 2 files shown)


tacaswell (approved):

tacaswell:
On one hand this fix looks correct, on the other this looks like a bot not a human and the :+1: in also from a bot.

I'm going to hold of merging for not to understand where this came from.

GreyforgeLabs:
Thanks for pausing on provenance. This PR came from GreyforgeLabs/ForgeHub, an AI-assisted open-source triage workflow run by a human operator on this account. The operator is watching this PR now and can make requested changes; it is not intended as an unattended drive-by merge request.

The patch was selected from #1987 because explicit falsy expected values like 0, False, and empty strings should be treated as expected values rather than as the absence of an expectation. Happy to adjust tests, wording, or implementation details if you want this shaped differently.

tacaswell:
I merged this because it is correct and simple enough that there is really no other way to do this so re-implementing as a human generated PR would just be out of spite.

We do not currently have a policy on fully automated PRs like this, for example Matplotlib fully bans this sort of PR https://matplotlib.org/devdocs/devel/contribute.html#use-of-generative-ai , which is my bias but as we currently lack agreement in the bluesky project about that it seems incorrect to enforce it.

PR #2005, opened 2026-04-10
ENH: add testing for python 3.14

jacopoabramo:
## Description

Add 3.14 to the CI testing.

## Motivation and Context

3.14 has been out since October 2025, would be nice to align - hopefully does not break anything (or it might be useful to catch some bugs).

## How Has This Been Tested?

Waiting for CI to do its thing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,10 @@ jobs:
       matrix:
         runs-on: ["ubuntu-latest", "windows-latest"] # can add macos-latest
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        exclude:
+          # duckdb (tiled dep) has no Windows wheels for 3.14 yet
+          - runs-on: "windows-latest"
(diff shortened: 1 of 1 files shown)


jacopoabramo:
building fails because of duckdb which i'm guessing is a tiled dependency. not sure what to make of it

tacaswell:
it looks like duckdb does not have wheels yet.  They build in linux, don't on windows.  The best option here is to take the run-time hit on linux and skip 3.14 on windows.

tacaswell:
On slightly closer inspection, duckdb does have 3.14 wheels available for all paltforms for current versions, but tiled pins back to <1.4 which does not.

jacopoabramo:
I'm wondering if the pin is due to some blocker on tiled side or it's just a safety precaution? @danielballan

jacopoabramo:
Traces back to https://github.com/bluesky/tiled/pull/1184

tacaswell (approved):

PR #2006, opened 2026-04-15
Update release history for v1.15.0

thopkins32:
Based on PR diff between `main` and previous tag

### DIFF
--- a/docs/api_changes.rst
+++ b/docs/api_changes.rst
@@ -2,21 +2,31 @@
  Release History
 =================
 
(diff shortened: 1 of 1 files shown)


mrakitin (approved):
Looks good to me!

mrakitin (changes requested):
Oh, I think we are missing the release notes for v1.14.6 here. Could you please update it with the information from https://github.com/bluesky/bluesky/releases/tag/v1.14.6?

mrakitin (approved):

PR #1951, opened 2025-09-10
Added `read_all` and `trigger_all` messages

evvaaaa:
Closes #1926 

## Description
New messages `"read_all"` and `"trigger_all"` to read or trigger a sequence of `Readable` or `Triggerable`, allowing for the asynchronous implementations to be ran together.

This allows for multiple ophyd-async devices to be triggered or read asynchronously alongside one another from the plan level. `Device`s with asynchronous `trigger` or `read` will always be triggered/read before synchronous `Device`s in the `Sequence`.

Though we could implement it with a dictionary of device instance to argument, for now a method of passing arguments through to the device's trigger or read is not supported.

## Motivation and Context
When repeatedly `one_shot`ing a large collection of devices, currently each ophyd-async device will be read asynchronously, however they will not be read asynchronously together. This leads to worse performance for beamlines where the majority of data is acquired through Epics/Tango.

PR #2007, opened 2026-04-24
Deprecate arguments to device `read` and allow `"read"` message to take multiple objects like `"locate"`

### LINKED ISSUE #1926: Allow multiple ophyd async devices passed into `one_shot` to be read together asynchronously
We are beginning the process of converting devices to ophyd-async.

ophyd-async `read`/`describe`'s asynchronously across all a device's subdevices:

https://github.com/bluesky/ophyd-async/blob/61186ffd4c7268cd6a47fbe3588be85afdb1e8be/src/ophyd_async/core/_readable.py#L116-L128

``` python
    async def describe_configuration(self) -> dict[str, DataKey]:
        return await merge_gathered_dicts(
            [func() for func in self._describe_config_funcs]
        )


    async def read_configuration(self) -> dict[str, Reading]:
        return await merge_gathered_dicts([func() for func in self._read_config_funcs])


    async def describe(self) -> dict[str, DataKey]:
        return await merge_gathered_dicts([func() for func in self._describe_funcs])


    async def read(self) -> dict[str, Reading]:
        return await merge_gathered_dicts([func() for func in self._read_funcs])
```

Meaining that for an ophyd-async device

``` python
class SomeDevice(StandardReadable):
    sig1 = soft_signal_rw(int, initial_value=42)
    sig2 = soft_signal_rw(str, initial_value="OK")
```

``` python
bps.one_shot([some_device])
```

will gather the read/describe on the signals asynchronously.

On the plan level however, the following will not be ran asynchronously:

``` python
bps.one_shot([some_device.sig1, some_device.sig2])
```

because each read message will be passed into the `RunEngine` separately.

https://github.com/bluesky/bluesky/blob/103accac0ebe3d2f0efcd0c5f7da87e8172f5b16/src/bluesky/plan_stubs.py#L1612

https://github.com/bluesky/bluesky/blob/103accac0ebe3d2f0efcd0c5f7da87e8172f5b16/src/bluesky/plan_stubs.py#L1458-L1459

# Problem

At bessy, our bluesky environment has all the devices instantiated/connected before the plan starts. Some devices are large with many sub devices. In some plans we may only want to read a small portion of the signals on such a device, but we want to do it asynchronously with signals across multiple devices. This is a problem because while the `RunEngine` `read` is `maybe_await` ing, other messages are blocked.

# Solution

The simplest solution I thought of was to make a wrapper device for all the async signals passed into a `trigger_and_read`:

``` python
class AsyncDeviceWrapper(StandardReadable):
    def __init__(self, devices_to_read: list[StandardReadable], *args, **kwargs):
        self._devices_to_read = devices_to_read
        kwargs["name"] = "_async_device_wrapper"
        super().__init__(*args, **kwargs)

    async def describe(self):
        describe_values = await asyncio.gather(
            *[obj.describe() for obj in self._devices_to_read]
        )
        value = {}
        for describe_value in describe_values:
            value.update(**describe_value)
        return value

    async def read(self):
        read_values = await asyncio.gather(
            *[obj.read() for obj in self._devices_to_read]
        )
        value = {}
        for read_value in read_values:
            value.update(**read_value)
        return value


def async_supporting_trigger_and_read(
    devices: Sequence[Readable], name: str = "primary"
) -> MsgGenerator[Mapping[str, Reading]]:
    """
    Trigger and read a list of detectors and bundle readings into one Event.

    Parameters
    ----------
    devices : list
        devices to trigger (if they have a trigger method) and then read
    name : string, optional
        event stream name, a convenient human-friendly identifier; default
        name is 'primary'

    Returns
    -------
    readings:
        dict of device name to recorded information

    Yields
    ------
    msg : Msg
        messages to 'trigger', 'wait' and 'read'
    """
    from bluesky.preprocessors import contingency_wrapper

    # If devices is empty, don't emit 'create'/'save' messages.
    if not devices:
        yield from bps.null()
    devices = separate_devices(devices)  # remove redundant entries
    rewindable = all_safe_rewind(devices)  # if devices can be re-triggered

    def inner_trigger_and_read():
        grp = _short_uid("trigger")
        no_wait = True
        for obj in devices:
            if isinstance(obj, Triggerable):
                no_wait = False
                yield from bps.trigger(obj, group=grp)
        # Skip 'wait' if none of the devices implemented a trigger method.
        if not no_wait:
            yield from bps.wait(group=grp)
        yield from bps.create(name)

        async_devices = {
            obj for obj in devices if asyncio.iscoroutinefunction(obj.read)
        }
        async_device_wrapper = AsyncDeviceWrapper(list(async_devices))
        sync_devices = list(set(devices) - async_devices)

        def read_plan():
            ret = {}  # collect and return readings to give plan access to them
            for obj in sync_devices + [async_device_wrapper]:
                reading = yield from bps.read(obj)
                if reading is not None:
                    ret.update(reading)
            return ret

        def standard_path():
            yield from bps.save()

        def exception_path(exp):
            yield from bps.drop()
            raise exp

        ret = yield from contingency_wrapper(
            read_plan(), except_plan=exception_path, else_plan=standard_path
        )
        return ret

    from bluesky.preprocessors import rewindable_wrapper

    return (yield from rewindable_wrapper(inner_trigger_and_read(), rewindable))


def async_supporting_one_shot(detectors):
    yield from bps.one_shot(
        detectors=detectors, take_reading=async_supporting_trigger_and_read
    )


```

This did seem to work for a single `one_shot` (with some janky metadata):

``` python
class DummySyncDevice(SyncDevice):
    sig1 = Cpt(Signal, value=42, name="sig1")
    sig2 = Cpt(Signal, value="OK", name="sig2")


class DummyAsyncDevice(StandardReadable):
    sig1 = soft_signal_rw(int, initial_value=42, name="sig1")
    sig2 = soft_signal_rw(str, initial_value="OK", name="sig2")


dummy_device_sync = DummySyncDevice(name="sync_name")
dummy_device_async = DummyAsyncDevice(name="async_name")


RE = RunEngine()


def run(detectors):
    yield from bps.open_run()
    yield from async_supporting_one_shot(detectors)
    yield from bps.close_run()


RE(
    run(
        [
            dummy_device_sync.sig1,
            dummy_device_sync.sig2,
            dummy_device_async.sig1,
            dummy_device_async.sig2,
        ]
    ),
    lambda x, y: print(x, "=============\n", pprint.pformat(y)),
)
```

```
start =============
 {'plan_name': 'run',
 'plan_type': 'generator',
 'scan_id': 1,
 'time': 1752055414.538485,
 'uid': '9ed35b23-ac67-42fa-a795-6973ad04d2d0',
 'versions': {'bluesky': '1.14.3.dev12+gc2b578aa',
              'ophyd': '1.10.6',
              'ophyd_async': '0.11'}}
descriptor =============
 {'configuration': {'_async_device_wrapper': {'data': {},
                                             'data_keys': {},
                                             'timestamps': {}},
                   'sync_name_sig1': {'data': {'sync_name_sig1': 42},
                                      'data_keys': {'sync_name_sig1': {'dtype': 'integer',
                                                                       'shape': [],
                                                                       'source': 'SIM:sync_name_sig1'}},
                                      'timestamps': {'sync_name_sig1': 1752055414.5368893}},
                   'sync_name_sig2': {'data': {'sync_name_sig2': 'OK'},
                                      'data_keys': {'sync_name_sig2': {'dtype': 'string',
                                                                       'shape': [],
                                                                       'source': 'SIM:sync_name_sig2'}},
                                      'timestamps': {'sync_name_sig2': 1752055414.5369074}}},
 'data_keys': {'sig1': {'dtype': 'integer',
                        'dtype_numpy': '<i8',
                        'object_name': '_async_device_wrapper',
                        'shape': [],
                        'source': 'soft://sig1'},
               'sig2': {'dtype': 'string',
                        'dtype_numpy': '|S40',
                        'object_name': '_async_device_wrapper',
                        'shape': [],
                        'source': 'soft://sig2'},
               'sync_name_sig1': {'dtype': 'integer',
                                  'object_name': 'sync_name_sig1',
                                  'shape': [],
                                  'source': 'SIM:sync_name_sig1'},
               'sync_name_sig2': {'dtype': 'string',
                                  'object_name': 'sync_name_sig2',
                                  'shape': [],
                                  'source': 'SIM:sync_name_sig2'}},
 'hints': {'_async_device_wrapper': {},
           'sync_name_sig1': {'fields': []},
           'sync_name_sig2': {'fields': []}},
 'name': 'primary',
 'object_keys': {'_async_device_wrapper': ['sig1', 'sig2'],
                 'sync_name_sig1': ['sync_name_sig1'],
                 'sync_name_sig2': ['sync_name_sig2']},
 'run_start': '9ed35b23-ac67-42fa-a795-6973ad04d2d0',
 'time': 1752055414.5406475,
 'uid': 'd54a2068-5cc6-455d-8f56-77b2e1f5de03'}
event =============
 {'data': {'sig1': 42,
          'sig2': 'OK',
          'sync_name_sig1': 42,
          'sync_name_sig2': 'OK'},
 'descriptor': 'd54a2068-5cc6-455d-8f56-77b2e1f5de03',
 'filled': {},
 'seq_num': 1,
 'time': 1752055414.5421457,
 'timestamps': {'sig1': 3487758.983936489,
                'sig2': 3487758.983954177,
                'sync_name_sig1': 1752055414.5368893,
                'sync_name_sig2': 1752055414.5369074},
 'uid': 'f76cf6ff-a65b-47cf-998e-017fdc06e028'}
stop =============
 {'exit_status': 'success',
 'num_events': {'primary': 1},
 'reason': '',
 'run_start': '9ed35b23-ac67-42fa-a795-6973ad04d2d0',
 'time': 1752055414.5424454,
 'uid': '0dd527f0-f896-4f65-a089-7ebd4859f352'}
```

however once a second

``` python
    yield from async_supporting_one_shot(detectors)
```

is added, we get failures in the bundlers:

```
  File "/home/a5497/bluesky/src/bluesky/bundlers.py", line 570, in save
    raise RuntimeError(f"Mismatched objects read, expected {frozenset(d_objs)!s}, got {objs_read!s}")
RuntimeError: Mismatched objects read, expected frozenset({Signal(name='sync_name_sig1', parent='sync_name', value=42, timestamp=1752055534.5201592), Signal(name='sync_name_sig2', parent='sync_name', value='OK', timestamp=1752055534.5201764), <__main__.AsyncDeviceWrapper object at 0x73c917846fc0>}), got frozenset({Signal(name='sync_name_sig1', parent='sync_name', value=42, timestamp=1752055534.5201592), Signal(name='sync_name_sig2', parent='sync_name', value='OK', timestamp=1752055534.5201764), <__main__.AsyncDeviceWrapper object at 0x73c916a58b30>})
```

This is presumably because `AsyncDeviceWrapper` is a different device by reference from the last read one here

https://github.com/bluesky/bluesky/blob/main/src/bluesky/bundlers.py#L568-L571

### Idea

We could make `AsyncDeviceWrapper` exist in bluesky and have the bundlers format the output (flattening the data read from `AsyncDeviceWrapper.devices_to_read`).

Whether or not we go with this kind of idea (I could probably get the same as the above with a `DeviceVector`), it would be nice to work on gathering asynchronous devices passed into the `RunEngine` separately.

evvaaaa:
Closes #1926 
Replaces #1951

## Description
New message `"read_all"` to read a sequence of `Readable`, allowing for the asynchronous implementations to be ran together.

This allows for multiple ophyd-async devices to be read asynchronously alongside one another from the plan level.

The plan stub `trigger_and_read` now uses this new message on it's sequence of devices.

## Motivation and Context
When repeatedly `one_shot`ing a large collection of devices, currently each ophyd-async device will be read asynchronously, however they will not be read asynchronously together. This leads to worse performance for beamlines where the majority of data is acquired through Epics/Tango.

These changes allowed for a speed up from 6s to 0.3s per `one_shot` from @PierreSchnizer.

## Consequences for Backwards Compatibility

The only place I can think of is if anyone has written a preprocessor which checks for `"read"` directly.


src/bluesky/run_engine.py
```diff
@@ -1980,23 +1999,55 @@ async def _read(self, msg):
...
         if (
             current_run := self._run_bundlers.get(run_key, key_absence_sentinel := object())
         ) is not key_absence_sentinel:
-            await current_run.read(msg, ret)
+            await current_run.read(msg, reading)
 
-        return ret
+        return reading
+
+    async def _read_all(self, msg):
```
evvaaaa: An idea which I think is slightly nicer:

One `Msg("read", ...` with either a single `Readable` as the object, or an `ObjTuple` of `Readable`.

The run engine will then check that it's one of these two things, in the former case it will do as it does now, in the latter it will gather the read.

Also happy keeping this to a new message `read_all`.

evvaaaa: We agreed that single `"read"` message with `ObjTuple` is indeed preferred. I will make the necessary change.

tacaswell:
As discussed on the call extending `read` to understand `ObjTuple` or `Readable` is nicer.

@evvaaaa 's noted a concern about a pre-processor that is looking at `read` messages go by is a good one, but I suspect rare.  To that end we should add an ENV like we have for requiring pre-declaration of streams that will disable the consolidation in `trigger_and_read` and `one_shot`.  It is a bit of complexity, but this is too nice of a consolidation to pass on by default.

src/bluesky/run_engine.py
```diff
@@ -1971,32 +1981,78 @@ async def _declare_stream(self, msg):
...
+
+        for obj in objs:
+            check_supports(obj, Readable)
+            if inspect.iscoroutinefunction(obj.read):
+                coro_objs.append(obj)
+            else:
+                non_coro_objs.append(obj)
+
+        coro_read_rets = await asyncio.gather(*[self._perform_single_reading(obj) for obj in coro_objs])
+        non_coro_read_rets = await asyncio.gather(*[self._perform_single_reading(obj) for obj in non_coro_objs])
```
evvaaaa: I considered `return_exceptions=True` and then raising a new exception with all the failures together, but I think we should probably keep to failing on the first device which has a problem. What's important is that the bundlers won't be appended to before all of them have had a chance to fail and this is achieved here already.

src/bluesky/tests/__init__.py
```diff
@@ -18,6 +22,20 @@
...
+        reason = "ophyd-async only supports more recent python releases"
+    else:
+        import ophyd_async  # type: ignore
+except ImportError as ie:
+    # pytestmark = pytest.mark.skip
+    ophyd_async = None
+    reason = str(ie)
+
+# define a skip condition based on if ophyd-async is available or not
+requires_ophyd_async = pytest.mark.skipif(ophyd_async is None, reason=reason)
```
coretl, line 37: I was sure we added ophyd-async to the dev dependencies of bluesky somewhere, but it's not in main, maybe there's a PR somewhere we forgot to merge?

src/bluesky/run_engine.py
```diff
@@ -88,6 +89,15 @@ class _RunEnginePanic(Exception): ...
 class WaitForTimeoutError(TimeoutError): ...
 
 
+@final
+class ObjTuple(tuple):
```
coretl: Why not use a plain tuple?

coretl (commented):
Couple of comments from a brief skim

src/bluesky/run_engine.py
evvaaaa: Felt cleaner to have a specific type for iterable containing readables because a subclass of tuple could also be readable. I'm happy to change it to `type(msg.obj) is tuple` instead:

``` python
is_non_subclassed_tuple_of_readable = type(msg.obj) is tuple and all(isinstance(obj, Readable) for obj in msg.obj)
```

probably nicer.

coretl: Another possibility is to use `*msg.args` for extra objects to read like `locate` does:
https://github.com/bluesky/bluesky/blob/d149731d6d4e34d2b6eb4ddcf6c5614b681a78d5/src/bluesky/run_engine.py#L2014

If there's an environment variable opt-in to this anyway then that could be the guard against the already deprecated usage of `*args`. What do you reckon @tacaswell ?

src/bluesky/tests/__init__.py
coretl: https://github.com/bluesky/bluesky/pull/1923/changes

We should really finish off that PR...

src/bluesky/run_engine.py
evvaaaa: This is much nicer imo!

evvaaaa: @tacaswell unless you decide otherwise I will swap the implementation to this and deprecate args passed to read?

src/bluesky/tests/__init__.py
evvaaaa: Happy to merge it in this one or wait, up to you

src/bluesky/run_engine.py
evvaaaa: This is now implemented.

src/bluesky/run_engine.py
```diff
@@ -1971,32 +1971,61 @@ async def _declare_stream(self, msg):
...
+        coro_objs, non_coro_objs = [], []
+
+        for obj in objs:
+            if inspect.iscoroutinefunction(obj.read):
+                coro_objs.append(obj)
+            else:
+                non_coro_objs.append(obj)
+
+        coro_read_rets = await asyncio.gather(*[self._perform_single_reading(obj) for obj in coro_objs])
+        non_coro_read_rets = await asyncio.gather(*[self._perform_single_reading(obj) for obj in non_coro_objs])
```
coretl: is this gather?

src/bluesky/plan_stubs.py
```diff
@@ -146,26 +147,46 @@ def drop() -> MsgGenerator:
     return (yield Msg("drop"))
 
 
+BLUESKY_FORCE_READ_ALL_ONE_MSG_PER_DEVICE = os.environ.get("BLUESKY_FORCE_READ_ALL_ONE_MSG_PER_DEVICE") == "1"
+
+
 @plan
-def read(obj: Readable) -> MsgGenerator[Reading]:
+def read(
+    *objs: Sequence[Readable[Any]], one_message_per_device: bool = BLUESKY_FORCE_READ_ALL_ONE_MSG_PER_DEVICE
```
coretl: I would lean to reading env var in the call rather than binding as a default

evvaaaa: I'll get rid of the kwarg and do this.

src/bluesky/run_engine.py
evvaaaa: You mean I should be using something else for the sync reads?

For the non awaitable reads I just use gather here to reuse code. I could probably be a little clearer?

src/bluesky/plan_stubs.py
```diff
@@ -179,7 +168,13 @@ def read(
         Reading object representing information recorded
     """
 
-    if one_message_per_device:
+    # If ``True`` then there will be one Msg per device, which means that
+    # asynchronous devices will not be gathered.;
+    legacy_one_message_per_device = os.environ.get("BLUESKY_FORCE_READ_ALL_ONE_MSG_PER_DEVICE", "").lower() in (
+        "1",
+        "true",
+    )
```
jacopoabramo: maybe i'm overthinking this but could it be better that this is set only once when importing the module rather than scoping it into the plan?

evvaaaa: I think this is probably better too.

I suppose with it inside the call the user could, for example, `os.environ["BLUESKY_FORCE_READ_ALL_ONE_MSG_PER_DEVICE"] = "1"` from an ipython thread inside their queueserver to toggle without having to restart everything... but I'd prefer to just leave the check outside the call and only once.

jacopoabramo: maybe it could be worthwhile to add a dedicated getter/setter? my only concern is calling multiple times `environ` for performance purposes but there might be also security issues to consider, dunno

evvaaaa: I think now we have three toggleable env vars across the codebase I'll pull them out to their own file and make them `bool`.

evvaaaa: Done in the below commit. If any of these variables need to be toggled during runtime then we shouldn't be depending on env vars for it.

src/bluesky/run_engine.py
coretl: Ah, I see what you're doing here, but it seems a bit unnecessary to filter out all the non coros and then gather them separately.

I'd say don't bother sorting them into 2 lists and do the gather on `objs`

evvaaaa: I had assumed it would be less performant to have the underlying synchronous `maybe_await`'s in the gather, but you're right it doesn't make a difference.

evvaaaa:
@tacaswell Would you be able to assign someone NSLSII-side for a review of this?

PR #2009, opened 2026-04-28
Convert CI to use pixi

jwlodek:
## Description

Add pixi project configuration. Add pixi lockfile. Modify CI to use pixi for environment management instead of tox.

## Motivation and Context

* Adding a lockfile will make ensuring repeatable environments between local development machines and CI much simpler.
* Lockfiles generally provide a much nicer development experience.

## How Has This Been Tested?

* Tests work locally.
* About to be tested in CI on PR.


tacaswell:
This seems to have completely broken CI?

PR #2015, opened 2026-05-07
RemoteDispatcher ZMQ connection handshake

### LINKED ISSUE #2014: RemoteDispatcher slow-joiner issue
Was looking into some timing issues in [Blop's new tutorial w/ bluesky-queueserver integration](https://blueskyproject.io/blop/main/tutorials/queueserver.html). It uses `RemoteDispatcher` in a separate thread to monitor when the plan we submitted to the queue has been completed (rather than polling the `REManagerAPI`'s history).

`bluesky.callbacks.zmq.RemoteDispatcher.start()` performs SUB socket setup (`connect` + `setsockopt(SUBSCRIBE)`) and the asyncio poll loop in the same call. In typical threaded usage, e.g. `Thread(target=dispatcher.start).start()` followed immediately by work that triggers a publisher, the caller has no way to know when the SUB has actually connected and registered its subscription with the proxy. Documents published in this window are silently dropped.

`RemoteDispatcher.__init__` does no socket work. `__factory` is invoked only inside `start()`, immediately before `loop.run_until_complete(self._task)`. Callers that put `start()` on a thread therefore have no synchronization point against socket readiness.

## Steps To Reproduce
Steps to reproduce the behavior. Generated by Claude, validated by me.

```python
import threading, time, uuid, pickle
from collections import Counter
import zmq
from bluesky.callbacks.zmq import Proxy, Publisher, RemoteDispatcher
from zmq.utils.monitor import recv_monitor_message
PROXY_IN, PROXY_OUT, N_TRIALS = 5577, 5578, 20
def fake_run():
    ru, du = str(uuid.uuid4()), str(uuid.uuid4())
    return [
        ("start", {"uid": ru, "time": time.time(), "scan_id": 1}),
        ("descriptor", {"uid": du, "time": time.time(), "run_start": ru,
                        "name": "primary", "data_keys": {}, "object_keys": {},
                        "hints": {}, "configuration": {}}),
        ("event", {"uid": str(uuid.uuid4()), "time": time.time(),
                   "descriptor": du, "seq_num": 1,
                   "data": {}, "timestamps": {}, "filled": {}}),
        ("stop", {"uid": str(uuid.uuid4()), "time": time.time(),
                  "run_start": ru, "exit_status": "success"}),
    ]
threading.Thread(target=Proxy(in_port=PROXY_IN, out_port=PROXY_OUT).start, daemon=True).start()
time.sleep(0.5)
pub = Publisher(f"localhost:{PROXY_IN}"); time.sleep(0.5)  # let PUB handshake
def trial_naive():
    got = []
    d = RemoteDispatcher(("localhost", PROXY_OUT))
    d.subscribe(lambda n, _: got.append(n))
    threading.Thread(target=d.start, daemon=True).start()
    for n, doc in fake_run(): pub(n, doc)
    time.sleep(0.5)
    try: d.stop()
    except Exception: pass
    return len(got)
print("naive:", Counter(trial_naive() for _ in range(N_TRIALS)))
```

This scripts regularly outputs `naive: Counter({0: 20})` meaning that the subscriber never saw any of the documents. Sometimes it gives `naive: Counter({0: 18, 4: 2})` meaning that 2 of the trials got all 4 expected documents and the subscriber joined before the docs were published.

## Acceptance Criteria
- `RemoteDispatcher` exposes some way of knowing when the socket has connected (possibly via [`get_monitor_socket`](https://pyzmq.readthedocs.io/en/latest/api/zmq.html#zmq.Socket.get_monitor_socket))
- The script above is made into a deterministic unit test and passes

thopkins32:
## Description
Adds a feature to the `RemoteDispatcher` that allows a user to wait for a handshake event on the ZMQ socket, even if the dispatcher is started on a separate thread.

## Motivation and Context

The only alternative to this was to sleep for a few seconds and assume the dispatcher is ready.

As described in #2014 , when used as part of a closed-loop decision making system, it can suffer from ZMQ's late-joiner problem, where messages were published before the sub socket was ready to receive them. In this scenario, the closed-loop decision making system would start the `RemoteDispatcher` to listen to events, then immediately launch an Bluesky plan. Without properly waiting for the `RemoteDispatcher` to connect, the start of the Bluesky plan could easily be missed.

Closes #2014 

## How Has This Been Tested?

- Unit tests were added with mocks for the ZMQ parts
- This should also fix flakiness in [Blop's queueserver tutorial](https://blueskyproject.io/blop/main/tutorials/queueserver.html)

## Notes

This late joiner issue is also a problem on the `Publisher` side when connecting to a `Proxy`. If this PR is merged I can follow-up with a handshake fix for that.

### DIFF
--- a/src/bluesky/callbacks/zmq.py
+++ b/src/bluesky/callbacks/zmq.py
@@ -442,7 +442,7 @@ class RemoteDispatcher(Dispatcher):
         optional event loop to use.  Default is to create a new event loop.
     deserializer: function, optional
         optional function to deserialize data. Default is pickle.loads
-    connection_timeout : float, optional
+    handshake_timeout : float, optional
         Configurable timeout to wait for connection handshake.
         Defaults to ``None``, no wait for handshake.
 
@@ -465,7 +465,7 @@ class RemoteDispatcher(Dispatcher):
         deserializer: Callable = pickle.loads,
         strict: bool = False,
         curve_config: ServerCurve | ClientCurve | None = None,
-        connection_timeout: float | None = None,
+        handshake_timeout: float | None = None,
     ):
         if isinstance(prefix, str):
             raise ValueError("prefix must be bytes, not string")
@@ -486,8 +486,9 @@ class RemoteDispatcher(Dispatcher):
         self._socket = None
         self._monitor = None
         self._wait_task = None
-        self._connection_timeout = connection_timeout
-        self._connected = threading.Event()
+        self._handshake_timeout = handshake_timeout
+        self._ready_event = threading.Event()
+        self._poll_ready = asyncio.Event()
 
         def __finish_setup():
             asyncio.set_event_loop(self.loop)
@@ -507,11 +508,11 @@ class RemoteDispatcher(Dispatcher):
                 server_key, _ = zmq.auth.load_certificate(curve_config.server_public_key)
                 sock.setsockopt(zmq.CURVE_SERVERKEY, server_key)
 
-            if self._connection_timeout is not None:
+            if self._handshake_timeout is not None:
                 self._monitor = self._socket.get_monitor_socket(
                     events=zmq.EVENT_HANDSHAKE_SUCCEEDED,
                 )
-                self._wait_task = self.loop.create_task(self._wait_for_handshake(self._connection_timeout))
+                self._wait_task = self.loop.create_task(self._wait_for_ready(self._handshake_timeout))
 
             self._socket.connect(self.address)
             self._socket.setsockopt_string(zmq.SUBSCRIBE, "")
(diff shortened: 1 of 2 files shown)


thopkins32:
Not sure if it makes sense to have a `connected` attribute if it's not actively checking. It might be misleading to include it.

thopkins32:
Re-worked this to be a clear readiness indicator, which has the following criteria at a minimum:
- Handshake event processed successfully within the configured timeout
- The dispatcher is actively in `await socket.recv()`

New API is `dispatcher.ready(timeout=5)` which returns `False` on timeout.

src/bluesky/callbacks/zmq.py
```diff
@@ -554,6 +579,60 @@ async def _poll(self):
...
+            raise
+        finally:
+            if self._socket is not None:
+                self._socket.disable_monitor()
+            if self._monitor is not None:
+                self._monitor.close(linger=0)
+
+    def ready(self, timeout: float | None = None) -> bool:
+        """
+        Block until the SUB socket until we are ready to receive messages.
```
jwlodek:
```suggestion
        Block until the SUB socket is ready to receive messages.
```

jwlodek (approved):
Looks good to me, minus one small grammar adjustment in a docstring.

thopkins32:
@jwlodek thanks for the review. Just one minor fix for the tests incoming and then I will merge.

thopkins32:
Never mind this doesn't solve the slow-joiner problem, TIL that handshakes are not a real readiness signal since subscriptions occur after them. The right way to do this is to have the publisher emit periodic heartbeats through the proxy.

PR #1894, opened 2025-03-14
[ENH] Allow fly plan to use collect_while_completing. Only collect collectable flyers in fly plan.

jwlodek:
## Description

Modifies the signature of `bluesky.plans.fly` to allow for optionally specifying a `stream_name` and a `flush_period`. If set, these will declare a stream for all flyers, and collect while completing every `flush_period` seconds. Thus, the `fly` plan can collect on a set time basis while the `complete` status is not marked as done, which is useful for any sort of live updates about the progress of the flyscan.

## Motivation and Context

Resolves: https://github.com/bluesky/bluesky/issues/1892

Currently, if you want to get any sort of live updates/datum docs emitted during a flyscan, you can't use the built in `fly` plan, and have to write out `kickoff`/`complete`/`collect` yourself. This adds a lot of unnecessary boilerplate.

Also, for detectors that collect datasets of the same length during the scan, or even ones that are scaled by a constant, we can add them into a single stream, for simpler access later. This PR also allows for specifying that stream name when executing a `fly` plan, and having it be declared for us.

This:

```python
def software_flyscan(dets, num_events, exposure_time = 1):
    uid = yield from bps.open_run()
    yield from bps.stage_all(*dets)

    for det in dets:
        yield from bps.prepare(det, TriggerInfo(number_of_events=num_events, trigger=DetectorTrigger.INTERNAL, livetime=exposure_time), wait=True)

    yield from bps.kickoff_all(*dets, wait=True)
    yield from bps.complete_all(*dets, group="complete_all")
    yield from bps.declare_stream(*dets, name="primary")

    done = False
    while not done:
        try:
            yield from bps.wait("complete_all", timeout = 0.5)
        except TimeoutError:
            pass
        else:
            done = True
        yield from bps.collect(*dets, name="primary")
        yield from bps.sleep(0.01)

    yield from bps.unstage_all(*dets)
    yield from bps.close_run()
    return uid
```

Can become this:

```python
def software_flyscan(dets, num_events, exposure_time = 1):
    yield from bps.stage_all(*dets)

    for det in dets:
        yield from bps.prepare(det, TriggerInfo(number_of_events=num_events, trigger=DetectorTrigger.INTERNAL, livetime=exposure_time), wait=True)

    uid = yield from bp.fly(dets, collect_flush_period=0.5, stream_name="primary")
    yield from bps.unstage_all(*dets)
    return uid
```

With this change.

[...]

PR #2017, opened 2026-05-08
Add Pixi and type fixes

thopkins32:
## Description
- Adds the smallest possible `pixi.toml` 
- Fixes type checking issues causing failed builds 

## Motivation and Context
Pixi is a useful tool (could have chosen `uv` as well, but I like Pixi for conda-forge non-python dependencies). It can be expanded further later, this was the minimal needed for development workflows.

Type checking tests were failing, so I took this opportunity to fix them.

## How Has This Been Tested?
`pixi run mypy src` runs without error.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# SCM syntax highlighting & preventing 3-way merges
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true -diff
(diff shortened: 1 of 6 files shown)


jwlodek (approved):
Ran into this when working on #1894 , and fixed these there as well, but I think making it a separate PR is a good idea. I was working on moving the CI to pixi in another PR as well. Per discussion with @tacaswell , he'd like for us to duplicate the dev dependency list as conda packages, but again we can do that in another PR.

PR #2019, opened 2026-05-12
BUG: make RemoteDispatcher.stop safe when called from another thread (#2012)

### LINKED ISSUE #2012: RemoteDispatcher stop method RuntimeError
`bluesky.callbacks.zmq.RemoteDispatcher.stop` method fails when called and the asyncio loop is still running. It needs to cancel the task in a thread-safe manner.

e.g.
```python
def stop(self):
    if self._task is not None:
        self.loop.call_soon_threadsafe(self._task.cancel)
        self._task = None
    if self._socket is not None:
        self._socket.close()
        self._socket = None
    if self._context is not None:
        self._context.destroy()
        self._context = None
    if not self.loop.is_running():
        self.loop.close()
    self.closed = True
```

If this `stop` method is meant to be internal-use only, [since it's executed inside of a try-finally block](https://github.com/bluesky/bluesky/blob/5e1423146939ced25413caea4058369bb2537390/src/bluesky/callbacks/zmq.py#L562-L570), then it should be a private member.

## Steps To Reproduce
```python
"""
Reproduces the RuntimeError("Cannot close a running event loop") bug in
bluesky's RemoteDispatcher.stop() when called from a different thread than
the one running the event loop.
No ZMQ broker is needed — the dispatcher will simply fail to connect and poll,
but the bug manifests during teardown regardless.
Usage:
    python reproduce_remote_dispatcher_stop.py
"""
import threading
import time
def main():
    # Any address will do — we just need the loop to be running.
    dispatcher = RemoteDispatcher("localhost:60610")
    thread = threading.Thread(target=dispatcher.start, daemon=True)
    thread.start()
    # Give the loop a moment to start.
    time.sleep(0.5)
    print("Calling dispatcher.stop() from main thread...")
    try:
        dispatcher.stop()
        print("stop() succeeded (bug is fixed)")
    except RuntimeError as e:
        print(f"RuntimeError: {e}")
        print("Bug confirmed: stop() cannot be called from another thread")
if __name__ == "__main__":
    main()
```

## Acceptance Criteria
- Script above is a unit tests that passes *or* `stop` is made private

SAY-5:
## Description

`bluesky.callbacks.zmq.RemoteDispatcher.stop` cancelled `self._task`, closed the ZMQ socket/context, and then called `self.loop.close()` directly from the caller's thread. When `start` is driven from a worker thread (a documented usage pattern), `stop` is invoked from a different thread than the one running the event loop. In that case the bare `self._task.cancel()` is not thread-safe and `self.loop.close()` raises `RuntimeError: Cannot close a running event loop`.

This patch:

* Schedules the task cancel via `loop.call_soon_threadsafe` when the loop is still running and returns early. Cleanup of the socket, context, and loop happens in the `start` method's `finally` block once `run_until_complete` has actually returned and the loop has stopped.
* When the loop is no longer running (the in-process / SIGINT case still exercised by `test_zmq`), it cancels the task, closes the socket/context, and closes the loop as before, but guards against double-close.

## Motivation and Context

Fixes #2012. Reproducer from the issue:

```python
dispatcher = RemoteDispatcher(\"localhost:60610\")
threading.Thread(target=dispatcher.start, daemon=True).start()
time.sleep(0.5)
dispatcher.stop()  # previously raised RuntimeError(\"Cannot close a running event loop\")
```

## How Has This Been Tested?

Added `test_remote_dispatcher_stop_from_other_thread_does_not_raise` in `src/bluesky/tests/test_zmq.py` that reproduces the issue scenario (start in a worker thread, stop from the main thread) and asserts the worker exits cleanly and `dispatcher.closed` becomes `True`. All non-integration tests in `test_zmq.py` continue to pass locally.

### DIFF
--- a/src/bluesky/callbacks/zmq.py
+++ b/src/bluesky/callbacks/zmq.py
@@ -570,11 +570,30 @@ class RemoteDispatcher(Dispatcher):
             self.stop()
 
     def stop(self):
+        # ``stop`` is documented as part of the public API and is also called
+        # from the ``finally`` block of :meth:`start`.  It may therefore be
+        # invoked from the same thread that is running the event loop (the
+        # in-process / interrupt case) or from a different thread (e.g. when
+        # the dispatcher's ``start`` is driven from a worker thread and the
+        # main thread tears it down).  When called from a non-loop thread we
+        # only schedule task cancellation here — the cleanup of socket,
+        # context and the loop itself happens in :meth:`start`'s ``finally``
+        # block once the loop has actually stopped.  Calling
+        # :py:meth:`asyncio.AbstractEventLoop.close` while the loop is still
+        # running raises ``RuntimeError: Cannot close a running event loop``.
+        if self.loop.is_running():
+            if self._task is not None:
+                self.loop.call_soon_threadsafe(self._task.cancel)
+            return
         if self._task is not None:
             self._task.cancel()
+            self._task = None
         if self._socket is not None:
             self._socket.close()
+            self._socket = None
         if self._context is not None:
             self._context.destroy()
-        self.loop.close()
+            self._context = None
+        if not self.loop.is_closed():
+            self.loop.close()
         self.closed = True
--- a/src/bluesky/callbacks/zmq.py
+++ b/src/bluesky/callbacks/zmq.py
@@ -590,17 +590,15 @@ class RemoteDispatcher(Dispatcher):
         self.closed = True
         self._stopped.set()
 
-    def stop(self, timeout=None):
+    def stop(self):
         """Stop the dispatcher and wait for it to finish.
 
         When called from a thread other than the one running the event loop,
         task cancellation is scheduled on the loop and this method blocks
-        until :meth:`start` has torn the dispatcher down. ``timeout`` is the
-        number of seconds to wait, or ``None`` to wait indefinitely.
+        until :meth:`start` has torn the dispatcher down.
         """
+
         if self.loop.is_running():
             if self._task is not None:
                 self.loop.call_soon_threadsafe(self._task.cancel)
-            self._stopped.wait(timeout=timeout)
-        else:
-            self._cleanup()
+            self._stopped.wait()
--- a/src/bluesky/callbacks/zmq.py
+++ b/src/bluesky/callbacks/zmq.py
@@ -16,6 +16,7 @@ import asyncio
 import copy
 import logging
 import pickle
+import threading
 import warnings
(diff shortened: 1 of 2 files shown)


src/bluesky/callbacks/zmq.py
```diff
@@ -570,11 +570,30 @@ def start(self):
...
+        # from the ``finally`` block of :meth:`start`.  It may therefore be
+        # invoked from the same thread that is running the event loop (the
+        # in-process / interrupt case) or from a different thread (e.g. when
+        # the dispatcher's ``start`` is driven from a worker thread and the
+        # main thread tears it down).  When called from a non-loop thread we
+        # only schedule task cancellation here — the cleanup of socket,
+        # context and the loop itself happens in :meth:`start`'s ``finally``
+        # block once the loop has actually stopped.  Calling
+        # :py:meth:`asyncio.AbstractEventLoop.close` while the loop is still
+        # running raises ``RuntimeError: Cannot close a running event loop``.
```
thopkins32: Can convert this to a proper docstring for the method.

src/bluesky/callbacks/zmq.py
```diff
@@ -570,11 +570,30 @@ def start(self):
...
+        if self.loop.is_running():
+            if self._task is not None:
+                self.loop.call_soon_threadsafe(self._task.cancel)
+            return
```
thopkins32: The early return here is misleading, in my opinion. A caller of `dispatcher.stop()` may think that the dispatcher has been stopped on return, but in this case it won't be until the task cancellation actually runs in the asyncio loop.

That's why in your test you had to wait for the thread to join before doing the assertions.

It may be simpler to have `stop` be something synchronous, like:
```python
def stop(self, timeout=None):
    if self.loop.is_running():
        if self._task is not None:
            self.loop.call_soon_threadsafe(self._task.cancel)
        self._stopped.wait(timeout=timeout)
```
where `self._stopped = threading.Event()` that does `self._stopped.set()` when the `start` method exits after cleanup (at the end of the `finally` block).

Then move the rest of `stop` into the `finally` block of `start` before setting the `self._stopped` event.

thopkins32 (changes requested):
Thanks for the PR.

I have one concern about the early return in `stop` and offered a potential solution. Let me know what you think.

thopkins32 (approved):
LGTM, I made some minor edits and will merge once CI passes. `stop` doesn't need a `timeout` (I realized) and it also doesn't need to cleanup if the loop is already stopped.

PR #2021, opened 2026-05-20
BUG: suppress OpenTelemetry import warning

### LINKED ISSUE #2020: Address DeprecationWarning from opentelemetry
In a [bluesky-tiled-plugins CI run](https://github.com/bluesky/bluesky-tiled-plugins/actions/runs/26103165571/job/76759637870#step:6:28) we saw a `DeprecationWarning` originating from `bluesky.run_engine`:

```
E   DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
```

Expand for full traceback:

<details>

```python
tests/conftest.py:4: in <module>
    from bluesky.run_engine import RunEngine, TransitionError
.venv/lib/python3.10/site-packages/bluesky/__init__.py:4: in <module>
    from .run_engine import RunEngine
.venv/lib/python3.10/site-packages/bluesky/run_engine.py:23: in <module>
    from opentelemetry import trace
.venv/lib/python3.10/site-packages/opentelemetry/trace/__init__.py:74: in <module>
    from opentelemetry import context as context_api
.venv/lib/python3.10/site-packages/opentelemetry/context/__init__.py:58: in <module>
    _RUNTIME_CONTEXT = _load_runtime_context()
.venv/lib/python3.10/site-packages/opentelemetry/context/__init__.py:50: in _load_runtime_context
    entry_points(  # type: ignore
.venv/lib/python3.10/site-packages/opentelemetry/util/_importlib_metadata.py:41: in entry_points
    return _original_entry_points_cached().select(**params)
.venv/lib/python3.10/site-packages/opentelemetry/util/_importlib_metadata.py:37: in _original_entry_points_cached
    return _as_entry_points(original_entry_points())
.venv/lib/python3.10/site-packages/opentelemetry/util/_importlib_metadata.py:32: in _as_entry_points
    return EntryPoints(ep for group_eps in eps.values() for ep in group_eps)
E   DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
```

</details>

Mirochill:
## Description

This suppresses the known OpenTelemetry/importlib-metadata `SelectableGroups dict interface is deprecated. Use select.` warning only while importing `opentelemetry.trace`. `run_engine.py` now obtains `get_current_span()` from the local tracing module instead of importing `opentelemetry.trace` directly at module import time.

It also adds a regression that imports `bluesky.run_engine` in a subprocess with `DeprecationWarning` treated as an error.

## Motivation and Context

Fixes #2020. Downstream projects that run tests with warnings as errors can fail during collection before any RunEngine code is used, because importing `bluesky.run_engine` imports OpenTelemetry and triggers this dependency warning.

## How Has This Been Tested?

Not run locally.
Static validation was limited to Git whitespace checks: `git diff --check`, `git diff --cached --check`, and `git diff --check HEAD~1..HEAD`.

## Disclosure

This PR is AI-assisted. I am responsible for the proposed change and for responding to review feedback.

### DIFF
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -20,8 +20,6 @@ from warnings import warn
 
 import event_model
 from event_model import DocumentNames
-from opentelemetry import trace
-from opentelemetry.trace import Span
 
 from bluesky._vendor.super_state_machine.errors import TransitionError
 from bluesky._vendor.super_state_machine.extras import PropertyMachine
@@ -44,7 +42,7 @@ from .protocols import (
     Triggerable,
     check_supports,
 )
-from .tracing import tracer
+from .tracing import get_current_span, tracer
 from .utils import (
     AsyncInput,
     CallbackRegistry,
@@ -70,6 +68,11 @@ from .utils import (
 )
 
 _SPAN_NAME_PREFIX = "Bluesky RunEngine"
+
+if typing.TYPE_CHECKING:
+    from opentelemetry.trace import Span
+else:
+    Span = typing.Any
 
 current_task: typing.Callable[[asyncio.AbstractEventLoop | None], asyncio.Task | None]
 try:
@@ -2175,7 +2178,7 @@ class RunEngine:
 
         where <GROUP> is a hashable identifier.
         """
-        _set_span_msg_attributes(trace.get_current_span(), msg)
+        _set_span_msg_attributes(get_current_span(), msg)
         kwargs = dict(msg.kwargs)
         group = kwargs.pop("group", None)
         obj = check_supports(msg.obj, Flyable)
@@ -2196,7 +2199,7 @@ class RunEngine:
             Msg('collect', flyer_object)
             Msg('collect', flyer_object, stream=True, return_payload=False, name="a_name")
         """
-        _set_span_msg_attributes(trace.get_current_span(), msg)
+        _set_span_msg_attributes(get_current_span(), msg)
         run_key = msg.run
         if (
             current_run := self._run_bundlers.get(run_key, key_absence_sentinel := object())
@@ -2233,7 +2236,7 @@ class RunEngine:
 
         where arguments are passed through to `obj.set(*args, **kwargs)`.
         """
-        _set_span_msg_attributes(trace.get_current_span(), msg)
+        _set_span_msg_attributes(get_current_span(), msg)
         obj = check_supports(msg.obj, Movable)
         kwargs = dict(msg.kwargs)
         group = kwargs.pop("group", None)
@@ -2280,7 +2283,7 @@ class RunEngine:
 
         where ``<GROUP>`` is any hashable key and ``<ERROR_ON_TIMEOUT>`` is a boolean.
         """
-        _set_span_msg_attributes(trace.get_current_span(), msg)
+        _set_span_msg_attributes(get_current_span(), msg)
         done = False  # boolean that tracks whether waiting is complete
         if msg.args:
             (group,) = msg.args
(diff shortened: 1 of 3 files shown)


tacaswell:
Do we understand why we are starting to see this now?  A quick look at upstream suggests that this deprecation is several years old.

Did this get reported upstream?  Can we fix this by bumping the minimum open telemetry version?

tacaswell:
This looks like at a minimum AI assisted.

Mirochill:
Thanks for checking. Yes, this PR is AI-assisted; I added that explicitly to the PR body. I am responsible for the proposed change and review follow-up.

On the OpenTelemetry side: yes, this is already reported upstream as open-telemetry/opentelemetry-python#5231. The failing downstream run linked from #2020 installed `opentelemetry-api==1.42.0`, and the traceback goes through the new stdlib `importlib.metadata` compatibility path added in 1.42.0. The current latest OpenTelemetry release, `v1.42.1`, does not change `opentelemetry-api/src/opentelemetry/util/_importlib_metadata.py`, so raising Bluesky's minimum to `>=1.42.0` or `>=1.42.1` would not avoid this warning on Python 3.10/3.11.

I also checked the remote CI here. The new import regression passed across the published matrix. The only red job is `test (windows-latest, 3.11)`, failing in existing `src/bluesky/tests/test_zmq.py::test_zmq[True]` with `_queue.Empty`; it does not touch the changed import path.

Given that, I see three realistic options:

1. Keep this local narrow filter, so `bluesky.run_engine` remains importable for downstream warning-as-error suites while upstream sorts out #5231.
2. Close this and wait for an OpenTelemetry release that fixes #5231.
3. Drop the filter and only move the `run_engine.py` direct import behind `tracing.py`, but that does not fully protect this import path if `tracing.py` still imports OpenTelemetry under `-W error`.

tacaswell:
What changed that we are now seeing this error?  Can we simply exclude certain versions of open telemetry?

Mirochill:
I traced this a bit further from the downstream CI and the OpenTelemetry tags.

What changed is OpenTelemetry, not Bluesky or the Python runner image. The failing `bluesky-tiled-plugins` run resolved `opentelemetry-api==1.42.0` on 2026-05-19. That release includes open-telemetry/opentelemetry-python@1a33c692f0248271023519fd52a1f6b51d787655 / #5203, which removed the third-party `importlib-metadata` dependency and switched this wrapper to stdlib `importlib.metadata` on Python >=3.10. On Python 3.10/3.11, the no-argument stdlib `entry_points()` path can return the deprecated dict-like `SelectableGroups`; OpenTelemetry then calls `eps.values()` in `_as_entry_points()`, which emits the warning under `-W error`. That also explains why the same downstream matrix passed on Python 3.13.

I checked `v1.42.1` too: it only fixes #5241, and `_importlib_metadata.py` is unchanged in `v1.42.1` and current `main` as of this check.

Excluding versions is a reasonable alternative if you prefer package metadata over a local import-time filter. The narrow known-bad constraint would be something like `opentelemetry-api!=1.42.0,!=1.42.1`, which would make fresh resolves use `1.41.1` until OpenTelemetry publishes a fixed release, while still allowing a future `1.42.2+` if it fixes #5231. The tradeoff is that it changes dependency resolution for all Bluesky installs, and the `dev` extra may also need to resolve an older matching `opentelemetry-sdk`, since the SDK typically pins the API version exactly. The current patch instead leaves dependency resolution alone and suppresses only this one upstream warning while importing the tracing shim.

I have not changed the branch yet. I can switch this PR to the version-exclusion approach if that is the direction you want.

PR #2022, opened 2026-05-23
Improve runtime of ZMQ tests

### LINKED ISSUE #1690: Rewrite the ZeroMQ tests to speed them up
The [ZeroMQ tests](https://github.com/bluesky/bluesky/blob/main/src/bluesky/tests/test_zmq.py) are long-running system tests that slow down CI and local testing. They should be rewritten.

The current tests throw up and connect a ZeroMQ socket in a subprocess and then sleep for several seconds, waiting for it to start. This is very slow and potentially unreliable.

## Possible Solution
Some ideas:
- More functionality could be tested with mocking and unit-style tests, which are much faster
- There could be a single system test alonside the unit tests, to test against a real socket while minimizing the long-running tests
- Rather than kickoff in a subprocess and wait 5 seconds, could make use of a multiprocessing `Event` to synchronise on

thopkins32:
## Description

Improves the runtime of the ZMQ tests since they were doing multiple 5 second sleeps which are also unreliable. If #2015 is merged, we can eliminate the remaining long sleeps.

ZMQ binds don't require any sleep, only ZMQ connect does (unless you monitor the socket for a handshake).

I rewrote some redundant tests with mocks to prove correctness. There still exist full integration tests that cover the round trip from `Publisher -> Proxy -> RemoteDispatcher`.

Closes #1690 
Closes #1738 

I also added some basic implementation of the Bluesky protocols to `conftest.py` that make testing very simple.

## Motivation and Context

The ZMQ tests were extremely slow because each instance would sleep for 5 seconds multiple times.

## How Has This Been Tested?

From `main`, running `pixi run pytest src/bluesky/tests/test_zmq.py` on my machine yields:
```
48 passed in 92.25s (0:01:32)
```

From this branch, running the same command on my machine (Linux) yields:
```
44 passed in 2.58s 
```

Code coverage has actually slightly improved as well.


jwlodek (approved):
LGTM. One note - I used `pytest_mock` for the mocking for some of the later zmq tests. I don't think there's any technical reason we can't use a mix of unittest.mock and pytest_mock, but maybe we should consider consolidating to one or the other?

src/bluesky/tests/test_zmq.py
```diff
@@ -73,315 +77,208 @@ def test_proxy_script():
...
+    RemoteDispatcher over 5568. The RemoteDispatcher will push them into
+    the queue, where we can verify that they round-tripped.
+    """
+    RE = RunEngine({})
+    RE.subscribe(publisher)
 
-    # Check that numpy stuff is sanitized by putting some in the start doc.
-    md = {"stuff": {"nested": np.array([1, 2, 3])}, "scalar_stuff": np.float64(3), "array_stuff": np.ones((3, 3))}
+    remote_stop_event = dispatcher[0]
+    remote_docs = dispatcher[1]
```
tacaswell:
```suggestion
    remote_stop_event, remote_docs = dispatcher
```

Super picky, take it or leave it.

tacaswell (approved):
Self merge after considering one suggestion.

thopkins32:
@jwlodek I'm used to `unittest.mock`. I particularly like `MagicMock` from that library. But I believe they should basically be equivalent? I'm not familiar with pytest's version.

PR #2024, opened 2026-05-26
Parse `object_classes` of described objects and add them to the event descriptor

### LINKED ISSUE #2013: Feature: Add import path of object class to desccriptor
Small feature:

A couple of times I've been writing callbacks, I would have found it useful to have the device class of the objects read in the event descriptor.

In the same way we have `object_keys` linking the device which will have `read` called, could add a key `object_classes` which adds the import uri of the device for which read was called?

evvaaaa:
Closes #2013 

The bluesky component of the above issue. This should be merged and released alongside event-model with https://github.com/bluesky/event-model/pull/375

### DIFF
--- a/src/bluesky/tests/test_external_assets_and_paging.py
+++ b/src/bluesky/tests/test_external_assets_and_paging.py
@@ -646,9 +646,9 @@ def test_object_classes(RE):
         def read(self):
             return {f"{self.name}-sig1": Reading(value=0, timestamp=123)}
 
-    class SomeStaticReadable:
+    class SomeClassReadable:
         parent = None
-        name = "some_static_readble"
+        name = "some_class_readble"
 
         @classmethod
         def describe(cls):
@@ -658,13 +658,28 @@ def test_object_classes(RE):
         def read(cls):
             return {f"{cls.name}-sig1": Reading(value=0, timestamp=123)}
 
+    class SomeStaticReadable:
+        parent = None
+        name = "some_static_readble"
+
+        @staticmethod
+        def describe():
+            return {"some_static_readble-sig1": DataKey(source="pv", dtype="number", shape=[])}
+
+        @staticmethod
+        def read():
+            return {"some_static_readble-sig1": Reading(value=0, timestamp=123)}
+
     docs = DocHolder()
 
-    RE(bp.count([SomeReadable(), axis, SomeStaticReadable]), docs.append)
+    RE(bp.count([SomeReadable(), axis, SomeStaticReadable, SomeClassReadable]), docs.append)  # type: ignore
     assert len(docs["descriptor"]) == 1
     descriptor = docs["descriptor"][0]
     assert descriptor["object_classes"] == {
         "det1": "ophyd.sim.SynAxis",
+        "some_class_readble": (
+            "bluesky.tests.test_external_assets_and_paging.test_object_classes.<locals>.SomeClassReadable"
+        ),
         "some_readble": "bluesky.tests.test_external_assets_and_paging.test_object_classes.<locals>.SomeReadable",
         "some_static_readble": (
             "bluesky.tests.test_external_assets_and_paging.test_object_classes.<locals>.SomeStaticReadable"
--- a/src/bluesky/tests/test_external_assets_and_paging.py
+++ b/src/bluesky/tests/test_external_assets_and_paging.py
@@ -24,6 +24,7 @@ from bluesky.protocols import (
     WritesExternalAssets,
     WritesStreamAssets,
 )
+from bluesky.tests import requires_ophyd
 
 
 class DocHolder(dict):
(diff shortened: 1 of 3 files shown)


evvaaaa:
CI will fail while the corresponding event model branch isn't being used.

evvaaaa:
@danielballan Feel free to assign a different reviewer, I'm never quite sure who to add from NSLS :)

danielballan:
Thanks, @evvaaaa! Looks good to me. I'll add a couple.

evvaaaa:
@jacopoabramo Would you be able to give a quick review?

It's a small change already accounted for in event-model and I would like it in the next release for our nexus writer.

src/bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -624,3 +625,48 @@ def test_multiple_declare_in_same_stream(RE):
     assert docs["descriptor"][1]["name"] == "main"
     assert frozenset(docs["descriptor"][1]["data_keys"]) == frozenset(data_keys)
     assert all(d["descriptor"] == docs["descriptor"][1]["uid"] for d in docs["stream_datum"][4:])
+
+
+@requires_ophyd
+def test_object_classes(RE):
+    from ophyd.sim import SynAxis  # type: ignore
```
jacopoabramo, line 632: Is it necessary to do a local import rather than putting it on top of the module?

src/bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -624,3 +625,48 @@ def test_multiple_declare_in_same_stream(RE):
...
+        parent = None
+        name = "some_static_readble"
+
+        @classmethod
+        def describe(cls):
+            return {f"{cls.name}-sig1": DataKey(source="pv", dtype="number", shape=[])}
+
+        @classmethod
+        def read(cls):
+            return {f"{cls.name}-sig1": Reading(value=0, timestamp=123)}
```
jacopoabramo: Just out of curiosity: supporting `staticmethod` is out of scope right?

jacopoabramo (approved):
Looks good, changes are minimal so I don't expect much trouble.

This forces `event-model` to be `1.24.0` minimum right? Should this requirement be pinned for safety?

src/bluesky/tests/test_external_assets_and_paging.py
evvaaaa: I was following the lead of how ophyd is currently used in tests.

Since `tests/` is a submodule of bluesky I can see the point. Avoids possible import errors from only dev group dependencies

jacopoabramo: Fair enough; it's just a minor nitpick based on taste anyway. Consider it resolved.

src/bluesky/tests/test_external_assets_and_paging.py
evvaaaa: I was about to comment that I was surprised a method with `classmethod` implementations passed, and that I didn't think that `staticmethod` would pass the protocol.

Both do. Added another device with `staticmethod`.

jacopoabramo: Yeah, after seeing your last commit I was about to say that I would strongly not reccomend such a pattern but you never know. Might as well test for it.

src/bluesky/tests/test_external_assets_and_paging.py
```diff
@@ -624,3 +625,63 @@ def test_multiple_declare_in_same_stream(RE):
...
+    assert len(docs["descriptor"]) == 1
+    descriptor = docs["descriptor"][0]
+    assert descriptor["object_classes"] == {
+        "det1": "ophyd.sim.SynAxis",
+        "some_class_readble": (
+            "bluesky.tests.test_external_assets_and_paging.test_object_classes.<locals>.SomeClassReadable"
+        ),
+        "some_readble": "bluesky.tests.test_external_assets_and_paging.test_object_classes.<locals>.SomeReadable",
+        "some_static_readble": (
+            "bluesky.tests.test_external_assets_and_paging.test_object_classes.<locals>.SomeStaticReadable"
```
jacopoabramo, line 685: I'm not very familiar with how `__qualname__` works. There's no way to alter this though, right? As in adding a method such as (for example) `__repr__` that would alter the output of `__qualname__`.

evvaaaa:
> This forces `event-model` to be `1.24.0` minimum right? Should this requirement be pinned for safety?

Good point

src/bluesky/tests/test_external_assets_and_paging.py
evvaaaa: You _can_ do this, and it passes typing, but it'd be pretty strange to do given the description of qualname on the [PEP](https://peps.python.org/pep-3155/#proposal).

``` python
class Y: ...

class X:
    __qualname__ = "Foo"

print(Y.__qualname__)
print(X.__qualname__)
```

If the user decides to do this then we take their qualname as it's given.

jacopoabramo: Fine by me!

jacopoabramo (approved):

jacopoabramo:
The failing tests seem to be unrelated, but I'm not sure. At any rate I don't have merging rights. Maybe @coretl can help

coretl:
I think the tests still need fixing. They seem to have a race condition between the suspender firing and the plan executing, maybe slackening off the timing might make them pass? Might be unrelated to this PR, or might be because of the tiny extra amount of time spend making the start document...

jacopoabramo:
I suspect the fix is included in #2026, probably

coretl (approved):

jwlodek (approved):
LGTM

PR #2026, opened 2026-06-10
Various test improvements

thopkins32:
## Description

This PR aims to reduce the runtime of the most time consuming tests. Each change is described below with justification (help from AI to get the per-test descriptions). All code was written by myself.

#### `test_in_plan_qt_callback`
- **Change:** `hw.motor` / `hw.det` replaced with `MovableSignal` / `ReadableSignal`; `motor.delay = 1` removed; uses a fresh local `RunEngine({})`.
- **Why:** The 1-second `motor.delay` on a 25-point scan added 25 seconds of wall-clock waiting. `MovableSignal.set()` returns an `AlwaysSuccessfulStatus` that's done immediately. The Qt callback wiring (the actual subject of the test) is unchanged.
---
#### `test_interrupted`
- **Change:** Installs a `SlowMovable` whose `set()` returns a status that never completes; a `msg_hook` fires the SIGINT the moment the RunEngine reaches its `wait` state; uses the existing `deterministic_sigint` fixture instead of `os.kill(os.getpid(), SIGINT)` directly.
- **Why:** The original deliberately set `motor.delay = 10` so a 1s timer could land mid-move — a wall-clock race. The new design short-circuits to "fire SIGINT as soon as RE actually starts waiting," giving sub-second runtime and removing the flake risk.
---
#### `test_pause_resume` — **REMOVED**
- **Why:** Pause-resume from another thread is covered in depth by `test_run_engine.py`. The deleted test also tail-padded with a meaningless `time.sleep(3)` after asserting `RE.state == "idle"`.
#### `test_pause_from_outside` — **REMOVED**
- **Why:** `request_pause()` from a separate thread is also covered in `test_run_engine.py`. The deleted test ran back-to-back `time.sleep(1)` + `time.sleep(2)` timer threads with no unique assertion.
---
#### `test_deferred_pause_from_suspend`
- **Change:** A `msg_hook` sets a `waiting_event` when the engine reaches `wait_for`; the deferred pause request and the `sig.put(0)` are then chained through `threading.Event`s. Also drops the `RE` fixture in favor of a local `RunEngine({})`.
- **Why:** The original `threading.Timer(1, ...)` / `threading.Timer(4, ...)` were guessed delays meant to land at specific RunEngine state transitions — 4s of wasted wall-clock and a real timing race. Dropping the `RE` fixture eliminates a duplicate invocation caused by the fixture's parametrization.
---
#### `test_many_motors`
- **Change:** Grid reduced from `5 × 5 × 5 × 5 = 625` points to `2 × 2 × 2 × 2 = 16` points; `hw.motor*` / `hw.ab_det` replaced with `MovableSignal` / `ReadableSignal`.
- **Why:** The test only validates "too many motors → no figure produced, warning issued, table still works." Full grid traversal is not part of the contract under test; the warning still triggers at 16 points.
---
#### `test_stage_all_and_unstage_all`
- **Change:** `time.sleep(0.5)` (old-style) and `threading.Timer(0.5, ...)` (new-style) replaced with `threading.Barrier(2)`s for stage and unstage. Removed the brittle `assert 3 < stop - start < 4` wall-clock check.
- **Why:** The sleeps merely *allowed* concurrent staging — they didn't prove it. With a barrier, if the new-style devices were staged sequentially, the first would block forever waiting for the second and hit the barrier timeout. The new test is both stronger (concurrency is *required*, not merely *possible*) and ~2 seconds faster.
---
#### `test_plans` (+ `xfail` variants) — **REMOVED**
- **Change:** The ~17-case parametrize matrix that re-ran every plan to verify `hints`, `dimensions`, and `uid` in the start doc is deleted. Equivalent assertions are added to existing tests in `test_scans.py` (`test_count`, `test_absolute_spiral`, `test_rel_spiral`, `test_absolute_fermat_spiral`, `test_relative_fermat_spiral`) via a small start-doc collector.
- **Why:** The same plans were already being executed in `test_scans.py` — the parametrize matrix was duplicate plan execution that only differed in which assertions it made. Folding the assertions into the existing tests removes the duplication. The two `xfail`-marked variants (motors without hints / with empty hints) were also deleted because they were marked to fail and contributed no signal.
---
#### `test_grid_scan` / `test_rel_grid_scan` parametrize cases
- **Change:** Every axis reduced from 3–6 points to 2 points, across all 10 parametrize cases.
- **Why:** The test validates *position ordering* (snake vs. non-snake, deprecated argument forms), which only requires ≥2 points per axis to be meaningful. Previously each case produced up to 108-position trajectories; now 8 positions per case.

## Motivation and Context

Tests take a long time when running the CI jobs, which is slow when trying to develop and maintain this project. 

The sheer quantity of tests is the other obvious culprit for the long runtime of the test suite, I think there is quite a bit of duplication here that could save a lot of time, but this will require more careful work, and should also be handled by smaller, focused PRs.

### Local test
Tested on my machine, I get the following improvements.

#### Before
```
2335 passed, 22 skipped, 34 xfailed in 324.14s (0:05:24) 
```

#### After
```
2293 passed, 22 skipped, 2 xfailed in 174.88s (0:02:54) 
```

### CI Runners
ubuntu-latest, dev
#### Before
```
2339 passed, 22 skipped, 34 xfailed in 698.26s (0:11:38)
```
#### After
```
2297 passed, 22 skipped, 2 xfailed in 387.34s (0:06:27)
```


jwlodek (approved):
These all look reasonable to me, but would appreciate a second pair of eyes.

src/bluesky/tests/test_plans.py
```diff
@@ -475,20 +475,20 @@ def _grid_scan_position_list(args, snake_axes):
...
     [
         # Calls using new arguments
-        (("motor", 1, 2, 3, "motor1", 4, 5, 6, "motor2", 7, 8, 9), None),
-        (("motor", 1, 2, 3, "motor1", 4, 5, 6, "motor2", 7, 8, 9), False),
-        (("motor", 1, 2, 3, "motor1", 4, 5, 6, "motor2", 7, 8, 9), True),
-        (("motor", 1, 2, 3, "motor1", 4, 5, 6, "motor2", 7, 8, 9), []),  # Empty list will disable snaking
-        (("motor", 1, 2, 3, "motor1", 4, 5, 6, "motor2", 7, 8, 9), ["motor1"]),
-        (("motor", 1, 2, 3, "motor1", 4, 5, 6, "motor2", 7, 8, 9), ["motor2"]),
-        (("motor", 1, 2, 3, "motor1", 4, 5, 6, "motor2", 7, 8, 9), ["motor1", "motor2"]),
+        (("motor", 1, 2, 2, "motor1", 4, 5, 2, "motor2", 7, 8, 2), None),
```
tacaswell: These were different to make in harder for transposes / flips to sneak through as working.

thopkins32: Should I change to 2, 3, 4 then? I thought this might be the case

src/bluesky/tests/test_suspenders.py
```diff
@@ -251,8 +251,9 @@ def test_suspend_when_changed_preserves_falsy_expected_value(hw):
     assert susp._should_suspend(1)
 
 
-def test_deferred_pause_from_suspend(RE, hw):
+def test_deferred_pause_from_suspend(hw):
```
tacaswell: This is one of the tests where I would expect the fixture is most important as we are messing with the RE state and if it fails it may fail in a dramatic way that leave the RE in weird state.

src/bluesky/tests/test_plans.py
tacaswell: I would go 2,3,5 so they are all (relatively) prime.

thopkins32:
TODO: Coverage dropped after removing `test_simple_api.py`, need to add tests to `test_bec.py` to cover this

PR #2028, opened 2026-06-17
ENH: Added during_task setter and getter to RunEngine

vshekar:
Added during_task setter and getter to the RunEngine to allow users to set it after initializing it

## Description
Added `@property` and `setter` to set private self._during_task. Added simple tests along with it

## Motivation and Context
This will allow a user to set during_task on an already initialized RE instance, for example a RunEngine initialized in a startup profile can later be integrated into a simple qt GUI, instead of pumping the Qt event loop via matplotlib imports one can write and attach their own DuringTask object. This also avoids directly assigning a DuringTask object to the private _during_task.

## How Has This Been Tested?
Added tests to test suite

### DIFF
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -613,7 +613,7 @@ class RunEngine:
         --------
         >>> from bluesky import RunEngine
         >>> RE = RunEngine(during_task=CustomDuringTask())
-        >>> RE.during_task 
+        >>> RE.during_task
         """
         return self._during_task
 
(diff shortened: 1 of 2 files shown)


src/bluesky/run_engine.py
```diff
@@ -604,6 +604,32 @@ def commands(self):
...
+        """
+        Property to set during_task after the RunEngine has initialized
+
+        Examples
+        --------
+        >>> from bluesky import RunEngine
+        >>> RE = RunEngine()
+        >>> RE.during_task = CustomDuringTask()
+        """
+        self._during_task = value
```
jacopoabramo, line 631: Just nit-picking really, but maybe if the run engine is not in idle it should raise an exception if there's an attempt to set `_during_task` while there's a plan running.

PR #2029, opened 2026-06-18
MNT: fix `Stoppable.stop` signature

### LINKED ISSUE #1999: Stoppable protocol is wrong
https://github.com/bluesky/bluesky/blob/c5fa3a30968c4ca4dd88eec5d629f60c637897ea/src/bluesky/protocols.py#L425

should match ophyd and ophyd-async

```python
    def stop(self, *, success=False) -> SyncOrAsync[None]:
```

jacopoabramo:
Closes #1999


jacopoabramo:
@coretl your ticket shows that the default value of `success` should be `False` but I sort of disagree with that. I think that it makes sense that the default condition of stopping should be a success, not a failure. Unless there's a valid reason for that.

coretl:
It's kind of a judgement call whether `stop` means `stop, something's wrong` or `stop, I'm finished`. Either way, this is not a user facing protocol, this is a RunEngine facing protocol, so the more important thing is matching what the implementations do:
https://github.com/search?q=repo%3Abluesky%2Fophyd%20stop&type=code
https://github.com/search?q=repo%3Abluesky%2Fophyd-async%20stop&type=code

The important thing is that the RunEngine always calls it with `success=True`, which gets you what you want:
https://github.com/bluesky/bluesky/blob/d276e2bf87ed6bc5cd8e1d20de605f7fec9812ca/src/bluesky/run_engine.py#L1461-L1470

coretl (approved):

PR #2030, opened 2026-06-18
ENH: add StatusWithResult

jacopoabramo:
## Description
Extend `Status` with the ability to return a value after an operation is complete.

## Motivation and Context
Should be used for a future `Executable` protocol, implemented in #1995 .

## How Has This Been Tested?
No tests added yet. Since it's a protocol, it's down to the implementer on the details of how it behaves. The `RE` shouldn't touch whatever results it carries anyway.


jacopoabramo:
@danielballan feel free to merge if this makes sense

danielballan:
I'm good with it going in. I'll miss Monday's call, but if anyone named "Tom" also agrees this is good to go, feel free to merge this then.

coretl (approved):

PR #2040, opened 2026-07-13
MNT: use `TypeIs` for type narrowing in `iterate_maybe_async`

jacopoabramo:
## Description

Uses `TypeIs` to correctly narrow the type of `SyncOrAsyncIterator` through `inspect.isasyncgen`.

## Motivation and Context

Noticed in the last batch of commits this type error was introduced (not sure how).

## How Has This Been Tested?

`tox -e type-checking` locally.


jwlodek (approved):
This looks correct to me, any objections with merging it?

PR #2041, opened 2026-07-13
Loosen RE metadata type hint

jwlodek:
Without this, `MutableMapping` subclasses like `RedisJSONDict` would be flagged by static analysis type checking as invalid inputs to `RunEngine.__init__`.

- **Loosen type hint for RE.md to be Mapping instead of dict**
- **Type must be mutable. Add unit test that confirms behavior, type checker passes**

### DIFF
--- a/src/bluesky/run_engine.py
+++ b/src/bluesky/run_engine.py
@@ -9,7 +9,7 @@ import threading
 import typing
 import weakref
 from collections import ChainMap, defaultdict, deque
-from collections.abc import Callable, Mapping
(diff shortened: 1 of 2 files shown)


jacopoabramo:
FYI there was a similar work going on in #1937 with some discussion on how to properly type annotate the key-value paring. It might be worthwhile to pick that one up again?

jwlodek:
Indeed, missed the previous PR. Closing this one as a duplicate.

PR #2045, opened 2026-08-11
DOC: replace use of removed coroutine decorator

tpoliaw:
The run_engine docs still referenced the @asyncio.coroutine decorator
approach to creating async functions used before the introduction of
async/await. This was deprecated in 3.7 and removed in 3.11.


jacopoabramo:
Thanks. At some point the missing sections of this part of the documentation should be updated too but that's for another day.

PR #2048, opened 2026-08-19
Replace assertion statements with explicit check + raise

jwlodek:
Assert statements are removed when running with a `-O` flag, and so should not be used in production code except for code quality checks / type narrowing. Replaced all instances outside of tests with explicit check + raise. Also adjusted some type hints, and fixed any tests that had watched for an AssertionError

- **Replace assert statements with explicit checks + raises in non test code. Update type hints in several places**
- **Fix some typing issues**
- **Apply ruff formatter**


jacopoabramo:
Thank you, big +1 for this on my side. I wanted to do this myself.

PR #2050, opened 2026-08-27
MNT: split RunEngine into PlanSession, PlanExecutor and RunEngine

coretl:
## Description

**Draft. Stacked on #2052 (which is stacked on #1806, which is stacked on #1923), so review those first; the diff here shows only the new work.**

Splits `RunEngine` into three objects, so that the plan-execution machinery can be used without a `RunEngine` at all — headless, entirely within an asyncio event loop, as needed by blueapi and queueserver.

| | Owns | Needs a `RunEngine`? |
|---|---|---|
| `PlanSession` | Metadata, document routing, suspenders, hooks, state machine, `RunBundler` factory, preprocessors, registered commands | No |
| `PlanExecutor` | One plan's execution: the run loop, message/plan stacks, per-plan caches, the live command vocabulary | No |
| `RunEngine` | Everything that exists *because a human is typing at a terminal on a main thread* | Is one |

After the split `RunEngine` keeps only `_loop`, `_th`, `_blocking_event`, `_task_fut`, `_during_task`, `context_managers` (SigintHandler), `max_depth`, `pause_msg`, `call_returns_result`, and the two children.

The headless interface, which works today on this branch:

```python
session = PlanSession(md={"beamline": "i22"})
session.subscribe(my_callback)

executor = session.new_executor()
result = await executor.run(plan)      # plain awaitable; no threads, no main thread
await executor.request_pause()         # control, from any coroutine on the same loop
```

`PlanExecutor.run` is the plain `async def` that #1654 asked for. One session may be used by many executors in turn, which is the shape #1652 needs.

### Design rules

1. **`PlanSession` and `PlanExecutor` own no locks and never block a thread.** Neither class holds a single threading primitive; every crossing of the thread boundary lives in `RunEngine`. Guarded by a test over both classes.
2. **`PlanExecutor`'s lifetime is one `__call__`** — constructed at the start of a plan, kept afterwards so a paused plan can be resumed and a finished one inspected, replaced when the next plan starts. Construction *is* cache clearing, so `_clear_call_cache` and `_clear_run_cache` are gone rather than maintained.
3. **The parent knows the child; the child does not write to the parent.** `PlanSession.new_executor()` builds an executor and adopts it. An executor reads its session — metadata, hooks, loop, document routing — but never assigns to it.
4. **Back-compatibility is by forwarding, but only for names something reaches for.** 24 private `RunEngine` attributes forward to the executor and the rest to the session. The rule covers what callers actually use, not every private name: `_state` and `_suspenders` are deleted rather than forwarded, because `state` and `suspenders` are the surface anyone was promised and keeping a private alias whose meaning had diverged is a trap. No existing test's call sites were rewritten.

### Public and private surface

Applied consistently across both new classes:

- Anything blueapi would reach for, or that `RunEngine` legitimately should, is **public**: `run`, `result`, `prepare_resume`, `request_pause`, `abort`, `stop`, `halt`, `request_suspend`, `state`, `resumable`, `rewindable_flag`, `run_start_uids`, `exit_status`, `exception`, `interrupted`, `reason`, `session`, `command_registry`, `emit` and `dispatcher` (a `RunBundler` is constructed with the first and a plan's subscriptions live in the second).
- Anything `RunEngine` reaches for that we would rather it did not is **private with a comment** saying so.
- Everything else is private: the per-plan caches (`_msg_cache`, `_plan_stack`, `_groups`, `_status_objs`, …) and the session's internal bookkeeping (`_suspenders`, `_registered_commands`, `_require_stream_declaration`, …).

`PlanSession` gains `register_command` / `unregister_command`, because those previously existed only on `RunEngine` and a headless session had no supported way to register one.

### `panicked` moves to the `RunEngine`, and the lock goes with it

`PlanSession` used to hold a `threading.RLock` for the state machine, because `_resume_task` wrote `self._state = "panicked"` from the main thread — the only main-thread state write in the codebase. `LoggingPropertyMachine.__get__` took that lock on *every read*, and the run loop reads the state several times per message.

That write could not simply be marshalled onto the loop, which is what made the lock look load-bearing. It executes precisely when we have concluded the loop thread is not responding: `PyThreadState_SetAsyncExc` failed, or it succeeded and the thread did not finish within a second. Handing the write to that loop would queue a callback on the very thing we have just declared wedged.

So it moves instead. `panicked` was the only state in the session's machine describing the `RunEngine` rather than the plan, and the whole panic mechanism is unimplementable without a second thread — signals only reach the main thread, and `SetAsyncExc` needs another thread's id to target. It is now `RunEngine._is_panicked`, a write-once `bool`; `RunEngine.state` reports `"panicked"` when it is set. Panic is unrecoverable, so a latch needs no ordering guarantees and no lock.

Deliberately unchanged:

- `States.PANICKED` and its transition table stay, so `test_panic_trap`'s guarantee that nothing transitions *out* of a panic still holds and still has teeth.
- `state_hook` and `state_logger` still fire on the transition, via `announce_state_change`. Same thread, same `("panicked", "running")` arguments as before: an observer driving a status display sees the panic rather than watching the state freeze.
- `PlanSession.__init__` now materialises its machine, because `PropertyMachine.__get__` builds it on first access by inserting into a `WeakKeyDictionary`. Doing that while only one thread holds a reference means no later cross-thread read can race the insertion — something the lock used to cover incidentally.

`PlanSession` now holds no threading primitives at all, and fires none of the user's hooks either: it holds `msg_hook`, `state_hook`, `waiting_hook` and `on_pause`, and the executor calls them when it has something to report.

### Two things run off the event loop, and only one is marshalled

A sync ophyd `Status` fires its done-callback on whichever thread completed it, and a sync ophyd signal fires a monitor callback on the device's own thread. bluesky treats these differently, and did before this PR:

- **Status callbacks are marshalled.** The callback handed to `Status.add_callback` does nothing but `loop.call_soon_threadsafe`, so everything touching executor state runs on the loop.
- **Monitor emission is not.** `RunBundler.monitor` hands `emit_event_for_subscribe` to `obj.subscribe()`, and it calls `emit` straight through to the dispatcher. **Subscribers to a monitored signal are invoked on the device's thread**, not the loop.

This PR changes neither. It is flagged because the split makes the asymmetry visible for the first time, and because it is the one place a `PlanExecutor` method is entered from a foreign thread.

Whether to marshal monitor emission too is left for maintainers, with arguments both ways:

- *For*: a subscriber can currently be re-entered concurrently from the device thread and the loop, and `LiveTable`, `LivePlot` and the tiled writers are not written to be thread-safe. Marshalling would give one total order over the document stream instead of two interleaving producers.
- *Against*: the device thread currently absorbs slow callbacks. On the loop, a slow monitor callback stalls plan execution instead, and monitor documents queue behind a busy plan rather than going out promptly.

Because it changes which thread every monitor subscriber runs on, it would want a guard — an `__init__` argument, or an environment variable for rollout without a code change — rather than a flag day.

### What forwards, and what uses it

The private `RunEngine` attributes that forward to the executor are split in two, so that the eighteen with no known caller can be dropped in one edit by anyone who would rather not carry them. Searched across bluesky, its tests and docs, bluesky-queueserver and blueapi:

| forward | used by |
|---|---|
| `_task` | `test_run_engine.py:1551,1569` |
| `_run_bundlers` | `test_run_engine.py:147,149` |
| `_run_start_uids` | `test_plan_executor.py` |
| `_seen_wait_and_move_on_keys` | `test_flyer.py:100` |
| `_command_registry` | `test_run_engine.py:2315,2324` |
| `_deferred_pause_requested` | bluesky-queueserver `worker.py:236` |

The eighteen with none: `_run_permit`, `_pardon_failures`, `_plan`, `_plan_stack`, `_response_stack`, `_msg_cache`, `_rewindable_flag`, `_metadata_per_call`, `_run_tracing_spans`, `_staged`, `_objs_seen`, `_movable_objs_touched`, `_groups`, `_status_objs`, `_exception`, `_interrupted`, `_exit_status`, `_reason`.

queueserver's use of `_deferred_pause_requested` is already only an `AttributeError` fallback behind the public property, with a TODO to remove it.

**That search was not proof.** It missed `RunEngine._request_pause_coro`, which bluesky-queueserver calls deliberately — the forwarding table covers attributes, and that is a method. 51 private methods left `RunEngine` in this branch; that was the only one with a caller anywhere, and it keeps a home on the `RunEngine` because there is no public non-blocking pause to send it to.

### Relationship to #1998

Supersedes #1998, harvesting its `_UNCACHEABLE_COMMANDS` hoist and its typing fixes, with attribution. That PR extracts the run loop but leaves the state behind in three shared cache dataclasses that both classes mutate, plus a callback bag of `RunEngine` bound methods — which makes its `SingleRunExecutor` impossible to construct without a `RunEngine`, and so unable to serve the blueapi use case. It also rewrites test call sites rather than preserving them.

This design also answers three of the four open questions in that PR's description: `task` belongs to the executor and `task_fut` to the engine (the thread boundary decides it, not taste); one run permit per executor; and no set-broadcasting for pause, since a plan names its own executor.

Addresses #1654. Paves the way for #1652.

## Commits

- [x] 1. `chore: drop dead pre-3.10 compatibility shims`
- [x] 2. `refactor: hoist NO_PLAN_RETURN and UNCACHEABLE_COMMANDS to module level`
- [x] 3. `refactor: extract PlanSession for state that outlives a plan`
- [x] 4. `refactor: move plan state into PlanExecutor with a forwarding shim`
- [x] 5. `refactor: take the interrupt result before resuming the main thread`
- [x] 6. `refactor: move the run loop and the vocabulary to PlanExecutor`
- [x] 7. `feat: let PlanExecutor.run take the plan to execute`
- [x] 8. `test: cover the seam between session, executor and engine`
- [x] 9. `refactor: move PlanSession and PlanExecutor to their own module`
- [x] 10. `refactor: announce state changes through one helper`
- [x] 11. `refactor: own panicked on the RunEngine and drop the session's lock`
- [x] 12. `refactor: split the executor forwarding table by whether anything calls it`
- [x] 13. `refactor: collapse RunBundler's emit pair into one sync call`
- [x] 14. `refactor: give each plan its own dispatcher`
- [x] 15. `refactor: delete RunEngine._state, leaving state as the public surface`
- [x] 16. `refactor: fire the user's hooks from the executor that has something to report`
- [x] 17. `refactor: build the session's executor eagerly`
- [x] 18. `refactor: untangle the suspend path`
- [x] 19. `test: pin the pause contract bluesky-queueserver depends on`
- [x] 20. `docs: note that NO_PLAN_RETURN should become a PEP 661 sentinel`
- [x] 21. `refactor: give the plan to run() rather than loading it first`

The `CallbackRegistry` fix that used to lead this branch is now #2052, immediately below it in the stack. `test_CallbackRegistry_1` was quietly dependent on when the cyclic collector last ran, because `_BoundMethodProxy` weakly references `cb.__self__` and for a **class method `__self__` is the class, not the instance** — so the registry held a weak reference to the class and unsubscribed the callback when the class was collected. Fixing it in the library rather than in the test made the workaround this branch used to carry unnecessary, so that commit is gone.

Commit 5 is a semantic change lifted out of commit 6 so that it is reviewable on its own rather than buried in two thousand lines of moved code.

Commits 1-11 are the split. Commits 12-21 are refinements from review: a smaller forwarded surface, one emit instead of two, a dispatcher per plan, hooks fired by the object with something to report, the suspend path collapsed from five layers to three, a note recording that `NO_PLAN_RETURN` wants to be a PEP 661 sentinel once mypy narrows on one, and `load_plan` folded into `run`.

Two planned commits are merged into 6: the run loop calls the halt coroutine when an external library raises `KeyboardInterrupt` at it, and the interrupt coroutines drive the loop through the state machine and the run permit, so moving them separately would have needed scaffolding that existed for one commit.

### Behaviour changes, all deliberate

- `__interrupter_helper` returned `task.result()` *after* `_resume_task()` had run. Resuming executes the plan's cleanup, which opens and closes runs of its own, so the result described the cleanup as much as the plan. It is now taken before resuming, with `task.result()` called first so an exception raised inside the interrupt coroutine still surfaces. Commit 6.
- `Msg('RE_class')` would have started reporting `PlanExecutor`. The session records the `RunEngine`'s class, so it answers as before, and answers for the executor when there is no `RunEngine`.
- The `"The object ... is done with status ..."` debug line is emitted on the event loop rather than on the device's callback thread. Same content.
- `RunEngine.resume()` made one threadsafe hop per `Pausable` device; it now makes one.
- `_status_object_completed` no longer takes the state lock. It is reached from a device's callback thread, and taking a lock there was the one place the executor could block a thread that was not its own.
- `panicked` is now reported by `RunEngine.state` from its own flag rather than by the session's state machine; see above. `RunEngine._state` is gone, since `state` is the surface callers were promised and the two had come to mean different things.
- `RE.unsubscribe` no longer removes a token issued by `Msg('subscribe')`. Plan subscriptions live in the plan's own dispatcher now. `RunEngine.unsubscribe` has always documented its argument as an ID issued by `RunEngine.subscribe`, so this matches the documented contract.
- `RunBundler.__init__` takes one `emit` rather than `emit` and `emit_sync`. It is public and overridable via `RunEngine.RunBundler`, so this is a break for anyone who subclassed it; nothing in bluesky, blueapi or bluesky-queueserver does.
- A `bluesky.suspenders.SuspenderBase` is installed against the `PlanSession`, not the `RunEngine` — as it has been since `PlanSession` was extracted, silently, because the attributes suspenders touch exist on both. It is now named: `.session` is the attribute, `.RE` remains as an alias.
- Every state-change log record carried the `LoggingPropertyMachine` descriptor in its `RE` tag rather than the object whose state changed, because `self` in a descriptor's `__set__` is the descriptor. Now fixed. Commit 11.
- `_interruptions_desc_uid` and `_interruptions_counter` are gone from the `RunEngine`. They were vestigial: the live ones belong to `RunBundler` and nothing read the `RunEngine`'s copies.

## Follow-ups

- [ ] A docs page: "How to use the PlanExecutor without a RunEngine in an asyncio application".

## How Has This Been Tested?

Full suite on **every commit**, not just at the tip, plus the base branch as a control, plus `mypy` and `ruff`.

Run in a clean virtualenv (`pip install -e ".[dev]"`) so that `ophyd-async` comes from PyPI rather than a local checkout, and in a separate git worktree so nothing else could disturb the tree under test.

**All twenty-one trees produce a failure set identical to `suspender-signature`'s** — the same three failures and the same twenty-two errors, none of them touching code this PR changes:

- `test_buffering::test_callback_logging_exceptions` (timing-dependent; it passes on some runs), `test_tiled_writer::test_imports_raise_warnings`, and `test_zmq::test_proxy_script`.
- 22 errors across `test_callbacks`, `test_plans::test_relative_pseudo`, `test_run_engine::test_filled` / `test_num_events`, and `test_vertical_integration`, all from the databroker/tiled stack in this dependency resolve.

Passing counts go 1811 on the base branch to 1828 at the tip, the difference being the seventeen tests added here.

`test_streams.py` is excluded: `streamz` fails to import against the `dask` version resolved here, and pytest aborts the whole session on a collection error. It fails identically on the base branch.

`mypy` reports no issues in `run_engine.py`, `plan_executor.py`, `bundlers.py` or `suspenders.py`.

Note for anyone reproducing locally: the SIGINT tests leak a signal that kills the pytest process partway through, so most runs abort early and report only the tests reached — three consecutive runs here gave 13, 2 and 2. That is pre-existing and behaves the same on the unmodified base branch. Run with `-k "not sigint"`, and the SIGINT tests separately and more than once, to get a usable signal.


src/bluesky/run_engine.py
```diff
@@ -72,6 +72,38 @@
 _SPAN_NAME_PREFIX = "Bluesky RunEngine"
 
 
+class _NoPlanReturn:
+    """The type of :data:`NO_PLAN_RETURN`."""
+
+    def __repr__(self) -> str:
+        return "NO_PLAN_RETURN"
```
coretl: I wonder if these actually belong in plan_executor.py

src/bluesky/run_engine.py
```diff
@@ -229,6 +229,26 @@ def default_scan_id_source(md: RunEngineMetadata) -> SyncOrAsync[int]:
...
+    back on the loop, where ``call_soon_threadsafe`` remains correct.
+
+    Any arguments the device passes are dropped: the handler is expected to
+    have closed over what it needs.
+    """
+
+    def callback(*args: typing.Any, **kwargs: typing.Any) -> None:
+        loop.call_soon_threadsafe(handler)
+
+    return callback
```
coretl: Wondering about whether we actually need this, wait for its usage in plan_executor to tell

src/bluesky/run_engine.py
```diff
@@ -258,6 +259,212 @@ def inner(self, *args, **kwargs):
...
+    dispatcher
+        The `Dispatcher` documents are emitted through.
+
+    suspenders
+        Read-only collection of `bluesky.suspenders.SuspenderBase` objects.
+    """
+
+    #: The bundler used to compose documents for each open run. Overridable
+    #: by subclasses, and by `RunEngine` subclasses that override their own.
+    RunBundler = RunBundler
```
coretl: Prefer to pass in to `__init__` rather than classvar, and spell as `run_bundler_cls: type[RunBundler] = RunBundler`. ditto with `run_engine_cls: type | None = None`

src/bluesky/run_engine.py
```diff
@@ -592,19 +773,133 @@ def __init__(
...
+        """Compose the live vocabulary from the built-ins and the session.
+
+        The session records what the user has registered and unregistered, so
+        that those survive being composed over a different set of built-ins.
+        """
+        registry = dict(self._default_command_registry)
+        registry.update(self._session.registered_commands)
+        for name in self._session.unregistered_commands:
+            registry.pop(name, None)
+        self._command_registry = registry
```
coretl: See how this looks when settled

src/bluesky/run_engine.py
```diff
@@ -1034,11 +1034,39 @@ def rewindable(self):
 
     @rewindable.setter
     def rewindable(self, v):
+        # Setting this can discard the message cache and reset every open
+        # bundler's checkpoint state, which belongs to the plan in progress.
+        # Do it on the loop, so it cannot race a plan that is paused rather
+        # than idle. The Msg('rewindable', ...) route is already on the loop.
+        self._on_loop(functools.partial(self._set_rewindable, bool(v)))
```
coretl: Is this fixing a bug that exists in the RE as currently written?

src/bluesky/run_engine.py
```diff
@@ -486,6 +491,102 @@ def notify_paused(self):
...
+        self.objs_seen: set[typing.Any] = set()  # every object seen in a Msg
+        self.movable_objs_touched: set[typing.Any] = set()  # everything we 'set'
+        self.groups: defaultdict[str, set[Callable[[], asyncio.Future]]] = defaultdict(set)
+        self.status_objs: defaultdict[typing.Any, set[typing.Any]] = defaultdict(set)
+        self.seen_wait_and_move_on_keys: set[typing.Any] = set()
+
+        self.exception: BaseException | None = None  # raised in the run loop
+        self.interrupted: bool = False  # paused, aborted or failed
+        self.exit_status: str = "success"  # optimistic default
+        self.reason: str = ""  # reason for an abort
```
coretl: In general we should follow this rule:
- anything blueapi would access from the PlanExectutor and PlanSession should be public
- anything the RunEngine *should access* from the PlanExecutor and PlanSession should be public
- anything that the RunEngine accesses from the PlanExectuor and PlanSession, but we would like to deprecate its access as there is a better way should be private with a comment
- anything else should be private

src/bluesky/run_engine.py
coretl: I assume all of [e2cf51f](https://github.com/bluesky/bluesky/commit/e2cf51f2843c52602c389af979ab9b0311728743) is mechanical so I did not read most of it. Draw my attention to anything I need to look at

docs/api_changes.rst
coretl: This needs a page of docs entitled "How to use the PlanExecutor without a RunEngine in an asyncio application". Don't add it yet, but don't forget it

jacopoabramo:
@coretl might be useful to bake in also a refactor of the state machine, probably as a future PR. I made an initial draft in #1992 

If this gets merged I'll close the previous refactor and work on that, unless you want to pick that up as well

src/bluesky/tests/test_run_engine.py
```diff
@@ -69,11 +69,29 @@ def test_states():
...
-    RE._state = "panicked"
+    RE._session._state = "panicked"
     for k in RunEngineStateMachine.States.states():
         if k != "panicked":
             with pytest.raises(TransitionError):
-                RE._state = k
+                RE._session._state = k
+
+
+def test_panicked_state_answers_like_any_other_state(RE):
```
jacopoabramo, line 79: is this test really necessary? are we testing that somebody access a private variable?

src/bluesky/tests/test_utils.py
```diff
@@ -139,6 +140,25 @@ def test_is_movable(hw):
         )
 
 
+@pytest.fixture
+def no_cyclic_gc():
```
jacopoabramo: I understand the text but is this something that effectively happens in real applications? Is it just a way to prevent some existing problems in testing, such as flaky tests in the CI?

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2450 @@
...
+    """The type of :data:`NO_PLAN_RETURN`."""
+
+    def __repr__(self) -> str:
+        return "NO_PLAN_RETURN"
+
+
+#: Returned in place of a plan's return value when the plan did not run to
+#: completion, and so never returned one. Distinguishable from a plan that
+#: completed and returned ``None``.
+NO_PLAN_RETURN = _NoPlanReturn()
```
jacopoabramo, line 117: Could you please add a comment stating that this should become a sentinel object in the future? mypy doesn't yet support them but they're already part of typing_extensions

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2450 @@
...
+    the loop, so that the message reaches the user when they asked rather
+    than whenever the loop gets to it. The ways in are disjoint -- a
+    `RunEngine` goes straight to its executor rather than by way of its
+    session -- so this is said once per suspension.
+    """
+    print("Suspending....To get prompt hit Ctrl-C twice to pause.")
+    print(f"Suspension occurred at {datetime.now().strftime('%Y-%m-%d %H:%M:%S')}.")
+
+
+def _panicked_state() -> ProxyString:
```
jacopoabramo, line 255: I don't like this very much. I understand the reasoning but it just strengthens the fact that the current state machine is a bit hacky.

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2450 @@
...
+        The session records what has been registered and unregistered, so that
+        those survive being composed over a different executor's built-ins.
+        """
+        registry = self._default_commands()
+        registry.update(self._session._registered_commands)
+        for name in self._session._unregistered_commands:
+            registry.pop(name, None)
+        self.command_registry = registry
+
+    def load_plan(self, plan, *, metadata=None, prologue=None):
```
jacopoabramo: I don't mind having this as part of the public API, I'm just curious why

src/bluesky/tests/test_plan_executor.py
```diff
@@ -0,0 +1,210 @@
...
+    threading.Lock().__class__,
+    threading.RLock().__class__,
+    threading.Condition,
+    threading.Semaphore,
+    threading.Barrier,
+    threading.Thread,
+)
+
+
+def test_executor_holds_no_threading_primitives():
```
jacopoabramo: How can there be a situation that leaks threading primitives into a PlanSession?

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2450 @@
...
+            else:
+                self._session.log.debug("No 'stop' method available on %r", obj)
+
+    def _destroy_open_run_tracing_spans(self):
+        while len(self._run_tracing_spans):
+            _span = self._run_tracing_spans.pop()
+            _span.set_attribute("exit_status", "aborted")
+            _span.end()
+
+    async def run(self, plan=None, *, metadata=None):
```
jacopoabramo: This is just for later but: instead of being an async function could it return an awaitable object? Like a task? Or is it too breaking?

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2450 @@
+"""The environment plans run in, and the execution of a single plan.
+
+See :class:`PlanSession` and :class:`PlanExecutor`. A `bluesky.run_engine.RunEngine`
+composes the two and drives them from a terminal on the main thread.
+"""
+
```
jacopoabramo, line 6: I'm wondering if we could start using `from __future__ import annotations`. I refrained from doing so but it we don't depend on annotations except for IDE interaction and we don't do fancy annotation manipulation. We could then move a bunch of imports under `if TYPE_CHECKING ...`

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2450 @@
...
+    "WaitForTimeoutError",
+    "autoawait_in_bluesky_event_loop",
+    "call_in_bluesky_event_loop",
+    "default_scan_id_source",
+    "get_bluesky_event_loop",
+    "in_bluesky_event_loop",
+    "set_bluesky_event_loop",
+]
+
+_SPAN_NAME_PREFIX = "Bluesky RunEngine"
```
jacopoabramo, line 97: This is still hardcoded as the RunEngine but technically a PlanExecutor could be used on its own, wouldn't it?

jacopoabramo (commented):
Made a first review pass. Looks good. Just some nits mostly. Would appreciate another set of eyes.

src/bluesky/tests/test_utils.py
coretl: yeah, moving things around changed the timing of the test that uses this, so had to disable gc to get that test to run. might dig into that test to see if it can be done in a different way

src/bluesky/plan_executor.py
coretl: The reasoning is that I wanted to do without a state lock. That means that only the `PlanExecutor` can mutate state as it operates within the event loop. But `panicked` is owned by the `RunEngine` because the event loop is unresponsive, so it needs its own override.

src/bluesky/tests/test_plan_executor.py
coretl: The LLM put this in as I stated as a requirement that `PlanSession` and `PlanExecutor` be pure asyncio, all operating from within the event loop. There are a couple of exceptions (ophyd v1 objects callback on a their own thread so need a trampoline), but this should hold in general

src/bluesky/plan_executor.py
coretl: I prefer coro, as it's trivial to wrap in a task yourself

coretl:
> @coretl might be useful to bake in also a refactor of the state machine, probably as a future PR. I made an initial draft in #1992
> 
> If this gets merged I'll close the previous refactor and work on that, unless you want to pick that up as well

Thanks, I think that probably is best in a PR after this one, so if you're happy to rebase over this when it settles then that would be great

src/bluesky/tests/test_utils.py
coretl: Will be fixed in #2052

src/bluesky/tests/test_run_engine.py
coretl: Partially. Because the `PlanExecutor` holds `state`, but the `RunEngine` trumps it with `_is_panicked`, this is the cheap way to test this works without having to make the event loop be unresponsive. `test_sigint_many_hits_panic` is the real way to test it, but that's skipped on windows and python < 3.12, so this test still adds something.

src/bluesky/plan_executor.py
coretl: Added

src/bluesky/plan_executor.py
tpoliaw: `from __future__ import annotations` [is now deprecated](https://peps.python.org/pep-0749/#the-future-of-from-future-import-annotations) and annotations are now (3.14+) always lazily evaluated. Until 3.14 is required, how many imports would be saved? If a type is passed/returned at runtime, wouldn't it need to be imported anyway?

jacopoabramo: Fair point

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2456 @@
...
+    repr. Replace both with a ``typing_extensions.Sentinel`` (PEP 661), which
+    ships today but which mypy does not yet narrow on; when it does, the
+    sentinel needs no companion class and this one goes away.
+    """
+
+    def __repr__(self) -> str:
+        return "NO_PLAN_RETURN"
+
+
+#: Returned in place of a plan's return value when the plan did not run to
```
tpoliaw: What is the `#:` syntax for in these comments?

src/bluesky/plan_executor.py
coretl: This was because `RunEngine` used to put a prologue in of `wait` messages, so the LLM split it into a separate function. Looking closer, the `run` method could just take the prologue itself. Changes made to delete this function.

src/bluesky/plan_executor.py
coretl: Correct, but as this PR is trying to maintain backwards compatibility then I think this should come in a separate issue

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2456 @@
...
+        "_start_suspender",
+    }
+)
+
+
+class WaitForTimeoutError(TimeoutError): ...
+
+
+@dataclass
+class RunEngineResult:
```
tpoliaw: This should be a `PlanResult` or similar now - there is no RunEngine involved at this point. It would allow the otherwise breaking changes below.

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2456 @@
...
+        A text description of the reason why the plan was aborted (if aborted)
+    exception :
+        The exception generated by the plan, if any
+    """
+
+    run_start_uids: tuple[str, ...]
+    plan_result: typing.Any
+    exit_status: str
+    interrupted: bool
+    reason: str
```
tpoliaw: While everything is up in the air, it would be good to improve the shape of these fields. At the moment there are four (five if the exception is ever used) that can't all be present at the same time. Would it make sense to make the `PlanResult` be something like

```python
class Complete:
    result: Any

class Aborted:
    reason: str | None

class Failed:
    exception: Exception

class PlanResult:
    run_ids: tuple[str, ...]
    outcome: Complete | Aborted | Failed
```

Again, I'm not sure when the Failed/exception case would ever be created. As far as I can make out, if the plan fails it throws the exception instead of returning it. It would be useful if the exception *could* be returned. The run engine could then raise it in the REPL case and headless runners could report errors without them having to be used as flow control when handling failures.

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2456 @@
...
+        ]
+
+
+def announce_state_change(obj, old_value, value) -> None:
+    """Log a state change, and tell ``obj``'s state hook about it.
+
+    Split out of :meth:`LoggingPropertyMachine.__set__` so that a state change
+    which does not go through the machine can be announced too, and announced
+    identically: there is one definition of what an observer is told rather
+    than two to drift apart.
```
tpoliaw: This is a commit message not a docstring

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2456 @@
...
+    ):
+        if loop is None:
+            loop = _default_event_loop()
+        self._loop = loop
+
+        self.log = log if log is not None else ComposableLogAdapter(logger, {"RE": self})
+
+        if md is None:
+            md = {}
+        self.md = md
```
tpoliaw:
```suggestion
        self.md = md or {}
```

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2456 @@
...
+        log: LoggerAdapter | None = None,
+        on_pause: typing.Callable[[], None] | None = None,
+        run_bundler_cls: type[RunBundler] = RunBundler,
+        run_engine_cls: type | None = None,
+    ):
+        if loop is None:
+            loop = _default_event_loop()
+        self._loop = loop
+
+        self.log = log if log is not None else ComposableLogAdapter(logger, {"RE": self})
```
tpoliaw:
```suggestion
        self.log = log or ComposableLogAdapter(logger, {"RE": self})
```

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2456 @@
...
+        md_normalizer: typing.Callable | None = None,
+        scan_id_source: typing.Callable[[RunEngineMetadata], SyncOrAsync[int]] = default_scan_id_source,
+        log: LoggerAdapter | None = None,
+        on_pause: typing.Callable[[], None] | None = None,
+        run_bundler_cls: type[RunBundler] = RunBundler,
+        run_engine_cls: type | None = None,
+    ):
+        if loop is None:
+            loop = _default_event_loop()
+        self._loop = loop
```
tpoliaw:
```suggestion
        self._loop = loop or _default_event_loop()
```

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2456 @@
...
+            self.md["versions"]["ophyd_async"] = ophyd_async.__version__
+        except ImportError:
+            self.log.debug("Failed to import ophyd_async.")
+
+        from ._version import __version__
+
+        self.md["versions"]["bluesky"] = __version__
+        self.md["versions"]["event_model"] = event_model.__version__
+
+        self.preprocessors = preprocessors if preprocessors is not None else []
```
tpoliaw: are these ever going to be different to checking truthiness directly? Do we need callers to be able to keep their own references to preprocessors?

```suggestion
        self.preprocessors = preprocessors or []
```

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2456 @@
...
+            import ophyd_async
+
+            self.md["versions"]["ophyd_async"] = ophyd_async.__version__
+        except ImportError:
+            self.log.debug("Failed to import ophyd_async.")
+
+        from ._version import __version__
+
+        self.md["versions"]["bluesky"] = __version__
+        self.md["versions"]["event_model"] = event_model.__version__
```
tpoliaw: This feels like it should be the `RunEngine`'s responsibility

src/bluesky/plan_executor.py
```diff
@@ -0,0 +1,2456 @@
...
+        """Announce that this plan has reached a paused resting state."""
+        if self._session.on_pause is not None:
+            self._session.on_pause()
+
+    def emit(self, name, doc) -> None:
+        """Give a document to the session's subscribers, then to this plan's.
+
+        The session goes first so that subscriptions outliving the plan see a
+        document before the ones that arrived with it, which is the order a
+        single shared registry gave by construction.
```
tpoliaw: History doesn't make sense in a docstring

tpoliaw:
A couple of higher level questions before I get distracted looking at the diff too closely.

Why does the executor need to know about the session?
When the executor is created it can be passed all the information it needs. Is there really a case for adding/removing hooks/commands *while a plan is running*?

---
Why does the user have to create the executor and then pass the plan coroutine to it as two steps?
This leaves an executor that can be reused multiple times which is either going to cause problems or suggests that there should be greater separation between the executor and session.
eg
```python 
exec = session.new_executor()
await exec.run(my_plan())
await exec.run(my_plan())
```

It would be clearer if the executor was handled by the session leaving the user to only deal with the plan. Something like

```python
plan_exec = session.run(my_plan()) # or session(my_plan()) to mimic the previous run engine interface?
await plan_exec.run() # or await plan_exec(), or even just await plan_exec?
```

---

There's currently a single reference to the executor but as far as I can see, no protection over overwriting it

```python
asyncio.create_task(session.new_executor().run(plan_one()))
asyncio.create_task(session.new_executor().run(plan_two()))
```

The second plan will fail with a `TransitionError` but the session will have lost its reference to `plan_one`'s executor and suspenders etc will stop working with it

---

If there is no per-plan state held in the session or engine, is there any reason why, at a code level, two plans can't run at once? If not, could the prevention logic be moved into the run engine and headless runners handle their own logic if it is required?

---

Lastly, and possibly off-topic but relevant for this PR, does bluesky have any kind of guidelines for LLM use and if not is it worth adding some going forward? Using generative tools for the code is one thing but having the wall of noise output as a PR message makes it more difficult to wade through what would already be a huge (5000+ line) review. [For instance](https://matplotlib.org/devdocs/devel/contribute.html#use-of-generative-ai) [other](https://devguide.python.org/getting-started/ai-tools/#considerations-for-success) [projects](https://forge.rust-lang.org/policies/llm-usage.html#-banned) [require](https://tiangolo.com/open-source/contributing/#closing-automated-and-ai-prs) [human](https://github.com/pydantic/pydantic?tab=contributing-ov-file#ai-policy) [written](https://github.com/astral-sh/.github/blob/main/AI_POLICY.md) [comments](https://numpy.org/doc/stable/dev/ai_policy.html#responsibility)/[descriptions](https://pandas.pydata.org/docs/dev/development/contributing.html#automated-contributions-policy)

coretl:
> Lastly, and possibly off-topic but relevant for this PR, does bluesky have any kind of guidelines for LLM use and if not is it worth adding some going forward? Using generative tools for the code is one thing but having the wall of noise output as a PR message makes it more difficult to wade through what would already be a huge (5000+ line) review. [For instance](https://matplotlib.org/devdocs/devel/contribute.html#use-of-generative-ai) [other](https://devguide.python.org/getting-started/ai-tools/#considerations-for-success) [projects](https://forge.rust-lang.org/policies/llm-usage.html#-banned) [require](https://tiangolo.com/open-source/contributing/#closing-automated-and-ai-prs) [human](https://github.com/pydantic/pydantic?tab=contributing-ov-file#ai-policy) [written](https://github.com/astral-sh/.github/blob/main/AI_POLICY.md) [comments](https://numpy.org/doc/stable/dev/ai_policy.html#responsibility)/[descriptions](https://pandas.pydata.org/docs/dev/development/contributing.html#automated-contributions-policy)

I'll comment more on this tomorrow, but I think this experiment has taught me 2 things:
- LLMs can do lots of things on their own, but on a repo I'm less familiar with it has variable results. E.g. here it has faithfully turned the `__call__` method into `run()`, and it's taken me 2 days to come to the same conclusion you've come to about someone calling `run()` twice because I spent that time focussing on the thread boundary issues...
- People read draft PRs more than I thought. Please don't read the diff yet, I put this up so @jacopoabramo and @tacaswell had advance notice of what was coming. That was a mistake, I should have put that up on the PR description rather than let the LLM write it. I'm rebasing this into scoped commits that are more reviewable, but I'll deal with the multiple `run()` issue first as that's pretty fundamental.

coretl:
Opened #2056, stacked on this branch: it drops the `prologue` keyword argument from `PlanExecutor.run` by having the wait for an already-tripped suspender go through the same `_start_suspender` machinery a mid-plan suspend uses, instead of the bespoke `wait_for` generator `RunEngine.__call__` used to build and push itself.

---
_Generated by [Claude Code](https://claude.ai/code)_

coretl:
Opened #2058, a draft PR on top of this branch that makes monitor-document emission optionally marshal onto the event loop, gated off by default. It exists so the monitor-marshalling question raised in this PR's "Two things run off the event loop" section can be judged on measurements rather than argument alone -- it includes a self-contained benchmark and the actual numbers for latency, plan throughput, and (the number that decides the "against" case) plan throughput with a deliberately slow monitor subscriber.

@tacaswell it's a decision aid, not a merge candidate -- see the PR body for the numbers and what broke when the guard was forced on for the whole test suite.

---
_Generated by [Claude Code](https://claude.ai/code)_

coretl:
Apologies for the AI spam. I'm going to close this off, start fresh, and bring it back here when it's ready

PR #2052, opened 2026-09-01
FIX: hold class method callbacks strongly in CallbackRegistry

### LINKED ISSUE #2051: CallbackRegistry: class method callbacks are weakly held by their class, and callbacks on non-weak-referenceable instances are broken
`_BoundMethodProxy.__init__` assumes `cb.__self__` is always a weak-referenceable instance. It is not, and two distinct bugs follow.

Both are in `bluesky/utils/__init__.py`:

```python
try:
    # This branch is successful if 'cb' bound method and class method,
    #   but destroy_callback mechanism works only for bound methods,
    #   since cb.__self__ points to class instance only for
    #   bound methods, not for class methods. Therefore destroy_callback
    #   will not be called for class methods.
    try:
        self.inst = ref(cb.__self__, self._destroy)
    except TypeError:
        self.inst = None
    self.func = cb.__func__
    self.klass = cb.__self__.__class__
```

## 1. A class method callback is weakly held by its class

For a class method, `cb.__self__` is **the class**, not an instance. A heap type is
weak-referenceable, so `ref()` succeeds and the registry ends up holding a weak reference
to the class, with a destroy callback that unsubscribes.

The comment above says this cannot happen — that the destroy callback "will not be called
for class methods". That is wrong. It is true in normal use only by accident, because
callback classes live in a module namespace and so are never collected. For a dynamically
created class, the callback disappears the moment the class is collected:

```python
import gc
from bluesky.utils import CallbackRegistry

def make():
    class C:
        @classmethod
        def cb(cls, *args, **kwargs): ...
    return C()

reg = CallbackRegistry()
inst = make()
reg.connect("sig", inst.cb)

print(len(reg.callbacks["sig"]))   # 1
inst = None
gc.collect()
print(len(reg.callbacks.get("sig", {})))   # 0  -- silently unsubscribed
```

This also contradicts the behaviour `test_CallbackRegistry_1` documents, which groups
class methods with functions and static methods as "strong reference is held internally,
deleting all external references does not influence the callbacks in CB registry".

As a smaller consequence of the same line, `self.klass` is set to
`cb.__self__.__class__`, which for a class method is the **metaclass** (`type`) rather
than the class. `klass` is never read anywhere in bluesky, so nothing depends on it today.

## 2. A callback on a non-weak-referenceable instance is silently broken

If `cb.__self__` cannot be weakly referenced — an instance of a class using `__slots__`
without `__weakref__` — the `except TypeError` sets `self.inst = None` but still stores
`self.func = cb.__func__`, the *unbound* function. `__call__` then takes the
"not a bound method" path and calls that function without its instance:

```python
from bluesky.utils import _BoundMethodProxy

class S:
    __slots__ = ()
    def cb(self, x): return f"got {x}"

_BoundMethodProxy(S().cb)("hello")
# TypeError: S.cb() missing 1 required positional argument: 'x'
```

`connect()` succeeds and the failure only surfaces later, when the signal is processed.

## Suggested fix

Branch on what `__self__` actually is, and hold the bound method strongly in the two cases
where a weak reference is either wrong or impossible:

- class method (`isinstance(cb.__self__, type)`) — strong reference, no destroy callback,
  and set `klass` to the class itself;
- unweak-referenceable instance — strong reference to the bound method, so the callback
  remains callable;
- ordinary bound method — unchanged: weak reference to the instance, with the destroy
  callback that makes auto-unsubscribe work.

Holding a class strongly costs nothing in practice, since classes are effectively immortal
in normal use, and it matches the behaviour that is already documented for group 2
callables.

Found while investigating a test that only passed depending on when the cyclic collector
last ran (bluesky#2050).

coretl:
Fixes #2051.

Two commits:
- Insert a `gc.collect()` in the test to show the bug, and add a couple of explicit tests
- Fix the code to match the tests

This means that we now hold a strong reference to classmethods, but this seems harmless as they are normally strongly held in a module namespace anyway. The alternative would require modifying the test and inserting a double `gc.collect()` so it has a chance to break all the cycles.

### DIFF
--- a/src/bluesky/tests/test_utils.py
+++ b/src/bluesky/tests/test_utils.py
@@ -1,4 +1,5 @@
 import asyncio
+import gc
 import operator
 import time
 import warnings
@@ -377,6 +378,12 @@ def test_CallbackRegistry_1(delete_objects, set_allowed_signals, callable_type):
         # Now delete all the callable objects one by one
         for n in range(len(obj_to_delete)):
             obj_to_delete[n] = None  # Overwriting the reference deletes the object
+            # Collect now, rather than leaving it to whenever the cyclic collector
+            # next runs. Without this the assertions below turn on collection
+            # timing rather than on what the registry holds: the class method
+            # cases passed for years while asserting the opposite of what the
+            # registry did, because the collector happened not to run in between.
+            gc.collect()
 
             # Check the function composition
             if callable_type in [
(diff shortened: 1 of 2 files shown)


tacaswell:
```
Coverage XML written to file cov.xml
=========================== short test summary info ===========================
FAILED src/bluesky/tests/test_new_examples.py::test_descriptor_layout_from_monitor[True] - AssertionError: assert {'rand': 0.1368670131857882} == {'rand': 0.2928482124836502}
  
  Differing items:
  {'rand': 0.1368670131857882} != {'rand': 0.2928482124836502}
  
  Full diff:
    {
  -     'rand': 0.2928482124836502,
  +     'rand': 0.1368670131857882,
    }
====== 1 failed, 1809 passed, 52 skipped, 2 xfailed in 628.77s (0:10:28) ======
--- Logging error ---
  File "C:\hostedtoolcache\windows\Python\3.13.15\x64\Lib\logging\__init__.py", line 1154, in emit
    stream.write(msg + self.terminator)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\hostedtoolcache\windows\Python\3.13.15\x64\Lib\site-packages\colorama\ansitowin32.py", line 47, in write
    self.__convertor.write(text)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "C:\hostedtoolcache\windows\Python\3.13.15\x64\Lib\site-packages\colorama\ansitowin32.py", line 177, in write
    self.write_and_convert(text)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "C:\hostedtoolcache\windows\Python\3.13.15\x64\Lib\site-packages\colorama\ansitowin32.py", line 205, in write_and_convert
    self.write_plain_text(text, cursor, len(text))
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\hostedtoolcache\windows\Python\3.13.15\x64\Lib\site-packages\colorama\ansitowin32.py", line 210, in write_plain_text
    self.wrapped.write(text[start:end])
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
```

This seems to have become a flaky test as I saw it yesterday as well, merging over this and will investigate.

tacaswell (approved):

PR #2057, opened 2026-09-02
ENH: choose the collect verb from what a device produced, not its static type

coretl:
## Description

`RunBundler.collect` picks which collect verb to use from the *static* protocol
membership of the collect object. After the asset docs have been packed, in
`src/bluesky/bundlers.py`:

```python
if len(collect_objects) == 1 and not isinstance(collect_objects[0], WritesStreamAssets):
    ...  # the EventPageCollectable / EventCollectable path
else:
    self._sequence_counters[stream_name] += indices_difference
```

So a device that satisfies **both** `WritesStreamAssets` and `EventPageCollectable`
never has `collect_pages()` called, whatever it actually produced. A paged fly scan
over such a device silently emits no data at all — no warning, no error:

```python
det = StreamAssetsOrPagesCollectable(name="det", produces="event_pages")
docs = DocHolder()
RE(collect_plan(det, pre_declare=True, stream_name="main"), docs.append)

# on main:
#   {'start': 1, 'descriptor': 1, 'stop': 1}
#   stop["num_events"] == {'main': 0}
# the descriptor is emitted, and then every frame is dropped
```

This PR replaces the static gate with one based on what the device produced.
The rule: **a single collect object is asked for Events or EventPages when it is
`EventCollectable` or `EventPageCollectable` and it has not reported stream assets
so far in this run; otherwise the sequence counter is advanced from the asset docs
as before.** `RunBundler` keeps a per-run record (`_reported_data_kinds`) of what
each collect object has reported, and a device that reports both ways in one run
raises a `RuntimeError` naming the device, both kinds, and what to do about it.

Reporting both ways has to be an error rather than something the bundler merges,
because the two paths advance the same counter. The stream-asset path advances
`self._sequence_counters[stream_name]` by the `indices_difference` returned from
`_pack_external_assets`, while the Event/EventPage path advances it through
`compose_event`/`compose_event_page`. A device doing both in one run would have
every frame counted twice, and the `seq_nums` written into its `stream_datum`
documents would no longer line up with the events in the stream.

**What is unchanged.** A device that satisfies only one of the two protocols takes
exactly the path it took before:

- The record is only kept for objects that satisfy `WritesStreamAssets`, since they
  are the only ones that can report two ways. Devices on the older
  `WritesExternalAssets` protocol have no index of their own and are expected to
  report Datums alongside the Events that reference them — the devices in
  `test_seq_num_indices_in_sync.py` that are `EventPageCollectable` *and* emit
  `stream_datum` documents keep taking the page path, with their `seq_nums` driven
  by the pages as before.
- A device that can only write stream assets is never asked for pages, whether or
  not it had anything to flush, because the gate also requires page/event
  collectability. Gating purely on "this collect produced no asset docs" would have
  changed that.
- The old-style doubly nested path (`stream_name is None`, `_describe_collect` has
  been called) takes the same branch as before.
- The `RuntimeError` for a `collect` on a device that is neither `EventCollectable`
  nor `EventPageCollectable` and has no `name=stream_name` has moved to the other
  side of the branch, so it is still raised for exactly the same messages.

Every existing test in `test_external_assets_and_paging.py` and
`test_seq_num_indices_in_sync.py` passes unchanged; no existing expectation was
edited.

## Motivation and Context

`ophyd_async.core.StandardDetector` produces stream assets or event pages depending
on what its data logics make for a given scan, which is not known until `prepare()`.
Because the bundler decides from the static type, ophyd-async currently has to bind
`collect_asset_docs` / `collect_pages` as *instance attributes* on every `prepare()`
and delete the other one (`StandardDetector._publish_collect_methods`; see ADR 0023
on the `detector-stack-4-flyable` branch and bluesky/ophyd-async#1408). With this
change those can be two ordinary methods.

ophyd-async will need a released bluesky containing this change before it can drop
that workaround. Nothing in ophyd-async is changed here.

## How Has This Been Tested?

New tests in `src/bluesky/tests/test_external_assets_and_paging.py`, built on the
existing helpers in that file. `StreamAssetsOrPagesCollectable` is
`Named, EventPageCollectable, WritesStreamAssets` and is told which kind of data to
produce:

- producing only stream assets → `stream_resource`/`stream_datum` emitted, no
  `event_page`, and `seq_nums`, `indices` and `stop.num_events` compared document
  for document against an equivalent `StreamDatumReadableCollectable` run;
- producing only event pages → `event_page` with the right data, `seq_num` `[1, 2]`
  and `stop.num_events == {"main": 2}`;
- reporting both ways in one run → raises, message asserted;
- a `WritesStreamAssets`-only device with nothing to flush;
- a both-protocol device with nothing to flush in its first collect, then stream
  assets in the second;
- a both-protocol device that reported pages, then collected alongside another
  object;
- `backstop_collect()` on a both-protocol device, in both modes.

The eight tests that cover new behaviour were confirmed to fail against the
unmodified `bundlers.py`; the twelve that assert unchanged behaviour pass against it.

Edge cases decided deliberately, and worth a reviewer's eye:

- **A device that reports pages and then stream assets raises; the reverse is not
  detected.** Once a device has reported stream assets it stays on the stream-asset
  path for the rest of the run, so if it then starts producing pages they are not
  collected. Detecting that would mean calling `collect_pages()` on a device already
  known to be streaming, which cannot be distinguished from the legitimate case of a
  streaming device with no new frames to flush.
  `test_both_protocols_reporting_stream_assets_stays_on_the_stream_asset_path`
  pins the current behaviour. Happy to change this if you would rather the bundler
  asked and raised.
- **A device is only recorded as having reported when it actually produced
  something.** A collect that flushes nothing and yields no pages leaves the device
  free to report either way later, which is what makes the "nothing to flush on the
  first collect" case work.
- **Multiple collect objects.** `collect()` can only compose pages for one object,
  and a both-protocol device passes the existing `check_supports(obj,
  WritesStreamAssets)` check, so an object that has reported pages earlier in the run
  now raises when it is collected alongside others rather than having its data
  dropped. The residual gap is an object that would produce pages but has never
  reported anything: there is no way to tell it apart from one with nothing to flush
  without calling `collect_pages()` on every object in a multi-object collect.
- **Nothing else was touched.** `protocols.py` is unchanged, and `collect()` was not
  refactored beyond what the rule needs.

The changelog entry in `docs/api_changes.rst` follows the convention of adding a new
version heading with a `yyyy-mm-dd` placeholder; the version number is a guess for
the release manager to correct.

Validation run locally, as CI runs it:

- `pre-commit run --all-files` — passed.
- `mypy src` — passed, no issues in 87 source files.
- `pytest` — 1855 passed, 24 skipped, 2 xfailed.
- `sphinx-build -EW --keep-going -T docs build/html` — could not be run as CI does,
  because this sandbox's network policy blocks `blueskyproject.io`, which
  `docs/conf.py` fetches at import time. With that one call stubbed out, the build
  completed and the only warnings were intersphinx inventories that the same network
  policy blocked; there were no content warnings. The `docs / build` job on CI here
  passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QcYkcAefJJUpGgjcLTkJb5

## AI Disclaimer

The code and this description were generated with AI and reviewed/tweaked by the author.

### DIFF
--- a/docs/api_changes.rst
+++ b/docs/api_changes.rst
@@ -2,6 +2,18 @@
  Release History
 =================
 
+v1.15.2 (yyyy-mm-dd)
+====================
+
+Changed
+-------
+
+- ``collect()`` now chooses whether to ask a device for Events, EventPages or stream
+  assets based on what that device produced, rather than on the protocols it statically
+  obeys, so a single device may obey both ``WritesStreamAssets`` and
+  ``EventPageCollectable`` and decide which of them it uses when it is prepared.
+  Reporting both ways in the same run raises a ``RuntimeError``.
+
 v1.15.1 (2026-05-05)
 ====================
 
--- a/src/bluesky/bundlers.py
+++ b/src/bluesky/bundlers.py
@@ -57,6 +57,11 @@ from .utils import (
 ObjDict = dict[Any, dict[str, T]]
 ExternalAssetDoc: TypeAlias = Datum | Resource | StreamDatum | StreamResource
 
+# The kinds of data a collect object can report, used to choose which collect verb
+# ``RunBundler.collect`` calls, and to reject a device reporting both in one run
+STREAM_ASSETS = "stream assets"
+EVENTS_OR_PAGES = "Events or EventPages"
+
 
 def _describe_collect_dict_is_valid(
     describe_collect_dict: Any | dict[str, Any],
@@ -168,6 +173,10 @@ class RunBundler:
         self._monitor_params: dict[Subscribable, tuple[Callback, dict]] = dict()  # noqa: C408  # cache of {obj: (cb, kwargs)}
         # a cache of stream_resource uid to the data_keys that stream_resource collects for
         self._stream_resource_data_keys: dict[str, Iterable[str]] = dict()  # noqa: C408
+        # what each collect object has reported so far in this run, either STREAM_ASSETS
+        # or EVENTS_OR_PAGES, so that collect() can choose the collect verb from what the
+        # device produced rather than from the protocols it statically satisfies
+        self._reported_data_kinds: dict[HasName, str] = dict()  # noqa: C408
         self.run_is_open = False
         self._uncollected: set[HasName] = set()  # objects after kickoff(), before collect()
         # we expect the RE to take care of the composition
@@ -961,7 +970,8 @@ class RunBundler:
         local_descriptors,
         return_payload: bool,
         message_stream_name: str | None,
-    ):
+    ) -> tuple[list, int]:
+        """Emit an EventPage per set of data keys, returning the payload and how many."""
         payload = []
         pages: dict[frozenset[str], list[Event]] = defaultdict(list)
 
@@ -1002,7 +1012,7 @@ class RunBundler:
                 self._run_start_uid,
                 extra={"doc_name": "event_page", "run_uid": self._run_start_uid},
             )
-        return payload
+        return payload, len(pages)
 
     async def _collect_event_pages(
         self,
@@ -1010,8 +1020,10 @@ class RunBundler:
         local_descriptors,
         return_payload: bool,
         message_stream_name: str | None,
-    ):
+    ) -> tuple[list, int]:
+        """Emit the EventPages the object produces, returning the payload and how many."""
         payload = []
+        emitted = 0
 
         if message_stream_name:
             compose_event_page = self._descriptors[message_stream_name].compose_event_page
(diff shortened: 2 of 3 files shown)


coretl:
CI note: 13 of 16 jobs are green. Three fail, all of them `test (ubuntu-latest, …) / run` — 3.11, 3.13 and dev. These are not this PR's:

- `test (ubuntu-latest, dev)`: `1 failed, 1856 passed, 22 skipped, 2 xfailed`, the failure being `test_run_engine.py::test_single_sigint_no_carry_over[False]` — a SIGINT/timing test that raised `RunEngineInterrupted` rather than pausing at the expected point.
- `test (ubuntu-latest, 3.11)`: `1374 passed` and then the same test tore the session down with a `KeyboardInterrupt` out of `threading.py`, so pytest exited with code 2.
- `test (ubuntu-latest, 3.13)`: the same job that fails on `main`.

`test (ubuntu-latest, 3.11)` and `test (ubuntu-latest, 3.13)` both fail on `main` itself at this PR's base commit 79e2d45 ([run 32841323537](https://github.com/bluesky/bluesky/actions/runs/32841323537)), so the base branch is red for these two before this change.

Nothing here touches `run_engine.py` or SIGINT handling — the diff is `bundlers.py` plus tests — and every collect/paging test passes on every job, including the three that failed. Locally the full suite is `1855 passed, 24 skipped, 2 xfailed`, with `pre-commit` and `mypy src` clean.

I have not re-run the failing jobs or tried to fix these tests, since they are pre-existing on `main`. Happy to if you would rather.

---
_Generated by [Claude Code](https://claude.ai/code)_

coretl:
Two follow-ups to the note above, neither changing the conclusion.

**`codecov/project` is green.** It briefly reported `89.31% (-0.22%) compared to 79e2d45` and failed, then resolved to success once the remaining coverage uploads landed (check run 100096093824). The three failing test jobs skip their `Upload coverage to Codecov` step, so the intermediate number was measured with their uploads missing. `codecov/patch` is also green.

**A fourth flaky failure, on the duplicate push-triggered run.** Commit a9460b7 has two CI runs — the `pull_request` run [33580259912](https://github.com/bluesky/bluesky/actions/runs/33580259912) and the `push` run [33580069764](https://github.com/bluesky/bluesky/actions/runs/33580069764). My earlier comment counted the `pull_request` run's 16 jobs. On the `push` run, `test (windows-latest, 3.11)` also failed:

```
FAILED src/bluesky/tests/test_suspenders.py::test_suspender_wrapper[False]
  - assert 0.8953683376312256 > ((0.5 + 0.2) + 0.2)
1 failed, 1826 passed, 52 skipped, 2 xfailed in 495.39s
```

That is a wall-clock assertion missing its threshold by 5 ms on a slow runner. **The same job at the same commit passed on the `pull_request` run**, so it is non-deterministic by direct observation rather than by inference — which is stronger evidence than a re-run would give, so I have not spent one.

Across both runs there are two distinct failing tests — `test_run_engine.py::test_single_sigint_no_carry_over` and `test_suspenders.py::test_suspender_wrapper` — both timing/signal sensitive, neither touching `bundlers.py` or anything else in the diff. Every collect and paging test passes on every job, including the ones that failed.

I am not pushing anything for these: the only way to make them pass inside this PR would be to loosen or disable tests unrelated to the change, which I would rather not do uninvited. If the timing thresholds in `test_suspenders.py` are worth widening, that seems like its own PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

coretl:
Apologies for the AI spam, will work it up locally and re-submit

PR #2060, opened 2026-09-02
TST: attempt to fix flaky test

jacopoabramo:
## Description

This is an attempt to fix a flaky test that randomly timeouts in the linux CI.

## Motivation and Context

`test_single_sigint_no_carry_over` was introduced in #1997 

The docstring says:

```python
"""A single SIGINT does not pause at the next plan's checkpoint"""
```

Basically the first SIGINT causing the pause of the first test plan is deferred to a background thread. `_tracked_request_pause` monkey patches `RE.request_pause` to set deferred_pause_done event. Since the first SIGINT is a deferred pause, it sets that event.

This patch removes the `first=False` guard (which was also faulty, since it was always false) and waits for the event to be set after *outside* the plan after the first deferred pause.

## How Has This Been Tested?

This happens only in Linux and randomly on CI. I did not have the chance to test it locally, I'm waiting for the CI to do it's thing.

### DIFF
--- a/src/bluesky/tests/test_run_engine.py
+++ b/src/bluesky/tests/test_run_engine.py
@@ -1020,18 +1020,16 @@ def test_single_sigint_no_carry_over(RE):
         os.kill(pid, signal.SIGINT)
 
     def test_plan():
-        first = False
         for _ in range(5):
             yield Msg("null")
-            if first:
-                deferred_pause_done.wait()
-                first = False
 
     # Single SIGINT defers a pause but plan finishes anyway
     sigint_thread = threading.Thread(target=send_sigint, daemon=True)
     sigint_thread.start()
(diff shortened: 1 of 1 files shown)


jacopoabramo:
Well, at least now all jobs fail.

@thopkins32 do you have maybe a better suggestion?

thopkins32:
Thanks for raising this. I think I see the reason why the original version failed. I was supposed to set `first = True` before the loop within the plan, but I think this wouldn't work because `threading.Event.wait` within a plan would block the RE event loop, meaning the sigint wouldn't be processed, causing a deadlock.

I just pushed a change, hopefully it resolves this. It uses `asyncio.Event` which won't block the `RE.loop`.

thopkins32 (approved):

PR #2063, opened 2026-09-02
ENH: support per-device stage and unstage groups

### LINKED ISSUE #1979: Allow specifying groups for each device in bluesky stage_decorator
[stage_wrapper](https://github.com/bluesky/bluesky/blob/aa191e31bea361310f3466e1cc8d9486a603f2fd/src/bluesky/preprocessors.py#L979) (and decorator) only let you pass in the plan and devices. 

I'd like to specify which devices to stage and unstage at the entry point to my plans and wait on them at specific times in the plan to minimise blocking waits. We can do this if we add optional arguments in `stage_wrapper`. 

For example the wrapper could change to something like
`stage_wrapper(plan: MsgGenerator, devices: list[Device] | list[tuple[Device, str | None, str | None]])`,
 where the two strings are staging and unstaging groups.

## Acceptance Criteria
- Stage wrapper/decorator allows optionally specifying groups for staging and unstaging for each device

niketh-keshavan:
## Description
Add per-device staging and unstaging groups to `stage_weapper` and `stage_decorator`.

Wrappers now accept collection of three-tuples:
```python3
(device, stage_group, unstage_group)
```
Each group can be any hashable vale or `None`. Devices are staged in input order and unstaged in reverse order. Callers manage the waits explicitly and plain-device API keeps auto. groups and waits. 

Root normalization and first-entry handling are unchanges. Both forms are documented in `stage_wrapper` docstring.

## Motivation and Context
Plans may need to begin work while some devices are still staging, waiting for each device only when it is first needed. Similarly, an enclosing plan may need to control when asynchronous unstaging completes.

closes: https://github.com/bluesky/bluesky/issues/1979

## How Has This Been Tested?
Added focused tests covering:
- existing plain-device behavior with async stage and unstage status
- automatically generated groups and wait messages
- exact propagation of string, non-string, and `None` groups
- input-order strings and reverse order unstaging
- caller positioned waits
- both `stage_wrapper` and `stage_decorator`
- unstaging cleanup when wrapped plan raises
- root-ancenstor normalization, duplicate removal, retention of first entry's groups

### DIFF
--- a/src/bluesky/tests/test_preprocessors.py
+++ b/src/bluesky/tests/test_preprocessors.py
@@ -11,23 +11,11 @@ from bluesky.preprocessors import (
     stage_decorator,
     stage_wrapper,
 )
-from bluesky.protocols import HasHints, HasParent, Movable, Stageable
+from bluesky.protocols import HasHints, HasParent, Movable, Stageable, Status
 from bluesky.run_engine import RequestStop, RunEngine
 
 
-class _StageableDevice:
-    def __init__(self, name, parent=None):
-        self.name = name
-        self.parent = parent
-
-    def stage(self):
-        return []
-
-    def unstage(self):
-        return []
-
-
-class _SuccessfulStatus:
+class _SuccessfulStatus(Status):
     def add_callback(self, callback):
         callback(self)
 
@@ -41,6 +29,18 @@ class _SuccessfulStatus:
     @property
     def success(self):
         return True
+
+
+class _StageableDevice(Stageable):
+    def __init__(self, name, parent=None):
+        self.name = name
+        self.parent = parent
+
+    def stage(self) -> Status:
+        return _SuccessfulStatus()
+
+    def unstage(self) -> Status:
+        return _SuccessfulStatus()
 
 
 def _collect_messages(plan, *, asynchronous_stage=False):
--- a/src/bluesky/tests/test_preprocessors.py
+++ b/src/bluesky/tests/test_preprocessors.py
@@ -8,9 +8,54 @@ from bluesky.preprocessors import (
     contingency_wrapper,
     lazily_stage_decorator,
     msg_mutator,
+    stage_decorator,
+    stage_wrapper,
 )
 from bluesky.protocols import HasHints, HasParent, Movable, Stageable
 from bluesky.run_engine import RequestStop, RunEngine
+
+
+class _StageableDevice:
+    def __init__(self, name, parent=None):
+        self.name = name
+        self.parent = parent
+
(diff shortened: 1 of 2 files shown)


niketh-keshavan:
I don't believe I can request reviews, so per:
> Hello!
> 
> This is no longer a priority for [@olliesilvester](https://github.com/olliesilvester) or myself as we've moved on to other projects but your approach sounds correct to me. I would suggest making a PR and tagging [@danielballan](https://github.com/danielballan) or [@tacaswell](https://github.com/tacaswell) as reviewers.

tagging @danielballan and @tacaswell here

src/bluesky/tests/test_preprocessors.py
```diff
@@ -8,11 +8,56 @@
...
+class _StageableDevice:
+    def __init__(self, name, parent=None):
+        self.name = name
+        self.parent = parent
+
+    def stage(self):
+        return []
+
+    def unstage(self):
+        return []
```
jacopoabramo: Is there a reason not to use ophyd devices for testing? Also why do stage/unstage methods return a list rather than a status?

src/bluesky/tests/test_preprocessors.py
```diff
@@ -8,11 +8,56 @@
...
+
+
+class _SuccessfulStatus:
```
jacopoabramo: Why not use [AlwaysSuccessfullStatus](https://github.com/bluesky/bluesky/blob/6a82217aa93e50e3af8eacbcb86107f21b8f4d67/src/bluesky/tests/conftest.py#L51)? Seems to match this 1-to-1

jacopoabramo (commented):
I don't know exactly the shape of the feature you're trying to implement so I apologize if these are silly questions.

src/bluesky/tests/test_preprocessors.py
niketh-keshavan: > Is there a reason not to use ophyd devices for testing?

`stage()` returns  a `list[Device]` rather than `Status`, so `stage_all` wont emit the `wait` message here
kinda following `test_lazily_stage_decorator` here
> Also why do stage/unstage methods return a list rather than a status?

ur right =D, fixed this.

jacopoabramo (commented):
To me it looks good, but I wouldn't mind have some more context on what is the use case for this, to help both @tacaswell and @danielballan have an easier time reviewing it.

I know it's maybe hard, but could you provide a snippet of *what* you're currently doing with a plan and *what is it you wish* you would be able to do?

I read the associated issue but it's uncharted territory for me.

PR #2065, opened 2026-09-16
BUG: allow suspenders to trip while the event loop is blocked

### LINKED ISSUE #1944: Suspender raises RuntimeError when under heavy load
Hi!

We are running Bluesky plans in [Prefect](https://www.prefect.io/) which generally works fine. However, when we use suspenders we often get a runtime error which prevents the suspender from working.

The error is raised on this line: https://github.com/bluesky/bluesky/blob/2d0d822a40fbedbf5e47322c61d5aa91ad13a4ba/src/bluesky/suspenders.py#L141

It seems like it is taking too long to schedule `really_make_the_event()` here https://github.com/bluesky/bluesky/blob/2d0d822a40fbedbf5e47322c61d5aa91ad13a4ba/src/bluesky/suspenders.py#L165 and timing out. That means that `__make_event()` exits before `self._ev` is set and the `RuntimeError` is raised.

I'm not 100% sure if this is a Prefect related thing, or just that the system is under a lot of load.

The easy fix is just to raise the timeout on the wait from 0.1 to 1. I've tested it in our code base and that fixes the issue.

I can put up a PR straight away that does that, along with some extra context that explains what is happening as it was hard to track down. Its in a fork here: https://github.com/bluesky/bluesky/compare/main...NoxHarmonium:bluesky:main

I know that this isn't really a satisfying fix, but I'm not sure what else we can do. Other ideas I have are:

1. Don't have a timeout - what happens if it can't schedule the thunk on the RE loop? Would it just raise an exception in that case? What guarantees does python/asyncio provide on `call_soon_threadsafe`?
2. Allow the timeout to be specified with an environment variable for people with out of the ordinary environments (like us!)

Let me know what you think!

CAOShurong:
## Description

- create the suspender's `asyncio.Event` immediately instead of waiting up to 100 ms for the RunEngine loop to create it
- keep event mutation on the RunEngine loop through the existing `call_soon_threadsafe` path
- add a regression test that trips a suspender from a device thread while synchronous plan code blocks the loop

## Motivation and Context

When the RunEngine loop is blocked for longer than 100 ms, the previous cross-thread creation path cancels its callback and drops the suspension request. The device callback then logs `RuntimeError: Could not create the suspender event`, and the plan continues instead of suspending.

`asyncio.Event` on the supported Python versions binds to the running loop when `wait()` needs it, so it can be constructed without waiting for the blocked loop. Raising the timeout would only move the same failure boundary.

Fixes #1944

## How Has This Been Tested?

- `python -m pytest -q src/bluesky/tests/test_suspenders.py` — 37 passed on Python 3.12 / Windows
- focused regression repeated five times — 10/10 parametrized runs passed
- Python 3.10 smoke test for constructing an event in a worker thread and awaiting it on the RunEngine-style loop — passed
- changed-file pre-commit hooks (Ruff lint/format and repository checks) — passed
- `mypy src/bluesky/suspenders.py src/bluesky/tests/test_suspenders.py` — passed

A full-suite run also reached 1,770 passed and 79 skipped. The remaining local failures were isolated to unavailable or version-sensitive optional PyQt/OpenTelemetry paths and an unrelated Windows ZMQ timing test; the suspender suite was fully green.

Implementation and regression testing were developed with OpenAI Codex assistance and reviewed through the checks above.

### Instructions to reviewer on how to test:

1. Run `pytest src/bluesky/tests/test_suspenders.py -k loop_is_blocked`.
2. Confirm both RunEngine loop configurations record `_start_suspender` after the loop-blocking plan segment.

### Checks for reviewer

- [x] Would the PR title make sense to a user on a set of release notes


tacaswell:
Hi @CAOShurong ! I think you are new here, can you tell us a bit about how you are using bluesky in your work?

This fix seems too good to be true. Can you please point me to the docs / change note when CPython changes this?  A quick look at the docs for asyncio.Event didn't have anything that jumped out at me.

In addition to the tests, can you do an analysis of why each of the bit of logic we had in there were put in and verify that we have a tests that actually covers those cases?

PR #2067, opened 2026-09-17
Fix/sigint hardening

tacaswell:
## Description
A second pass at fixing the sigint handling.

Aided by an LLM.

## Motivation and Context

We left a race conditions in the sigint handling

## How Has This Been Tested?

All existing tests pass + added a new test.


coretl (approved):
This seems fine to me. Won't fix the flaky `test_sigint_three_hits` test though, is that coming in a later PR?

src/bluesky/utils/__init__.py
```diff
@@ -377,9 +397,8 @@ def handler(signum, frame):
                     self._request = PauseRequest.HARD
                     self._request_event.set()
                 else:
-                    self._released = True
                     self._request_event.set()
-                    self._original_handler(signum, frame)
+                    self._restore_and_reraise(signum, frame)
```
thopkins32, line 401: `self._released` needs to be set to `True` before `self._request_event.set()` is called, otherwise, the `self._request_thread` running `self._watch_request` may never terminate.

This thread, blocked on `self._request_event.wait()` may observe:
```
_request_event.set()
-> exits self._request_event.wait()
-> calls self._request_event.clear()
-> observes self._released = False
-> calls self._request_event.wait() and waits forever
```

thopkins32 (changes requested):
Resolve one race condition. Otherwise, looks good.

src/bluesky/utils/__init__.py
thopkins32: The fix is to restore `self._released = True` before `self._request_event.set()` and remove `self._released = True` from `_restore_and_reraise()`

thopkins32 (approved):

PR #2068, opened 2026-09-17
Mnt/drop old python

tacaswell:
## Description
Builds on #2067 

## Motivation and Context

Cleans up a number pre py3.10 back-compat that we were running.

## How Has This Been Tested?

existing tests still pass.

PR #2069, opened 2026-09-17
Tst/harden and speedup test

tacaswell:
## Description

This is another pass (with aid of an LLM) to clean up anything that leaks out of the tests, make speed up the tests by default (by not running the uid return branch), and make all of the tests parrallel safe.

Builds on #2068

## Motivation and Context
Our tests have had a long history of random hangs and being a bit slow, this is intended to address both of those issues.

## How Has This Been Tested?

Should be no user facing changes, only work on making the tests more reliable.


coretl (approved):

thopkins32 (approved):
LGTM. Feel free to merge after #2068

PR #2070, opened 2026-09-17
Mnt/lots of bugfixes

tacaswell:
## Description

Builds on #2069 

Assisted by an LLM

## Motivation and Context

I asked an LLM to review bluesky looking for bugs and modernizations and it found a lot of tiny edge cases.

Each commit should be able to be review atomically.

## How Has This Been Tested?

Most bugs have a new test added, some are being more correct about an unreachable branch.


coretl (approved):

src/bluesky/callbacks/core.py
```diff
@@ -434,7 +434,8 @@ def event(self, doc):
         super().event(doc)
 
     def stop(self, doc):
-        if ensure_uid(doc["run_start"]) != self._start["uid"]:
+        # Ignore a 'stop' with no preceding 'start', or one from another run.
+        if self._start is None or ensure_uid(doc["run_start"]) != self._start["uid"]:
             return
```
thopkins32: Should this ignore or raise? Before this would raise with a partially unclear error. A `RemoteDispatcher` that joins an event stream mid-run will have missed the start document, so maybe ignoring here is correct. Just want to double check with you.

src/bluesky/tests/test_callbacks.py
```diff
@@ -593,6 +624,18 @@ def test_broken_table():
         assert ln.strip() == "failed to format row"
 
 
+def test_livetable_stop_without_start():
+    """A 'stop' document with no preceding 'start' must be ignored, not crash.
```
thopkins32: Double checking the intention of ignoring this instead of raising.

src/bluesky/tests/test_run_engine.py
```diff
@@ -2190,6 +2248,32 @@ def bad_plan():
...
+    def bad_plan():
+        yield Msg("pause")
+
+    with pytest.raises(RunEngineInterrupted):
+        RE(bad_plan())
+
+    rs = getattr(RE, bail_func)()
+    assert isinstance(rs.exception, exc_type)
+    # Guard against the class being stored instead of an instance.
+    assert not isinstance(rs.exception, type)
```
thopkins32, line 2274: Small nitpick: `issubclass(rs.exception, exc_type)` preferred here to be more explicit?

thopkins32 (approved):
Meant to approve. Left some comments, but they do not block merging. Feel free to merge once #2069 is merged.

src/bluesky/callbacks/core.py
tacaswell: That was the intent, but I'll pull this commit out (I have another API change commit waiting in the wings, I'll do those two together).

src/bluesky/tests/test_run_engine.py
tacaswell: The issue is that we were setting `RE._exception = RequestAbort` rather than `RequestAbort()`.  From the point of view of Python's machinery this is fine as it will auto-create the instance when raise (and I suspect `.throw`).  Not sure of the exact history of that, but I suspect it is a niceity to save typing and a hold-over from Python 2 syntax).

all class's are instances of type type so this is checking that it is not an instance of type type to verify that the RE instantiated the exception class.

This is indeed redundent with the previous line though as

```
>>> isinstance(ValueError, ValueError)
False
```

We can either leave it as a guide to future readers or drop it as redundant.

PR #2071, opened 2026-09-17
Add function that can serialize Msg objects into json-safe containers

jwlodek:
Allows for serializing messages into json-safe structures, which would allow things like creating a msg hook that can stream message information over a socket.

### Instructions to reviewer on how to test:
1. Create a `Msg`
2. Call `msg_to_json_safe_dict` and make sure the result is a json-compatible dict.

### Checks for reviewer
- [X] Would the PR title make sense to a user on a set of release notes


src/bluesky/utils/__init__.py
```diff
@@ -82,6 +118,21 @@ def __repr__(self):
         return f"Msg({self.command!r}, obj={self.obj!r}, args={self.args}, kwargs={self.kwargs}, run={self.run!r})"
 
 
+    def serialize(self) -> dict[str, Any]:
```
tacaswell: This is all public state, I think a free-function `msg_to_json` or something similar is a better path.  There may be more than one reasonable way to serialize `Msg` and this is taking a very unique name for that.

jwlodek: Makes sense to me, I'll rework it into a free function.

tacaswell (approved):

PR #2076, opened 2026-09-18
TST: stop the sigint tests leaking SIGINT into the pytest session

coretl:
# Description

`test_sigint_three_hits` was flaky, and the sigint tests could kill the whole pytest session. This fixes those test by making them depend on a single fixture.

# Details

The tests fire more Ctrl+C than the RunEngine needs. Mostly that worked, but a late hit could land after the RunEngine had handed SIGINT back to pytest, killing the run rather than being ignored the way a Ctrl+C typed at an idle prompt is. On main, 30 repetitions of `-k sigint -n 2` lose 4 runs that way.

SigintOwner owns SIGINT for a with block: hits arriving after the RunEngine releases it are absorbed, sender threads are joined on exit, sending is then refused, and only then is pytest's disposition restored so no hit can escape. Every sigint test now sends inside such a block, through background(), with its own sender body.

The old fixture also faked the clock the handler debounces against, which the timing tests could not use. SigintHandler now takes its clock as a constructor argument, injected through the documented context_managers hook, so the two concerns are separate fixtures. That also removes a `patch("bluesky.utils.time.monotonic", …)` which was patching stdlib time.monotonic process-wide.

The remaining race in `test_sigint_three_hits` was the 0.5 s motor move; it now blocks on a Status the test completes.


src/bluesky/tests/test_run_engine.py
```diff
@@ -728,44 +736,46 @@ def simple_plan():
 
 
 @uses_os_kill_sigint
-def test_sigint_three_hits(RE, hw, deterministic_sigint):
-    motor = hw.motor
-    motor.delay = 0.5
+@requires_ophyd_async
```
tacaswell: Is the switch to ophyd-async part of fixing the escaping sigint?  If not the churn makes it a bit harder to review.

coretl: Not part of it, replaced with a `Status` producing Device instead

src/bluesky/tests/conftest.py
```diff
@@ -210,62 +210,156 @@ def cleanup_any_figures(request):
...
+        # does not run until the plan ends for some other reason.
+        # The lock spans the check and the signal, and __exit__ takes it before
+        # restoring the real disposition, so a sender cannot pass the check and
+        # then fire into pytest's handler.
+        with self._lock:
+            if self._closed:
+                raise RuntimeError("SIGINT sent after the owner gave the signal back")
+            signal.pthread_kill(threading.main_thread().ident, signal.SIGINT)
+        if not self._handler_done.wait(timeout=10):
+            raise RuntimeError("SIGINT was never handled")
```
coretl, line 273: Note that `DuringTask.block` and `DefaultDuringTask.block` do an untimed wait in 3 places:
https://github.com/bluesky/bluesky/blob/8bccfde7875aeb35748d3c9597ece32ce39ccd1e/src/bluesky/utils/__init__.py#L1817
https://github.com/bluesky/bluesky/blob/8bccfde7875aeb35748d3c9597ece32ce39ccd1e/src/bluesky/utils/__init__.py#L1857
https://github.com/bluesky/bluesky/blob/8bccfde7875aeb35748d3c9597ece32ce39ccd1e/src/bluesky/utils/__init__.py#L2022

This means that if the kernel delivers SIGINT to a different thread to main then it won't wake the main thread to run the handler until the event is set by the plan ending for some other reason.

We could fix it by polling like:
https://github.com/bluesky/bluesky/blob/8bccfde7875aeb35748d3c9597ece32ce39ccd1e/src/bluesky/utils/__init__.py#L2014

Then the test can go back to `os.kill` which is a more faithful test of what happens in production.

This looks like a "I pressed Ctrl-C but nothing happened" bug to me. It is present if matplotlib is never imported or has a backend that is not qt/ipympl/nbagg. Should I fix it in this PR, or raise another issue for it?

PR #2078, opened 2026-09-18
Add additional message types & plan stubs for declaring and updating plan-level progress. Update default plans to leverage new feature.

### LINKED ISSUE #2047: Introduce a mechanism for encoding plan progress updates, with support for nested progress for adaptive plans.
At the moment, there is no mechanism for plans to display progress information. Devices can support watcher updates for `set` or `trigger`, tied to the `RE.waiting_hook` which allows for progress to be displayed for individual device operations, but plans themselves have no way to indicate progress.

I propose the following as a solution:

* Two new message types - one to declare a new plan progress indicator, one to increment the progress indicator.
* Plans can produce one or more progress indicators. Typically, a plan would make one for the overall progress, and then start/close one every time it goes into an inner plan/loop. This covers the case of an adaptive experiment that may iterate an undefined number of times - have the inner progress indicator created for every iteration, so that there is still progress feedback available.

For example, for a `count` plan, we would spin up a progress indicator with the plan name + some metadata, and specify that the target number of progress updates is `num`. Then, after each `trigger_and_read`, increment the progress indicator.

Similarly to the `waiting_hook`, there should be a `progress_hook` (or something similar) for clients to be able to hook into this information and produce progress bars and the like.

## Acceptance Criteria
- There is a mechanism for indicating plan-level progress.
- Inner plans can produce progress updates irrespective of the outer progress indicators.
- The schema for progress updates is well defined.
- A hook exists for clients to tap into the progress update stream to display progress bars and the like.

jwlodek:
Adds support for indicating plan-level progress via an additional RE hook with an identical API to the `waiting_hook`.

Changes listed out:

* Added two new message commands: `declare_progress` and `update_progress`.
* Added two new equivalent plan stubs.
* Added a new RE hook `progress_hook` for streaming progress updates. Signature should be identical to `waiting_hook`.
* Add `Watchable` and `WatchableStatus` protocols for improved typing.
* Add `PlanProgress` `Watchable` that encodes the necessary information for a plan update.
* Add methods to the `RE` that call the `progress_hook` at appropriate on recieving a progress message.
* Thread through a `progress_scope` and `parent_progress_scope` optional kwarg through all the builtin plans.
* If provided progress_scope, the builtin plans declare and update progress based on steps.
* Resolve issue with the terminal progress bar where a print at the same time that it was trying to draw something would leak characters.
* Update `ProgressBarManager` to accept multiple streams of status objects.
* Add unit tests for all of the new functionality.

A consideration: as-is, progress updates are entirely opt-in. When calling a default plan directly, the user must pass `progress_scope` to get any plan-level progress updates. I propose that this is behavior is changable via an environment variable, so that while the default remains the current behavior, users can opt-in to enable plan progress automatically if desired.

## Motivation and Context
Resolves: #2047 

## How Has This Been Tested?
* Unit tests
* Manually with the below example:

```Python
"""Demo: plan-driven progress reporting with ProgressBarManager."""

from ophyd.sim import SynAxis, SynGauss

import bluesky.plans as bp
from bluesky import RunEngine


RE = RunEngine({})

# Updates to the pbar manager in this PR allow for one instance to be used for both hooks.
progress_bar_manager = ProgressBarManager()
RE.waiting_hook = progress_bar_manager.new_stream()
RE.progress_hook = progress_bar_manager.new_stream()

# Also subscribe the BEC, to demonstrate that the new progress indicators don't conflict with other output.
RE.subscribe(BestEffortCallback())

motor = SynAxis(name="motor", value=0, delay=0.1)
slow_motor = SynAxis(name="slow_motor", value=0, delay=1.0, events_per_move=10)
det = SynGauss("det", motor, "motor", center=0, Imax=100, sigma=1)

# Basic scan with scan level progress
RE(bp.scan([det], slow_motor, -5, 5, 10, progress_scope="scan"))

# Count with progress in counts
RE(bp.count([det], num=10, delay=0.05, progress_scope="count"))

# Grid scan with both overall and per-axis progress
RE(bp.grid_scan(
    [det], motor, -2, 2, 5, slow_motor, -2, 2, 5,
    progress_scope="grid",
    per_dim_progress=True,
))

# Custom plan with a nested scan progress bar
def my_experiment():
    yield from bps.declare_progress(name="experiment")
    yield from bps.update_progress("experiment", current=0, initial=0, target=3, unit="scan")
    for i in range(3):
        yield from bp.scan(
            [det], slow_motor, -5, 5, 10,
            progress_scope=f"scan_{i}",
            parent_progress_scope="experiment",
        )
        yield from bps.update_progress("experiment", current=i + 1, initial=0, target=3, unit="scan")
    yield from bps.update_progress("experiment", done=True)


RE(my_experiment())
```

image

AI disclosure: Assisted by GH copilot, Claude Opus 4.8.


src/bluesky/protocols.py
```diff
@@ -59,6 +59,25 @@ class Reading(Generic[T], ReadingOptional):
...
+    def done(self) -> bool: ...
+
+    @abstractmethod
+    def watch(self, func: Callable) -> None:
+        """Register a callback to receive progress updates.
+
+        The callback will be called with keyword arguments:
+        ``name``, ``current``, ``initial``, ``target``, ``unit``,
+        ``precision``, ``fraction``, ``time_elapsed``, ``time_remaining``.
+        """
```
jacopoabramo, line 77: Could `func` be annotated more explicitly? Something like ophyd-async's [Watcher](https://github.com/bluesky/ophyd-async/blob/b400cb1e767ce01ad1d5043f5e254ef90613370c/src/ophyd_async/core/_protocol.py#L124)

src/bluesky/plan_stubs.py
```diff
@@ -1529,6 +1529,93 @@ def exception_path(exp):
...
+    current: Any = None,
+    initial: Any = None,
+    target: Any = None,
+    unit: str = "unit",
+    precision: int | None = None,
+    fraction: float | None = None,
+    time_elapsed: float | None = None,
+    time_remaining: float | None = None,
+    done: bool = False,
+) -> MsgGenerator:
```
jacopoabramo, line 1573: Ditto for the annotation, you could reuse the T `TypeVar` from the `utils`

jacopoabramo:
By the way I love this, thank you. It's exactly what I needed to implement a UI progress bar.

thopkins32 (commented):
How does this work with RE pause/resume? Not sure these new RE messages can be replayed. I think adding tests for those cases would be good. These messages may have to be put in `_UNCACHEABLE_COMMANDS`.

For example,
```python
def plan():
    yield Msg("checkpoint")
    yield Msg("declare_progress", name="scan")
    yield Msg("pause")  # could be a suspender trip or hard pause
    yield Msg("update_progress", name="scan", done=True)
```
`Msg("declare_progress", name="scan")` is cached and then replayed on `RE.resume()`, meaning that the `"scan"` scope is declared twice. I think there may be a similar issue if `"update_progress"` is replayed with `done=True`.

src/bluesky/plans.py
```diff
@@ -1479,7 +1674,86 @@ def _set_snaking(chunk, value):
...
+        dim_names = [motor.name for motor, start, stop, num, snake in chunk_args]
+        total_steps = len(full_cycler)
+
+        def _per_dim_wrapper(inner_per_step):
+            """Wrap per_step to update per-dimension progress bars."""
+            flat_index = 0
+
+            def wrapped(detectors, step, pos_cache):
+                nonlocal flat_index
+                yield from inner_per_step(detectors, step, pos_cache)
```
thopkins32, line 1688: Wrapping `per_step` breaks the function signature inspection done here: https://github.com/jwlodek/bluesky/blob/c2040e739043a060a0e8890275dfe84d9e59d86f/src/bluesky/plans.py#L1260-L1324

It doesn't allow for one to specify `per_step` with this signature anymore: https://github.com/jwlodek/bluesky/blob/c2040e739043a060a0e8890275dfe84d9e59d86f/src/bluesky/plans.py#L1322

So if I write a 1-D per-step hook like this and use a grid scan with 1 motor, then it passes validation but the parameter names no longer match the contents.
```python
def per_step(detectors, motor, step):
    print(motor)  # is actually `step` dict
    print(step)  # is actually `pos_cache` dict

RE(grid_scan([hw.det], hw.motor, -1, 1, 2, per_step=per_step, progress_scope="per-dim", per_dim_progress=True))
```

src/bluesky/utils/__init__.py
```diff
@@ -1396,6 +1398,86 @@ def apply_to_dict_recursively(d, f):
...
+    ``ProgressBarManager`` / ``TerminalProgressBar`` infrastructure.
+    Created by the ``declare_progress`` message and updated by
+    ``update_progress``.
+    """
+
+    def __init__(self, name: str, *, parent: "PlanProgress | None" = None):
+        self.name = name
+        self.parent = parent
+        self._done = False
+        self._watchers: list[Callable] = []
```
thopkins32, line 1414: I don't think anything is clearing this callback list. This will accumulate callbacks when the manager rebuilds and it uses a `TerminalProgressBar`, no?

```
ProgressBarManager._update_stream()
-> ProgressBarManager._rebuild()
    -> self.pbar_factory(statuses)
        -> TerminalProgressBar.__init__()
            -> status.watch(...)
```

The manager probably needs to clear this `self._watchers` list when appropriate.

